PR #3, opened 2016-03-23
Update instructions for Firebird 3 compiler msvc12

mariuz:


dyemanov:
It's actually about Firebird 4 compiler ;-) If backported to the B3_0_Release branch, the official MSVC version should be changed to msvc10.

PR #5, opened 2016-03-25
Add .gitignore to windows builds (Visual Studio files) and VC14 Solution

mariuz:


aafemt:
One .gitignore file in tree root is enough, there is no point to spread this garbage everywhere.

PR #9, opened 2016-03-26
Fix CORE-5173

aafemt:
Commented out assertion because execution flow in nodes is not what I expected when wrote this assertion. Added comment in code to remember that.

### DIFF
--- a/src/dsql/StmtNodes.cpp
+++ b/src/dsql/StmtNodes.cpp
@@ -569,6 +569,7 @@ const StmtNode* BlockNode::execute(thread_db* tdbb, jrd_req* request, ExeState* 
 					// The savepoint of this block will be dealt with below.
 					// Do this only if error handlers exist. If not - leave rollbacking to caller node
 					while (transaction->tra_save_point &&
+						count < transaction->tra_save_point->sav_number &&
 						transaction->tra_save_point->sav_next &&
 						count < transaction->tra_save_point->sav_next->sav_number)
 					{
--- a/src/dsql/StmtNodes.cpp
+++ b/src/dsql/StmtNodes.cpp
@@ -558,6 +558,31 @@ const StmtNode* BlockNode::execute(thread_db* tdbb, jrd_req* request, ExeState* 
 
 			if (handlers && handlers->statements.getCount() > 0)
 			{
+				// First of all rollback failed work
+				if (transaction != sysTransaction)
+				{
+					count = *request->getImpure<SLONG>(impureOffset);
+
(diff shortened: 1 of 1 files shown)


src/dsql/StmtNodes.cpp
```diff
@@ -647,9 +647,16 @@ const StmtNode* BlockNode::execute(thread_db* tdbb, jrd_req* request, ExeState*
...
 					// If not - mirror copy of CORE-4424 or CORE-4483 is around here.
-					fb_assert(transaction->tra_save_point && transaction->tra_save_point->sav_number == count);
+					// Except the case of nested exception handlers and throwing in inner one.
+					// In this case execution flow is like this:
+					//		inner before (block that rollbacking savepoints above)
+					//		outer before
+					//		outer after  (this block)
+					//		inner after
+					// Because of this following assert is commentd out
+					//fb_assert(transaction->tra_save_point && transaction->tra_save_point->sav_number == count);
```
dyemanov: What is "before" and "after" here?

aafemt: These are parts of code managing savepoints that are placed before and after call of looper for exception handlers, as written in parentheses.

dyemanov: I'd prefer to see the bad execution flow fixed rather than working around with a disabled assert.

aafemt: I'm not sure that recursive handling is a bad thing. In any case this is a bigger work for different ticket.

aafemt:
Added more protection as agreed in discussion in #10

src/dsql/StmtNodes.cpp
```diff
@@ -569,6 +569,7 @@ const StmtNode* BlockNode::execute(thread_db* tdbb, jrd_req* request, ExeState*
 					// The savepoint of this block will be dealt with below.
 					// Do this only if error handlers exist. If not - leave rollbacking to caller node
 					while (transaction->tra_save_point &&
+						count < transaction->tra_save_point->sav_number &&
 						transaction->tra_save_point->sav_next &&
 						count < transaction->tra_save_point->sav_next->sav_number)
 					{
```
dyemanov: In tra.cpp, one condition is ">" and another one is ">=". Is it correct that both conditions in this patch are strict inequalities?

aafemt: Yes. In tra.cpp rolling forward must stop immediatelly before given savepoint, here - before one savepoint before the given one because there we must rollback the savepoint, here - rollback everything but the savepoint as it must be applied afterward.

PR #10, opened 2016-03-27
Fix CORE-5174

aafemt:
Take into account possible reverse of savepoint number sequence in short chains.
Example: 
For savepoints sequence "12 -> 14 -> 7 -> 1" request to rollback savepoints up to 12, must result in "14 -> 7 -> 1", not "7 -> 1" as it might happen before this fix.

### DIFF
--- a/src/jrd/tra.cpp
+++ b/src/jrd/tra.cpp
@@ -3560,8 +3560,8 @@ void jrd_tra::rollbackToSavepoint(thread_db* tdbb, SLONG number)
  **************************************/
 {
 	// merge all but one folowing savepoints into one
-	while(tra_save_point && tra_save_point->sav_next &&
-		  tra_save_point->sav_next->sav_number >= number)
+	while(tra_save_point && tra_save_point->sav_number > number &&
+		  tra_save_point->sav_next && tra_save_point->sav_next->sav_number >= number)
 	{
 		rollforwardSavepoint(tdbb);
 	}


dyemanov:
How could savepoint 7 be started after savepoint 14?

aafemt:
Stack is shown from new savepoints to old one. Savepoint 1 is transaction savepoint, savepoint 7 is statement savepoint.
If you had on mind "12", not "7", then yes, it is result of suspend trick.

src/jrd/tra.cpp
```diff
@@ -3560,8 +3560,8 @@ void jrd_tra::rollbackToSavepoint(thread_db* tdbb, SLONG number)
  **************************************/
 {
 	// merge all but one folowing savepoints into one
-	while(tra_save_point && tra_save_point->sav_next &&
-		  tra_save_point->sav_next->sav_number >= number)
+	while(tra_save_point && tra_save_point->sav_number > number &&
+		  tra_save_point->sav_next && tra_save_point->sav_next->sav_number >= number)
 	{
```
dyemanov: 1) AFAIU, you assume here that only one savepoint in the stack may have its number out of order, supposedly this is the looper-generated one. Couldn't there be more? E.g. started by BlockNode or SavePointNode?
2) Shouldn't the same check be applied to other places like BlockNode::execute()? I see there the kissing cousin of the loop you've fixed in tra.cpp.

aafemt: 1) No, but I assume that there cannot be savepoints with smaller number before the given one in stack. I cannot imagine how such case can happen.
2) Probably yes, but so far no testcase activated similar assert there.

PR #15, opened 2016-04-16
Basic support for Unicode file names and some bugfixes

aafemt:
Unicode file names and pathes work at API level. Little glitches can be fixed after Pavel Zotov create test suite.
Much more must be done for unicode support in utilities command lines.

### DIFF
--- a/src/auth/SecurityDatabase/LegacyManagement.epp
+++ b/src/auth/SecurityDatabase/LegacyManagement.epp
@@ -148,6 +148,8 @@ void SecurityDatabaseManagement::start(Firebird::CheckStatusWrapper* st, Firebir
 		if (role && role[0])
 			dpb.insertString(isc_dpb_sql_role_name, role, fb_strlen(role));
 
+		dpb.insertTag(isc_dpb_utf8_filename);
+
 		ISC_STATUS_ARRAY status;
 		if (isc_attach_database(status, 0, secDbName, &database,
 								dpb.getBufferLength(), reinterpret_cast<const char*>(dpb.getBuffer())))
@@ -565,7 +567,9 @@ int SecurityDatabaseManagement::execute(Firebird::CheckStatusWrapper* st, Firebi
 							if (!U.PLG$GROUP_NAME.NULL)
 								a3.printf("GroupName=%s\n", U.PLG$GROUP_NAME);
 
-							attr = a1 + a2 + a3;
+							attr = a1;
+							attr += a2;
+							attr += a3;
 							user->attributes()->set(s, attr.c_str());
 							check(s);
 							user->attributes()->setEntered(s, attr.hasData() ? 1 : 0);
@@ -648,7 +652,9 @@ int SecurityDatabaseManagement::execute(Firebird::CheckStatusWrapper* st, Firebi
 						if (!U.PLG$GROUP_NAME.NULL)
 							a3.printf("GroupName=%s\n", U.PLG$GROUP_NAME);
 
-						attr = a1 + a2 + a3;
+						attr = a1;
+						attr += a2;
+						attr += a3;
 						user->attributes()->set(s, attr.c_str());
 						check(s);
 						user->attributes()->setEntered(s, attr.hasData() ? 1 : 0);
--- a/src/auth/SecurityDatabase/LegacyManagement.epp
+++ b/src/auth/SecurityDatabase/LegacyManagement.epp
@@ -567,9 +567,7 @@ int SecurityDatabaseManagement::execute(Firebird::CheckStatusWrapper* st, Firebi
 							if (!U.PLG$GROUP_NAME.NULL)
 								a3.printf("GroupName=%s\n", U.PLG$GROUP_NAME);
 
-							attr = a1;
-							attr += a2;
-							attr += a3;
+							attr = a1 + a2 + a3;
 							user->attributes()->set(s, attr.c_str());
 							check(s);
 							user->attributes()->setEntered(s, attr.hasData() ? 1 : 0);
@@ -652,9 +650,7 @@ int SecurityDatabaseManagement::execute(Firebird::CheckStatusWrapper* st, Firebi
 						if (!U.PLG$GROUP_NAME.NULL)
 							a3.printf("GroupName=%s\n", U.PLG$GROUP_NAME);
 
-						attr = a1;
-						attr += a2;
-						attr += a3;
+						attr = a1 + a2 + a3;
 						user->attributes()->set(s, attr.c_str());
 						check(s);
 						user->attributes()->setEntered(s, attr.hasData() ? 1 : 0);
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -658,7 +658,7 @@ void add_files(BurpGlobals* tdgbl, const char* file_name)
 
 	for (burp_fil* file = tdgbl->gbl_sw_files; file; file = file->fil_next)
 	{
-		if (!(file->fil_name == file_name))
+		if (file->fil_name != file_name)
 		{
 			if (start > MAX_ULONG)
 			{
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -562,6 +562,7 @@ int RESTORE_restore (const TEXT* file_name, const TEXT* database_name)
 
 	// set forced writes to the value which was in the header
 	dpb.insertByte(isc_dpb_force_write, tdgbl->hdr_forced_writes ? 1 : 0);
+	dpb.insertTag(isc_dpb_utf8_filename);
 
 	FB_API_HANDLE db_handle = 0;
 	if (isc_attach_database(tdgbl->status_vector, 0, database_name, &db_handle,
@@ -657,7 +658,7 @@ void add_files(BurpGlobals* tdgbl, const char* file_name)
 
 	for (burp_fil* file = tdgbl->gbl_sw_files; file; file = file->fil_next)
 	{
-		if (file->fil_name != file_name)
+		if (!(file->fil_name == file_name))
 		{
 			if (start > MAX_ULONG)
 			{
@@ -884,6 +885,7 @@ void create_database(BurpGlobals* tdgbl, const TEXT* file_name)
 
 	dpb.insertInt(isc_dpb_page_size, page_size & 0xff00);
 	dpb.insertString(isc_dpb_gbak_attach, FB_VERSION, fb_strlen(FB_VERSION));
+	dpb.insertTag(isc_dpb_utf8_filename);
 
 	if (sweep_interval != MAX_ULONG)
 	{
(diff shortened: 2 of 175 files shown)


src/common/classes/fb_string.h
```diff
@@ -411,10 +411,10 @@ namespace Firebird
 		}
 		void insert(size_type p0, const_pointer s, const size_type n)
 		{
```
AlexPeshkoff: IMO this fix should be done as separate pull request and backported to 2.5 and 3.0.

aafemt: No, this is regression of this branch, problem does not exist in other branches.

AlexPeshkoff: On 06/16/2016 07:58 PM, Dimitry Sibiryakov wrote:
>> @@ -411,10 +411,10 @@ namespace Firebird
>>   		}
>>   		void insert(size_type p0, const_pointer s, const size_type n)
>>   		{
> No, this is regression of this branch, problem does not exist in other branches.

OK, in that case no problems.

src/common/os/posix/os_utils.cpp
```diff
@@ -393,4 +395,147 @@ void CtrlCHandler::handler(void*)
...
+			outBuf = NULL;
+		}
+	};
+
+#ifdef HAVE_ICONV_H
+	class IConverter : public Converter
+	{
+	public:
+
+		IConverter(const char* sysLocale) : ic(NULL)
```
hvlad: Class name "IConverter" points that class is (kind of) interface while this is not so.
Base class "Converter" is pure virtual, i.e. it is good candidate to be "IConverter" :)
Please, avoid confusion when choosing names.

aafemt: This is just a copy-paste from isc_file.cpp::IConv class. "I" in this case has nothing to do with interfaces.

dyemanov: Then it should better be renamed to IntlConverter (or whatever "I" stands for there).

aafemt: I leave it to one who will finish this POSIX port.

hvlad: > This is just a copy-paste from isc_file.cpp::IConv class.
This is not true. 
I see "class IConv" in isc_file.cpp (yes, bad name too) while you introduced 3 new classes and none of them is "IConv".
I.e. you already made changes in copied code.
If you don't want to do something - say "i don't want", no need to create empty reasoning...

> I leave it to one who will finish this POSIX port.
So, you ask us to merge half-made code and to "finish" it after you ? Cool...

aafemt: No. I ask for a help from someone with POSIX knowledge. Create pull request to this 
branch, I'll check and merge it. Then whole work can be merged into master branch.

hvlad: >  I ask for a help from someone with POSIX knowledge

It is not related with my original comment re. class names

AlexPeshkoff: > This is just a copy-paste from isc_file.cpp::IConv class.

class IConv is a helper for calling function iconv(): http://linux.die.net/man/3/iconv
It has nothing to do with interfaces but exactly in this form I treat it's name as good.

What about copy-paste - this is what I do not understand. Why have 2 same (may be almost same) classes with different names? You should have single class performing all actions required from iconv() and use them in all c++ modules where you need. I'm strongly against copyNpaste techology!

AlexPeshkoff: Re posix port. I do not see big problems with having incomplete OS-A build provided author is using OS-B. This can be fixed a bit later (specially at pre-alpha stage). Certainly it does not mean that code for OS-A may get poisoned with bad programming style.

aafemt: They are not same. This one is consuming 4 times less memory, but do conversion from UTF-8 
only. Different name to be avoid Linux symbol hell.

AlexPeshkoff: On 06/29/2016 12:59 PM, Dimitry Sibiryakov wrote:
>> +		char* convert(const char* fromBuf, size_t fromLen) override
>> +		{
>> +			return const_cast<char*>(fromBuf);
>> +		}
>> +		void destroy(char* &outBuf)
>> +		{
>> +			outBuf = NULL;
>> +		}
>> +	};
>> +
>> +#ifdef HAVE_ICONV_H
>> +	class IConverter : public Converter
>> +	{
>> +	public:
>> +
>> +		IConverter(const char* sysLocale) : ic(NULL)
> They are not same. This one is consuming 4 times less memory, but do conversion from UTF-8
> only. Different name to be avoid Linux symbol hell.

In that case one class can be derived from another.
Anyway please avoid code duplication!

egorpugin:
Why just not use boost::filesystem and remove tons of code?
Especially when boost::filesystem was accepted into C++17.

hvlad:
What issue the last commit (2f89d89) solved ? Just for fun ?

aafemt:
It got rid of magic numbers before Adriano requested that.

egorpugin:
Size of `dllExt` string is known at compile time, so it's better to use `sizeof()` operator.

aafemt:
So, strlen also will be calculated at compile time. But in contrast with sizeof it will 
return right value, not size of pointer.

egorpugin:
`sizeof` for `static const char dllExt[] = ".dll";`, not for `const char *`.
Why `strlen()` is compile time?

aafemt:
Because compilers are clever nowadays.

egorpugin:
It will probably require modern `constexpr` keyword.
By the way on line 195 of 2f89d89 you can use also `dllExt` and not `".dll"`.

src/auth/SecurityDatabase/LegacyManagement.epp
```diff
@@ -565,7 +567,9 @@ int SecurityDatabaseManagement::execute(Firebird::CheckStatusWrapper* st, Firebi
 							if (!U.PLG$GROUP_NAME.NULL)
 								a3.printf("GroupName=%s\n", U.PLG$GROUP_NAME);
 
-							attr = a1 + a2 + a3;
+							attr = a1;
```
AlexPeshkoff: This change has nothing to do with unicode support

aafemt:
> This change has nothing to do with unicode support


Yes, this is result of string refactoring which you demanded.

aafemt:
> What's a reason in using here plain char pointer instead class reference?


This routine used to be called using string literal. Creating of temporary class is 
unnecessary overhead. In rare cases when it is called with data contained in string, call 
of c_str() has no overhead.

aafemt:
> Specially strange taken together with introduced by you assignment operator using rvalue
> reference.


This operator has no advantage on short strings which fit into internal buffer. That's why 
control of assignments is still important for performance.

src/common/classes/fb_string.h
```diff
@@ -430,108 +314,217 @@ namespace Firebird
...
+				}
+				else // fallback to copy
+				{
+					assign(v);
+				}
+			}
+		}
 
-		AbstractString& append(const AbstractString& str)
+		void append(const AbstractString& str)
```
AlexPeshkoff: Rollback this massive changes from 'AbstractString&' to 'void' return type. Not related with unicode support (and violate syntax of std::basic_string).

AlexPeshkoff:
On 10/18/16 14:00, Dimitry Sibiryakov wrote:
>> Specially strange taken together with introduced by you assignment operator using rvalue
>> reference.
>
> This operator has no advantage on short strings which fit into internal buffer. That's why
> control of assignments is still important for performance.
>
>

Anyway rollback that "optimization".

src/common/classes/init.cpp
```diff
@@ -102,7 +102,7 @@ namespace
 			if (dladdr((void*) &allClean, &path))
 			{
 				name = path.dli_fname;
-				name += ".memdebug.log";
```
AlexPeshkoff: Use of function call instead redefined operator does not make code better. 
Rollback such changes too.

aafemt:
> Use of function call instead redefined operator does not make code better.


It makes code working. As was agreed in firebird-devel, the function doesn't add directory 
separator before appended string.

AlexPeshkoff:
Dmitry, I stop review.
Please remove _yourself_ optimizations not related to unicode support.
Some of them may be discussed later in separate pull requests.

AlexPeshkoff:
On 10/18/16 14:19, Dimitry Sibiryakov wrote:
>> Use of function call instead redefined operator does not make code better.
>
> It makes code working. As was agreed in firebird-devel, the function doesn't add directory
> separator before appended string.
>
>

OK, agreed.

src/common/classes/fb_string.h
aafemt: These routines are private. They cannot be called from outside code and have nothing to do with public interface of std::basic_string except name.

AlexPeshkoff:
On 10/18/16 15:31, Dimitry Sibiryakov wrote:
> aafemt commented on this pull request.
>
>
>
>>   
> -		AbstractString& append(const AbstractString& str)
> +		void append(const AbstractString& str)
>
> These routines are private.

They are protected. Am I undestand it right that
         using AbstractString::append;
in public section of derived class makes it visible from the outside world?

> They cannot be called from outside code and have nothing to do with public interface of std::basic_string except name.
>

Class string was initially designed to be able to serve as a replacement 
of basic_string<char> with extended functionality and I see no reasons 
to change this ability as long as we have no serious reasons to do it.

I.e. to be precise - not only name but name and functionality.

aafemt:
> in public section of derived class makes it visible from the outside world?

Yes.

> I see no reasons to change this ability as long as we have no serious reasons to do it.

Isn't chained code like "a.append("abc").append("def").c_str()" prohibited by project's style guides?..

AlexPeshkoff:
On 10/18/16 15:48, Dimitry Sibiryakov wrote:
>> in public section of derived class makes it visible from the outside world?
> Yes.
>
>> I see no reasons to change this ability as long as we have no serious reasons to do it.
> Isn't chained code like "a.append("abc").append("def").c_str()" prohibited by project's style guides?..
>

May be - but I do not know about it.
Chains of redefined operators are actively used.

aafemt:
> Chains of redefined operators are actively used.

But Adriano called them "ugly and non-sense", didn't he?..

AlexPeshkoff:
On 10/18/16 15:55, Dimitry Sibiryakov wrote:
>> Chains of redefined operators are actively used.
> But Adriano called them "ugly and non-sense", didn't he?..
>

Do not think so - he is actively them. For example, in SQL builder.
But that all goes offtopic here.

src/burp/restore.epp
```diff
@@ -657,7 +658,7 @@ void add_files(BurpGlobals* tdgbl, const char* file_name)
 
 	for (burp_fil* file = tdgbl->gbl_sw_files; file; file = file->fil_next)
 	{
-		if (file->fil_name != file_name)
```
AlexPeshkoff: Why this change?
Are we missing operator!=()?

aafemt:
> Are we missing operator!=()?

  Yes.

PR #17, opened 2016-04-20
If fakeroot available use it for creating linux install image

artyom-smirnov:
These changes allow to execute 'make dist' without need of root. 'make install' of course still need root.

This piece of code is same in 2.5/3.0 so it can be used there to.

### DIFF
--- a/builds/install/arch-specific/linux/Makefile.in
+++ b/builds/install/arch-specific/linux/Makefile.in
@@ -86,7 +86,7 @@ $(DebugFile) :  buildImageDir
 	(cd $(GEN_ROOT)/$(DebugDir); tar -czf ../$(DebugFile) .)
 
 buildRoot:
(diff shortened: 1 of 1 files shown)


AlexPeshkoff:
Artyom, your usage of fakeroot is not completely correct - all process of package build should happen under same fakeroot, not only makeInstallImage. Currently all root-specific changes, made under fakeroot, get lost before .tgz file is created.

But due to the fact that most of required to build as non-root changes were already done you've succeeded to get installable binary (see CORE-5180). I've missed some changes cause did them in a hurry of 3.0 release which is definitely not good excuse for me.

Currently process of non-root build is tested up to package install. No need in fakeroot for head and 3.0, but if you prepare pull request for 2.5 I'll be glad to merge it. Changes should be made (fakeroot invoked) in builds/posix/Makefile.in.firebird - targets "package packages dist" should be made under fakeroot.

PR #18, opened 2016-04-28
mk/4.0/gcc6

mkubecek:
First patch allows to build with gcc6 in C++14 mode without -fno-sized-deallocation option. Second fixes a misleading indentation gcc6 warns about.


asfernandes:
@mkubecek what about my comment regarding -fno-delete-null-pointer-checks in CORE-5099?

-fno-delete-null-pointer-checks: becase the ->as ->is methods that we call in null pointers. May have others similar things.

mkubecek:
That one is still needed. I plan to look into it as well.

PR #19, opened 2016-04-30
Fix CORE-5223

aafemt:
A little more clever parsing of path allow to use double dots in names but still avoid "up dir hack".

### DIFF
--- a/src/common/config/dir_list.cpp
+++ b/src/common/config/dir_list.cpp
@@ -43,7 +43,7 @@ void ParsedPath::parse(const PathName& path)
 		{
 			continue;
 		}
-		if (elem == ".") // Skip current dir reference
+		if (elem == PathUtils::curr_dir_link) // Skip current dir reference
 		{
 			continue;
 		}
--- a/src/common/config/dir_list.cpp
+++ b/src/common/config/dir_list.cpp
@@ -33,25 +33,44 @@ void ParsedPath::parse(const PathName& path)
 {
 	clear();
 
-	if (path.length() == 1) {
-		add(path);
-		return;
-	}
-
 	PathName oldpath = path;
+	int toSkip = 0;
 	do {
 		PathName newpath, elem;
 		PathUtils::splitLastComponent(newpath, elem, oldpath);
 		oldpath = newpath;
+		if (elem.isEmpty() && !oldpath.isEmpty()) // Skip double dir separator
+		{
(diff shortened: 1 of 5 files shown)


src/common/config/dir_list.cpp
```diff
@@ -33,25 +33,44 @@ void ParsedPath::parse(const PathName& path)
...
 		oldpath = newpath;
+		if (elem.isEmpty() && !oldpath.isEmpty()) // Skip double dir separator
+		{
+			continue;
+		}
+		if (elem == ".") // Skip current dir reference
+		{
+			continue;
+		}
+		if (elem == PathUtils::up_dir_link) // skip next up dir
```
asfernandes: Does FB still run (compiles for) in Win 9X based kernels? I do remember in this OS "..." (or more dots) was valid part of paths, meaning "..\.." and so. Is this compatible? Or is the old code (2.X) compatible with this?

aafemt: AFAIR, Firebird doesn't support Win9X anymore.
Neither Windows 8 nor Linux support three dots in path.

mkubecek:
To be more precise, "..." is a valid name for a directory or a file in Linux, it doesn't have any special meaning:

    mike@unicorn:/tmp> mkdir ...
    mike@unicorn:/tmp> touch .../...
    mike@unicorn:/tmp> ls -la ...
    total 0
    drwxr-xr-x  2 mike users  60 May  4 23:20 .
    drwxrwxrwt 22 root root  720 May  4 23:20 ..
    -rw-r--r--  1 mike users   0 May  4 23:20 ...

src/common/config/dir_list.cpp
```diff
@@ -33,25 +33,44 @@ void ParsedPath::parse(const PathName& path)
...
+	int toSkip = 0;
 	do {
 		PathName newpath, elem;
 		PathUtils::splitLastComponent(newpath, elem, oldpath);
 		oldpath = newpath;
+		if (elem.isEmpty() && !oldpath.isEmpty()) // Skip double dir separator
+		{
+			continue;
+		}
+		if (elem == ".") // Skip current dir reference
```
dyemanov: Worth adding PathUtils::curr_dir_link, I believe.

aafemt: As you wish.

PR #20, opened 2016-05-05
Allow to enforce IPv4 or IPv6 in URL-like connection string.

mkubecek:
Extend the new (URL like) connection string format by protocols inet4 and inet6 which work exactly like inet except the resolver is asked for only IPv4 or only IPv6 addresses. As discussed on firebird-devel mailing list, this can be used as a workaround on systems where the resolver returns an IPv6 address for a name but an attempt to connect to it times out rather than failing quickly.

### DIFF
--- a/doc/README.IPv6
+++ b/doc/README.IPv6
@@ -22,6 +22,19 @@ If a domain name is used in connection string, all addresses (IPv4 and IPv6)
 are tried in the order returned by resolver until a connection is established.
 Only if all attempts fail, the client fails to connect.
 
+New URL-style connection string format (see README.connection_strings) allows
+to restrict name lookup to only IPv4 or IPv6 addresses:
+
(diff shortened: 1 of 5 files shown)


aafemt:
I'd suggest to make legacy connection string host:database a synonym to inet4://

mkubecek:
That would be highly unusual. There are client applications that provide an option to enforce IPv4/IPv6, mostly with '-4' or '-6' option (e.g. ssh, netcat or wget). There are client applications with special IPv6 variant (mostly low level utilities like ping or traceroute). And there are, of course, programs that do not support IPv6 at all as they were written long ago and nobody fixed them yet. But I'm not sure if I have ever seen a client application that would support IPv6 connections but would require a special syntax (or parameter) to be allowed to use it if remote host is passed by name.

PR #21, opened 2016-05-07
Fix locking on big-endian architectures

jrtc27:
4e4d8002e5fe9968b4d5a493fdb567ed773ccbab extended locks to have 64-bit keys in most cases, but some were left as 32-bit. However, code using these 32-bit locks assumed that the significant bytes of the key's long value started from lck_string[0], which is false on big-endian architectures.  This commit adds Lock::getKeyString(), which gets a pointer to the first used byte of lck_string, and should be used in place of accessing lck_string directly.

### DIFF
--- a/src/jrd/GlobalRWLock.cpp
+++ b/src/jrd/GlobalRWLock.cpp
@@ -78,7 +78,7 @@ GlobalRWLock::GlobalRWLock(thread_db* tdbb, MemoryPool& p, lck_t lckType,
 
 	cachedLock = FB_NEW_RPT(getPool(), lockLen)
 		Lock(tdbb, lockLen, lckType, this, lockCaching ? blocking_ast_cached_lock : NULL);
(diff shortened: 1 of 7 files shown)


dyemanov:
James, do you have an account at tracker.firebirdsql.org? I'd like to have this issue recorded there as well.

jrtc27:
Thanks; I just created an account and filed http://tracker.firebirdsql.org/browse/CORE-5232.

dyemanov:
I will polish the getKeyString() function a little and then backport this bugfix into v3.0.1. Thanks for contribution.

jrtc27:
Great. The `lck_length <= 8` check (I guess could be `lck_length < 8`, depends on which is more favourable for branch prediction) falling back on `&lck_string[0]` is very deliberate, as `jrd_rel::getRelLockKeyLength()` gives a 12-byte key. I missed that case and was returning `&lck_string[8-lck_length]` unconditionally, hence this warning.

PR #22, opened 2016-05-09
Fixed CORE-5233: Fix unaligned memory access in MET_format

jrtc27:
This works around the issue by copying into temporary correctly-aligned data structures. It would be nice if this wasn't needed and the data was guaranteed to be correctly aligned, but I don't know if that is possible to do, as it seems this is stored on disk.

### DIFF
--- a/src/jrd/met.epp
+++ b/src/jrd/met.epp
@@ -1480,16 +1480,21 @@ Format* MET_format(thread_db* tdbb, jrd_rel* relation, USHORT number)
 		count = p[0] | (p[1] << 8);
 		p += 2;
 
+		UCharBuffer tmpArray;
 		while (count-- > 0)
 		{
 			USHORT offset = p[0] | (p[1] << 8);
 			p += 2;
 
-			const Ods::Descriptor* odsDflDesc = (Ods::Descriptor*) p;
-			p = (UCHAR*) (odsDflDesc + 1);
+			Ods::Descriptor odsDflDesc;
+			memcpy(&odsDflDesc, p, sizeof(odsDflDesc));
+			p += sizeof(Ods::Descriptor);
 
-			dsc desc = *odsDflDesc;
-			desc.dsc_address = const_cast<UCHAR*>(p);
+			dsc desc = odsDflDesc;
+
+			tmpArray.getBuffer(desc.dsc_length);
+			desc.dsc_address = tmpArray.begin();
+			memcpy(desc.dsc_address, p, desc.dsc_length);
 			EVL_make_value(tdbb, &desc, &format->fmt_defaults[offset], relation->rel_pool);
 
 			p += desc.dsc_length;


aafemt:
May be it would be better to teach EVL_make_value() to work with unaligned data using rooutines from memory_routines.h instead of copying data back and forth?..

asfernandes:
@jrtc27 you can reuse the logic just above the failing code. You can reuse variables odsDescs and odsDesc, after "count = p[0] | (p[1] << 8);", then you can memcpy at once (also like above the failing code).

jrtc27:
@asfernandes unless I'm mistaken, the above code doesn't handle inline data; in this particular case, there is a descriptor with 3-bytes of dtype_dtext after it, so memcpy'ing the entire rest of the buffer isn't going to aline the descriptor following that.

asfernandes:
@jrtc27 you're right.

But now about this part:

`			USHORT align = type_alignments[desc.dsc_dtype];
			if (!align)
				align = 1;

			UCHAR tmp_buffer[desc.dsc_length + align - 1];
			desc.dsc_address = FB_ALIGN(&tmp_buffer[0], align);
`

First, this "UCHAR tmp_buffer[desc.dsc_length + align - 1]" is not very portable (standard) AFAIK.

I think the code would be better if you allocate an Array<UCHAR> (an array always have the max [64 bit] alignment) outside the while. then you can assign tmpArray.getBuffer(desc.dsc_length) to desc.dsc_address, so it will reuse tmpArray as possible and realloc only when necessary.

And as Array returns aligned data, you should not care about FB_ALIGN usage.

jrtc27:
@asfernandes sorry for disappearing for so long. You're right that `UCHAR tmp_buffer[desc.dsc_length + align - 1]` isn't standard; it's available in C99, and gcc supports it in C++ as an extension, but even C++14 doesn't have it as far as I know. Should I go ahead and allocate an array outside the loop like you suggest, or follow @aafemt's proposal of teaching `EVL_make_value` to correctly handle unaligned data?

asfernandes:
@jrtc27 the problem is here, not in EVL_make_value.

jrtc27:
@asfernandes updated and tested

asfernandes:
I merged with changes in master. Please test it so we can backport to 3.0 branch.

jrtc27:
Yep, as expected, the cleaned up version still works. Thanks!

asfernandes:
UCharBuffer does not enforce the 64-bit alignment.

jrtc27:
Oh, interesting, I guess either nothing needed the 64-bit alignment or I got lucky.

AlexPeshkoff:
We have classes specially designed to solve align problems
Look at common/classes/Aligner.h

asfernandes:
They are for static buffers. I think a mix of HalfStaticArray with Align would be good.

asfernandes:
Sorry, they're not for static buffer. I'll make a change.

aafemt:
For static buffers there is __attribute__ ((aligned(X))).

PR #25, opened 2016-05-13
Fixed output of show grants

romansimakov:


dyemanov:
Does ISQL metadata extraction use the same routine for grants? I.e. whether ISQL -X also needs corrections or not.

AlexPeshkoff:
On 05/13/2016 04:08 PM, Dmitry Yemanov wrote:
> Does ISQL metadata extraction use the same routine for grants?

yes

>   I.e. whether ISQL -X also needs corrections or not.

no more corrections needed

PR #33, opened 2016-06-01
Example of database backup via services

aafemt:
Also made example 01 bult on Windows where there is no setenv().

### DIFF
--- a/examples/interfaces/10.backup.cpp
+++ b/examples/interfaces/10.backup.cpp
@@ -1,0 +1,175 @@
+/*
+ *	PROGRAM:	Object oriented API samples.
+ *	MODULE:		10.backup.cpp
+ *	DESCRIPTION:	A sample of making backup of database using service maanger.
+ *
+ *					Example for the following interfaces:
+ *					IService - interface to work with service maanger.
+ *
+ *  The contents of this file are subject to the Initial
+ *  Developer's Public License Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the
+ *  License. You may obtain a copy of the License at
+ *  http://www.ibphoenix.com/main.nfs?a=ibphoenix&page=ibp_idpl.
+ *
+ *  Software distributed under the License is distributed AS IS,
+ *  WITHOUT WARRANTY OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing rights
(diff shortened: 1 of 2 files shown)


examples/interfaces/10.backup.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+		// Output buffer
+		unsigned char results[1024];
+
+		printf("** Demo of making backup using service manager...\n");
+
+		// Build service start SPB
+		spb2 = utl->getXpbBuilder(&status, IXpbBuilder::SPB_START, NULL, 0);
+		spb2->insertTag(&status, isc_action_svc_backup);
+		spb2->insertString(&status, isc_spb_dbname, "fbtests.fdb");
```
AlexPeshkoff: Most of samples use employee database which tends to be present on a server after installation. 
Why not use it here?

aafemt: I, actually, thought about sequence of examples: create - backup - restore. Employee is just an alias, it could have a problem on last step. But if you wish, I'll change the name.

asfernandes: @AlexPeshkoff what happens if file name used in isc_spb_dbname is different than one used in isc_spb_expected_db? Could someone authenticate with one plugin and backup a database that uses another authentication plugin?

AlexPeshkoff: @aafemt I think it will be correct to use employee in backup and what else you wish at restore. employee_copy.fdb or anything else you wish. It's good idea when samples run out of the box.

AlexPeshkoff: @asfernandes If databases in isc_spb_dbname and isc_spb_expected_db have same security database service will definitely work. If not - it depends upon mapping rules in a database isc_spb_dbname. 

What about ability to authenticate with one plugin and backup a database that uses another authentication plugin - yes, it's possible if configuration and/or mapping enable attach to that database with such credentials. For example if both databases use same security database that's definitely possible.

aafemt:
As you wish.

examples/interfaces/10.backup.cpp
aafemt: Doesn't backup service use embedded access and thus bypass database security completely?..

PR #34, opened 2016-06-08
Example of shutting database down

aafemt:
And fix of a couple of mistypes.

### DIFF
--- a/examples/interfaces/09.service.cpp
+++ b/examples/interfaces/09.service.cpp
@@ -1,11 +1,11 @@
 /*
  *	PROGRAM:	Object oriented API samples.
  *	MODULE:		09.service.cpp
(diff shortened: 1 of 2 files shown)


AlexPeshkoff:
I'm not sure - do we need so much almost same samples that differ only in a few bytes in SPB?
May be better collect them in one?

aafemt:
May be. I use these examples to test single pieces of engine, so they must be narrow for 
this purpose. You can combine them into super-example if you wish.

AlexPeshkoff:
I've fixed comments in separate commit.
BTW, what prevents from using fbsvcmgr for engine tests? Re.services old/new API gate performs minimum (more precise - no) actions.

aafemt:
To use fbsvcmgr for tests I would have to make it understand unicode first.

PR #36, opened 2016-06-12
Ignore Visual Studio generated files specific to user workspace

mrotteveel:
I have made the pull request against B3_0_Release. I think it should also be applied to master. Let me know if you need another pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ config.status
 configure
 libtool
 *.log
+builds/win32/*/*.sdf
+builds/win32/*/*.VC.db
+builds/win32/*/*.VC.opendb


.gitignore
```diff
@@ -11,3 +11,6 @@ config.status
 configure
 libtool
 *.log
+builds/win32/*/*.sdf
+builds/win32/*/*.VC.db
+builds/win32/*/*.VC.opendb
```
hvlad: We already have .gitignore at builds/win32. I think such changes shouild be there.

mrotteveel: I hadn't noticed that one. I will create a new pull request this weekend.

hvlad: Mark, i can do necessary changes, if you wish

mrotteveel: If you can do that, that would be great. Thanks!

hvlad: Done

PR #39, opened 2016-07-12
Add platform support for m68k

glaubitz:
Hello!

This pull request adds generic Linux/m68k platform support to Firebird 2.5. The first patch by me just adds the necessary platform definitions for Linux/m68k while the second patch by Michael Karcher fixes the alignment of 'sem_t' on m68k which is different to the alignment of 'long long' which resulted in the futex system call failing with "The futex facility returned an unexpected error code.Aborted".

See Debian bug report 828141 [1] for more details.

These two patches currently address Firebird 2.5 only since this is the currently used version of Firebird in Debian unstable. We have 3.0 in Debian experimental, but I haven't found the time yet to test Firebird 3.0 on Linux/m68k. However, I am planning to do this within the next days so that I can provide patches against the 3.0 branch and hopefully also master.

Thanks,
Adrian

> [1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=828141

### DIFF
--- a/src/jrd/inf_pub.h
+++ b/src/jrd/inf_pub.h
@@ -217,6 +217,7 @@ enum  info_db_implementations
 	isc_info_db_impl_linux_alpha = 83,
 	isc_info_db_impl_linux_arm64 = 84,
 	isc_info_db_impl_linux_ppc64el = 85,
+	isc_info_db_impl_linux_m68k = 86,
 
 	isc_info_db_impl_last_value   // Leave this LAST!
 };
--- a/configure.in
+++ b/configure.in
@@ -354,6 +354,17 @@ dnl CPU_TYPE=ppc64
    SHRLIB_EXT=so
    ;;
 
+  m68k*-*-linux*)
+   MAKEFILE_PREFIX=linux_generic
+   INSTALL_PREFIX=linux
+   PLATFORM=LINUX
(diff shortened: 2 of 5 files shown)


AlexPeshkoff:
The only problem remaining to merge your pull request is same job for FB3

glaubitz:
Of course, I just haven't got around yet working on FB3.0 and the master branch. I just opened this PR already because this code has already been verified to work :).

AlexPeshkoff:
We usually do not accept branch-only ports if there is no port for the 
master

glaubitz:
I understand that and never asked for that. It's just that I had tested 2.5 first because that's what Debian unstable currently has. As I said, 3.0 and master will follow suit.

src/jrd/inf_pub.h
```diff
@@ -217,6 +217,7 @@ enum  info_db_implementations
 	isc_info_db_impl_linux_alpha = 83,
 	isc_info_db_impl_linux_arm64 = 84,
 	isc_info_db_impl_linux_ppc64el = 85,
+	isc_info_db_impl_linux_m68k = 86,
```
glaubitz: Hi Mariuz!

This was for Firebird2.5 which did not have any support for ppc64 back then, but it was added to FB3.0.

I can backport the ppc64 changes to 2.5 if you want to and update this PR accordingly.

Adrian

PR #41, opened 2016-07-13
Add platform support for m68k

glaubitz:
Referencing #40, here's the same PR for the master branch.

Thanks,
Adrian

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -361,6 +361,17 @@ dnl CPU_TYPE=ppc64
    SHRLIB_EXT=so
    ;;
 
+  m68k*-*-linux*)
+   MAKEFILE_PREFIX=linux_generic
+   INSTALL_PREFIX=linux
+   PLATFORM=LINUX
+   AC_DEFINE(LINUX, 1, [Define this if OS is Linux])
+   AC_DEFINE(M68K, 1, [Define this if CPU is M68k])
+   LOCK_MANAGER_FLG=Y
+   EDITLINE_FLG=Y
+   SHRLIB_EXT=so
+   ;;
+
   *-*-linux* | *-*-gnu*)
     MAKEFILE_PREFIX=linux_generic
     INSTALL_PREFIX=linux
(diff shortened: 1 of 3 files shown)


AlexPeshkoff:
Sorry, but you did not build firebird with your patch (or at least did not check how is resulting database marked)

glaubitz:
Hello Alex!

I'm not sure whether I can follow. What do you mean with "how the database is marked"? I'm willing to improve my code, so there is no need to close the PR. You don't need to close a PR and reopen a new PR for every change, one can just force-push changes, that's how systemd upstream does it.

Anyway, if you can provide a little more detailed explanation, I'm happy to make the necessary changes.

Thanks,
Adrian

AlexPeshkoff:
On 07/13/2016 02:20 PM, John Paul Adrian Glaubitz wrote:
> Hello Alex!
>
> I'm not sure whether I can follow. What do you mean with "how the database is marked"? I'm willing to improve my code, so there is no need to close the PR. You don't need to close a PR and reopen a new PR for every change, one can just force-push changes, that's how systemd upstream does it.
>
> Anyway, if you can provide a little more detailed explanation, I'm happy to make the necessary changes.

./gstat -h <database-name>
should return correct implementation line. In amd64 I get:

         Implementation          HW=AMD/Intel/x64 little-endian OS=Linux 
CC=gcc

In your patch you define CPU to something like CpuM68k, but there is no 
such symbol at all (and no textual description for it). Therefore I 
suppose that your #ifdef does not work at all (and you get some other HW 
identifier active), or firebird will not compile at all - you will get 
syntax error with CpuM68k.

Ports make sense when they are at least minimally tested on target HW.

glaubitz:
Hi Alex!

Thanks for the additional hints and very sorry for the late reply, I was very busy with work!

You're absolutely right, my patch is incomplete. I'm working on an updated version now and will update the PR once I'm ready. We can also drop the patch for Firebird2.5 as Debian just switched to Firebird3.0.

Thanks,
Adrian

PR #47, opened 2016-09-02
Memory elimination

Logash:


asfernandes:
Why is the code so inconsistently formated?

It's full of things without spaces "if(" instead of "if (", or extra spaces "MoveFile( ... )" instead of "MoveFile(...)".

It's not even a different formating, it's even inconsistent with itself.

AlexPeshkoff:
Why do you explicitly turn off wipe for boot build?

src/common/classes/TempFile.cpp
```diff
@@ -229,7 +230,7 @@ void TempFile::init(const PathName& directory, const PathName& prefix)
 
 	if (doUnlink)
 	{
-		::unlink(filename.c_str());
+		::do_unlink(filename.c_str());
```
aafemt: Temporary files on Windows are created with "delete on close" flag and never deleted explicitly. So, erasing of files don't, actually, works there.

AlexPeshkoff: On 09/02/2016 05:38 PM, Dimitry Sibiryakov wrote:
>> @@ -229,7 +230,7 @@ void TempFile::init(const PathName& directory, const PathName& prefix)
>>   
>>   	if (doUnlink)
>>   	{
>> -		::unlink(filename.c_str());
>> +		::do_unlink(filename.c_str());
> Temporary files on Windows are created with "delete on close" flag and never deleted explicitly. So, erasing of files don't, actually, works there.
>

On posix things are looking more or less same - file is unlinked right 
after open, after it file can't be accessed from OS but descriptor 
remains opened in a process, can be used (and is used) before close() 
call. Looks like for temporary files wipe should be also done on close 
for all platforms.

aafemt: Also, is there point to erase randomly generated file name from directory?

builds/win32/preprocess.bat
```diff
@@ -18,10 +18,10 @@
...
 @if not defined FB_BIN_DIR (@call set_build_target.bat %*)
 
-@if "%1"=="BOOT" (set BOOTBUILD=1) else (set BOOTBUILD=0)
+@if "%1"=="BOOT" (set FIREBIRD_BOOT_BUILD=1) else (set FIREBIRD_BOOT_BUILD=0)
 @echo.
-@if "%BOOTBUILD%"=="1" (call :BOOT_PROCESS) else (call :MASTER_PROCESS)
-@set BOOTBUILD=
+@if "%FIREBIRD_BOOT_BUILD%"=="1" (call :BOOT_PROCESS) else (call :MASTER_PROCESS)
+@set FIREBIRD_BOOT_BUILD=
 @set GPRE=
```
aafemt: What is the point of these changes? Just for fun?

src/common/classes/WipeFile.cpp
```diff
@@ -0,0 +1,659 @@
...
+		SetFilePointer(hFile, dwSizeLow, (PLONG) &dwSizeHigh, FILE_BEGIN);
+
+		// Write one zero byte, which causes the file system to fill the entire
+		// file's on-disk contents with 0.
+		if (!SecureOverwrite(hFile, 1)) {
+			fatal_exception::raiseFmt("IO error (%d) overwriting file: %s",
+				GetLastError(),	FileName);
+			CloseHandle(hFile);
+			return;
+		}
```
aafemt: Fix me: the code after throwing an exception is really supposed to work?..
And why would Windows overwrite whole file because of one byte?..

builds/win32/preprocess.bat
Logash: These changes were necessary to ignore the clear memory at the time of assembly. But these changes can be excluded due to the changes of the task CORE-5348.

src/common/classes/WipeFile.cpp
Logash: Does not work, we must close the handler and then throw an exception. I will fixed it.

aafemt:
> Does not work


Yes. Because it is never called.

aafemt:
Why to ignore it?

Logash:
Because the value is read from the non-existent config.

aafemt:
That's what default values for in config.cpp.

Logash:
Wipe of temporary files was done by rewind + write using a file descriptor before closing it. Because temporary files are created in Windows with the flag delete_on_close, and on POSIX unlink()  called before closing the handle.
Compressed temp files cannot be wiped. Rewind + write can not be used since after rewriting the compressed file its size will be changed and stored to other parts of the hard disk. Wipe of file through direct access to the disk will not work. Direct access to the disk has been limited starting with Windows Vista and Windows Server 2008. As a result, if the wipe of files is enabled the use of compressed temporary files are not allowed.

aafemt:
> Compressed temp files cannot be wiped. Rewind + write can not be used since after
> rewriting the compressed file its size will be changed and stored to other parts of the
> hard disk.


May I ask for a proof from MSDN?..

romansimakov:
At least here you can find description of the original problem. Encrypted and compressed files may consist of parts located in other places afer rewriting.
https://technet.microsoft.com/en-gb/sysinternals/bb897443.aspx
Starting with Vista+ MS prohibited such access. But we cannot find direct eplanation of it. The link has info about ability to run on Vista+ but the talk is exactly about Windows NT/2000's. We check and it seems that SDelete leave uncompressed files untouched. Probably it rewrites file parts but then override free'd space.
I suggest to raise an exception for now in this case and make good comment. It could be improved later.

AlexPeshkoff:
Does it make sense to care about such ancient operating systems? (I'm sure that our posix builds will not run on OS from previous century.)

romansimakov:
I'd care about it if someone will really need it.

livius2:
Why is it looping to wipe byte by byte? This should be done by preallocating three "static" pages: the first filled with 0x00, the second with 0xFF, and the third with random values. Then, write or copy them as a whole.

PR #51, opened 2016-10-13
Add platform support for Linux/m68k

glaubitz:
Hello!

Here's my V2 patch series to add platform support for Linux/m68k. I will open another PR shortly for the 3.0 branch.

Thanks,
Adrian

### DIFF
--- a/builds/posix/vers.sh.in
+++ b/builds/posix/vers.sh.in
@@ -57,6 +57,7 @@ platform_linux() {
 		echo "${TAB}$i;"
 	done
 
+	echo "${TAB}_IO_stdin_used;"
 	echo 'local:'
 	echo "${TAB}*;"
 	echo '};'
--- a/configure.ac
+++ b/configure.ac
@@ -363,6 +363,17 @@ dnl CPU_TYPE=ppc64
    SHRLIB_EXT=so
    ;;
 
+  m68k*-*-linux*)
+   MAKEFILE_PREFIX=linux_generic
+   INSTALL_PREFIX=linux
+   PLATFORM=LINUX
+   AC_DEFINE(LINUX, 1, [Define this if OS is Linux])
+   AC_DEFINE(M68K, 1, [Define this if CPU is M68k])
+   LOCK_MANAGER_FLG=Y
+   EDITLINE_FLG=Y
+   SHRLIB_EXT=so
+   ;;
+
   *-*-linux* | *-*-gnu*)
     MAKEFILE_PREFIX=linux_generic
     INSTALL_PREFIX=linux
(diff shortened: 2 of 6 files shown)


glaubitz:
I just added another patch on top to fix the generator script for the linker version script.

The version scripts must include _IO_stdin_used when using the glibc as otherwise the application can crash or caused unexpected behavior on some architectures, see:

> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=840666
> http://lists.gnu.org/archive/html/bug-glibc/2001-12/msg00203.html

AlexPeshkoff:
Looks like fix with _IO_stdin_used will be useful for any platform?

glaubitz:
Hi Alex!

Yes, this fix should be included regardless the architecture. Without the additional symbol, Firebird will crash or misbehave on PowerPC, MIPS, Alpha among others and the toolchain therefore always includes the symbol by default when not using a linker script, e.g. amd64:

glaubitz@ikarus:~$ dpkg --print-architecture
amd64
glaubitz@ikarus:~$ objdump -T /bin/bash |grep _IO_stdin_used
00000000004c5300 g    DO .rodata        0000000000000004  Base        _IO_stdin_used
glaubitz@ikarus:~$

Damyan has also included this fix in the Debian package [1] and firebird3.0 now builds fine everywhere which it didn't before.

> [1] https://anonscm.debian.org/cgit/pkg-firebird/3.0.git/commit/?id=57d5f607e9c083438d108e25daf5e8a3d3c225de

themaddoctor:
With gcc 7.1.0 on x86_64, and firebird 2.5.8 sources, I'm also getting this:

make[2]: *** [../gen/Makefile.refDatabases:68: empty.fdb] Segmentation fault

glaubitz:
@themaddoctor This is a different issue. Version 2.5.8 doesn't use any version scripts, so it cannot be affected by the missing _IO_stdin_used on Linux.

themaddoctor:
Then what is this issue, which has the same symptoms, and seg faults on the same executable?

glaubitz:
A segfault is a very generic error message. Just because you see two cases of a segfault, does not mean you are seeing the same bug. As I said before, Firebird 2.5.x does not use version scripts, so this cannot be the same bug.

If you want to know what your problem is, please build Firebird with debug symbols enabled and use gdb to generate a usable backtrace.

themaddoctor:
```
gdb create_db
(gdb) run test.fdb
Starting program: /building/libreoffice/create_db test.fdb
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7ffff458d700 (LWP 21842)]
[New Thread 0x7ffff3b7a700 (LWP 21843)]
[New Thread 0x7ffff3379700 (LWP 21844)]

Thread 1 "create_db" received signal SIGSEGV, Segmentation fault.
0x00000000006d2eb0 in RSE_open (tdbb=tdbb@entry=0x7fffffffdc90, rsb=0x7ffff2b459c8)
    at ../src/jrd/rse.cpp:380
380       if (rsb->rsb_invariants)
```

glaubitz:
Can you please open a new issue for that? This isn't related.

themaddoctor:
Do you mean a new "pull request"? Because there is no "issues" tab in this project, as far as I can see.

hvlad:
Bug tracker is here
http://tracker.firebirdsql.org/

PR #52, opened 2016-10-13
Add platform support for Linux/m68k for the 3.0 branch

glaubitz:
Hello!

This changeset adds Linux/m68k support to the 3.0 branch of Firebird.

Thanks,
Adrian

### DIFF
--- a/builds/posix/vers.sh.in
+++ b/builds/posix/vers.sh.in
@@ -57,6 +57,7 @@ platform_linux() {
 		echo "${TAB}$i;"
 	done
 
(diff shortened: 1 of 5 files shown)


themaddoctor:
Is there a similar fix for 2.5.8? It does not have the file that you patch, nor does it have a file containing the line that you change (in that directory).

glaubitz:
@themaddoctor No, there is no similar fix for 2.5.8 because 2.5.8 isn't affected by this particular bug.

PR #54, opened 2016-10-15
Replace SVN with Git tree

mariuz:


doruilasi (changes requested):
VCS (standing for version control system) instead of SVN/GIT? (please cat README.license.usage.txt | grep SVN to catch all lines needing this change)

mariuz:
@doruilasi thanks i will do another pull request

PR #58, opened 2016-10-24
fix gcc build with -std=c++11 on i386

mkubecek:
After commit 2f3dc8bca8f4 ("Added -std=c++11 flag to Linux builds."),
i386 (32-bit) linux builds started to fail with "#error Define FB_CPU
for your platform". This is because gcc doesn't define "i386" macro with
-std=c++11, only "**i386**". While current gcc seems to define both
macros without -std=c++11, let's play it safe and check for both.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -126,7 +126,7 @@
 #define FB_CPU CpuPowerPc
 #endif
 
-#ifdef i386
+#if defined(i386) || defined(__i386__)
 #define I386
 #define FB_CPU CpuIntel
 #endif /* i386 */
@@ -230,7 +230,7 @@
 #define powerpc
 #define FB_CPU CpuPowerPc
 #endif
-#ifdef i386
+#if defined(i386) || defined(__i386__)
 #define I386
 #define FB_CPU CpuIntel
 #endif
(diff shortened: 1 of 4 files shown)


asfernandes:
Wouldn't it much simple to do instead:

```
#if !defined(i386) && defined(__i386__)
#define i386 __i386__
#endif
```

?

AlexPeshkoff:
For some other HW we have own define (like -DARM, -DAMD64). May be it will be better going this way for x86 too - instead of guessing what else define will be changed by compiler?

mkubecek:
We already define "I386" in src/common/common.h anyway so doing this outside of the source (like we do for AMD64) and checking it everywhere would make sense.

aafemt:
Using external macros will tie us to the death with autoconf for recognizing platform/HW. Autoconf was good 15 years ago, but now GCC has wider set of macros for this. According to standard, these macros must have two underscores in the begin and end. Macros without them are deprecated. If we are changing sources to use new standard, we must use new macros.

mariuz:
I would go with a more readable solution 

if defined(__i386__)
#define ARCH_X86_32
#endif


if defined(__x86_64__)
#define ARCH_X86_64
#endif


But we all know that I386 is x86_32 and AMD64 is x86_64 so i guess this can be merged 

http://stackoverflow.com/a/152034/66242

mkubecek:
Obsoleted by commit 4107794ac29c

PR #59, opened 2016-10-24
create android x86/arm64/x86_64 scripts and ports

mariuz:
all builds ok like in the arm 32 bits port 

file libfbclient.so.3.0.2 
libfbclient.so.3.0.2: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, not stripped

### DIFF
--- a/builds/posix/make.android.arm64
+++ b/builds/posix/make.android.arm64
@@ -1,0 +1,74 @@
+
+ifeq ($(NDK),)
+$(error Must export NDK location before building for Android)
+endif
+NDK_ROOT:=$(NDK)
+#NDK_LOG:=1
+include $(NDK)/build/core/init.mk
+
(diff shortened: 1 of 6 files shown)


mariuz:
Added Android x86_64 cross and makefile

file libfbclient.so.3.0.2 
libfbclient.so.3.0.2: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped

mariuz:
Added also Android x86 32 bit makefile and cross 

here is teh result 
file gen/Release/firebird/lib/libfbclient.so.3.0.2 
gen/Release/firebird/lib/libfbclient.so.3.0.2: ELF 32-bit LSB shared object, Intel 80386, version 1 (SYSV), dynamically linked, not stripped

PR #62, opened 2016-11-01
Fix android include paths related to c++11 

mariuz:
This one fixed ::posix_fadvise' has not been declared for arm32

(Had to upgrade android platform to 24)

Also include paths needed to be updated to include the c++11 headers


mariuz:
All builds ok with the above patches 

file gen/Release/firebird/lib/libfbclient.so.4.0.0 
gen/Release/firebird/lib/libfbclient.so.4.0.0: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, not stripped

PR #67, opened 2016-11-14
Fixed: gbak can overwrite database if it is excecuted with arguments where path to backup corresponds to path to this database

ilya071294:
Example:
`gbak -user SYSDBA -password * -b dbalias db.fdb`
where dbalias refers to db.fdb

Solution:
If the output file exists, then its header is checked. The existing file isn't overwritten if it is similar to database.

### DIFF
--- a/builds/posix/make.shared.variables
+++ b/builds/posix/make.shared.variables
@@ -81,9 +81,10 @@ AllObjects += $(GFIX_Objects)
 
 # gbak
 Svc_GBAK_Objects:= $(call dirObjects,burp)
-GBAK_Objects:= $(Svc_GBAK_Objects) $(call dirObjects,burp/main)
+GBAK_Own_Objects:= $(Svc_GBAK_Objects) $(call dirObjects,burp/main)
+GBAK_Objects:= $(GBAK_Own_Objects) $(call makeObjects,jrd,ods.cpp)
 
-AllObjects += $(GBAK_Objects)
+AllObjects += $(GBAK_Own_Objects)
 
 # gsec
 Svc_GSEC_Objects:= $(call dirObjects,utilities/gsec)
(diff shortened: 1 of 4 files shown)


hvlad:
Hmm... gbak v3 will not overwrite database in ODS 12, but will overwrite database in other ODS
Looks like very strange decision...
If i wanted to prevent gbak from overwrite existing backup file, i'd add special option (like we already have for -replace switch)

romansimakov:
It's not about overwrite existing backup. It's about overwrite existing database :) @ilya071294 already made a fix in parsing command line arguments which lead to an error after which gbak thought that db is a backup and clean it. But currently it's still possible to specify DB instead of backup. Yes. It's human factor but this change is an attempt to understand this case and prevent it.
IMO it's not so bad to prohibit any overwriting but it can break some existing solutions. But I cannot imagin situation where gbak overwrites a file with the first 4K very likes to DB header page :)

hvlad:
> It's not about overwrite existing backup.
It was a typo, two lines above i was more correct.

> But I cannot imagin situation where gbak overwrites a file with the first 4K very likes to DB header page :)
But proposed solution checks for **supported ODS only**. And i consider it is wrong, sorry.
Agree - "check for supported ODS" and "first 4K very likes to DB header page" - far not the same

AlexPeshkoff:
On 11/14/16 18:53, Roman Simakov wrote:
>> Agree - "check for supported ODS" and "first 4K very likes to DB header page" - far not the same
> Exactly. I made a point to explain that special option is useless. A way to check the file looks like DB should be improved.
>

But it's close to unreal to guess how will header change in the future...

romansimakov:
New backup will know about the changes and change recognision) IMO it's enough to recognize that a file is not backup.

AlexPeshkoff:
You suggest to overwrite backups? Take a look at http://tracker.firebirdsql.org/browse/CORE-2037. I suggest very old solution for such cases - rename existing file to something like .bak.

aafemt:
> IMO it's not so bad to prohibit any overwriting but it can break some existing solutions.

   I wouldn't care about these solutions much. Remember, that not only databases can be 
overwritten by mistake, but almost any file in system.

romansimakov:
I totally agree. Even if such solutions are it's easely to write wrapper and move or delete existing files. Or add an option to call of gbak like -rep. It's easy and IMO makes sense.

PR #69, opened 2016-12-08
gbak continues to write to stdout if write function returns -1

ilya071294:


src/burp/mvol.cpp
```diff
@@ -610,12 +610,11 @@ UCHAR MVOL_write(const UCHAR c, int* io_cnt, UCHAR** io_ptr)
 			}
 			else
 			{
-				if (!cnt ||
 #ifndef WIN_NT
-					errno == ENOSPC || errno == EIO || errno == ENXIO ||
+				if (!cnt || errno == ENOSPC || errno == EIO || errno == ENXIO ||
 					errno == EFBIG)
```
aafemt: This condition is basically wrong for POSIX, but there is no point to fix it as multivolume backup is expected to be cleaned out in version 4.

AlexPeshkoff:
Ilya, if backport to 3.0 and/or 2.5 is needed please prepare appropriate PRs

PR #71, opened 2016-12-14
Adjusted tokens array to be consistent to non_reserved_word of grammar

romansimakov:


src/yvalve/keywords.cpp
```diff
@@ -466,16 +466,16 @@ static const TOK tokens[] =
 	{TOK_VARIABLE, "VARIABLE", false},
 	{TOK_VARYING, "VARYING", false},
 	{TOK_VIEW, "VIEW", false},
-	{TOK_WAIT, "WAIT", false},
-	{TOK_WEEK, "WEEK", false},
+	{TOK_WAIT, "WAIT", true},
+	{TOK_WEEK, "WEEK", true},
```
dyemanov: WEEK is actually reserved. There may be other mistakes, please double-check.

PR #72, opened 2016-12-24
clang/clang++(LLVM) build on FreeBSD11 and macOS

nakagami:
FreeBSD11
--------------

Install tools and libraries.

    sudo pkg install libtool autoconf automake icu libtommath

Build

    export LD_LIBRARY_PATH=/usr/local/lib
    export LIBRARY_PATH=/usr/local/lib
    export CC=clang
    export CXX=clang++
    export CPPFLAGS="-I/usr/local/include"
    export LIBS="-L/usr/local/lib -ltommath"

    ./autogen.sh --disable-raw-devices
    gmake

macOS Sierra (10.12) + XCode 8.2
-----------------------------------

Install tools and libraries. (I use MacPorts.)

    sudo port install gmake libtool autoconf automake icu libtommath
    sudo ln -s /opt/local/lib/ibicui18n.dylib /usr/local/lib/libicui18n.dylib
    sudo ln -s /opt/local/lib/libicuuc.dylib /usr/local/lib/libicuuc.dylib

Build

    export LIBRARY_PATH=/opt/local/lib
    export LIBTOOLIZE=glibtoolize
    export CC=clang
    export CXX=clang++
    export CPPFLAGS="-I/opt/local/include -I/opt/local/include/libtommath"
    export LIBS="-L/opt/local/lib"

    ./autogen.sh --disable-raw-devices
    gmake


builds/posix/make.rules
```diff
@@ -96,15 +96,18 @@ $(OBJ)/%.cpp: $(SRC_ROOT)/%.epp
 
 $(OBJ)/%.o: $(SRC_ROOT)/%.c
 	$(CC) $(WCFLAGS) -c $(firstword $<) -o $@
```
AlexPeshkoff: Why do you replace single sed call using inline edit with 2 calls?

src/common/StatementMetadata.cpp
```diff
@@ -167,7 +167,7 @@ const char* StatementMetadata::getPlan(bool detailed)
 
 	if (plan->isEmpty())
 	{
-		UCHAR info[] = {detailed ? isc_info_sql_explain_plan : isc_info_sql_get_plan};
```
AlexPeshkoff: It's very bad idea to poison sources here and there with explicit types casts.
That should be solved using correct types, not casts

builds/posix/make.rules
nakagami: BSD sed does not have -i option, and seems can't place multiple -e options.

src/common/StatementMetadata.cpp
nakagami: Oh, sorry, I don't know c++ so much.
explicit types casts means modify

from

    (UCHAR)ics_info_sql_explain_plan

to

    UCHAR(isc_info_sql_explain_plan)

like this ?

builds/posix/make.rules
AlexPeshkoff: In that case this should better be done in OS-specific manner 
BTW, looking at:
https://www.freebsd.org/cgi/man.cgi?sed%281%29 
I see that -i is present

src/common/StatementMetadata.cpp
AlexPeshkoff: Please no.
I understand that you solve some problem (not sure error or warning) with integer to unsigned char conversion.
IMO it should have better solution than placing explicit casts (does not matter C or C++ style) here and there.
If you can't instruct compiler to accept existing style please wait a bit - your solution works but it is not good from sources readability POV

AlexPeshkoff: Try (for this particular item) the following:


diff --git a/src/jrd/inf_pub.h b/src/jrd/inf_pub.h
index 9c3b0c2..c223ec4 100644
--- a/src/jrd/inf_pub.h
+++ b/src/jrd/inf_pub.h
@@ -417,7 +417,7 @@ enum info_db_provider
 #define isc_info_sql_records             23
 #define isc_info_sql_batch_fetch         24
 #define isc_info_sql_relation_alias            25
-#define isc_info_sql_explain_plan              26
+#define isc_info_sql_explain_plan              26u
 #define isc_info_sql_stmt_flags         27
 
 /*********************************/

builds/posix/make.rules
nakagami: Oh, I should use '-i ""'  option and It is in $(INLINE_EDIT_SED) variable.

src/common/StatementMetadata.cpp
nakagami: /Users/nakagami/firebird/src/common/StatementMetadata.cpp:170:19: error:
          non-constant-expression cannot be narrowed from type 'unsigned int' to
          'UCHAR' (aka 'unsigned char') in initializer list [-Wc++11-narrowing]
      ...info[] = {detailed ? isc_info_sql_explain_plan : isc_info_sql_get_plan};
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

AlexPeshkoff: Did you try -Wno-narrowing ?

nakagami: I try it, But same error

AlexPeshkoff: Sorry if I missed it - but what compiler/version are you using?

nakagami: Both are so calld LLV clang/clang++. They are default c/c++ compiler now.

FreeBSD11
 
    [nakagami@freebsd11 ~]$ clang++ --version
    FreeBSD clang version 3.8.0 (tags/RELEASE_380/final 262564) (based on LLVM 3.8.0)
    Target: x86_64-unknown-freebsd11.0
    Thread model: posix
    InstalledDir: /usr/bin

macOS (XCode 8.2.1)

    hajime:~ nakagami$ clang++ --version
    Apple LLVM version 8.0.0 (clang-800.0.42.1)
    Target: x86_64-apple-darwin16.3.0
    Thread model: posix
    InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

AlexPeshkoff:
Most of suggested changes are not needed - narrowing protection is fixed in clang and our code cause no errors. Fixed related with freebsd were placed in separate commit 77437e443bb2e402f91aeb0a337142ff35249c4e.

PR #73, opened 2017-01-05
Implementation of CORE-5064

aafemt:
Whatever I tried, I failed to avoid increased number of conflicts. Can you suggest a solution for that?


dyemanov:
You shouldn't include new tokens into non_reserved_word. Fundamental data types should be reserved. Now we have a conflict between BINARY as type and BINARY as user-defined domain. SQL specification also defines BINARY and VARBINARY as reserved words.

aafemt:
Oh, that was the trick. Thanks.

src/yvalve/why.cpp
```diff
@@ -330,7 +330,7 @@ int SQLDAMetadata::getSubType(CheckStatusWrapper* status, unsigned index)
 		fb_assert(sqlda->sqld > (int) index);
 		ISC_SHORT sqltype = sqlda->sqlvar[index].sqltype & ~1;
 		if (sqltype == SQL_VARYING || sqltype == SQL_TEXT)
-			return 0;
+			return sqlda->sqlvar[index].sqlsubtype == 1 ? 1 : 0;
 		return sqlda->sqlvar[index].sqlsubtype;
```
dyemanov: Some named constant should be introduced for the subtype.

dyemanov: Also, an entry in types.h / RDB$TYPES would be appreciated for subtype 1.

aafemt: Unfortunatelly, RDB$TYPES already contains entries for BLOB subtypes. New entry for text will clash with them, because the same field is used.

dyemanov: Damn, you're right.

aafemt: I wonder if names "fb_text_subtype_text" and "fb_text_subtype_binary" are ok for these constants. Do you have a better name on mind?

doc/sql.extensions/README.data_types
```diff
@@ -74,3 +74,30 @@ BOOLEAN (FB 3.0)
...
+  Example(s):
+    1. DECLARE VARIABLE VAR1 VARBINARY(10);
+    2. CREATE TABLE TABLE1 (FIELD1 BINARY(16),
+                            FIELD2 VARBINARY(100),
+                            FIELD3 BINARY VARYING(1000));
+
+  Note(s):
+    1. If length is omitted for type BINARY, it is considered to be 1.
+    2. Can be distinguished from text types by value 1 in RDB$FIELD_SUB_TYPE.
+    3. Character set is set to OCTETS for backward compatibility.
```
asfernandes: What means "backward compatibility" here? If it's the fact that a thing called BINARY works like and exposes this charset number via API, I don't think this is "backward compatibility".

aafemt: Ok, I can change wording if you suggest a better variant why field described as "char subtype 1" has also set charset to OCTETS.

src/isql/isql.epp
```diff
@@ -1822,63 +1822,68 @@ bool ISQL_printNumericType(const char* fieldName, const int fieldType, const int
...
-			case DOUBLE_PRECISION:
-				isqlGlob.printf("NUMERIC(15, %d)", -fieldScale);
 				break;
-			default:
-				isqlGlob.printf("%s", Column_types[i].type_name);
 			}
 		}
-		else
+		return true;
+	case T_CHAR:
```
asfernandes: Handling CHAR/VARCHAR in a function named ISQL_printNumericType??

doc/sql.extensions/README.data_types
```diff
@@ -74,3 +74,30 @@ BOOLEAN (FB 3.0)
...
+    4. In API are similar to corresponding text types, getSubType() returns 0.
```
asfernandes: That is not what I see in why.cpp. It's quite the contrary, you changed the old behavior:

```
return sqlda->sqlvar[index].sqlsubtype == CS_BINARY ? fb_text_subtype_binary : fb_text_subtype_text;
```

aafemt: You just looking at wrong class. SQLDAMetadata is an envelope for ISC API, and this particular function is not used anywhere for now, so whatever is written here, it has no effect in current engine.

src/isql/isql.epp
aafemt: I keep patch as small as possible. Refactoring or cleanup unrelated things will cause rejection of the patch.

doc/sql.extensions/README.data_types
asfernandes: Why a function not used is maintained in the code? Why it's maintained with a behavior different than the one being used?

aafemt: Because interface in a contract. Every function in interface must be implemented and work as documented. Even if it is never called in the rest of code. getSubType() is supposed to return subtype of field. In this class it is possible (with a little assumption), in other classes it is impossible for now and fixing that is out of scope for this patch.

PR #77, opened 2017-01-30
Do not enforce -std=c+11 on Linux.

mkubecek:
Commit 52d9a05a0f3d ("Backport from master: Optimized hash function for
lock manager and hash join") adds "-std=c++11" to CXXFLAGS on Linux
unconditionally. This doesn't seem to be necessary (looks rather like an
omission) and breaks the build on distributions with old gcc versions
(e.g. SLES 11 SP4).

### DIFF
--- a/builds/posix/prefix.linux
+++ b/builds/posix/prefix.linux
@@ -28,5 +28,3 @@ DEV_FLAGS=-p $(COMMON_FLAGS) $(WARN_FLAGS)
 
 # This file must be compiled with SSE4.2 support
 %/CRC32C.o: COMMON_FLAGS += -msse4
-
-CXXFLAGS := $(CXXFLAGS) -std=c++11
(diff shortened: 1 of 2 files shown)


AlexPeshkoff:
I'm not sure this commit solves all problems. Before initial backport we did not require SSE4 support for running FB3. I'm afraid people can meet problems upgrading to 3.0.2. May be conditional compilation making it possible to build 'official' binaries not using SSE should be added keeping ability to build optimized binaries for users who have SSE4 support?

aafemt:
SSE4 support is not required for running. Using of CRC command is conditional.

mkubecek:
If I understand correctly, the helper using SSE 4.2 instructions should only be used if CPU support for it is detected on run time. But there may be some problem with that as someone reported an "illegal instruction" fault in boot_isql on firebird-devel list recently.

aafemt:
Before commiting into master branch I tested it on both SSE4.2 capable and incapable CPUs. 
If no mistakes was made during backport, it must not produce "illegal instruction" error. 
In that thread I asked a couple of question and got no answer. Besides, project's testbed 
server uses CPU of the same architecture as was mentioned in that report and no problem 
occured.

PR #78, opened 2017-01-30
Fix for CORE-5469: Handles allocated with isc_dsql_alloc_statement2() are not cleaned on disconnect

aafemt:


hvlad (commented):
Why massive move of code ? It is impossible to see what was changed in moved lines.

aafemt:
IscStatement cannot be in anonymous namespace anymore since is it used in declaration of YAttachment. Thus it has been moved into Why namespace.

hvlad:
It could be done in different commits to separate the logic from refactoring

aafemt:
It could, yes.
The only changes pieces are IscStatement::~IscStatement() and IscStatement::destroy(). The rest of IscStatement code was copy-pasted untouched.

asfernandes:
Observations for all PRs: please put the ticket description, not only the number.

AlexPeshkoff:
Dmitry please prepare same PR for 3.0 - it should be backported before closing ticket

aafemt:
> Dmitry please prepare same PR for 3.0


Can't be this commit just cherry-picked there? why.cpp wasn't changed much in ver.4, so in 
theory it could be possible.

PR #79, opened 2017-02-08
Hiding password in fbtracemgr arguments from ps

ilya071294:


AlexPeshkoff:
Ilya, please prepare same PR for FB3 too

PR #88, opened 2017-03-30
Minimal support for dtype_varying in GPRE for C++

aafemt:
Just in case if someone was going to work with VARBINARY fields from engine in a right way.

### DIFF
--- a/src/gpre/cmp.cpp
+++ b/src/gpre/cmp.cpp
@@ -788,7 +788,7 @@ static void cmp_field( gpre_req* request, const gpre_fld* field,
 		break;
 
 	case dtype_varying:
-		if (!(field->fld_flags & FLD_charset) && field->fld_ttype)
+		if (!(field->fld_flags & FLD_charset) && field->fld_ttype >= dsc_text_type_metadata)
 		{
 			request->add_byte(blr_varying);
 			request->add_word(field->fld_length);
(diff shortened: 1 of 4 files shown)


src/gpre/cmp.cpp
```diff
@@ -1731,17 +1731,20 @@ static gpre_port* make_port( gpre_req* request, ref* reference)
 			CPR_bugcheck("missing prototype field for value");
 		if (temp->ref_value && (temp->ref_flags & REF_array_elem))
 			field = field->fld_array;
-		if ((field->fld_length & 7) == 0)
+		FLD_LENGTH l = field->fld_length;
```
asfernandes: Please don't use variable named just `l`.

aafemt:
Ok. Renamed.

AlexPeshkoff:
I hardly believed my eyes looking at this PR and internals of legacy security DB - they (suppose it was that people who added isc4.gdb to interbase) declared PASSWD field as VARCHAR but always used to work with it as plain CHAR. And I did not notice this bug when used to rework something around for almost 15 years.

First of all - I'm almost sure (did not test) that with your PR legacy authentication will be broken. The problem is that in auth server code field is treated as CHAR - also no VAR at the beginning. Yep, we can fix authenticating code but it will become incompatible with old hashes stored in databases. And that compatibility is the only excuse for existence of kegacy plugin... On my mind we should not raise the hell with changing hash storage format in legacy plugin.

So next question - if we apply the rest of your PR and get support of VAR objects in GPRE will it generate compatible (with old hashes) code for legacy plugin?

aafemt:
> First of all - I'm almost sure (did not test) that with your PR legacy authentication will be broken.

I did test. It worked.

AlexPeshkoff:
Committed PR changes by hand

PR #90, opened 2017-05-10
Fix 32-bit gcc build

mkubecek:
Resolves ambiguity in dtype_long case of CVT_get_dec64() and CVT_get_dec128().


AlexPeshkoff:
Michal - I've committed a simpler solution for a problem therefore close this pull request

PR #91, opened 2017-05-18
Remove initialization of p outside of while loop

doychin:
Also remove the unnecessary second initialization of p after while loop


AlexPeshkoff:
I close it not because PR is bad a wrong but due to committed better fix in core-5549

PR #92, opened 2017-05-22
have constant link order

bmwiedemann:
to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Without this patch g++ would order functions in libEngine12.so and fbintl
depending on random order of files in the build system's filesystem.

partial fix for http://tracker.firebirdsql.org/browse/CORE-5548


AlexPeshkoff:
Not to say I'm much impressed with an idea of binary reproducible builds but must agree that this PR does not break anything and if it's useful for someone - let it be...

PR #93, opened 2017-05-23
allow to override current time

bmwiedemann:
to enable reproducible builds.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

partial fix for http://tracker.firebirdsql.org/browse/CORE-5548

### DIFF
--- a/src/common/classes/NoThrowTimeStamp.cpp
+++ b/src/common/classes/NoThrowTimeStamp.cpp
@@ -67,6 +67,12 @@ NoThrowTimeStamp NoThrowTimeStamp::getCurrentTimeStamp(const char** error) throw
 	seconds = time_buffer.time;
 	milliseconds = time_buffer.millitm;
 #endif
+	char *source_date_epoch;
+	time_t override_seconds;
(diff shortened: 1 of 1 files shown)


hvlad:
Looks like dirty hack, sorry. It opens way to manipulate current time.
It must use fb_utils::bootBuild() also, but i still not sure we should allow such things in database engine.

AlexPeshkoff:
I also dislike this PR. Not due to theoretical ability to manipulate current time (how can environment of running server be changed), but due to breaking old useful feature in the header format. I do not accept seriously arguments in  https://reproducible-builds.org/. There are a lot of reasons, but main one is that when you build from sources you should not try to emulate some foreign build environment, you build something that will be optimal for your environment (both hard & soft). The only possible exception are public builds with known checksums - but this has nothing to do with what is called "reproducible builds" here. I suggest to close this PR and tracker item.

bmwiedemann:
In the openSUSE distribution, we sort of do 'public builds'... e.g. we usually do not optimize for specific machines but generic i586 or x86_64 to allow everyone to run it on his hardware.

PR #95, opened 2017-07-07
Conversion of statements, procedures and functions parameters (CHAR and VARCHAR) to UTF8 in trace log

ilya071294:


AlexPeshkoff:
If you want to add a function to already released cloop interface you should as minimum:
- add version label before it,
- make it have Status interface as first parameter
- after calling it check for isc_interface_version_too_old error (in a case when old trace plugin was used - in this particular case) and handle it some way (for example output non-utf text)

PR #97, opened 2017-07-14
Fix make clean - "make install" now works again after "make clean"

samofatov:
"make clean" deleted files generated during "configure" stage.
This broke "make install" operation. This pull request corrects that.


aafemt:
One must be completely insane to do "make install" after "make clean".

samofatov:
This is a fix for regression. "make clean && make && make install" worked before and now it's not.

PR #98, opened 2017-07-18
CORE-5586

cincuranet:


src/msgs/history2.sql
```diff
@@ -1144,7 +1144,7 @@ message to appear.', 'Your program encountered an input or output error.', 'c_pg
 (1149, 'hvlad', '2006-02-02 19:15:11', 17, 44, 'Current memory = !c
 Delta memory = !d
 Max memory = !x
-Elapsed time= !e sec', NULL, NULL, 'c_pg');
+Elapsed time = !e sec', NULL, NULL, 'c_pg');
 (1150, 'hvlad', '2006-02-03 19:17:10', 17, 45, 'Cpu = !u sec
```
hvlad: No need to change history records :)

cincuranet: I saw it there, so I played safe card. Should I remove this change?

romansimakov: @hvlad Is there a reason to keep them?

hvlad: history2.sql contains old versions of changed messages. Formally, you should put pre-edited version of message as new record there, and never change existing records. But i really not sure it (history2.sql) is used somehow these days...

dyemanov: I'd vote for deleting the history file. We have VCS logs anyway.

asfernandes: I'd go further and delete more things not useful, as the fields for file and function name in messages2.sql. We have grep and others tools anyway.

hvlad: No objection fro my side, but this is offtop here ;)

cincuranet: So. Should I revert the `history2.sql` for this PR or could it stay as it's going to be deleted anyway?

dyemanov: Revert, plz.

cincuranet: Done.

PR #99, opened 2017-07-19
New interface Batch helping to efficiently implement JDBC prepared statement batches

AlexPeshkoff:


asfernandes:
A readme with a detailed description of the feature is needed. I don't remember it being discussed in the lists.

AlexPeshkoff:
Certainly will do - documenting new interfaces is anyway required step.

src/common/classes/vector.h
```diff
@@ -135,7 +157,11 @@ class Vector
 
 protected:
 	FB_SIZE_T count;
-	T data[Capacity];
+	union
+	{
+		T data[Capacity];
+		A align;
+	};
 };
```
hvlad: This doesn't compiles on MSVC12:
error C2621: 'Firebird::Vector<Object,16,char>::data' : illegal union member; type 'Jrd::PageNumber' has a copy constructor
...
\src\jrd\../jrd/Record.h(169) : see reference to class template instantiation 'Firebird::Stack<Jrd::PageNumber,16>' being compiled

hvlad: See also: https://msdn.microsoft.com/en-us/library/bd149yt8(v=vs.100).aspx

AlexPeshkoff: @asfernandes Instead readme added a chapter to OO API documentation. Usage sample also reworked to cover various batch aspects.

hvlad: Looks OK now, thanks!

AlexPeshkoff:
@asfernandes - want to make sure you've paid attention that description is complete in doc/Using_OO_API.html

asfernandes:
@AlexPeshkoff I read it, but did not reviewed the code changes yet. I plan to start today or tomorrow.

At first, I miss a way to pass BPB in addBlob.

AlexPeshkoff:
@asfernandes 
You are absolutely right - for Alpha1 I've decided to ignore BPB.
Blob parameters is a bit unclear issue. Currently the only supported type of BLOBs is stream blob but is's obvious that due to at least backward compatibility issues we need to support segmented too. And here main question comes - how to separate segments? For BLOB stream it's more or less clear - with segmented blob user must place segment size + segment data of appropriate length in a stream, one thing to decide - should that size be in host or network order. What about addBlob - I have no final idea what to do. Separate appendBlobData() for each segment is the simplest way for users but it produces additional overhead. Any ideas? (As far as I can see after deciding with this question the rest of BPB is just technical job)

hvlad:
BPB also used to specify blob filter and blob storage destination (persistent\temporary table space). 
As for segmented blob's - i can live without it at all :)

AlexPeshkoff:
On 01.08.2017 13:06, hvlad wrote:
>
> BPB also used to specify blob filter and blob storage destination 
> (persistent\temporary table space).
>

Certainly, and unlike segmented blobs I see no problems with this items.

> As for segmented blob's - i can live without it at all :)
>

I too - but afraid that gbak restore can't.

src/common/classes/BatchCompletionState.h
```diff
@@ -0,0 +1,185 @@
...
+		}
+
+		void dispose()
+		{
+			delete this;
+		}
+
+		~BatchCompletionState()
+		{
+			for (unsigned i = 0; i < rare.getCount() && rare[i].second; ++i)
```
asfernandes: @AlexPeshkoff can't a `rare` element have a NULL `second` and a subsequent element have non-NULL `second`? If it can, there is a leak here.

src/common/classes/vector.h
```diff
@@ -118,6 +126,20 @@ class Vector
...
 
+	void push(const T* items, const FB_SIZE_T itemsSize)
+	{
+		fb_assert(count <= FB_MAX_SIZEOF - itemsSize);
+		fb_assert(count + itemsSize <= Capacity);
+		memcpy(data + count, items, sizeof(T) * itemsSize);
+		count += itemsSize;
+	}
+
+	void append(const T* items, const FB_SIZE_T itemsSize)
```
asfernandes: @AlexPeshkoff I think to identical methods with different names is not a good thing. It's better to stablish on one or another. Also, itemsCount would be a better parameter name than itemsSize. When I read, I thought itemsSize was something about sizeof a single items buffer.

src/common/classes/BatchCompletionState.h
AlexPeshkoff: No, it can not. In rare array only first N (default 64) status vectors are kept - rest of the elements always have a NULL second. This is done to avoid server OOM state in case when crazy user has error in all messages.

src/common/classes/vector.h
AlexPeshkoff: In general I agree, but in this particular place - not. We have 2 ways/styles (don't know how to call it better) to work with Vector:
A. add/insert + remove
B. push + pop
Yes, push and add perform exactly same action but logically code looks better when in pair with push pop (not add) is used.

What about itemsSize - fixed.

PR #101, opened 2017-07-21
Excess conversion of sql statement to UTF8 is removed from fbtrace

ilya071294:
Сonversion is not necessary because statements already in UTF8. Also, this commit fixes exception in matcher when statement contains 2-byte characters. Excess conversion "damages" the statement in this case.


AlexPeshkoff:
Where is same change for master? Looks like it's also needed.

PR #102, opened 2017-07-24
Support for different hash algorithms in HASH system function (CORE-4436)

asfernandes:
Can please someone adjust Windows build?

I've used subset of libtomcrypt (public domain, same people from libtommath) disabling assembly for portability.


asfernandes:
Algorithms imported: MD5, SHA1, SHA256, SHA512.

hvlad:
Windows build is  OK now

asfernandes:
@hvlad thank you!

asfernandes:
Build is consistently failing (segfault executing our code) in travis with g++ 4.8.4. @AlexPeshkoff is this compiler version supported by us? The only thing I could imagine is failing is the usage of lambda and std::function.

dyemanov:
Adriano, we have libtommath inside /extern (as other non-FB-mastered code) but you have imported libtomcrypt into /common. I believe this violates project rules for external libraries. Could you please re-consider this location.

asfernandes:
Ok, will change it.

asfernandes:
There is an important thing to decide, as `extern` usage is not standard for all packages.

Should libtomcrypt be used as libicu and libtommath (in `extern` only for Windows and rely on OS installed library for POSIX)?

Or should it be like decNumber and used for both?

AlexPeshkoff:
On 09/04/17 18:21, Adriano dos Santos Fernandes wrote:
> There is an important thing to decide, as `extern` usage is not standard for all packages.
>
> Should libtomcrypt be used as libicu and libtommath (in `extern` only for Windows and rely on OS installed library for POSIX)?
>
> Or should it be like decNumber and used for both?
>

decNumber has one interesting specific - it's not recommended by authors 
to be used as dynamic library, probably due to use of macros instead 
some functions. And as it's built static there is no way to use OS 
installed library in our binaries.

But how is it related to placing of foreign code into extern, not 
src/common?

asfernandes:
Initial idea was to import subset of files important for us. It was added to src/common and integrated in the build.

Following libtommath, the extern files for libtomcrypt will generally not be important in common Linux build. It will use OS files.

So what should I do? Just move the current common/tomcrypt files to extern and add a custom Windows build? Or add full (or not) libtomcrypt?

Libtommath seems to be imported partially.

zlib and icu was added as compressed exe.

So much standards for the same thing. Since I do not do Windows build anymore for a long time, I have no idea how should it be done.

PR #105, opened 2017-08-18
Read consistency

hvlad:
The patch introduces:
- statement-level read consistency in read-committed transactions
- intermediate garbage collection - ability to clean up unused record versions at any place of versions chain (i.e. not necessarily at the tail of the chain)
- common TIP cache for all architectures (SS\CS)

More detailed description is preparing currently.


hvlad:
Alex, in general I agree. 
I found just one usage of AtomicPointer - at BufferControl::bcb_lru_chain. 
Is it make sence to use std::atomic there and completely remove AtomicPointer ?

PR #106, opened 2017-09-04
Fixed conversion of parameters to UTF8 in trace: reading string length depending on descriptor type

ilya071294:


AlexPeshkoff:
Also don't forget to backport to FB3

PR #108, opened 2017-09-04
High precision datatype support - Numeric(34,x)

AlexPeshkoff:
New datatype is based on decQuad from decNumber library - same as DecFloat(34). Fixed scale is enhanced programmatically by additional code around decNumber library. This resolves CORE-4409.

### DIFF
--- a/src/common/DecFloat.cpp
+++ b/src/common/DecFloat.cpp
@@ -904,6 +904,8 @@ DecimalFixed DecimalFixed::add(DecimalStatus decSt, DecimalFixed op2) const
 	DecimalContext context(this, decSt);
 	DecimalFixed rc;
 	decQuadAdd(&rc.dec, &dec, &op2.dec, &context);
+	context.checkForExceptions();
+	decQuadQuantize(&rc.dec, &rc.dec, &c1.dec, &context);
 	return rc;
 }
 
@@ -912,6 +914,8 @@ DecimalFixed DecimalFixed::sub(DecimalStatus decSt, DecimalFixed op2) const
 	DecimalContext context(this, decSt);
 	DecimalFixed rc;
 	decQuadSubtract(&rc.dec, &dec, &op2.dec, &context);
+	context.checkForExceptions();
+	decQuadQuantize(&rc.dec, &rc.dec, &c1.dec, &context);
 	return rc;
 }
 
@@ -920,6 +924,8 @@ DecimalFixed DecimalFixed::mul(DecimalStatus decSt, DecimalFixed op2) const
 	DecimalContext context(this, decSt);
 	DecimalFixed rc;
 	decQuadMultiply(&rc.dec, &dec, &op2.dec, &context);
+	context.checkForExceptions();
+	decQuadQuantize(&rc.dec, &rc.dec, &c1.dec, &context);
 	return rc;
 }
 
--- a/src/common/DecFloat.cpp
+++ b/src/common/DecFloat.cpp
@@ -71,7 +71,7 @@ public:
 		init(DEC_INIT_DECIMAL64);
 	}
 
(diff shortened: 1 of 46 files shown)


AlexPeshkoff:
@asfernandes the branch is ready for review

src/common/DecFloat.cpp
```diff
@@ -117,7 +117,9 @@ class DecimalContext : public decContext
 };
 
 CDecimal128 dmax(DBL_MAX, DecimalStatus(0)), dmin(-DBL_MAX, DecimalStatus(0));
+CDecimal128 dzup(DBL_MIN, DecimalStatus(0)), dzlw(-DBL_MIN, DecimalStatus(0));
```
asfernandes: @AlexPeshkoff can't these variables be marked as const?

src/common/DecFloat.cpp
```diff
@@ -582,12 +664,34 @@ void Decimal128::toString(string& to) const
...
 	if (compare(decSt, dmin) < 0 || compare(decSt, dmax) > 0)
 		decContextSetStatus(&context, DEC_Overflow);
+	else if ((!decQuadIsZero(&dec)) && compare(decSt, dzlw) > 0 && compare(decSt, dzup) < 0)
+	{
+		decContextSetStatus(&context, DEC_Underflow);
+		return 0.0;
+	}
 	else
 	{
 		char s[IDecFloat34::STRING_SIZE];
```
asfernandes: @AlexPeshkoff here I see:

```
		char s[IDecFloat34::STRING_SIZE];
		memset(s, 0, sizeof(s));
		decQuadToString(&dec, s);
		return atof(s);
```

It seems not necessary to initialize the buffer before call decQuadToString. In utl.cpp you did not it.

src/common/dsc.h
```diff
@@ -296,6 +301,14 @@ typedef struct dsc
 		dsc_address = (UCHAR*) address;
 	}
 
+	void makeDecimalFixed(Firebird::DecimalFixed* address = NULL)
```
asfernandes: @AlexPeshkoff shouldn't you set the scale in this function?

src/dsql/ExprNodes.cpp
```diff
@@ -467,6 +467,80 @@ void ArithmeticNode::genBlr(DsqlCompilerScratch* dsqlScratch)
 	GEN_expr(dsqlScratch, arg2);
 }
 
+namespace
+{
+
+const UCHAR f64 = 0;
```
asfernandes: @AlexPeshkoff these names seems very common to be in the file namespace, also I think constants should be all upper-cased. I believe some PREFIX_ and upper-case names would be clear.

src/include/firebird/Message.h
```diff
@@ -141,6 +141,10 @@
 	builder->setType(status, index, SQL_DEC34);	\
 	builder->setLength(status, index, sizeof(FB_DEC34));
 
+#define FB__META_FB_DEC_FIXED	\
```
asfernandes: @AlexPeshkoff this macro needs a scale parameter.

src/jrd/PreparedStatement.cpp
```diff
@@ -109,6 +109,11 @@ namespace
 				item.length = sizeof(Decimal128);
 				break;
 
+			case dtype_dec_fixed:
```
asfernandes: @AlexPeshkoff I think the scale should be set here too.

src/common/dsc.h
AlexPeshkoff: Hops - this function is not used at all. Removed it.

src/jrd/PreparedStatement.cpp
AlexPeshkoff: Thank you, fixed all suggested places

PR #109, opened 2017-09-06
Put libtomcrypt files in extern and use/build as libtommath in POSIX

asfernandes:
Please someone do the Windows changes.


hvlad:
Patch for Windows is in progress

asfernandes:
> MSVC10 and MSVC14 builds will follow

@hvlad should I wait these commits before merge what we have now?

hvlad:
As far as we build snapshots (and alpha) with MSVC12 - you shouldn't wait for other MSVC project files.
Sorry, did not say it before.

PR #110, opened 2017-09-22
Fix crash on external engine trigger executing for transaction triggers

artyom-smirnov:
trg->relation can be null, so need checking


artyom-smirnov:
Found this bug when working with fbjava. I've have fixed it straightforward, but may be better solution, exists.

artyom-smirnov:
Also same crash in 3.0 branch

AlexPeshkoff:
Your fix is OK - please provide fix for 3.0 too

artyom-smirnov:
My mistake, 3.0 branch not affected, we had crash in Red Database 3.0 because of 4.0 backported code.

PR #111, opened 2017-09-25
Do not scan function if compiling not needed

artyom-smirnov:
Argument 'compile' wasn't negated in this place, for argument 'noscan'

### DIFF
--- a/src/jrd/dfw.epp
+++ b/src/jrd/dfw.epp
@@ -695,7 +695,7 @@ namespace
 					getDependencies(work, compile, transaction);
 
 					T* routine = lookupByName(tdbb,
-						QualifiedName(work->dfw_name, work->dfw_package), compile);
+						QualifiedName(work->dfw_name, work->dfw_package), !compile);
 
 					if (!routine)
 						return false;


dyemanov:
I'm letting Adriano to verify the fix, but as far as I see v2.5 is also affected.

asfernandes:
Apart from confusion with variable names, I think the current code is correct.

If dfw_arg_check_blr is present, compile = false and passed as noscan = false, with means to scan.

```
const bool compile = !work->findArg(dfw_arg_check_blr);
getDependencies(work, compile, transaction);

T* routine = lookupByName(tdbb,
	QualifiedName(work->dfw_name, work->dfw_package), compile);
```

But I see problem in another place. Here inside getDependencies:

```
if (compile)
	compile = !tdbb->getAttachment()->isGbak();
```

But this compile also means noscan and then seems wrong.

Opinions?

artyom-smirnov:
I see. Then shouldn't this code be something like
`const bool compile = !(work->findArg(dfw_arg_check_blr) && tdbb->getAttachment()->isGbak());`
to avoid scanning during database restoring?

asfernandes:
I'm looking more deeply in this problem.

PR #112, opened 2017-09-25
Do not scan function if compiling not needed

artyom-smirnov:
Argument 'compile' wasn't negated in this place, for argument 'noscan'


asfernandes:
Related discussion in #111.

PR #114, opened 2017-10-03
Fixed CORE-5611 - Higher memory consumption for prepared statements.

asfernandes:
- Remove some fields from nodes, including the pool reference and list of children nodes
- Organize some fields for better alignment and possible reduction on node size
- Delete DSQL scratch pool for DML statements

### DIFF
--- a/src/dsql/dsql.cpp
+++ b/src/dsql/dsql.cpp
@@ -1418,25 +1418,28 @@ static dsql_req* prepareStatement(thread_db* tdbb, dsql_dbb* database, jrd_tra* 
 
 	try
 	{
-		Parser* parser = FB_NEW_POOL(*scratchPool) Parser(tdbb, *scratchPool, scratch, clientDialect,
-			scratch->getAttachment()->dbb_db_SQL_dialect, text, textLength,
-			tdbb->getAttachment()->att_charset);
+		string transformedText;
 
-		{	// scope
+		{	// scope to delete parser before the scratch pool is gone
 			Jrd::ContextPoolHolder scratchContext(tdbb, scratchPool);
 
+			Parser parser(tdbb, *scratchPool, scratch, clientDialect,
+				scratch->getAttachment()->dbb_db_SQL_dialect, text, textLength,
+				tdbb->getAttachment()->att_charset);
+
 			// Parse the SQL statement.  If it croaks, return
-			request = parser->parse();
+			request = parser.parse();
+
+			if (parser.isStmtAmbiguous())
+				scratch->flags |= DsqlCompilerScratch::FLAG_AMBIGUOUS_STMT;
+
+			transformedText = parser.getTransformedString();
 		}
 
 		request->req_dbb = scratch->getAttachment();
 		request->req_transaction = scratch->getTransaction();
 		request->statement = scratch->getStatement();
 
-		if (parser->isStmtAmbiguous())
-			scratch->flags |= DsqlCompilerScratch::FLAG_AMBIGUOUS_STMT;
-
-		string transformedText = parser->getTransformedString();
 		SSHORT charSetId = database->dbb_attachment->att_charset;
 
 		// If the attachment charset is NONE, replace non-ASCII characters by question marks, so
--- a/src/dsql/dsql.cpp
+++ b/src/dsql/dsql.cpp
@@ -1520,11 +1520,8 @@ static dsql_req* prepareStatement(thread_db* tdbb, dsql_dbb* database, jrd_tra* 
 		}
 		else
 		{
-			if (scratchPool)
-				database->deletePool(scratchPool);
-
-			if (statementPool)
-				database->deletePool(statementPool);
+			database->deletePool(scratchPool);
+			database->deletePool(statementPool);
 		}
 
 		throw;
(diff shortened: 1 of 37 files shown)


asfernandes:
Can someone please review?

asfernandes:
Anyone to review? @dyemanov should all these changes later be backported to 3.0?

The commit fa821815fe808052ea58eea75fd5cdd268b473fc is an one that should be rewriten from scratch to backport.

dyemanov:
I'm reviewing it right now.

src/dsql/dsql.cpp
```diff
@@ -1405,22 +1412,25 @@ static dsql_req* prepareStatement(thread_db* tdbb, dsql_dbb* database, jrd_tra*
 
 	try
 	{
-		// Parse the SQL statement.  If it croaks, return
-
-		Parser parser(tdbb, *tdbb->getDefaultPool(), scratch, clientDialect,
+		Parser* parser = FB_NEW_POOL(*scratchPool) Parser(tdbb, *scratchPool, scratch, clientDialect,
 			scratch->getAttachment()->dbb_db_SQL_dialect, text, textLength,
 			tdbb->getAttachment()->att_charset);
```
dyemanov: What is the point in allocating the parser instance dynamically?

src/jrd/Attachment.cpp
```diff
@@ -111,10 +111,11 @@ void Jrd::Attachment::destroy(Attachment* const attachment)
...
-MemoryPool* Jrd::Attachment::createPool()
+MemoryPool* Jrd::Attachment::createPool(MemoryPool* parent)
 {
-	MemoryPool* const pool = MemoryPool::createPool(att_pool, att_memory_stats);
-	att_pools.add(pool);
+	MemoryPool* const pool = MemoryPool::createPool(parent ? parent : att_pool, att_memory_stats);
+	if (!parent)
+		att_pools.add(pool);
 	return pool;
 }
```
dyemanov: Why is the parent-based new pool not accounted in att_pools? It's still a proper pool and must be destroyed via deletePool() in ~Attachment() if not deleted explicitly.

asfernandes: When the scratch pool is manually deleted (DML statement compiled without error) the thing works with or without registering it in att_pools, as deletePool will be called and remove it.

But for the non-DML case or DML with exception in compilation, that means only the statement pool is explicitly destroyed (scratch is child of statement). In this case, it will try to destroy an already deallocated pool in detach.

Is there any real problem in deleting a child pool via its parent?

asfernandes: After some tests I see that current usage (let child pool die with parent) does not decrement child pool's memory statistics when parent dies.

I think parent pools should maintain a list of its active children and when parent is destroyed it should decrement statistics from children.

AlexPeshkoff: @asfernandes Does child pool unmap memory got from OS when parent is destroyed? As far as I remember we always used to explicitly destroy child pools before parent.

dyemanov: > Is there any real problem in deleting a child pool via its parent?
AFAIK, it just never happened in our code, so I cannot answer for sure. "Parent" concept was used to redirect allocations up to some amount, but the child pool may have its own memory blocks that the parent is not aware of (and thus cannot deallocate).

asfernandes: @AlexPeshkoff if I manually destroy parent then follow and destroy child, a segfault happens and this seems very correct, as should be the concept of child pool.

But if I only destroy parent, child does not decrement its usage in the memory stats object.

So IMO parent should do this for its active children when parent is deleted.

asfernandes: When a child pool is going to allocate large amounts of memory, are these blocks requested from parent or are mapped directly with the OS?

dyemanov: > But if I only destroy parent, child does not decrement its usage in the memory stats object.

AFAIU, the idea is that the parent should never die before any of its children. Once they're dead, it makes no sense to adjust their counters.

dyemanov: In other words, "delete by pool" concept does not apply to child pools. And I'm not sure it should - it would complicate things without any real gain.

asfernandes: Then how should you prefer? Making the scratch pool independent (not as child of statement) or use parent-child implementing some logic related to att_pools to never delete child after parent?

AlexPeshkoff: > When a child pool is going to allocate large amounts of memory, are these blocks requested from parent or are mapped directly with the OS?

They are mapped directly from OS.

dyemanov: I'd prefer to make the scratch pool independent but please ensure that it dies when necessary (prepare failure?).

asfernandes: @dyemanov is it ok to merge in master now?

asfernandes:
@dyemanov is it ok to merge in master now?

dyemanov:
I don't have other complaints, so it can be merged. Let's wait a bit before considering a backport into v3 though.

dyemanov:
@asfernandes , can we get back to this issue and backport into v3?

asfernandes:
> @asfernandes , can we get back to this issue and backport into v3?

I'm rebasing and completing the old patch.

PR #116, opened 2017-10-09
Recalculate computed fields

vasiliy-yashkov:
#CORE-5633


hvlad (commented):
It looks dangerous to me to run separate transaction in DWF. And, of course, dynamic statements also better to be avoided in such place.
Usually, if you need to perfrom additional actions for user DDL statement, you should add it at DYN or VIO level. I don't look in depth in this case, though.

asfernandes:
Also, it uses object sources in a place Firebird never used it, as Firebird internals currently just works with BLR.

romansimakov:
So what options are possible? Just prohibit altering type of field if there are some dependencies? Leave it to user to clear dependency, alter field type, restore dependend objects?

hvlad:
Roman, 
it just should be implemented properly :) Contact me and we'll discuss how it could be done in correct way.

asfernandes:
The same problem happens with views too.

dyemanov:
Closed as unmergeable.

PR #118, opened 2017-10-12
Services API simplification

aafemt:
If you have specified database in isc_service_attach() call, you can omit it in following isc_service_start() calls.


AlexPeshkoff:
I do not understand - why waste space on github with with a change which was discussed and rejected on fb-devel?

PR #122, opened 2017-10-26
Update .travis.yml

mariuz:
add B2_5_Release to travis config


asfernandes:
I think this is not going to be necessary, as travis will get the file from B2_5_Release.

PR #128, opened 2017-11-14
Stop database restore with -o option when no space left on disk

ilya071294:
When this error occurs, gbak tries to restore all remained records. Restore log and firebird.log are growing up to several GB because of errors. The same occurs if connection was lost. I think it's better to allow to continue restore for some errors (currently isc_not_valid and isc_malformed_string), but prohibit for all others.

### DIFF
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -3056,17 +3056,8 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
 					BURP_error_redirect(status_vector, 342);	// isc_gbak_invalid_data
 			}
 			else
-			{
-				if (tdgbl->gbl_sw_incremental)
-				{
-					BURP_print (false, 114, relation->rel_name);
-					// msg 114 restore failed for record in relation %s
-					BURP_print_status (false, status_vector);
-				}
-				else
-					BURP_error_redirect (status_vector, 48);
-					// msg 48 isc_send failed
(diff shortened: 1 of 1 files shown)


ilya071294:
Please, review it.

ilya071294:
There are many cases when -o option might help to restore database with ignoring errors. For this reason gbak should stop restore only when unrecoverable errors occur (such as isc_io_error, isc_att_shutdown, isc_net_write_err).

src/burp/restore.epp
```diff
@@ -3057,7 +3059,10 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
 			}
 			else
 			{
-				if (tdgbl->gbl_sw_incremental)
+				if (tdgbl->gbl_sw_incremental &&
```
AlexPeshkoff: I agree with an idea to make error processing during restore more flexible. But list of fatal errors appears to be incomplete. Except isc_net_write_err isc_net_read_err can happen, except isc_att_shutdown - isc_shutdown, etc. On the other hand all such gds codes are already classified as SQL code -902. Why not use it instead?

AlexPeshkoff:
I see PR to B3_0_Release. Same changes in master are also required.

ilya071294:
If PR is OK now, I'll prepare the same for master.

AlexPeshkoff:
I see no more problems with this PR

PR #130, opened 2017-11-16
pcmet was removed

mapopa:
fixes http://tracker.firebirdsql.org/browse/CORE-5662

PR #131, opened 2017-11-16
CMake Linux fixes

vadz:
With these changes, build using CMake succeeds under Debian x86_64.

NB: this extends and replaces #130.

### DIFF
--- a/builds/cmake/Configure.cmake
+++ b/builds/cmake/Configure.cmake
@@ -200,6 +200,7 @@ set(functions_list
     accept4
     AO_compare_and_swap_full
     clock_gettime
+    ctime_r
     dirname
     fallocate
     fchmod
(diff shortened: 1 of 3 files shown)


egorpugin:
`ctime_r` fix also proposed for master https://github.com/FirebirdSQL/firebird/pull/133

vadz:
Just a warning: I've abandoned the idea of using CMake as while the build succeeds, the generated binary crashes with non obvious error during run-time when the same version built with configure+make works fine. So there is still something very wrong with CMake build and perhaps it's better for it to remain broken...

egorpugin:
It's also a known issue. It will be investigated some time later.

egorpugin:
Btw, do you have a simple repro of this crash? I'd like to check it.

vadz:
I saw it just by running `isql` with a simple SQL script creating some tables, so I think it should happen with anything.

mapopa:
Can be closed , #133 is now merged

egorpugin:
No, this is for 3.0 branch.

mariuz:
@asfernandes @AlexPeshkoff  could  you merge this for 3.0 branch ?

PR #134, opened 2017-12-05
Fixed CORE-5675: isc_vax_integer() and isc_portable_integer() wrong work with short negative numbers

aafemt:


aafemt:
Backports for 3.0 and 2.5 also ready.

dyemanov:
Does ClumpletReader also need the same fix or ClumpletWriter stores negatives differently?

aafemt:
For integers most likely yes, but looking at ClumpletReader.getDouble() I'm not sure that the same fix will work with it.

PR #137, opened 2017-12-07
Backport for CORE-5675: isc_vax_integer() and isc_portable_integer() wrong work with short negative numbers

aafemt:


dyemanov:
Why the ClumpletReader fix is missing in this PR?

PR #139, opened 2017-12-27
Gbak

AlexPeshkoff:
Use of IBatch in gbak seriously increases performance of restore over the wire. In a case of localhost connection using TCP/IP restore runs 2 - 6 times faster. First case is with a real database, containing a lot of indices and some blobs, second - specially created test database without indices but with a lot of small blobs. 

This also fixes http://tracker.firebirdsql.org/browse/CORE-5637 due to use of SQL instead BLR to restore tables like PLG$USERS which helps avoid conflict between old field length and new UNICODE_FSS rules.

### DIFF
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -2842,6 +2842,7 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
 
 	if (tdgbl->gbl_network_protocol == 0)
 	{
+//		bool sysDomFlag = true;		// Debug !!!!!!!!!!!!!!
 		bool sysDomFlag = false;
 
 		for (field = relation->rel_fields; field && !sysDomFlag; field = field->fld_next)
@@ -2886,8 +2887,10 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
 
 	try
 	{
+		Firebird::string name(relation->rel_name);
+		BURP_makeSymbol(tdgbl, name);
 		Firebird::string sqlStatement;
-		sqlStatement.printf("insert into %s(", relation->rel_name);
+		sqlStatement.printf("insert into %s(", name.c_str());
 
 		Firebird::RefPtr<Firebird::IMetadataBuilder> builder(Firebird::REF_NO_INCR,
 			Firebird::MasterInterfacePtr()->getMetadataBuilder(fbStatus, count));
@@ -2944,7 +2947,9 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
 			builder->setScale(&tdgbl->throwStatus, count, sqlScale);
 			builder->setCharSet(&tdgbl->throwStatus, count, characterSetId);
 
-			sqlStatement += field->fld_name;
+			name = field->fld_name;
+			BURP_makeSymbol(tdgbl, name);
+			sqlStatement += name;
 			sqlStatement += ", ";
 
 			field->fld_parameter = count++;
@@ -3006,7 +3011,6 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
 			// Possible reason of fail - use of keywords as fields in old backup
 			// Try to roll back to use of old version (fieldnames independent)
 			BURP_print(false, 27/*, sqlStatement.c_str()*/);	// msg 27 isc_compile_request failed     !!!!!!!! new WARNING
-			BURP_print_warning(fbStatus);
 			return get_data_old(tdgbl, relation);
 		}
 
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -2721,7 +2721,7 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
 	// Start by counting the interesting fields
 
 	RCRD_OFFSET offset = 0;
-	ULONG length = 0;
+	RCRD_LENGTH length = 0;
 	USHORT count = 0;
 
 	burp_fld* field;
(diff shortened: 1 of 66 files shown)


dyemanov:
Any test results using the embedded connection, e.g. Services API? It should also be faster, AFAIU.

AlexPeshkoff:
No, it's not faster. Moreover, an attempt to use SQL batch instead traditional BLR-based restore causes approximately 5% slowdown. That's because of need to put messages into buffer, perform additional map_in_out(), etc. And after it invoke that same EXE_send() that is called from gbak almost directly. At least that's not new fact that BLR-based data processing is slightly faster compared with SQL.
Therefore I've kept old way for embedded processing (with some exceptions but they are not important when talking about performance). May be sometimes later when we have deeper optimization of batch processing inside engine it will make sense to review that solution.

src/burp/restore.epp
```diff
@@ -2710,18 +2827,511 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
...
+		if (!(field->fld_flags & FLD_computed))
+			++count;
+	}
+
+	// Time to generate SQL and message metadata to store data.  Whoppee.
+
+	try
+	{
+		Firebird::string sqlStatement;
+		sqlStatement.printf("insert into %s(", relation->rel_name);
```
asfernandes: @AlexPeshkoff this will not work for non-ASCII names or keywords.

src/burp/restore.epp
```diff
@@ -2710,18 +2827,511 @@ rec_type get_data(BurpGlobals* tdgbl, burp_rel* relation, bool skip_relation)
...
+				sqlScale = 0;
+			}
+
+			builder->setType(&tdgbl->throwStatus, count, sqlType);
+			builder->setSubType(&tdgbl->throwStatus, count, sqlSubType);
+			builder->setLength(&tdgbl->throwStatus, count, sqlLength);
+			builder->setScale(&tdgbl->throwStatus, count, sqlScale);
+			builder->setCharSet(&tdgbl->throwStatus, count, characterSetId);
+
+			sqlStatement += field->fld_name;
```
asfernandes: @AlexPeshkoff this will not work for non-ASCII names or keywords.

src/burp/restore.epp
```diff
@@ -10498,26 +11053,25 @@ void store_blr_gen_id(BurpGlobals* tdgbl, const TEXT* gen_name, SINT64 value, SI
...
 		fb_print_blr(blr_buffer, blr_length, NULL, NULL, 0);
-		BURP_error_redirect (status_vector, 42); // msg 42 Failed in store_blr_gen_id
+		BURP_error_redirect(&status_vector, 42); // msg 42 Failed in store_blr_gen_id
 	}
 
 	BURP_verbose (185, SafeArg() << gen_name << value);
 	// msg 185 restoring generator %s value: %ld
 
-	isc_release_request (status_vector, &gen_id_reqh);
+	gen_id_reqh->release();
```
asfernandes: @AlexPeshkoff isn't better to use RefPtr here?

src/common/classes/auto.h
```diff
@@ -168,6 +168,16 @@ class AutoPtr
 };
 
 
+template <typename ID>
```
asfernandes: @AlexPeshkoff maybe we can use C++11's template ... using here?

src/include/firebird/FirebirdInterface.idl
```diff
@@ -599,6 +599,8 @@ version:	// 3.0 => 4.0
 	// Batch API
 	Batch createBatch(Status status, Transaction transaction, uint stmtLength, const string sqlStmt,
 		uint dialect, MessageMetadata inMetadata, uint parLength, const uchar* par);
+
+	uint getRemoteProtocolVersion(Status status);
```
asfernandes: @AlexPeshkoff can't this be a method of another thing, for example IUtil receiving the attachment as parameter? It does not look good for me in IAttachment... Also, looks like a layer violation. Maybe something about consult capabilities information would be much better to consult in gbak.

src/jrd/exe.cpp
```diff
@@ -692,7 +692,7 @@ void EXE_receive(thread_db* tdbb,
 
 			if (desc->isBlob())
 			{
-				const bid* id = (bid*) (buffer + (ULONG)(IPTR)desc->dsc_address);
+				const bid* id = (bid*) (reinterpret_cast<UCHAR*>(buffer) + (ULONG)(IPTR)desc->dsc_address);
```
asfernandes: @AlexPeshkoff can'tbe static_cast here?

src/remote/client/interface.cpp
```diff
@@ -5600,7 +5615,7 @@ void Request::send(CheckStatusWrapper* status, int level, unsigned int msg_type,
 
 		RMessage* message = request->rrq_rpt[msg_type].rrq_message;
 		// We are lying here, but the interface shows for years this param as const
-		message->msg_address = const_cast<UCHAR*>(msg);
+		message->msg_address = const_cast<unsigned char*>(reinterpret_cast<const unsigned char*>(msg));
```
asfernandes: This also seems possible to be static_cast.

src/remote/client/interface.cpp
```diff
@@ -5893,7 +5908,7 @@ void Request::startAndSend(CheckStatusWrapper* status, Firebird::ITransaction* a
 
 		REMOTE_reset_request(request, 0);
 		RMessage* message = request->rrq_rpt[msg_type].rrq_message;
-		message->msg_address = const_cast<unsigned char*>(msg);
+		message->msg_address = const_cast<unsigned char*>(reinterpret_cast<const unsigned char*>(msg));
```
asfernandes: And here too...

src/common/classes/auto.h
AlexPeshkoff: @asfernandes On my mind what is needed here is template template parameter, i.e.

template<typename Where, template<typename> class Clear = SimpleDelete>
class AutoPtr { .................

but I'm not sure have we decided to use them in FB4 or not.

src/include/firebird/FirebirdInterface.idl
AlexPeshkoff: Do not see layer violation - but agree that this call is looking bad here. Avoid it in favor of 'good old' getFbVersion().

PR #140, opened 2018-01-01
ARM needs MSG_NOSIGNAL in FB_SEND_FLAGS, too

maciejsszmigiero:
We need to add MSG_NOSIGNAL to FB_SEND_FLAGS for ARM, too, otherwise
clients get killed with SIGPIPE when server connection breaks, just like
commit 37bcd38740a187dad79a14a47241945fcd193b82 did for other platforms.

Use this opportunity to also split out common build flags to COMMON_FLAGS
variable, as the above commit did.

This change will also need to be applied to 2.1, 2.5 and 3.0 branches, with a small adjustment:
2.1 and 2.5 have additional -DNDEBUG in PROD_FLAGS.

### DIFF
--- a/builds/posix/prefix.linux_arm
+++ b/builds/posix/prefix.linux_arm
@@ -20,10 +20,9 @@
 
 #LD=@CXX@
 
-#PROD_FLAGS=-ggdb -O3 -fno-omit-frame-pointer -DLINUX -pipe -MMD -fPIC
-#DEV_FLAGS=-ggdb -DLINUX -DDEBUG_GDS_ALLOC -pipe -MMD -p -fPIC -Wall -Wno-switch
+COMMON_FLAGS=-DLINUX -DARM -pipe -MMD -fPIC -fsigned-char -fmessage-length=0 -DFB_SEND_FLAGS=MSG_NOSIGNAL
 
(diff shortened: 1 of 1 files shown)


mariuz:
@AlexPeshkoff @asfernandes could you merge this , seems to be trivial

AlexPeshkoff:
On 02/15/18 15:09, Popa Marius Adrian wrote:
>
> @AlexPeshkoff <https://github.com/alexpeshkoff> @asfernandes 
> <https://github.com/asfernandes> could you merge this , seems to be 
> trivial
>
done

maciejsszmigiero:
Thanks for merging this into master.

However:
> This change will also need to be applied to 2.1, 2.5 and 3.0 branches, with a small adjustment:
> 2.1 and 2.5 have additional -DNDEBUG in PROD_FLAGS.

This is a bug fix for clients getting killed with SIGPIPE so it should go into stable branches, too.

maciejsszmigiero:
Ping about applying this bug fix to stable branches, too.

AlexPeshkoff:
Where is pull request for 2.5 & 3.0. 2.1 is unsupported long ago.

maciejsszmigiero:
You want two additional pull requests for 2.5 and 3.0?
Unfortunately, GitHub seems to only support targeting one branch for one pull request.

In a similar case in PR 74 you simply applied that patch to these branches, too.

PR #141, opened 2018-02-21
Fixed CORE-5758: Database is corrupted when conflicting "starting at page" values is specified for secondary files

ilya071294:
Example:
`create database '127.0.0.1:C:\Tasks\db.fd1'
user 'SYSDBA' password 'masterkey';

alter database add file 'C:\Tasks\db.fd2' starting at page 1001;

alter database add file 'C:\Tasks\db.fd3' starting at page 501;`

The database will be corrupted after creating tables and inserting some data in the current connection.

### DIFF
--- a/src/jrd/dfw.epp
+++ b/src/jrd/dfw.epp
@@ -1785,7 +1785,7 @@ static bool add_file(thread_db* tdbb, SSHORT phase, DeferredWork* work, jrd_tra*
  *
  **************************************/
 	USHORT section, shadow_number;
-	SLONG start, max;
+	SLONG start, min_start;
 
 	SET_TDBB(tdbb);
 	Database* const dbb = tdbb->getDatabase();
@@ -1807,7 +1807,7 @@ static bool add_file(thread_db* tdbb, SSHORT phase, DeferredWork* work, jrd_tra*
 
 	case 4:
 		CCH_flush(tdbb, FLUSH_FINI, 0);
-		max = PageSpace::maxAlloc(dbb) + 1;
+		min_start = PageSpace::maxAlloc(dbb) + 1;
 		AutoRequest handle;
 		AutoRequest handle2;
 
@@ -1839,20 +1839,19 @@ static bool add_file(thread_db* tdbb, SSHORT phase, DeferredWork* work, jrd_tra*
 			// too low a value, make a stab at assigning one based on
 			// the indicated preference for the previous file length.
 
-			if ((start = X.RDB$FILE_START) < max)
-			{
-				FOR(REQUEST_HANDLE handle2 TRANSACTION_HANDLE transaction)
-					FIRST 1 Y IN RDB$FILES
-						WITH Y.RDB$SHADOW_NUMBER EQ X.RDB$SHADOW_NUMBER
-						AND Y.RDB$FILE_SEQUENCE NOT MISSING
-						SORTED BY DESCENDING Y.RDB$FILE_SEQUENCE
-				{
-					start = Y.RDB$FILE_START + Y.RDB$FILE_LENGTH;
-				}
-				END_FOR
-			}
+			start = MAX(min_start, X.RDB$FILE_START);
 
-			start = MAX(max, start);
+			FOR(REQUEST_HANDLE handle2 TRANSACTION_HANDLE transaction)
+				FIRST 1 Y IN RDB$FILES
+					WITH Y.RDB$SHADOW_NUMBER EQ X.RDB$SHADOW_NUMBER
+					AND Y.RDB$FILE_SEQUENCE NOT MISSING
+					SORTED BY DESCENDING Y.RDB$FILE_SEQUENCE
+			{
+				min_start = Y.RDB$FILE_START + (Y.RDB$FILE_LENGTH ? Y.RDB$FILE_LENGTH : 1);
+				start = MAX(min_start, start);
+			}
+			END_FOR
+
 			shadow_number = X.RDB$SHADOW_NUMBER;
 			if ((shadow_number &&
 				(section = SDW_add_file(tdbb, X.RDB$FILE_NAME, start, shadow_number))) ||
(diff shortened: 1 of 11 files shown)


src/jrd/dfw.epp
```diff
@@ -1839,20 +1839,19 @@ static bool add_file(thread_db* tdbb, SSHORT phase, DeferredWork* work, jrd_tra*
...
-				FOR(REQUEST_HANDLE handle2 TRANSACTION_HANDLE transaction)
-					FIRST 1 Y IN RDB$FILES
-						WITH Y.RDB$SHADOW_NUMBER EQ X.RDB$SHADOW_NUMBER
-						AND Y.RDB$FILE_SEQUENCE NOT MISSING
-						SORTED BY DESCENDING Y.RDB$FILE_SEQUENCE
-				{
-					start = Y.RDB$FILE_START + Y.RDB$FILE_LENGTH;
-				}
-				END_FOR
+				min_start = Y.RDB$FILE_START + (Y.RDB$FILE_LENGTH ? Y.RDB$FILE_LENGTH : 1);
```
asfernandes: I cannot comment on the fix, but is the test for `Y.RDB$FILE_LENGTH` about it being 0 or about it being `NULL`? If `NULL`, you should test `Y.RDB$FILE_LENGTH.NULL`...

hvlad:
Register bug at tracker, please

ilya071294:
CORE-5758

src/jrd/dfw.epp
ilya071294: If the length of the previous file is not specified, `Y.RDB$FILE_LENGTH` is 0.

romansimakov: It's nullable and it's better to check it explicitly. Right?

ilya071294: Yes, `RDB$FILE_LENGTH` is a nullable field, but actually it can't be NULL here.

AlexPeshkoff: In that case check for this will be very useful. IMHO this code is very far from performance critical paths, i.e. extra check does not cause troubles.

hvlad:
Is it better to raise error instead of "correcting" user error ? 
Show contents of RDB$FILES when engine with patch run your test case, please.

ilya071294:
Result of `select * from rdb$files;`
[out.txt](https://github.com/FirebirdSQL/firebird/files/1747807/out.txt)

hvlad:
Thanks, it looks like

RDB$FILE_NAME | RDB$FILE_SEQUENCE | RDB$FILE_START | RDB$FILE_LENGTH 
-------- | -------- | -------- | -------- 
C:\RDB\TASKS\24682\DB.FD2 | 1 | 1001 | 1 
C:\RDB\TASKS\24682\DB.FD3 | 2 | 1002 | 0 

is it OK that second file (db.fd2) length is just 1 page ? 
Does it have sence ? 
I think, it is better to avoid such cases and raise error.

AlexPeshkoff:
On 02/22/18 17:14, hvlad wrote:
>
> Thanks, it looks like
>
> RDB$FILE_NAME 	RDB$FILE_SEQUENCE 	RDB$FILE_START 	RDB$FILE_LENGTH
> C:\RDB\TASKS\24682\DB.FD2 	1 	1001 	1
> C:\RDB\TASKS\24682\DB.FD3 	2 	1002 	0
>
> is it OK that second file (db.fd2) length is just 1 page ?
> Does it have sence ?
> I think, it is better to avoid such cases and raise error.
>

+1

ilya071294:
I have updated the branch. Please, review it.

src/msgs/messages2.sql
```diff
@@ -897,6 +897,7 @@ Data source : @4', NULL, NULL)
 ('already_opened', 'lockDatabaseFile', 'unix.cpp', NULL, 0, 787, NULL, 'Database already opened with engine instance, incompatible with current', NULL, NULL);
 ('bad_crypt_key', NULL, 'CryptoManager.cpp', NULL, 0, 788, NULL, 'Invalid crypt key @1', NULL, NULL);
 ('encrypt_error', NULL, 'CryptoManager.cpp', NULL, 0, 789, NULL, 'Page requires encyption but crypt plugin is missing', NULL, NULL);
```
hvlad: There is a (traditional) problem with error codes already used in next version. Codes 790-884 of JRD facility is already used at master branch. We could add code 885 (as in master), creating big gap in numeration for v3, or, use isc_random in v3. Opinions ?

AlexPeshkoff:
On 03/20/18 13:49, hvlad wrote:
>
> *@hvlad* commented on this pull request.
>
>
> In src/msgs/messages2.sql 
> <https://github.com/FirebirdSQL/firebird/pull/141#discussion_r175722580>:
>
> > @@ -897,6 +897,7 @@ Data source : @4', NULL, NULL)
>   ('already_opened', 'lockDatabaseFile', 'unix.cpp', NULL, 0, 787, NULL, 'Database already opened with engine instance, incompatible with current', NULL, NULL);
>   ('bad_crypt_key', NULL, 'CryptoManager.cpp', NULL, 0, 788, NULL, 'Invalid crypt key @1', NULL, NULL);
>   ('encrypt_error', NULL, 'CryptoManager.cpp', NULL, 0, 789, NULL, 'Page requires encyption but crypt plugin is missing', NULL, NULL);
>
> There is a (traditional) problem with error codes already used in next 
> version. Codes 790-884 of JRD facility is already used at master 
> branch. We could add code 885 (as in master), creating big gap in 
> numeration for v3, or, use isc_random in v3. Opinions ?
>

I prefer isc_random in such cases. Small additional proof for it - with 
minor version upgrade people often tend to forget about clients.

ilya071294:
I'm going to prepare same PR for master. Should I use isc_random, like in v3, or add new message (code 885)?

AlexPeshkoff:
for master definitely new code

ilya071294:
PR is OK now?

AlexPeshkoff:
Ilya sorry - right now I have no time to review PRs. Very many new (and 
serious) bugs in FB3 last week.
Will return to it a bit later.

PR #142, opened 2018-02-26
add modulo operator '%'

nakagami:


nakagami:
No problem.
Certainly Oracle is not implemented % operator,
but PostgreSQL and MySQL have both MOD function and modulo operator.

So I want both MOD and % if it possible.

mariuz:
https://www.postgresql.org/docs/16/functions-math.html

asfernandes:
I think this PR should be closed and not merged.
PostgreSQL way is different than Firebird way, they use much more non standard operators.

PR #143, opened 2018-03-07
Redirect stdout and stderr to /dev/null if possible

artyom-smirnov:
Guardian closes stdin/stdout/stderr/... before
launching server binary so server can reuse those
descriptors when opening files or using shmem for
example. So some stray stdout can introduce
unwanted data or even crash server. Server do not
emit any stdout/stderr in release build but
UDRs or external libraries can print to stdout.

### DIFF
--- a/builds/install/misc/firebird.conf.in
+++ b/builds/install/misc/firebird.conf.in
@@ -811,6 +811,17 @@
 #
 #EventMemSize = 64K
 
+# ----------------------------
+#
+# File to redirect stdout and stderr output of server
+#
+# Default '/dev/null' for *nix and 'nul' for Windows
+# Empty value or '-' keeps stdout and stderr as is.
+#
+# Type: string
+#
+#OutputRedirectionFile = /dev/null
+
 
 # ===========================
 # Engine Settings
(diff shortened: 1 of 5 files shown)


src/common/os/posix/divorce.cpp
```diff
@@ -113,6 +113,12 @@ void divorce_terminal(int mask)
 		}
 	}
 
+	if (int dev_null = open("/dev/null", O_RDWR) >= 0)
+	{
+		dup2(dev_null, fileno(stdout));
+		dup2(dev_null, fileno(stderr));
+	}
```
AlexPeshkoff: What if dev_null is 1 or 2? What will do dup2(1, 1)? 
If dev_null is not 1 or 2 - should it be closed after dup2 calls?
May be let users put stdout/stderr to some not-null file - for example as a tool to debug UDR?

artyom-smirnov: >What if dev_null is 1 or 2? What will do dup2(1, 1)?
>If dev_null is not 1 or 2 - should it be closed after dup2 calls?

Will fix it.

>May be let users put stdout/stderr to some not-null file - for example as a tool to debug UDR?

Originally this commit intended to fix *nix only problem caused by guardian. If we want to use output redirecting as some debugging tool it should be more unified and work on all platforms and without guardian. May be then server binary should have ability to redirect its output to /dev/null or any other file from config?

artyom-smirnov: @AlexPeshkoff I have updated branch. Please review. I have added config option to allow user define path. Also both classic and super now will behave same way when handling stdout/stderr. Changes are only for posix build for now.

AlexPeshkoff: For posix everything appears good.

PR #145, opened 2018-03-18
CORE-5612: Gradual slowdown compilation (create, recreate or drop) of views

abzalov:


hvlad:
The build failed with
-key size exceeds implementation restriction for index "RDB$INDEX_56"

PS i thought you want to relax limitation on creating user index of system objects...

abzalov:
While 3.0 build was passed.
And the creation of a similar index in 4.0 in the database with the 4K page is created normally.

>>PS i thought you want to relax limitation on creating user index on system objects...
Did not quite understand what is meant.

hvlad:
> While 3.0 build was passed.

I doubt we will accept different solutions for v3 and v4 for the same case

> And the creation of a similar index in 4.0 in the database with the 4K page is created normally.

SQL> create database 's:\temp\i.fdb' page_size 4096;
SQL>
SQL> create table t (
CON>  f1 char(63) character set utf8,
CON>  f2 char(63) character set utf8,
CON>  f3 char(63) character set utf8,
CON>  f4 smallint,
CON>  f5 smallint,
CON>  f6 char(63) character set utf8);
SQL>
SQL> create index t_idx on t (f1, f2, f3, f4, f5, f6);
unsuccessful metadata update
-key size exceeds implementation restriction for index "T_IDX"

>> PS i thought you want to relax limitation on creating user index on system objects...

> Did not quite understand what is meant.

http://tracker.firebirdsql.org/browse/CORE-5746

abzalov:
>>I doubt we will accept different solutions for v3 and v4 for the same case

It's clear.
I thought that this is a regression - I did not see the increase the size of the fields (31 -> 63), and the change of charset (UNICODE_FSS -> UTF8).

>>And the creation of a similar index in 4.0 in the database with the 4K page is created normally.

Tested with UNICODE_FSS charset.
Of course, now we will not go through this restriction, because in fact, increased the size 8 times, while leaving the old limit on the size of the index.

>>PS i thought you want to relax limitation on creating user index on system objects...

I thought to go in parallel ways - to solve the particular and general tasks separately.
But now it turns out that none of them can not be solved without increasing the index limit ...

abzalov:
In version 4.0, there are no plans to use the minimum page size as 8K?
This would be a logical step after changing the encoding of the system tables to UTF8 and increasing the maximum length of the identifiers.

PR #149, opened 2018-03-21
Windows changes for PR #143

artyom-smirnov:
I've added some support for windows similar to *nix changes made in #143.
Changes described in commit d8096a5.
Code have some copy-paste for inet/wnet/xnet implementations. Not sure wrap it in some function or not.

### DIFF
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -1835,6 +1835,7 @@ static int fork(SOCKET old_handle, USHORT flag)
 	string cmdLine;
 	cmdLine.printf("%s -i -h %" HANDLEFORMAT"@%" ULONGFORMAT, name, new_handle, GetCurrentProcessId());
 
+
 	STARTUPINFO start_crud;
 	start_crud.cb = sizeof(STARTUPINFO);
 	start_crud.lpReserved = NULL;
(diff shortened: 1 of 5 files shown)


AlexPeshkoff:
Artyom, I can't say something about this PR. That must come from our win developers.

src/remote/inet.cpp
```diff
@@ -1844,8 +1845,23 @@ static int fork(SOCKET old_handle, USHORT flag)
...
+		{
+			start_crud.hStdOutput = stdout_dup;
+			start_crud.hStdError = stdout_dup;
+		}
+		start_crud.dwFlags |= STARTF_USESTDHANDLES;
+	}
+
 	PROCESS_INFORMATION pi;
-	if (CreateProcess(NULL, cmdLine.begin(), NULL, NULL, FALSE,
+	if (CreateProcess(NULL, cmdLine.begin(), NULL, NULL, TRUE,
```
hvlad: We make child process to not inherit parent handles when fixed CORE-2356.
This patch undo that change.
Not sure it is good.

src/remote/inet.cpp
```diff
@@ -1844,8 +1845,23 @@ static int fork(SOCKET old_handle, USHORT flag)
...
+	HANDLE stdout_handle = GetStdHandle(STD_OUTPUT_HANDLE);
+	if (GetFileType(stdout_handle) == FILE_TYPE_DISK)
+	{
+		if (DuplicateHandle(GetCurrentProcess(), GetStdHandle(STD_OUTPUT_HANDLE), GetCurrentProcess(),
+			&stdout_dup, 0, TRUE, DUPLICATE_SAME_ACCESS))
+		{
+			start_crud.hStdOutput = stdout_dup;
+			start_crud.hStdError = stdout_dup;
+		}
+		start_crud.dwFlags |= STARTF_USESTDHANDLES;
```
hvlad: Should this flag be set if DuplicateHandle failed ?

src/remote/server/os/win32/srvr_w32.cpp
```diff
@@ -252,6 +252,65 @@ int WINAPI WinMain(HINSTANCE hThisInst, HINSTANCE /*hPrevInst*/, LPSTR lpszArgs,
...
+	{
+		//We want emit output to console, just open console for each process
+		//Looks like there is no easy way to redirect output to console from 
+		//child to parent (only by using pipes).
+		if (strcmp(redirection_file, "-") == 0 || stricmp(redirection_file, "CON") == 0)
+		{
+			AllocConsole();
+			AttachConsole(GetCurrentProcessId());
+			freopen("CON", "w", stdout);
+			freopen("CON", "w", stderr);
```
hvlad: Should we check return value ? 
Would it be better to use freopen_s here ?

src/remote/server/os/win32/srvr_w32.cpp
```diff
@@ -252,6 +252,65 @@ int WINAPI WinMain(HINSTANCE hThisInst, HINSTANCE /*hPrevInst*/, LPSTR lpszArgs,
...
+	//Preparations for redirection between child and parent processes
+	const char* redirection_file = Config::getOutputRedirectionFile();
+	HANDLE redirection_handle = INVALID_HANDLE_VALUE;
+	if (redirection_file)
+	{
+		//We want emit output to console, just open console for each process
+		//Looks like there is no easy way to redirect output to console from 
+		//child to parent (only by using pipes).
+		if (strcmp(redirection_file, "-") == 0 || stricmp(redirection_file, "CON") == 0)
+		{
```
hvlad: If run as a service - is it make sense to alloc console ?

PR #150, opened 2018-03-24
CORE-5710 make precision optional for DECFLOAT

mrotteveel:
Fixes [CORE-5710](http://tracker.firebirdsql.org/browse/CORE-5710)

This commit increases the number of reduce/reduce conflicts, but I couldn't find a way to fix that.

### DIFF
--- a/doc/sql.extensions/README.data_types
+++ b/doc/sql.extensions/README.data_types
@@ -116,23 +116,25 @@ DECFLOAT (FB 4.0)
     Alex Peshkoff <peshkoff@mail.ru>
 
   Syntax rules:
+    DECFLOAT
     DECFLOAT(16)
     DECFLOAT(34)
 
   Storage:
-    64-bit / 128-bit, format according to IEEE 754.
+    64-bit / 128-bit, format according to IEEE 754 Decimal64/Decimal128
 
   Example(s):
     1. DECLARE VARIABLE VAR1 DECFLOAT(34);
     2. CREATE TABLE TABLE1 (FIELD1 DECFLOAT(16));
 
   Note(s):
-    1.	A number of standard functions can be used with DECFLOAT datatype. It is:
+    1.	If no precision has been specified in the type declaration, the precision is 34.
+    2.	A number of standard functions can be used with DECFLOAT datatype. It is:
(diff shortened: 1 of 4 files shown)


AlexPeshkoff:
Well - you've noticed it. That's the only reason why I've removed this feature from initial implementation. Such huge growth of conflicts is definitely not good and presence of default precision is not worth it.

asfernandes:
Was this conflicts investigated, i.e., FLOAT and DECIMAL uses the same kind of construction. Are they also adding conflicts? Or is it because DECFLOAT is not a reserved word?

dyemanov:
Given that DECFLOAT is a standard datatype, I suppose we need to define this token as a reserved word.

AlexPeshkoff:
On 03/26/18 13:50, Dmitry Yemanov wrote:
>
> Given that DECFLOAT is a standard datatype, I suppose we need to 
> define this token as a reserved word.
>

This is the simplest solution (and probably long-term correct one). But 
we all remember problems people had when we declared some words reserved.

AlexPeshkoff:
On 03/26/18 13:45, Adriano dos Santos Fernandes wrote:
>
> Was this conflicts investigated, i.e., FLOAT and DECIMAL uses the same 
> kind of construction. Are they also adding conflicts? Or is it because 
> DECFLOAT is not a reserved word?
>

That's because it's not a reserved word. Conflicts happens with words 
also used in column declararations like CHECK, COLLATE, UNIQUE, 
GENERATED, etc.

mrotteveel:
Removing `DECFLOAT` from the `non_reserved_word` reverts the reduce/reduce conflicts back to 17. Personally I don't think it is a problem to make it reserved. I would guess (and hope) this isn't a name that would come up by itself in a normal database. It is also specified in SQL:2016 as a reserved word. Shall I commit my change?

AlexPeshkoff:
On 03/26/18 16:33, Mark Rotteveel wrote:
>
> Removing |DECFLOAT| from the |non_reserved_word| reverts the 
> reduce/reduce conflicts back to 17. Personally I don't think it is a 
> problem to make it reserved. I would guess (and hope) this isn't a 
> name that would come up by itself in a normal database. It is also 
> specified in SQL:2016 as a reserved word. Shall I commit my change?
>

yes

mrotteveel:
Added to `keyword_or_column`

PR #153, opened 2018-03-30
incomplete pull request for mapping Windows groups to Firebird roles

livius2:
Hi,

this is my second attempt to understand Firebird codebase - and it is not an easy task for non C++ user.
It is really incomplete pull request and i need help from someone to bring this to life and 
if someone have time to teach me how to make this better, easier...

It is incomplete because i have finished windows group reads, i put it to writer interface but i do not know how to consume it in the "mapping.cpp"? or somewhere else to map windows group to Firebird role. And especially to not override role provided during login process.

if someone can help or use this code to finish the job i am really appreciated.

### DIFF
--- a/src/auth/trusted/AuthSspi.cpp
+++ b/src/auth/trusted/AuthSspi.cpp
@@ -196,22 +196,30 @@ bool AuthSspi::checkAdminPrivilege(PCtxtHandle phContext, ObjectsArray<string>& 
 	groupNames.clear();
 	// Finally we'll iterate through the list of groups for this access
 	// token looking for a match against the SID we created above.	
+	
 	for (DWORD i = 0; i < ptg->GroupCount; i++)
 	{
 		// consider denied ACE with Administrator SID
 		if (((ptg->Groups[i].Attributes & SE_GROUP_USE_FOR_DENY_ONLY) != SE_GROUP_USE_FOR_DENY_ONLY) && ((ptg->Groups[i].Attributes & SE_GROUP_ENABLED) == SE_GROUP_ENABLED))
 		{
 			DWORD dwSize = 256;
-			if (!LookupAccountSid(NULL, ptg->Groups[i].Sid, groupName, &dwSize, domainName, &dwSize, &snu))
+			if (LookupAccountSid(NULL, ptg->Groups[i].Sid, groupName, &dwSize, domainName, &dwSize, &snu))
 			{
-				return false;
+				string sumName = domainName;
+				sumName += "\\";
+				sumName += groupName;
+
+				groupNames.add(sumName);				
+			}
+			else
+			{
+				// "No mapping between account names and security IDs was done"
+				// some groups mapping between names can not exists and returning from this point is not ok
+				// simply ignore "error" and go to the next group name
+				// return false;					
 			}
 
-			string sumName = domainName;		
-			sumName += "\\";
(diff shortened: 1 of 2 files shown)


src/auth/trusted/AuthSspi.cpp
```diff
@@ -363,7 +394,8 @@ int WinSspiServer::authenticate(Firebird::CheckStatusWrapper* status,
 		{
 			bool wheel = false;
 			string login;
-			sspi.getLogin(login, wheel);
+			Firebird::ObjectsArray<Firebird::string> grNames;
+			sspi.getLogin(login, wheel, grNames);
 			ISC_systemToUtf8(login);
```
AlexPeshkoff: I suppose grNames use system encoding (something like cp1251 in case of Russia)? If yes they should be converted to utf8 like it's done for login.

AlexPeshkoff:
To map windows group to firebird role you should create appropriate mapping. Something like this:
CREATE MAPPING WIN_GROUP USING PLUGIN WIN_SSPI FROM ANY Windows_Group TO ROLE;
That will work if user has single group assigned!

What else worth deciding on my mind:
- should we call that objects Windows_Group or just Group?
- may be it's worth to have both Windows_Group (in a form DOMAIN\GROUP) and Group not containing domain name? That can help to write domain-independent mappings like:
CREATE MAPPING Managers USING PLUGIN WIN_SSPI FROM Group Managers to ROLE Manager;
and so on.

src/auth/trusted/AuthSspi.cpp
livius2: Ok i have fixed this - thank you

livius2:
>That will work if user has single group assigned!

Yes, i have thinked previously about this but i do not see possibility to provide some prioritizing (e.g. sorting and first only taken)
without modification in RDB$AUTH_MAPPING table - or maybe you have some other concept?

>should we call that objects Windows_Group or just Group?

it say what it is, more then only "Group" self.

about second - do you think that this is common multi domain?
I suppose it is rather seldom and in this case someone can create same mapping for another domain.
What do you think?

PS> i can answer with delay because of Easter.

livius2:
Hi @AlexPeshkoff 

>That will work if user has single group assigned!

one question - is `map` in `mapping.cpp` something like Dictionary or it preserve order something like OrderedDictionary?
If it can preserve order :
then not very clean solution but user can specify role name with prefix e.g. 
```
0001_xCorp\SomeUsefullGroup
0002_xCorp\SomeOtherGroup
...
```
it will be optional (as domain name and group name cannot contain number at first place it will be safe and simple to indentity)

And in mapping.cpp for WIN_SSPI plugin will be this number removed 
It will be used only for sorting (prioritizing roles).
there will be added `order by` in fetch from `RDB$AUTH_MAPPING`

in FB4 it will be done clearer by adding field to  `RDB$AUTH_MAPPING`.

If above is not acceptable then maybe some exception should be raised if two or more groups are 
specified something like `Ambiguous role assignment by Windows group mapping. Please specify role name during connection.`

AlexPeshkoff:
I close pull request cause it could not be used directly for some reasons (bad formatting, bad style sometimes), but it was taken as a base for fixing CORE-5639.

livius2:
Great to hear that you have finished mapping rules for Windows groups.
Thank you.

PR #154, opened 2018-03-31
Update keywords for FB4 and check against SQL:2016

mrotteveel:


AlexPeshkoff:
Want to make sure that at the same time when DECFLOAT becomes reserved word according to SQL specification related words like TOTALORDER and some others remain firebird extension.

mrotteveel:
@AlexPeshkoff  Not entirely sure what you mean. `TOTALORDER` is listed under non-reserved words and marked with `*` to indicate it is Firebird-specific.

AlexPeshkoff:
Will try to explain better. As you've said DECFLOAT is reserved word in sql 2016. I wonder - did you check in sql standard for some other _tightly_ related with DECFLOAT words, namely: TOTALORDER, COMPARE_DECFLOAT, NORMALIZE_DECFLOAT & QUANTIZE. That are builtin functions, needed to serve various decfloat-related needs.

mrotteveel:
@AlexPeshkoff I checked. None of those functions are defined in the SQL standard.

asfernandes:
This latest standard seems to be a problem for us. The version I have does not even mention DECFLOAT.

mrotteveel:
@asfernandes I own a copy of SQL:2016-2 (Foundation), I used that to check it. It defines `DECFLOAT`, but it does not define the extra functions we added.

AlexPeshkoff:
It was mentioned (not sure where - may be in devel?) that latest sql 2016 does have DECFLOAT as reserved word. I wonder - if it's really so why do they not add related works as at least non-reserved into specification...

mrotteveel:
@AlexPeshkoff `DECFLOAT` is a reserved word in SQL:2016 (see screenshot below). The reason those related functions aren't, is because those related functions **are not defined** in the standard (maybe DB2 has them too, doesn't mean they are in the standard).

image

PR #162, opened 2018-06-21
Fixed CORE-5823: No permission for SELECT access to blob field in stored procedure.  And CORE-5841: No permission for SELECT access to TABLE PLG$SRP in newer snapshot.

dmitry-starodubov:


src/jrd/tra.cpp
```diff
@@ -3651,11 +3652,45 @@ void jrd_tra::checkBlob(thread_db* tdbb, const bid* blob_id)
...
+
+						if (punt)
+							throw;
+
+						// but someone else has (SP, view)
+						// store Blob ID as allowed in this transaction
+						tra_fetched_blobs.add(*blob_id);
+						tdbb->tdbb_status_vector->clearException();
+					}
+				}
```
AlexPeshkoff: Why not use switch() here instead if / else if ?

src/jrd/tra.cpp
```diff
@@ -3651,11 +3652,45 @@ void jrd_tra::checkBlob(thread_db* tdbb, const bid* blob_id)
...
+					{
+						SCL_check_access(tdbb, s_class, 0, 0, NULL, SCL_select, SCL_object_table, false,
+							blb_relation->rel_name);
+						s_class->scl_blb_access = SecurityClass::BA_SUCCESS;
+					}
+					catch (const Exception&)
+					{
+						if (tdbb->tdbb_status_vector->getErrors()[1] != isc_no_priv)
+							throw;
+
```
hvlad: Should we check catched exception here, not tdbb_status_vector ?

src/jrd/tra.cpp
dmitry-starodubov: Did not notice this when reworked the logic.

src/jrd/tra.cpp
dmitry-starodubov: OK, I will fix it.

hvlad: Thanks, but it seems not enough - few lines below you clear tdbb_status_vector.
It could clear some previous error\warning.
Usually, if code handle exception and not re-throw it, it set temporary status-vector to not overwrite tdbb's one.

AlexPeshkoff:
Does anyone see any problems preventing to merge this PR? I suggest to merge it.

dyemanov:
I don't mind merging it.

PR #164, opened 2018-06-26
Fixed CORE-5861: GRANT OPTION is not checked for new object.

romansimakov:
Checks moved to code from trigger1 and all new objects and checked.
Checks of object's owned is not necceddary because of owner has all required privileges to pass regular checks.
Also now GRANT OPTION can be checked for roles too.


src/dsql/DdlNodes.epp
```diff
@@ -11978,32 +11981,35 @@ void GrantRevokeNode::checkGrantorCanGrantRelation(thread_db* tdbb, jrd_tra* tra
...
-			else
-			{
-				if (fieldName.hasData() && fieldName == PRV.RDB$FIELD_NAME)
-					goFld = 1;
+				if (goFld == -1)
+					goFld = hasGrantOption ? WITH_GRANT_OPTION : 0;
+				else if ((goFld == 0) && hasGrantOption)
+					goFld = WITH_GRANT_OPTION;
 			}
 		}
```
AlexPeshkoff: I do not like an idea of leaving FOR/END_FOR loop using break. This keeps busy resources used by request.

src/dsql/DdlNodes.epp
```diff
@@ -12183,15 +12189,19 @@ void GrantRevokeNode::checkGrantorCanGrantDdl(thread_db* tdbb, jrd_tra* transact
 	{
 		if ( (PRV.RDB$USER_TYPE == obj_sql_role) && !attachment->att_user->roleInUse(tdbb, PRV.RDB$USER))
 			continue;
+
 		if (PRV.RDB$GRANT_OPTION == WITH_GRANT_OPTION)
+		{
 			grantable = true;
```
AlexPeshkoff: I do not like an idea of leaving FOR/END_FOR loop using break. This keeps busy resources used by request.

dyemanov: It was in the past (explicit isc_unwind_request call was needed), it's no longer a problem since FB3.

src/dsql/DdlNodes.epp
romansimakov: AFAIU destructor of AutoCacheRequest should be called both for exit after the last iteration of loop and for some intermediate one. I see no any special handles after END_FOR. Are there releasing resources after the last call EXE_receive? Moreover I see "returns" from loop in functions: isItSqlRole, getIndexRelationName, getTriggerRelationName.

dyemanov (approved):

PR #168, opened 2018-08-30
Fix external engine memory leaks

artyom-smirnov:
Some engine objects never destructing explicitly, but hold resources of external engine which should be released. As I see udr_engine do not affected leaking, but architecture of fbjava prone to leaking if dispose/release never called. I've added straightforward solution to avoid memory leaks from external engine, is it ok? 3.0 have same problem.

### DIFF
--- a/src/jrd/ExtEngineManager.cpp
+++ b/src/jrd/ExtEngineManager.cpp
@@ -887,7 +887,7 @@ ExtEngineManager::Trigger::Trigger(thread_db* tdbb, MemoryPool& pool, ExtEngineM
 
 ExtEngineManager::Trigger::~Trigger()
 {
-	// hvlad: shouldn't we call trigger->dispose() here ?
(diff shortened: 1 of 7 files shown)


src/jrd/ExtEngineManager.cpp
```diff
@@ -888,6 +888,7 @@ ExtEngineManager::Trigger::Trigger(thread_db* tdbb, MemoryPool& pool, ExtEngineM
 ExtEngineManager::Trigger::~Trigger()
 {
 	// hvlad: shouldn't we call trigger->dispose() here ?
```
dyemanov: The comment looks confusing after the change, please remove it and I'll merge this PR.

artyom-smirnov: Done. Squished this fix to first commit.

PR #175, opened 2018-10-28
Fix for returning wrong time after TZ changes

aafemt:


hvlad:
I can't find in documentation if Windows notifies processes when DST status is changed by the system itself. 
Did you test it ?

Does POSIX builds suffers of this issue too ? 
If not - why ? 
If yes - this PR should be completed or explicitly show it is for Windows only.

aafemt:
> Did you test it ?

Yes. The last DST change was yesterday.

> Does POSIX builds suffers of this issue too ?

No. POSIX builds don't use MSVC runtime.

src/remote/server/os/win32/cntl.cpp
```diff
@@ -195,11 +196,16 @@ static void WINAPI control_thread( DWORD action)
...
+		// Refresh RTL static variables
+		_tzset();
 		break;
+
+	default:
+		return ERROR_CALL_NOT_IMPLEMENTED;
 	}
 
-	report_status(state, NO_ERROR, 0, 0);
+	return NO_ERROR;
```
hvlad: Why this change ?

aafemt:
> Why this change ?

Quoting MSDN: "When the control handler function is invoked with a control request, the 
service must call SetServiceStatus to report status to the service control manager only if 
the service status has changed, such as when the service is processing stop or shutdown 
controls. If the service status has not changed, the service should not report status to 
the service control manager."

hvlad:
Very interesting: 
- when i ask you to confirm something that is directly related with PR subject - you ignored important part of my question (about process notification) and give no proof for another part (about usage of MSVC CRT),
- when i ask you about change that have nothing common with PR subject - you show it is correct and refers to MSDN to prove it.
It could be correct, but it is not related with this PR, isn't is ?

aafemt:
> It could be correct, but it is not related with this PR, isn't is ?

It is related. Only HandlerEx can receive TIMECHANGE notification. Rules and signatures 
are different between previously used Handler and newly used HandlerEx. That's why these 
changes are unavoidable.

PS: Do you seriously need confirmation that Linux doesn't use MSVC RTL?..

dyemanov:
It's not about using MSVC RTL, it's about confirmation whether GCC RTL caches the TZ at startup or not (like MSVC RTL does).

aafemt:
AFAIK, CGG on POSIX has no RTL as such. It uses system's LIBC.
I never saw a single report that Linux version has any problem with DST change, only for 
Windows they were.

aafemt:
How did you try to reproduce it when DST change time is only twice a year and autumn one 
is already passed?

hvlad:
Run firebird, attach and query both CURRENT_TIMESTAMP and NOW, ensure it is equal.
Change system time to be 5 sec less than DST change time.
Query both CURRENT_TIMESTAMP and NOW, ensure it is equal and the same as system clock show.
Wait 5 sec and ensure system clock changed and DST change happens (system time shifted accordingly).
Query both CURRENT_TIMESTAMP and NOW, ensure it is equal and the same as system clock show.
Repeat few times for both DST change events - spring and autumn.
Firebird always show the same timestamp value as i see at system clock.
Firebird is not restared and the same connection is used.

hvlad:
I tried FB3 built with VC10 and official FB 2.5.0 (build with VC8).
No i didn't change TZ manually as it is not related to the "bug with DST" we talk about.

aafemt:
> No i didn't change TZ manually as it is not related to the "bug with DST" we talk about.

Ok, I'll change the header.

hvlad:
You may change the header, but what did you test when DST was changed ? ;)

aafemt:
> I tried FB3 built with VC10 and official FB 2.5.0 (build with VC8).

https://groups.yahoo.com/neo/groups/firebird-support/conversations/topics/133211 mentioned that the issue happen with Firebird 3. Perhaps, it is also Windows version-depended.

hvlad:
I suggest to create corresponding ticket and then we could accept the PR

aafemt:
CORE-5959

hvlad:
I still not merge this PR because it have one weakness - it not works with embedded.

So, i have a question - why we use CRT  (ftime() + localtime()) but not WinAPI (GetLocalTime() or a bit more complex GetSystemTimeAsFileTime() etc) here (at NoThrowTimeStamp::getCurrentTimeStamp()) ?

Does anybody knows the reason to not use WinAPI ?

aafemt:
So far nobody complained about this issue in embedded mode. Perhaps, it is because 
application that use it don't work continuously.

hvlad:
> So far nobody complained about this issue in embedded mode

This is not an excuse for the partial fix

AlexPeshkoff:
On 11/06/18 22:01, Vlad Khorsun wrote:
>
> I still not merge this PR because it have one weakness - it not works 
> with embedded.
>
> So, i have a question - why we use CRT (ftime() + localtime()) but not 
> WinAPI (GetLocalTime() or a bit more complex GetSystemTimeAsFileTime() 
> etc) here (at NoThrowTimeStamp::getCurrentTimeStamp()) ?
>
> Does anybody knows the reason to not use WinAPI ?
>

IMHO should be rewritten if that helps avoid issues with msvcrt.

hvlad:
Fix is backported into B3_0_Release, PR is closed

PR #176, opened 2018-11-05
When creation of trace log file fails, there is no error message in firebird.log

ilya071294:


src/utilities/ntrace/TracePluginImpl.cpp
```diff
@@ -1907,6 +1910,24 @@ bool TracePluginImpl::checkServiceFilter(ITraceServiceConnection* service, bool
...
+#else
+	char buff[256];
+	strerror_r(errno, buff, sizeof(buff));
+	strErr = buff;
+#endif
+	fatal_exception::raiseFmt("TracePluginImpl: operation \"%s\" failed on file \"%s\". Error is : %s",
+		operation, config.log_filename.c_str(), strErr);
+}
+
+
```
hvlad: Why fatal_exception ? It kill whole Firebird process, iirc. Too much, i think.
Why not OS native error code and corresponding exception ?
Explain, please: how write failure is related with "creation of trace log file fails" ?

ilya071294: The exception is caught in TraceFactoryImpl::trace_create(...), like in FB 2.5. Trace log file is created during the first write attempt.

PR #179, opened 2018-11-29
Reload firebird.conf after spawning CS server process.

artyom-smirnov:
In pre 3.0 era classic had sometimes useful ability: it not required restart after config changes (because it was spawned by xinetd). Now if server's child process was spawned by parent firebird process config will not be reloaded, because it loaded once and inherited from parent. How about reloading it forcefuly?

This is partial solution: ServerMode and UdfAccess will not be updated,
because they cached (actually we do not need to change ServerMode, but
change UdfAccess will be useful).

### DIFF
--- a/src/common/config/config.cpp
+++ b/src/common/config/config.cpp
@@ -79,14 +79,10 @@ public:
 		}
 	}
 
-	/***
-	It was a kind of getting ready for changing config remotely...
-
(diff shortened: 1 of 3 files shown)


aafemt:
You'd better reload it on demand only as it is done for databases.conf.

PR #181, opened 2018-11-30
Cryptographic functions according to CORE-5970

AlexPeshkoff:


src/dsql/parse.y
```diff
@@ -7985,13 +8010,110 @@ system_function_special_syntax
...
+%type <metaNamePtr> opt_mode
+opt_mode
+	: MODE valid_symbol_name
+		{ $$ = $2; }
+	| /* nothing */
+		{ $$ = newNode<MetaName>(""); }
+	;
+
+%type <valueExprNode> opt_iv
+opt_iv
```
asfernandes: @AlexPeshkoff I prefer to see that `opt_*` things prefixed with something like `crypt_`

src/dsql/parse.y
```diff
@@ -8682,6 +8804,21 @@ non_reserved_word
 	| TOTALORDER
 	| TRAPS
 	| ZONE
+	| MODE				// crypt functions
```
asfernandes: @AlexPeshkoff on this area every keyword should be listed alphabetically for the FB version in general, not per feature.

src/jrd/SysFunction.cpp
```diff
@@ -203,6 +218,10 @@ void makeBin(DataTypeUtilBase* dataTypeUtil, const SysFunction* function, dsc* r
 void makeBinShift(DataTypeUtilBase* dataTypeUtil, const SysFunction* function, dsc* result, int argsCount, const dsc** args);
 void makeCeilFloor(DataTypeUtilBase* dataTypeUtil, const SysFunction* function, dsc* result, int argsCount, const dsc** args);
 void makeDateAdd(DataTypeUtilBase* dataTypeUtil, const SysFunction* function, dsc* result, int argsCount, const dsc** args);
+void makeEncrypt(DataTypeUtilBase* dataTypeUtil, const SysFunction* function, dsc* result, int argsCount, const dsc** args);
```
asfernandes: @AlexPeshkoff every other function in SysFunction.cpp is listed generally in alphabetical order, not per feature.

PR #187, opened 2019-02-11
If stdin is FILE_TYPE_PIPE then we should treat it as non-interactive

artyom-smirnov:
With current check isql always in interactive mode, even when input is piped (like echo connect 'test'; | isql.exe ...)

### DIFF
--- a/src/isql/isql.epp
+++ b/src/isql/isql.epp
@@ -8753,7 +8753,7 @@ static bool stdin_redirected()
 #ifdef WIN_NT
 	HANDLE in = GetStdHandle(STD_INPUT_HANDLE);
 	const DWORD file_type = GetFileType(in);
-	if (file_type == FILE_TYPE_CHAR || file_type == FILE_TYPE_PIPE)
+	if (file_type == FILE_TYPE_CHAR)
 		return false;
 #else
 	if (isatty(fileno(stdin)))


aafemt:
Fine. You have just broke every frontend for isql that use pipe redirection including MSYS.

asfernandes:
> Fine. You have just broke every frontend for isql that use pipe redirection including MSYS.

You mean, evey broken software that didn't fixed themself in decades.

Windows terminal is working fine, as well cmd.

aafemt:
Of course they do. They use console, not pipes.

asfernandes:
This issue was been discussed in fb-devel thread "Stdin redirection with pipe in Windows ISQL", where last message was more than a year ago where I asked for objections. You had more than one year for that.

I also showed in this topic that ISQL in MSYS was very broken in general.

Nobody is using Firebird v6 and it's not going to be released soon.

Let the ecosystem do they necessary work.

Fact is, ISQL was bugged in Windows and there is no reason to let it that for so long time while the others parts want to be lazy and not do their jobs correctly.

hvlad:
If issue described by @aafemt really exists, we could introduce a way to explicitly control interactive mode.
It could be a command-line switch, SET command or even environment variable to use with legacy apps that can't be modified.
BTW, sample of such broken app also will not harm.

dyemanov:
I already mentioned in the past that GLDS has such an app (GUI frontend for ISQL based on pipes) and distributes it to their clients.

AlexPeshkoff:
On 11/24/23 15:01, Vlad Khorsun wrote:
>
> If issue described by @aafemt <https://github.com/aafemt> really 
> exists, we could introduce a way to explicitly control interactive mode.
> It could be a command-line switch, SET command or even environment 
> variable to use with legacy apps that can't be modified.
>

Agreed, that to be fixed somehow.

asfernandes:
I see no problem in add workaround and in this case looks like environment variable would be the way, as we are talking about software that maintainers is not going to change.

asfernandes:
> I already mentioned in the past that GLDS has such an app (GUI frontend for ISQL based on pipes) and distributes it to their clients.

I know and I asked after it about objections about this PR. At same time we have bug. And workaround like is being discussed now is possible.

PR #188, opened 2019-02-12
Improved module search

aafemt:
1) It can find not only "module" and "libmodule.so", but also "module.so".
2) No second attempt if doctor made no changes in name.

### DIFF
--- a/src/common/os/posix/mod_loader.cpp
+++ b/src/common/os/posix/mod_loader.cpp
@@ -69,25 +69,39 @@ bool ModuleLoader::isLoadableModule(const Firebird::PathName& module)
 	return true;
 }
 
-void ModuleLoader::doctorModuleExtension(Firebird::PathName& name)
+bool ModuleLoader::doctorModuleExtension(Firebird::PathName& name, int& step)
 {
 	if (name.isEmpty())
-		return;
+		return false;
 
-	Firebird::PathName::size_type pos = name.rfind("." SHRLIB_EXT);
-
-	if (pos != name.length() - 3)
+	switch (step++)
 	{
-		pos = name.rfind("." SHRLIB_EXT ".");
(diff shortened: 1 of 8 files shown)


src/common/os/posix/mod_loader.cpp
```diff
@@ -69,25 +69,39 @@ bool ModuleLoader::isLoadableModule(const Firebird::PathName& module)
 	return true;
 }
 
-void ModuleLoader::doctorModuleExtension(Firebird::PathName& name)
+bool ModuleLoader::doctorModuleExtension(Firebird::PathName& name, int& step)
```
AlexPeshkoff: I like an idea of patch but would not it be better to use
class ModuleName : public PathName {.......
containing step number instead adding artificial parameter to doctorModuleExtension function which must be in sync with module's PathName?

aafemt:
> I like an idea of patch but would not it be better to use
> class ModuleName : public PathName {.......
> containing step number instead adding artificial parameter to doctorModuleExtension 
> function which must be in sync with module's PathName?

No. It would make only troubles and no advatages. I won't jump into this trap again.

AlexPeshkoff:
One advantage is obvious - related data is stored in single data structure and there is no chance to mix unrelated module name and doctor's step.
What troubles do you see?

aafemt:
fixAndLoad() should return fixed name. Here are two possibilities:
1) A tiny class with single member "step" and reference/pointer to PathName which is used in exactly two places. Too little result for such effort: this change will increase amount of code without increasing of functionality.
2) Snowball of type changes starting from fixAndLoad's parameter rolling over whole call stack. This gocha I already had with Unicode branch, so - no, thanks.

AlexPeshkoff:
OK, let's try to choose minimum evil

PR #189, opened 2019-02-19
SQL control of table's replication

aafemt:


aafemt:
Why CI tests disappeared from here?

dyemanov:
I've committed my own implementation to the replication-ddl branch. I'm open for discussions.

PR #192, opened 2019-02-21
Backported CORE-6004: Add a switch to disable the "TCP Loopback Fast …

KarloX2:
…Path" option

### DIFF
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -481,7 +481,7 @@ static int		send_partial(rem_port*, PACKET *);
 
 static int		xdrinet_create(XDR*, rem_port*, UCHAR *, USHORT, enum xdr_op);
 static bool		setNoNagleOption(rem_port*);
-static bool		setFastLoopbackOption(rem_port*, SOCKET s = 0);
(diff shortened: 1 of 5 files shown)


src/remote/inet.cpp
```diff
@@ -1486,10 +1486,11 @@ static rem_port* aux_connect(rem_port* port, PACKET* packet)
 		port->auxAcceptError(packet);
 		inet_error(false, port, "socket", isc_net_event_connect_err, savedError);
 	}
+	new_port->port_handle = n;
```
hvlad: It seems not fully correct to assign not connected socket here.

KarloX2: Why not? n has the new port handle here. In the previous version, n was assigned to new_port->port_handle just a few lines below, I just moved it up so the setFastLoopbackOption() method can already use the port_handle value. If the method runs into an exception e.g. on connect, the new_port object won't be returned to the caller anyway.

hvlad: A lot of reasons: in case of error, `n` is closed but `port_handle` will still point to the incorrect socket. Another thread could allocate new socket and it could get same value. Also, `new_port` is still linked with parent port.

KarloX2: I see your point.

PR #195, opened 2019-02-25
Windows build with msbuild to also work with VS Build Tools.

asfernandes:
Tested: VS 2017 / VS 2017 Build Tools (in same machine with VS 2017 installed).

That pull request has a problem that needs to be investigated.

Incremental build is not working well.

That may be due to non-existing (headers?) files referenced or something else in the project files that I wan't able to fix.

My tlogs shows a dependency against `C:\WINDOWS\SYSTEM32\TZRES.DLL`. I found this https://developercommunity.visualstudio.com/content/problem/159331/up-to-date-projects-compiled-again-because-of-tzre.html but my `tzres.dll` does not have a future timestamp.

Update: the problem was due to using the Windows short path for the solution file name.

### DIFF
--- a/builds/win32/compile.bat
+++ b/builds/win32/compile.bat
@@ -8,6 +8,7 @@
 
 setlocal
 set solution=%1
+set output_path=%~dp2
 set output=%2
 set projects=
 
@@ -17,14 +18,6 @@ set projects=
 	set config=debug
 )
 
-set config="%config%|%FB_TARGET_PLATFORM%"
-
-if "%VS_VER_EXPRESS%"=="1" (
-	set exec=vcexpress
-) else (
-	set exec=devenv
-)
-
 shift
 shift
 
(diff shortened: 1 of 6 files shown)


asfernandes:
Travis now supports Windows with Visual Studio 2017 Build Tools, so once this branch is integrated we could hope for faster automated Windows builds.

hvlad:
It seems to work (VS2017 build is running at my machine currently) but i not see it is faster. 
And i wonder - why it should be faster if compiler and linker is the same ?

asfernandes:
The thing about faster is about automated builds. Travis is generally faster than AppVeyor. But Linux build is faster than Windows build anyway, so the assumption may not be really valid.

But build with msbuild allows to build with the command line build tools, without Visual Studio installed, so that is the primary improvement.

hvlad:
Ok, understand now. No objection from my side.

dyemanov (approved):
Fine with me.

reevespaul:
run_all.bat seems to still work OK, at least up until make_examples.bat

Unfortunately I then get this error:

  Building employee.fdb
  creating database C:/FB40/CLEAN_~1/FIREBI~1/gen/examples/employee.fdb
  Turning forced writes off
  Creating tables
  Turning  off indices and triggers
  Loading  column data
  Turning  on indices and triggers
  Dynamic SQL Error
  -SQL error code = -104
  -Token unknown - line 1, column 17
  -role_tree

    Error  - empbuild.exe failed - see make_examples.log for details

However, I don't think this is related to the changes to the build
process.


Paul
-- 
Paul Reeves
http://www.ibphoenix.com
Supporting users of Firebird

PR #196, opened 2019-03-11
Fixed CORE-5892: SQL SECURITY DEFINER context is not properly evaluated for monitoring tables

romansimakov:
Now we take into account the call hierarchy when use SQL SECURITY option.
Added new context variable in SYSTEM namespace - EFFICIENT_USER which is
returns user name in which context a code works.
We change efficient user before call procedure and function, fetch a
record from selective procedure and before execute a trigger.


reevespaul:
What exactly does EFFICIENT_USER mean? Is it documented somewhere? The name ought to be self explanatory and I'm having trouble understanding  the reason behing choosing it.

asfernandes:
Is `EFFICIENT_USER` standard? I found it weird. Don't `ACTUAL_USER` would be a better name?

dyemanov:
I suspect it should have been named EFFECTIVE rather than EFFICIENT.

AlexPeshkoff:
On 3/11/19 6:18 PM, Dmitry Yemanov wrote:
>
> I suspect it should have been named EFFECTIVE rather than EFFICIENT.
>

You are right :-)
It was my suggestion and I just misspelled.

But Roman - adding a few lines to appropriate README is definitely good 
idea.

aafemt:
User A created procedure A with SECURITY DEFINER which calls procedure B created by user B with SECURITY DEFINER which updates a table created by user C with trigger created by user D. Which user is effective in the trigger?

I see no point in this system variable.

AlexPeshkoff:
On 3/11/19 6:23 PM, Dimitry Sibiryakov wrote:
>
> User A created procedure A with SECURITY DEFINER which calls procedure 
> B created by user B with SECURITY DEFINER which updates a table 
> created by user C with trigger created by user D. Which user is 
> effective in the trigger?
>

Certainly D. It's not cumulative.

aafemt:
> Certainly D. It's not cumulative.

Even if the trigger was created with default SECURITY INVOKER? In this case the variable has even less meaning because it is effectively CURRENT_OBJECT_OWNER that can be simple hardcoded literal.

AlexPeshkoff:
From your question I've understood that you talk about SECURITY DEFINER. What about SQL SECURITY INVOKER - in this mode EFFECTIVE_USER == CURRENT_USER. What a problem?

aafemt:
As I said: this is the same level as "how to know which table a trigger is called for". If a routine was created with SECURITY DEFINER then EFFECTIVE_USER is known at the moment of creation and requires no context variable to be found.

romansimakov:
Yes. It should be EFFECTIVE_USER. ACTUAL_USER is also good for me. EFFECTIVE word is from Unix. I agreed with Alex to use EFFECTIVE. Any ideas how to resolve the collision? :)

romansimakov:
> User A created procedure A with SECURITY DEFINER which calls procedure B created by user B with SECURITY DEFINER which updates a table created by user C with trigger created by user D. Which user is effective in the trigger?
> 
> I see no point in this system variable.

A table as well as the trigger of a table can have their own SQL SECURITY. In your example you do not tell about SQL SECURITY of table and trigger. In this case EFFECTIVE_USER should be user B.

AlexPeshkoff:
On 3/11/19 6:47 PM, Dimitry Sibiryakov wrote:
>
> As I said: this is the same level as "how to know which table a 
> trigger is called for". If a routine was created with SECURITY DEFINER 
> then EFFECTIVE_USER is known at the moment of creation and requires no 
> context variable to be found.
>

Yes, one can live without that context variable, but I've agreed with 
Roman that it makes code better in a case when SQL security is changed 
_after_ writing procedure/trigger code. Certainly code review solves a 
problem but if in some aspect we can avoid it - why not?

aafemt:
> In this case EFFECTIVE_USER should be user B.

You said B, Alex said CURRENT_USER. Which really?

asfernandes:
> Yes. It should be EFFECTIVE_USER. ACTUAL_USER is also good for me. EFFECTIVE word is from Unix. I agreed with Alex to use EFFECTIVE. Any ideas how to resolve the collision? :)

No problem for me with `EFFECTIVE_USER`

romansimakov:
> > In this case EFFECTIVE_USER should be user B.
> 
> You said B, Alex said CURRENT_USER. Which really?

I guess he does not take into account that both table and trigger may have SQL SECURITY but your example leave them by default.

More simple case may show the need of context variable. If procedure P_DEF defined by SYSDBA with SQL SECURITY DEFINER calls a procedure P_INV defined by default or with SQL SECURITY INVOKER you cannot know at definition time who is effective caller (invoker). In ths case the effective caller will SYSDBA. But in case you will call P_INV directly it will be CURRENT_USER. This rule works for all objects that handle SQL SECURITY.

romansimakov:
> > Yes. It should be EFFECTIVE_USER. ACTUAL_USER is also good for me. EFFECTIVE word is from Unix. I agreed with Alex to use EFFECTIVE. Any ideas how to resolve the collision? :)
> 
> No problem for me with `EFFECTIVE_USER`

Good. I'll fix the typo tomorrow.

romansimakov:
> > > In this case EFFECTIVE_USER should be user B.
> > 
> > 
> > You said B, Alex said CURRENT_USER. Which really?

```
roman:bin$ rm -f /tmp/z.fdb; ./isql -q -i ~/a
create database 'localhost:/tmp/z.fdb';

create or alter user a password 'a';
create or alter user b password 'b';
create or alter user c password 'c';
create or alter user d password 'd';
commit;

grant create table to c;
grant alter any table to d;
grant create procedure to a;
grant create procedure to b;
commit;

connect 'localhost:/tmp/z.fdb' user 'c' password 'c';

create table t(invoker varchar(32), cur varchar(32), eff varchar(32));
grant select, insert on t to a;
grant select, insert on t to b;
commit;

set term ^;

connect 'localhost:/tmp/z.fdb' user 'd' password 'd'^

create trigger trig for t before insert
as
declare variable n integer;
begin
  if (NEW.eff is NULL) then
  begin
    NEW.eff = RDB$GET_CONTEXT('SYSTEM', 'EFFECTIVE_USER');
    NEW.cur = RDB$GET_CONTEXT('SYSTEM', 'CURRENT_USER');
  end
end^
commit^

connect 'localhost:/tmp/z.fdb' user 'b' password 'b'^

create procedure p_b (invoker varchar(32)) sql security definer
as
begin
  insert into t(invoker) values('trigger from p_b from ' || :invoker);  -- to run trigger
  insert into t values('p_b from ' || :invoker, RDB$GET_CONTEXT('SYSTEM', 'CURRENT_USER'), RDB$GET_CONTEXT('SYSTEM', 'EFFECTIVE_USER'));
end^
grant execute on procedure p_b to a^
commit^

connect 'localhost:/tmp/z.fdb' user 'a' password 'a'^

create procedure p_a sql security definer
as
begin
  insert into t(invoker) values('trigger from p_a');  -- to run trigger
  insert into t values('p_a', RDB$GET_CONTEXT('SYSTEM', 'CURRENT_USER'), RDB$GET_CONTEXT('SYSTEM', 'EFFECTIVE_USER'));
  execute procedure p_b('p_a');
end^

set term ;^

commit;

execute procedure p_a;
select * from t;

INVOKER                          CUR                              EFF                              
================================ ================================ ================================ 
trigger from p_a                 A                                A                                
p_a                              A                                A                                
trigger from p_b from p_a        A                                B                                
p_b from p_a                     A                                B                                


```

romansimakov:
I've committed fixes.

AlexPeshkoff:
On 3/12/19 5:12 PM, Roman Simakov wrote:
>
>             In this case EFFECTIVE_USER should be user B.
>
>         You said B, Alex said CURRENT_USER. Which really?
>

I was wrong here, hurried a bit. In fact - when SQL security is set to 
caller and caller is some procedure/trigger where effective user is B, 
correct answer is B.

PR #197, opened 2019-04-17
Extract VS version from VisualStudioVersion variable

artyom-smirnov:
This simplify version extraction and fix it on VS with non-english language packs.


asfernandes:
We still have usage of `devenv` in `examples/build_win32/setenvvar.bat`, but not in `builds/win32/setenvvar.bat` anymore.

PR #198, opened 2019-04-18
Build fixes.

egorpugin:
@asfernandes

### DIFF
--- a/src/common/os/win32/mod_loader.cpp
+++ b/src/common/os/win32/mod_loader.cpp
@@ -101,7 +101,7 @@ public:
 				"msvcr110.dll",
 #elif _MSC_VER == 1800
 					"msvcr120.dll",
-#elif _MSC_VER >= 1900 && _MSC_VER < 1920
+#elif _MSC_VER >= 1900 && _MSC_VER < 1930
 					"vcruntime140.dll",
 #else
                     #error Specify CRT DLL name here !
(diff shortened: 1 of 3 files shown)


src/common/os/win32/mod_loader.cpp
```diff
@@ -101,7 +101,7 @@ class ContextActivator
 				"msvcr110.dll",
 #elif _MSC_VER == 1800
 					"msvcr120.dll",
-#elif _MSC_VER >= 1900 && _MSC_VER < 1920
+#elif _MSC_VER >= 1900 && _MSC_VER < 1930
```
asfernandes: @egorpugin isn't VS 2019 _MSC_VER 1920? In this case the correct should be `<= 1920`.

egorpugin: Like before we set considerable interval here to not send patches every update.
VS2019 will have frequent updates with 1921, 1922 etc. numbers.

VS2017 has >=1910 <1920 or <= 1916.
VS2019 will have >= 1920.

asfernandes: That may be ok for me then.

I just wan't to point out that when I tested Firebird with VS 2019 few weeks ago, even when not retargeting the projects to use the new compiler (i.e., using the old compiler) `_MSC_VER` was `1920`. That was interesting for me.

PR #204, opened 2019-05-27
Core 4029

AntonNIKIT:
Made Core-4029: Add build in function BIN_BIT, BIN_BIT_STR, BIN_TO_STR and BIN_TO_STR_LIST


aafemt:
Could you also add some documentation about these functions: purpose, parameters, result etc.

src/jrd/SysFunction.cpp
```diff
@@ -5991,12 +5991,222 @@ dsc* evlSystemPrivilege(thread_db* tdbb, const SysFunction*, const NestValueArra
...
+
+	impure->vlu_misc.vlu_int64 = bit;
+	impure->vlu_desc.makeInt64(0, &impure->vlu_misc.vlu_int64);
+	return &impure->vlu_desc;
+}
+
+void makeBIN_BIT_STR(DataTypeUtilBase*, const SysFunction* function, dsc* result,
+	int argsCount, const dsc** args)
+{
+	(argsCount == function->minArgCount);
```
aafemt: Isn't something missed i this line?

asfernandes:
Why you changed `extern/icu/icudt63l.dat`?

asfernandes:
Please don't invent new code conventions. Almost every source part you changed you invented a new one (way to space `*` in pointers, upper case and underline in SysFunctions identifiers, alphabetical order of things, spaces after braces, etc).

asfernandes:
Please verify Linux build logs. It's broken.

AntonNIKIT:
Platform independent, starts with the least bit.

livius2:
@AntonNIKIT thank you for implementing this functions.
This is also good overview how to write custom function for Firebird in the source code.

May i ask about implementation of `evlBIN_TO_STR_LIST`, is there same separator for input and output value list? In my example in CORE-4029 i have 2 different separators ";' for input and definable output separator list. 
I use always ";" as input separator but other users can use different one. Do you think that this will be good to add one more parameter to this function and tell explicitly what is input separator list?

livius2:
Any update about merge this pull request?

asfernandes:
> Any update about merge this pull request?

The code has many problems.

More important, IMO I don't think these functions should exist in Firebird - and they names are weird for worldwide usage.

They may be easily be implemented as PSQL or UDR for who need them.

I would close this and https://github.com/FirebirdSQL/firebird/issues/4359.

livius2:
The names could be dapted but why do you think that they should not be in the Firebird? As developer i can say more functions is better.
I like to have as many as possible to do not start from scratch in new project, but i do not decide here.
And bit operations are common in the scientific usage of Firebird.

asfernandes:
Please name a program language or DBMS that come with these functions.

livius2:
??

dyemanov:
Adriano means that these functions (under any name) seems to be missing in all programming languages, as well as in all popular DBMS. It implies they're not so important. Given we already have bitwise AND/OR/XOR and both shifts, every other operation becomes easy to implement by hands.

livius2:
@dyemanov More understandable :)

PR #206, opened 2019-05-29
SHA2 Refactoring [B3_0].

ghost:
General normalization of class structures.

### DIFF
--- a/src/common/sha2/sha2.cpp
+++ b/src/common/sha2/sha2.cpp
@@ -252,8 +252,7 @@ void sha256_ctx::transf(const unsigned char *message,
 
 		for (j = 0; j < 64; j++)
 		{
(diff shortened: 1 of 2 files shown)


ibprovider:
I vote to remove test code from sha2.h(cpp).

1. They will saved in git history.
2. IBProvider tests has the verifications for SRP224-SRP512
3. I think, Jaybird also has these verification

Of course, in future these tests need restore in normal module test for FB source code.

ibprovider:
get_BlockSize methods does not used in code.

Can I remove they?

AlexPeshkoff:
Certainly - please do remove unused code

ibprovider:
Hello,

I builded Firebird with these changes and test again (SRP224-SRP512) - all works as expected.

What about removing of garbage (test) from sha2.h(cpp)?

It will allows create more clean code.

PR #209, opened 2019-05-30
Reorganization of code to reduce the number of compiler warnings (VC++) [B3_0]

ghost:
Example of warnings:
 alloc.h(292): warning C4595: 'operator new': non-member operator new or delete functions may not be declared inline
 alloc.h(296): warning C4595: 'operator new[]': non-member operator new or delete functions may not be declared inline
 alloc.h(310): warning C4595: 'operator delete': non-member operator new or delete functions may not be declared inline
 alloc.h(314): warning C4595: 'operator delete[]': non-member operator new or delete functions may not be declared inline

### DIFF
--- a/src/common/classes/alloc.cpp
+++ b/src/common/classes/alloc.cpp
@@ -2870,6 +2870,70 @@ void AutoStorage::ProbeStack() const
 
 } // namespace Firebird
 
+void* operator new(size_t s ALLOC_PARAMS) throw (OOM_EXCEPTION)
+{
+	return MemoryPool::globalAlloc(s ALLOC_PASS_ARGS);
+}
+
+void* operator new[](size_t s ALLOC_PARAMS) throw (OOM_EXCEPTION)
+{
+	return MemoryPool::globalAlloc(s ALLOC_PASS_ARGS);
+}
+
+void* operator new(size_t s, Firebird::MemoryPool& pool ALLOC_PARAMS) throw (OOM_EXCEPTION)
+{
+	return pool.allocate(s ALLOC_PASS_ARGS);
+}
+
+void* operator new[](size_t s, Firebird::MemoryPool& pool ALLOC_PARAMS) throw (OOM_EXCEPTION)
+{
+	return pool.allocate(s ALLOC_PASS_ARGS);
+}
+
+void operator delete(void* mem ALLOC_PARAMS) throw()
+{
+	MemoryPool::globalFree(mem);
+}
+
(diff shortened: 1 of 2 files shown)


AlexPeshkoff:
That changes are already present in HEAD, as for me I see no big reasons to apply them to FB3. At least Mac build should be checked with them in order not to break it.

ibprovider:
I asked to make these changes (in FB3) three or four years ago.

hvlad:
Sorry for been too late, @AlexPeshkoff - we still have a LOT of this warnings in DEBUG build:

>fb40\src\remote\../common/classes/alloc.h(404): warning C4595: 'operator delete': non-member operator new or delete functions may not be declared inline (compiling source file ..\..\..\src\remote\inet.cpp)

>fb40\src\remote\../common/classes/alloc.h(408): warning C4595: 'operator delete[]': non-member operator new or delete functions may not be declared inline (compiling source file ..\..\..\src\remote\inet.cpp)

AlexPeshkoff:
Vlad, I do not see big reasons to deal with warnings in non-HEAD branches. On the other hand I have no problems if you fix them. Mac build is always checked now i.e. this problem can be easily solved.

hvlad:
Sorry again, did not notice it was about v3. But I talked about the master

AlexPeshkoff:
In master it's fixed

hvlad:
Please, read **compiler** messages above. DEBUG build is still not fixed

AlexPeshkoff:
Vlad, I see that messages. But I do not understand - why they begin just from line 404, not earlier? Or may be the sample is incomplete?

hvlad:
```
#ifdef DEBUG_GDS_ALLOC

#ifdef __clang__
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Winline-new-delete"
#endif

inline void operator delete(void* mem) FB_NOTHROW
{
	MemoryPool::globalFree(mem);
}
inline void operator delete[](void* mem) FB_NOTHROW
{
	MemoryPool::globalFree(mem);
}

#ifdef __clang__
#pragma clang diagnostic pop
#endif

#endif // DEBUG_GDS_ALLOC

```
You see - there are two inline operators left. I tried to comment it out and build is OK.

AlexPeshkoff:
Take a look at the code 30 lines before this.
Why no warnings on that inlines?

AlexPeshkoff:
Feel free to convert them to extern with body in alloc.cpp.

hvlad:
Done

PR #210, opened 2019-06-25
Allow to load system ICU (when shared modules names have no version t…

hvlad:
Allow to load system ICU (when shared modules names have no version tag).
Refactor ICU load\initialization code a bit to avoid code duplication and dumb try\catch.
Change ICU versions search range to avoid looking for never existing versions and significantly reduce search time.
Now ICU search order is: preferred version (63), system-wide version, 79-49, 4.8-3.0
Improve error reporting a bit - include ICU version number that failed to load or init.

### DIFF
--- a/src/common/unicode_util.cpp
+++ b/src/common/unicode_util.cpp
@@ -1168,7 +1168,7 @@ UnicodeUtil::ConversionICU& UnicodeUtil::getConversionICU()
 
 	// Try "favorite" (distributed on windows) version first
 	const int favMaj = 63;
-	const int favMin = 1;
+	const int favMin = 0;
 	try
 	{
 		if ((convIcu = ImplementConversionICU::create(favMaj, favMin)))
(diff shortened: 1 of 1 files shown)


asfernandes:
What is "system ICU (when shared modules names have no version tag)"? Is ICU natively distributed with Windows 10?

hvlad:
http://userguide.icu-project.org/design#TOC-ICU-Binary-Compatibility:-Using-ICU-as-an-Operating-System-Level-Library

hvlad:
And yes, ICU is distributed with Windows 10 Creators Update: 
https://docs.microsoft.com/en-us/windows/desktop/intl/international-components-for-unicode--icu-

I found version 59 in SxS and System32 folders, all built as system-level library.
Resources is present in one instance (icudtl.dat) at separate folder

asfernandes (approved):

AlexPeshkoff (approved):

PR #214, opened 2019-07-06
 Fixed CORE-6055: Xnet crypt

tonal:


AlexPeshkoff:
@tonal I wonder with what plugin did you test your code.

tonal:
@AlexPeshkoff The plugin is our internal development.
But the code has not yet been tested. Only the syntax correctness of copy-paste was checked.

AlexPeshkoff (changes requested):
Not tested... That explains everything. That's bad idea to submit patch requests with untested changes.

Unfortunately this would not work. Yep, this change is required part of making xnet pass db crypt keys over the wire but that is not enough - one should also learn xnet to handle async operations in async way (like TCP does). And that's definitely more complex task.

tonal:
see #6305

PR #218, opened 2019-07-27
CORE-6109 SQL standard FLOAT type syntax support

mrotteveel:
See also http://tracker.firebirdsql.org/browse/CORE-6109

### DIFF
--- a/src/dsql/parse.y
+++ b/src/dsql/parse.y
@@ -5095,11 +5095,19 @@ decimal_keyword
 
 %type <legacyField> float_type
 float_type
-	: FLOAT precision_opt
+	: FLOAT precision_opt_nz
 		{
+		    // Precision is binary digits of the significand: 1-24 for 32 bit single precision, 25-53 for 64 bit double precision
+			// Precision 0 is the 'no precision specified' case, which defaults to 32 bit single precision
+			SLONG precision = $2;
+
+			if (precision != 0 && (precision < 1 || precision > 53))
+				yyabandon(YYPOSNARG(2), -842, Arg::Gds(isc_precision_err2) << Arg::Num(1) << Arg::Num(53));
+																// Precision must be between 1 and 53
+
 			$$ = newNode<dsql_fld>();
 
-			if ($2 > 7)
+			if (precision > 24)
 			{
 				$$->dtype = dtype_double;
 				$$->length = sizeof(double);
@@ -5110,8 +5118,16 @@ float_type
 				$$->length = sizeof(float);
 			}
 		}
-	| LONG FLOAT precision_opt
+	| LONG FLOAT precision_opt_nz
 		{
+			// Precision is binary digits of the significand: 1-53 for 64 bit double precision
+			// Precision 0 is the 'no precision specified case', which defaults to 64 bit double precision
+			SLONG precision = $3;
+
+			if (precision != 0 && (precision < 1 || precision > 53))
+				yyabandon(YYPOSNARG(3), -842, Arg::Gds(isc_precision_err2) << Arg::Num(1) << Arg::Num(53));
+																// Precision must be between 1 and 53
+
 			$$ = newNode<dsql_fld>();
 			$$->dtype = dtype_double;
 			$$->length = sizeof(double);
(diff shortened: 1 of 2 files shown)


mrotteveel:
I'm not entirely happy with the error reporting on the precision. I have a fix, but I want some feedback on that, see _"Current handling of precision in syntax"_ on the firebird-devel mailing list.

src/dsql/parse.y
```diff
@@ -5095,11 +5095,19 @@ decimal_keyword
...
 		{
+		    // Precision is binary digits of the significand: 1-24 for 32 bit single precision, 25-53 for 64 bit double precision
+			// Precision 0 is the 'no precision specified' case, which defaults to 32 bit single precision
+			SLONG precision = $2;
+
+			if (precision != 0 && (precision < 1 || precision > 53))
+				yyabandon(YYPOSNARG(2), -842, Arg::Gds(isc_precision_err2) << Arg::Num(1) << Arg::Num(53));
+																// Precision must be between 1 and 53
+
 			$$ = newNode<dsql_fld>();
```
dyemanov: const SLONG, please. And maybe it's worth defining 53 and 24 as named constants in the beginning of this file.

mrotteveel: I followed what `decfloat_type` does. Should I change it there as well?

dyemanov: Then leave your code as is. I will change both in a separate commit.

mrotteveel:
@dyemanov I noticed that with my change, the rule fragment `precision_opt` is no longer used. Should it be removed?

src/dsql/parse.y
mrotteveel: Ok, I'll leave that as is.

dyemanov:
> @dyemanov I noticed that with my change, the rule fragment `precision_opt` is no longer used. Should it be removed?

I'd preserve it (for possible future use).

asfernandes:
IFAIR an unused rule will cause a warning.

mrotteveel:
@asfernandes 
> IFAIR an unused rule will cause a warning.

Yes, btyacc produces a warning _"2 rules never reduced"_

dyemanov:
> @asfernandes
> 
> > IFAIR an unused rule will cause a warning.
> 
> Yes, btyacc produces a warning _"2 rules never reduced"_

We can comment that rule out then.

mrotteveel:
> > Yes, btyacc produces a warning _"2 rules never reduced"_
> 
> We can comment that rule out then.

In that case, why not just delete it and if it is ever necessary again, it can be retrieved from the git history.

dyemanov:
> > > Yes, btyacc produces a warning _"2 rules never reduced"_
> > 
> > 
> > We can comment that rule out then.
> 
> In that case, why not just delete it and if it is ever necessary again, it can be retrieved from the git history.

Generally, I don't mind. Although I'm pretty sure it wil be re-created from scratch when required, nobody will be digging through the history ;-)

mrotteveel:
I deleted the unused rule. The only question I have left is if I should improve the error handling for zero or negative precision, or leave it as is.

AlexPeshkoff:
Mark, sorry for being late with discussion - may be I've missed something. In all other places where precision of some numeric value is specified we use decimal digits (NUMERIC, DECIMAL, DECFLOAT) to specify precision. Isit required by standard?
PS. It's clear that the only exact way to specify precision is use of number of binary digits. But would not it be confusing to use different ways for this in different datatypes?

mrotteveel:
@AlexPeshkoff 
> Mark, sorry for being late with discussion - may be I've missed something. In all other places where precision of some numeric value is specified we use decimal digits (NUMERIC, DECIMAL, DECFLOAT) to specify precision. Is it required by standard?
> PS. It's clear that the only exact way to specify precision is use of number of binary digits. But would not it be confusing to use different ways for this in different datatypes?

Yes, the SQL standard requires that `FLOAT` uses precision in binary digits. See also the quote from SQL:2016 on http://tracker.firebirdsql.org/browse/CORE-6109 and in the topic _"Changing FLOAT to a SQL standard compliant FLOAT datatype"_ on firebird-devel.

PR #221, opened 2019-08-23
Resolves: tdf#125284 config entries not substituted under Turkish locale

caolanm:
https://bugs.documentfoundation.org/show_bug.cgi?id=125284

This leads to the inability of LibreOffice to create embedded firebird
databases under a Turkish locale.

strcasecmp is used with LC_CTYPE honoured, and...

"These functions shall behave as if the strings had been converted to
lowercase and then a byte comparison performed".

Turkish has the infamous problem that the lowercase of I is the
dotless ı not i.

so strcasecmp("i", "I") != 0 means dir_FOO doesn't caseless match DIR_FOO

firebird/databases.conf contains $(dir_sampleDb) and the string it's compared
to is DIR_SAMPLEDB so a simple fix is to changed DIR_SAMPLEDB to dir_sampledb
and the defalt config will then substitute fine.

### DIFF
--- a/src/common/config/config_file.cpp
+++ b/src/common/config/config_file.cpp
@@ -530,16 +530,14 @@ bool ConfigFile::substituteStandardDir(const String& from, String& to) const
 		unsigned code;
 		const char* name;
 	} dirs[] = {
-#define NMDIR(a) {Firebird::IConfigManager::a, "FB_"#a},
-		NMDIR(DIR_CONF)
-		NMDIR(DIR_SECDB)
-		NMDIR(DIR_PLUGINS)
-		NMDIR(DIR_UDF)
-		NMDIR(DIR_SAMPLE)
-		NMDIR(DIR_SAMPLEDB)
-		NMDIR(DIR_INTL)
-		NMDIR(DIR_MSG)
-#undef NMDIR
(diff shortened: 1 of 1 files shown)


AlexPeshkoff:
That fix just hides a problem, not fixes it. Macro in conf file should be case-insensitive, i.e. one can type in a databases.conf: $(DIR_SAMPLEDB), $(dir_sampledb) or even $(DiR_sAmPlEdB) and that all should work.

AdrianBunk:
@AlexPeshkoff The problem is that uppercase/lowercase (like sort order) depends on the locale even for ASCII.

The lowercase of `DIR_SAMPLEDB` might be `dir_sampledb` or `dır_sampledb` (dotless `ı`) depending on the locale.

aafemt:
ASCII is a very specific American charset. Only American locale rules can be applied to it.

AdrianBunk:
> ASCII is a very specific American charset. Only American locale rules can be applied to it.

To make you nitpicker happy, let me correct myself:

The problem is that uppercase/lowercase (like sort order) depends on the locale even for **the 26 characters that are also in** ASCII.

@aafemt Is this a correct statement, so that you can now fix the bug?

PR #225, opened 2019-09-16
Fixed CORE-6144: Inconsistent behaviour of the NEW context variable in AFTER UPDATE OR DELETE triggers

ilya071294:


hvlad (approved):
Looks correct

src/jrd/exe.cpp
```diff
@@ -1044,11 +1044,11 @@ void EXE_execute_triggers(thread_db* tdbb,
...
 	{
-		const Record* const record = old_rec ? old_rec : new_rec;
-		fb_assert(record && record->getFormat());
+		record_param* rpb = old_rpb ? old_rpb : new_rpb;
+		fb_assert(rpb && rpb->rpb_relation->rel_current_format);
 		// copy the record
 		MemoryPool& pool = *tdbb->getDefaultPool();
-		null_rec = FB_NEW_POOL(pool) Record(pool, record->getFormat());
+		null_rec = FB_NEW_POOL(pool) Record(pool, rpb->rpb_relation->rel_current_format);
 		// initialize all fields to missing
```
dyemanov: I'm wondering wouldn't it be more robust to use MET_current() call instead of using rel_current_format directly?

dyemanov (approved):

dyemanov:
I've just frontported the fix to master.

PR #229, opened 2019-10-29
CORE-5538 implementation

aafemt:
A new switch for gbak -INCLUDE(_DATA) with the same semantic as -SKIP_D(ATA).


AlexPeshkoff:
I see no problems with code, just add a few words in documentation about behavior when both filters are set. Telling true I've supposed that include/skip options are exclusive. On my mind logic when setting both is not obvious i.e. should be documented.

aafemt:
> On my mind logic when setting both is not obvious i.e. should be documented.

That's why I used the table: to be able to tune the logic if someone consider it wrong. 
Now it support case "include all tables starting from A but not ABC" which appeared useful 
for me.

PR #231, opened 2019-11-06
Fix warning on Win64 build

aafemt:
Also according to MSDN _beginthreadex() return error code in errno, not LastError.

### DIFF
--- a/src/common/ThreadStart.cpp
+++ b/src/common/ThreadStart.cpp
@@ -309,13 +309,16 @@ ThreadId Thread::start(ThreadEntryPoint* routine, void* arg, int priority_arg, H
 	 * Advanced Windows by Richter pg. # 109. */
 
 	unsigned thread_id;
-	unsigned long real_handle =
-		_beginthreadex(NULL, 0, THREAD_ENTRYPOINT, THREAD_ARG, CREATE_SUSPENDED, &thread_id);
-	if (!real_handle)
+	HANDLE handle =
+		reinterpret_cast<HANDLE>(_beginthreadex(NULL, 0, THREAD_ENTRYPOINT, THREAD_ARG, CREATE_SUSPENDED, &thread_id));
+	if (!handle)
 	{
+		// Though MSDN says that _beginthreadex() returns error in errno,
+		// GetLastError() still works because RTL call no other system
+		// functions after CreateThread() in the case of error.
+		// Watch out if it is ever changed.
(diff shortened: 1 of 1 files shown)


src/common/ThreadStart.cpp
```diff
@@ -309,13 +309,12 @@ ThreadId Thread::start(ThreadEntryPoint* routine, void* arg, int priority_arg, H
...
-	unsigned long real_handle =
-		_beginthreadex(NULL, 0, THREAD_ENTRYPOINT, THREAD_ARG, CREATE_SUSPENDED, &thread_id);
-	if (!real_handle)
+	HANDLE handle =
+		reinterpret_cast<HANDLE>(_beginthreadex(NULL, 0, THREAD_ENTRYPOINT, THREAD_ARG, CREATE_SUSPENDED, &thread_id));
+	if (!handle)
 	{
-		Firebird::system_call_failed::raise("_beginthreadex", GetLastError());
+		Firebird::system_call_failed::raise("_beginthreadex", errno);
 	}
```
dyemanov: IIRC, system_call_failed / system_error exceptions expect Win32 error codes rather then CRT ones.

aafemt:
> IIRC, system_call_failed / system_error exceptions expect Win32 error codes rather then 
> CRT ones.

errno is used with system_call_failed everywhere on POSIX and in a couple of places on 
Win32 so I don't see difference.

AlexPeshkoff:
On POSIX use of errno in system_call_failed is as designed, on win32 - sooner of all a couple of bugs

aafemt:
Which exception class should be used here then?
As I said, 
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex?view=vs-2019 
clearly states that error is returned in errno.

dyemanov:
> clearly states that error is returned in errno
Because it's a CRT function, they are somewhat isolated from the system calls. But it doesn't necessarily mean that SetLastError() is not called under the hood. Google shows a number of recommendations to check GetLastError() after failing _beginthreadex(), so it might actually work. So I'd revert this your change back (or replace errno with ERRNO which is defined as GetLastError() on Windows).

aafemt:
In current versions of CRT it works because no other system functions are called after 
CreateThread() as you can see in ucrt sources but they can change it any time.

src/common/ThreadStart.cpp
aafemt: Ok, I've reverted to GetLastError() with comment.

PR #232, opened 2019-11-07
Fixed CORE-5902 - Add Firebird Event fails with error

asfernandes:
"While isc_que_events - Failed to establish a secondary connection for event processing".

### DIFF
--- a/src/remote/SockAddr.h
+++ b/src/remote/SockAddr.h
@@ -59,6 +59,8 @@ private:
 	void checkAndFixFamily();
 
 public:
+	void convertFromMacOsToPosixWindows();
+	void convertFromPosixWindowsToMacOs();
 	void clear();
 	const SockAddr& operator = (const SockAddr& x);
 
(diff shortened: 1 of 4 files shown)


hvlad:
If we can't rely on port_client_arch, probably we could change MacOS build (using #ifdef) to always read\write data in POSIX format ?

src/remote/SockAddr.h
```diff
@@ -120,6 +122,44 @@ inline void SockAddr::checkAndFixFamily()
...
+		fb_assert(false);
+	else
+	{
+		sockAddrMacOs.sa_len = length();
+		sockAddrMacOs.sa_family = ptr()->sa_family;
+		memcpy(sockAddrMacOs.sa_data, ptr()->sa_data, length() - 2);
+		memcpy(ptr(), &sockAddrMacOs, length());
+	}
+}
+
```
hvlad: It will not work witn IPv6 addresses.

You don't need to copy sa_data at all ! 
Remove it from sockAddrPosixWindows\sockAddrMacOs and copy just 2 bytes. 
This is all you need.

asfernandes: > It will not work witn IPv6 addresses.

Why?

hvlad: Because sizeof(sockaddr_in6) == 28 in both Windows and MacOS (i'm sure - on Linux too).
And you copy 22 bytes into 14.
And you don't need to copy more than first 2 bytes.

hvlad:
Great !

PR #234, opened 2019-11-17
Use static_assert to check ODS layout

aafemt:
It works on any platform and require no external tools. Also protects from unintentional local changes at module level.


aafemt:
No comments anyone?

AlexPeshkoff:
What I dislike is that code becomes twice bigger and every change requires manual sync of parts. I've tried to avoid that.

aafemt:
GIThub disregards lines removed from ods.txt. With them total balance of lines is negative.

Yes, double work is intentional as a double check.
On the other hands the checks are as close to declaration as possible so you don't need to 
jump between several files.
Also fortunately ODS changes are seldom.

dyemanov (approved):
I tend to support this PR. Having definitions and offset checks close to each other improves readability (as compared to checking ods.h vs ods.txt) and allows to avoid the [unnecessary] preliminary runtime checks.

PR #235, opened 2019-11-18
Describe as

AlexPeshkoff:
Replace various 'SET something BIND' statements with generic 'SET BIND OF type1 TO type2'.

### DIFF
--- a/doc/sql.extensions/README.set_bind.md
+++ b/doc/sql.extensions/README.set_bind.md
@@ -20,23 +20,45 @@ SET BIND OF type NATIVE;
 Makes it possible to define rules of describing types of returned to the client columns in non-standard way -
 `type-from` is replaced with `type-to`, automatic data coercion takes place.
 
-Except SQL-statement there are two more ways to specify data coercion - tag `isc_dpb_bind` in DPB
-and `SetBind` parameter in firebird.conf & databases.conf. The later the rule is introduced (.conf->DPB->SQL)
-the higher priotiy it has. I.e. one can override any preconfigured settings from SQL statement.
-
 When incomplete type definition is used (i.e. `CHAR` instead `CHAR(n)`) in left part of `SET BIND` coercion
 will take place for all `CHAR` columns, not only default `CHAR(1)`.
 When incomplete type definiton is used in right side of the statement (TO part) firebird engine will define missing
 details about that type automatically based on source column.
 
 Special `TO` part format `LEGACY` is used when datatype, missing in previous FB version, should be represented in
-a way, understandable by old client software (may be with some data losses). For example, `NUMERIC(38)` in legacy
-form will be represented as `NUMERIC(18)`.
+a way, understandable by old client software (may be with some data losses). The following coercions are done for
+legacy datatypes:
+
+| Native datatype | Legacy datatype |
+|---|---|
+| BOOLEAN | CHAR(5) |
+| DECFLOAT | DOUBLE PRECISION |
+| NUMERIC(38) | NUMERIC(18) |
+| TIME WITH TIME ZONE | TIME WITHOUT TIME ZONE |
+| TIMESTAMP WITH TIME ZONE | TIMESTAMP WITHOUT TIME ZONE |
 
 Setting `NATIVE` means `type` will be used as if there were no previous rules for it.
 
+Except SQL-statement there are two more ways to specify data coercion - tag `isc_dpb_bind` in DPB
+and `DataTypeCompatibility` parameter in firebird.conf & databases.conf. The later the rule is introduced
+(.conf->DPB->SQL) the higher priotiy it has.
+I.e. one can override .conf in any other way any DPB from SQL statement.
 
-### Samples:
+Value of clumplet with `isc_dpb_bind` tag in DPB should be specified as a set of partially
+formed SET BIND statements, i.e. with prefix SET BIND OF is omitted, separated by semicolons.
+For example:
+```c++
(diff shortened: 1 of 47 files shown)


aafemt:
I don't see where actual data mapping happens. I would expect it to be in DSC_make_descriptor() or in dsc::getSqlInfo(), but common/dsc.cpp isn't touched at all.

AlexPeshkoff:
Dimitry, start with dsql/gen.cpp

aafemt (commented):
IMHO, you perform data transformation too early. Imagine a DBA set mapping DOUBLE->INTEGER which los fractions at database or server level and a programmer set sqlvar.sqltype to SQL_DOUBLE at isc_dsql_fetch() call. AFAIU, with current code there will be two data transformations DOUBLE->INTEGER->DOUBLE and the result will be wrong.

IMHO, data type mapping is better to be performed at isc_describe(_bind)/inf time and do not touch DSC at all.

src/dsql/parse.y
```diff
@@ -5272,18 +5283,35 @@ set_decfloat_traps
...
-%type <setHighPrecBindNode> set_decfloat_bind
-set_decfloat_bind
-	: SET bind_to_type BIND
-			{ $$ = newNode<SetHighPrecBindNode>($2); }
-		decfloat_bind_clause($4)
-			{ $$ = $4; }
+%type <setBindNode> set_bind
+set_bind
+	: SET BIND OF set_bind_from
+			{ $$ = newNode<SetBindNode>(); $$->from = $4; }
```
asfernandes: @AlexPeshkoff why do you need to create intermediary action rule here, if the intermediary rule is not passed as parameter later?

doc/sql.extensions/README.set_bind.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Description:
+
+Makes it possible to define rules of describing types of returned to the client columns in non-standard way -
+`type-from` is replaced with `type-to`, automatic data coercion takes place.
+
+When incomplete type definition is used (i.e. `CHAR` instead `CHAR(n)`) in left part of `SET BIND` coercion
+will take place for all `CHAR` columns, not only default `CHAR(1)`.
+When incomplete type definiton is used in right side of the statement (TO part) firebird engine will define missing
+details about that type automatically based on source column.
```
aafemt: It doesn't work:

> SQL> set bind of integer to varchar;
> Statement failed, SQLSTATE = 42000
> Dynamic SQL Error
> -SQL error code = -104
> -Unexpected end of command - line 1, column 24

Also doesn't work:

> SQL> set bind of blob sub_type text to varchar(2000);
> Statement failed, SQLSTATE = 42000
> Dynamic SQL Error
> -SQL error code = -104
> -Token unknown - line 1, column 13
> -blob

aafemt:
Engine must reject SET BIND command if data coercion is not supported for this pair:

> SQL> set bind of integer to boolean;
> SQL> select a from t;
> OUTPUT message field count: 1
> 01: sqltype: 32764 BOOLEAN Nullable scale: 0 subtype: 0 len: 1
>   :  name: A  alias: A
>   : table: T  owner: SD
> 
>       A
> =======
> Statement failed, SQLSTATE = 22018
> conversion error from string "1"

aafemt:
AFAIR databases.conf is reloaded (if changed) on every configuration read. What should be engine behaviour if compatibility parameter was changed in the middle of working attachment?
a) Attachment must use rules that were set when it started
b) Attachment must use new rules starting from the next statement

AFAICS currently b) is the case.

AlexPeshkoff:
No, attachment currently is using rules that were set when it started.
Moreover - to be precise rules used when DBB was started will be in 
effect, exactly like with other parameters from databases.conf.
Stable pointer to existing database-specific configuration is stored in DBB.

aafemt:
In this case I don't understand this three nested ifs. It looks like coercion rules are 
acquired on first request. Could you add some comments about the logic there to allow 
readers get it without tracing whole execution tree, please?

aafemt:
Ability to set mappings in DPB dosn't look much useful for me (because work for changing DPB in an application is not much different from work for supporting new data types) but I think that some utilities of gbak-class would like to have a way to disable all user data mapping for the connection using something simplier than explicit mapping of every data type to "native".

AlexPeshkoff:
On 2019-11-30 13:24, Dimitry Sibiryakov wrote:
>
> Ability to set mappings in DPB dosn't look much useful for me (because 
> work for changing DPB in an application is not much different from 
> work for supporting new data types)
>

Unfortunately our driver developers have another mind - according to 
that logic having DPB support is 'must-be' feature for Java/C#.

> but I think that some utilities of gbak-class would like to have a way 
> to disable all user data mapping for the connection using something 
> simplier than explicit mapping of every data type to "native".
>

Gbak does not need it.
If/when one really needs that support it's always possible to add such 
improvement in dot release.

aafemt:
> Gbak does not need it.

How does it bypass mapping set in .conf file then?
Or may be it just doesn't care and write to backup file values with backward-compatible types?

AlexPeshkoff:
On 2019-12-02 14:29, Dimitry Sibiryakov wrote:
> > Gbak does not need it.
>
> How does it bypass mapping set in .conf file then?

According to your previous suggestion - passing explicit message format 
to engine. And that messages are built based on system tables, not 
getInfo().

> Or may be it just doesn't care and write to backup file values with 
> backward-compatible types?

No :) I've checked that.

PR #240, opened 2019-12-04
In case of direct IO use 4K buffer alignment because on some devices …

dmitry-starodubov:
…pread() with less alignment returns EINVAL


AlexPeshkoff:
Same PR for master?

dmitry-starodubov:
> 
> 
> Same PR for master?

In progress.

dyemanov:
Versions 4-6 already contain the fix, merging this PR is impossible and v3 is close to its end-of-life, thus closing this PR. Those who's affected should upgrade to newer versions.

PR #241, opened 2019-12-04
In case of direct IO use 4K buffer alignment because on some devices …

dmitry-starodubov:
…pread() with less alignment returns EINVAL


hvlad:
Replaced by #7890

PR #245, opened 2019-12-20
Windows file description

hvlad:
Currently (and since the first release of Firebird) "Firebird SQL Server" is used as a description in version info at all binary files (.exe and .dll). In this pull request I offer to set distinct and more meaningful descriptions. 
It allows users to see additional info in file properties and at Windows explorer. 
Also this description string is put by the Windows in some system error messages and it is very useful for support team to see "Firebird Interactive Query tool..." or "Firebird Gbak tool..." instead of the same "Firebird SQL Server" for all executables.

I set some short descriptions for all modules (see jrd/version.h) and added also bitness and debug\release info of how module was built. For example, for x64 release build of engine13.dll description is set to "Firebird Engine plugin. 64-bit release build".

Also, fileds InternalName and OriginalFilename of version info is set according to MS requirements.

### DIFF
--- a/src/jrd/version.h
+++ b/src/jrd/version.h
@@ -1,0 +1,151 @@
+/*
+ *  The contents of this file are subject to the Initial
+ *  Developer's Public License Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the
+ *  License. You may obtain a copy of the License at
+ *  http://www.ibphoenix.com/main.nfs?a=ibphoenix&page=ibp_idpl.
+ *
+ *  Software distributed under the License is distributed AS IS,
+ *  WITHOUT WARRANTY OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing rights
+ *  and limitations under the License.
+ *
+ *  The Original Code was created by Vladyslav Khorsun
+ *  for the Firebird Open Source RDBMS project.
+ *
+ *  Copyright (c) 2019 Vladyslav Khorsun <hvlad@users.sourceforge.net>
+ *  and all contributors signed below.
+ *
+ *  All Rights Reserved.
+ *  Contributor(s): ______________________________________.
+ */
+
+#ifndef JRD_VERSION_H
+#define JRD_VERSION_H
+
+#define STRINGIZE_AUX(x)	#x
+#define STRINGIZE(x)		STRINGIZE_AUX(x)
+
+
+#ifdef RC_TARGET_engine13
+#define VER_FILEDESC "Engine plugin"
+
+#elif defined RC_TARGET_fb_lock_print
+#define VER_FILEDESC "Lock Print tool"
+
+#elif defined RC_TARGET_fbguard
+#define VER_FILEDESC "Guardian"
+
+#elif defined RC_TARGET_fbrmclib
+#define VER_FILEDESC "RM/COBOL client interface"
+
+#elif defined RC_TARGET_firebird
+#define VER_FILEDESC "Server main executable"
+
+#elif defined RC_TARGET_fbsvcmgr
+#define VER_FILEDESC "Service Manager"
+
+#elif defined RC_TARGET_fbtrace
+#define VER_FILEDESC "Trace plugin"
+
+#elif defined RC_TARGET_fbtracemgr
+#define VER_FILEDESC "Trace Manager"
+
+#elif defined RC_TARGET_gbak
+#define VER_FILEDESC "Gbak tool"
+
+#elif defined RC_TARGET_gfix
+#define VER_FILEDESC "Gfix tool"
+
+#elif defined RC_TARGET_gpre
+#define VER_FILEDESC "Gpre tool"
+
+#elif defined RC_TARGET_gsec
+#define VER_FILEDESC "Gsec tool"
+
+#elif defined RC_TARGET_gsplit
+#define VER_FILEDESC "Gsplit tool"
+
+#elif defined RC_TARGET_gstat
+#define VER_FILEDESC "Gstat tool"
+
+#elif defined RC_TARGET_ib_util
+#define VER_FILEDESC "UDF support library"
+
+#elif defined RC_TARGET_instclient
+#define VER_FILEDESC "Install Client tool"
+
+#elif defined RC_TARGET_instreg
+#define VER_FILEDESC "Install Registry tool"
+
+#elif defined RC_TARGET_instsvc
+#define VER_FILEDESC "Install Service tool"
+
+#elif defined RC_TARGET_fbintl
+#define VER_FILEDESC "International Characters support library"
+
+#elif defined RC_TARGET_isql
+#define VER_FILEDESC "Interactive Query tool"
+
+#elif defined RC_TARGET_legacy_auth
+#define VER_FILEDESC "Legacy Auth plugin"
+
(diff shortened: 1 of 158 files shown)


aafemt:
Wouldn't it be easier to add "-D VER_FILEDESC="XXXX"" to project files directly instead of endless "elif" in version.h?

hvlad:
I prefer to see all descriptions in the single place.
It is much more easy to change description in .h file than in project file.
In project file one should change "PreprocessorDefinitions" section 4 times - for every kind of build.

aafemt:
Well, if MSVS cannot have common define section for all project's builds then you are right.

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+#define VER_FILEDESC "Lock Print tool"
+
+#elif defined RC_TARGET_fbguard
+#define VER_FILEDESC "Guardian"
+
+#elif defined RC_TARGET_fbrmclib
+#define VER_FILEDESC "RM/COBOL client interface"
+
+#elif defined RC_TARGET_firebird
+#define VER_FILEDESC "Server main executable"
```
dyemanov: Maybe just "Server executable"? I don' see any other "executable" in the descriptions.

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+#define VER_FILEDESC "Engine plugin"
+
+#elif defined RC_TARGET_fb_lock_print
+#define VER_FILEDESC "Lock Print tool"
+
+#elif defined RC_TARGET_fbguard
+#define VER_FILEDESC "Guardian"
+
+#elif defined RC_TARGET_fbrmclib
+#define VER_FILEDESC "RM/COBOL client interface"
```
dyemanov: RM/Cobol Helper library?

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+
+#elif defined RC_TARGET_firebird
+#define VER_FILEDESC "Server main executable"
+
+#elif defined RC_TARGET_fbsvcmgr
+#define VER_FILEDESC "Service Manager"
```
dyemanov: "Service Manager tool" or "Service Management tool" perhaps?

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+
+#elif defined RC_TARGET_fbtrace
+#define VER_FILEDESC "Trace plugin"
+
+#elif defined RC_TARGET_fbtracemgr
+#define VER_FILEDESC "Trace Manager"
```
dyemanov: "Trace Manager tool" or "Trace Management tool"?

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+#define VER_FILEDESC "Gsec tool"
+
+#elif defined RC_TARGET_gsplit
+#define VER_FILEDESC "Gsplit tool"
+
+#elif defined RC_TARGET_gstat
+#define VER_FILEDESC "Gstat tool"
+
+#elif defined RC_TARGET_ib_util
+#define VER_FILEDESC "UDF support library"
```
dyemanov: UDF Helper library? UDF Utility library?

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+#define VER_FILEDESC "Interactive Query tool"
+
+#elif defined RC_TARGET_legacy_auth
+#define VER_FILEDESC "Legacy Auth plugin"
+
+#elif defined RC_TARGET_legacy_usermanager
+#define VER_FILEDESC "Legacy User Manager plugin"
+
+#elif defined RC_TARGET_nbackup
+#define VER_FILEDESC "Physical Backup Manager"
```
dyemanov: Physical Backup tool?

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+#define VER_FILEDESC "UDF compatibility library"
+
+#elif defined RC_TARGET_udr_engine
+#define VER_FILEDESC "User Defined Routines engine"
+
+#elif defined RC_TARGET_fbclient
+#define VER_FILEDESC "Client library"
+
+#elif defined RC_TARGET_build_msg
+#define VER_FILEDESC "Build Message file tool"
```
dyemanov: file -> File ?

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+
+#elif defined RC_TARGET_codes
+#define VER_FILEDESC "Generate Error Codes tool"
+
+#elif defined RC_TARGET_gpre_boot
+#define VER_FILEDESC "Gpre boot tool"
```
dyemanov: Bootstrap Gpre tool? Or just ignore its specifics and also name it "Gpre tool".

src/jrd/version.h
hvlad: I took it from src\gpre\languages\fbrmclib.cpp:

> DESCRIPTION:	RM/Cobol / fbclient interface

src/jrd/version.h
hvlad: nbackup usage screen say : "Physical Backup Manager"

src/jrd/version.h
```diff
@@ -0,0 +1,151 @@
...
+
+#endif
+
+
+#ifdef NDEBUG
+#define VER_DBG "release"
+#else
+#define VER_DBG "debug"
+#endif
+
```
dyemanov: I seriously don't think we need that. Debug builds can be seen on developer machines only and the file size/performance usually explains it ;-) 

But if you insist, then please show it for debug versions only. No need to see "release" in 99.999% of all builds.

src/jrd/version.rc
```diff
@@ -35,9 +36,10 @@ BEGIN
     BEGIN
       VALUE "Comments", "This product created by the Firebird - All Copyright (c) retained by the individual contributors - original code Copyright (c) 2000 Inprise Corporation and predecessors.\0"
       VALUE "CompanyName", "Firebird Project\0"
-      VALUE "FileDescription", "Firebird SQL Server\0"
+      VALUE "FileDescription", "Firebird " VER_FILEDESC ". " VER_ARCH " " VER_DBG " build"
```
dyemanov: May I suggest "Firebird " VER_FILEDESC " (" VER_ARCH " " VER_DBG ")" instead? I'd prefer something as simple as:
Firebird Gbak tool (64-bit)
Firebird Gbak tool (64-bit DEBUG)

src/jrd/version.h
hvlad: Ok, I'll remove "release" and leave "debug"

src/jrd/version.rc
hvlad: Ok

src/jrd/version.h
hvlad: Ok

src/jrd/version.h
dyemanov: We have "executable", "library", "plugin", "tool" -- they're well understood. What is "interface" and how it differs from "library"? What is "manager" and how it differs from "tool"? Why GBAK is "tool" but NBACKUP is "manager"? ;-) I know this is nitpicking and I surely accept the current names too. But perhaps we could be more consistent.

mrotteveel:
This fixes http://tracker.firebirdsql.org/browse/CORE-5691

PR #246, opened 2020-01-09
[B3_0] Review of wirecompression code

ghost:
Changed functions:
 - REMOTE_inflate
 - REMOTE_deflate
 - rem_port::initCompression
 - ServerAuth::authenticate
 - accept_connection
 - ConnectAuth::accept


dmitry-lipetsk:
@AlexPeshkoff I understood that this PR is incorrect...

I was wrong about packet processing.

REMOTE_inflate may be called before setup of PORT_compressed flag.

So REMOTE_inflate looks on the port->port_compressed only and initCompression need call before authPort->send(send);

One question - why REMOTE_deflate checks both:
 (1) "port->port_compressed"
and
 (2) "port->port_flags & PORT_compressed" ?

Checking "port->port_flags & PORT_compressed" is not enough?

AlexPeshkoff:
I do not see what problems does suggested PR solve.

aafemt:
> I do not see what problems does suggested PR solve.


If you have a circular reference of such pointers, then deleting of pointed object may 
cause this object to be deleted as well, crashing at member access.

This is crazy and should never happen, but CORE-6224 imply that it is possible.

AlexPeshkoff:
@aafemt looks like you post at the wrong place

aafemt:
Yep, sorry.

dmitry-lipetsk:
This PR does not solve the problems. And, I hope, does not create them.

This PR enhances code cleanliness and quality.

src/remote/remote.cpp
```diff
@@ -1425,7 +1425,13 @@ bool REMOTE_inflate(rem_port* port, PacketReceive* packet_receive, UCHAR* buffer
 {
 #ifdef WIRE_COMPRESS_SUPPORT
 	if (!port->port_compressed)
+	{
 		return packet_receive(port, buffer, buffer_length, length);
+	}//if
+
+	fb_assert(port->port_compressed);
+
+	// Attention: PORT_compression flag may still not setup!
```
AlexPeshkoff: All this block does not look good.
1. Adding { } for single line block does not follow FB code standard (both styles possible but no cleanup from one to other).
2. "assert(expr)" right after "if (!expr) return" look wierd
3. Syntax error in comment, not PORT_compressION but PORT_compressED

src/remote/remote.cpp
```diff
@@ -1503,19 +1509,17 @@ bool REMOTE_deflate(XDR* xdrs, ProtoWrite* proto_write, PacketSend* packet_send,
 {
 #ifdef WIRE_COMPRESS_SUPPORT
 	rem_port* port = (rem_port*) xdrs->x_public;
-	if (!(port->port_compressed && (port->port_flags & PORT_compressed)))
+	if (!(port->port_flags & PORT_compressed))
+	{
 		return proto_write(xdrs);
+	}//if
+
+	fb_assert(port->port_compressed);
```
AlexPeshkoff: Same issue (do not mention them later) with { } around single line comment.

src/remote/server/server.cpp
```diff
@@ -709,11 +709,30 @@ class ServerAuth : public GlobalStorage, public ServerAuthBase
...
-				if (send->p_acpt.p_acpt_type & pflag_compress)
+				if ((send->p_acpt.p_acpt_type & pflag_compress) == 0)
+				{
+					authPort->send(send); //throw?
+				}
+				else
+				{
+					// Configuration of port for compression usage (1-3).
+					fb_assert(send->p_acpt.p_acpt_type & pflag_compress);
+
```
AlexPeshkoff: Once again assert() right after 'if'

src/remote/server/server.cpp
```diff
@@ -709,11 +709,30 @@ class ServerAuth : public GlobalStorage, public ServerAuthBase
...
+					// ATTENTION:
+					//  In this moment server may already got the compressed
+					//  packet from client and is waiting for finish of this
+					//  not completed answer to previous packet!
+
+					// 3. Enabled deflate.
 					authPort->port_flags |= PORT_compressed;
+				}//else
+
```
AlexPeshkoff: You remove duplication of condition check but add one for function call. Very doubtfully...
And code becomes 5 times longer than before. That's also inconvenient - we still do not have video walls from Bradbury's 451 as monitors :) Personally I dislike this change.

AlexPeshkoff:
In that case please close it and create PR for master, cleaning branches in this way does not make much sense.

dmitry-lipetsk:
ok

PR #247, opened 2020-01-10
Removed TDBB_trusted_ddl flag and refactored related code.

romansimakov:
Removed TDBB_trusted_ddl flag which switched off all SCL_checks globally.
Also removed SCL_checks from VIO_store, VIO_modify, VIO_erase which were
switched off by TDBB_trudted_ddl after checkPermission in DdlNode.
Other ways to run VIO_xxx are prohibited and users cannot modify system
tables directly.
Now error includes an object name when user does not have the privilege CREATE for it.


romansimakov:
Flags which allow to modify system tables are:
Attachment::isGbak()
or
INSERT: JrdStatement::FLAG_INTERNAL in statement->flags.
UPDATE и DELETE: FLAG_SYS_TRIGGER | FLAG_INTERNAL | FLAG_IGNORE_PERM in statement->flags.

FLAG_SYS_TRIGGER - allows to do anything for system triggers.
FLAG_IGNORE_PERM - almost the same but for cascade or consistent integruty triggers.
FLAG_INTERNAL - set in AutoRequest, AutoCacheRequest, AutoPreparedStatement.
Attachment::prepareUserStatement doesn't use them.

So DDL operations don't have to check privileges on an access to the system tables. It's enough only  to check DDL operation check itself.
Direct changes to the system tables are also prohibited now.
I left SCL_check_XXX calls if protect_system_... is skept.
Note!!! All deleted SCL_check_XXX actually were switched off by the TDBB_trusted_ddl flags when DDL operator was working.

Checking privileges for package items I just switched off by additional parameter to executeDdl.
Now code look more obvious and must be less error prone in case of DDL trigger execution for example.

I tested it in compating to master and also ran all possible tests I have for RedDatabase. There are no fails.
But I'm asking for additional review)

src/dsql/Nodes.h
```diff
@@ -194,16 +194,14 @@ class DdlNode : public Node
...
 		// dsqlScratch should be NULL with CREATE DATABASE.
 		if (dsqlScratch)
 			dsqlScratch->setTransaction(transaction);
 
-		const ULONG flag = checkPermission(tdbb, transaction) ? TDBB_trusted_ddl : 0;
-
-		Firebird::AutoSetRestoreFlag<ULONG> trustedDdlFlag(&tdbb->tdbb_flags, flag, true);
-
+		if (!trusted)
+			checkPermission(tdbb, transaction);
```
dyemanov: Shouldn't checkPermission() return void now? And the same question for SCL_check_* routines.

src/include/gen/msgs.h
```diff
@@ -968,6 +968,7 @@ Data source : @4"},		/* eds_statement */
 	{335545261, "Can not convert @1 to @2"},		/* bind_convert */
 	{335545262, "cannot update old BLOB"},		/* cannot_update_old_blob */
 	{335545263, "cannot read from new BLOB"},		/* cannot_read_new_blob */
+	{335545264, "There is no privilege CREATE @1"},		/* dyn_no_create_priv */
```
dyemanov: "User has no CREATE @1 privilege" ?
"Privilege CREATE @1 is not granted to user" ?
"No permission for CREATE @1 operation" ?

src/jrd/jrd.h
```diff
@@ -485,7 +485,6 @@ const ULONG TDBB_use_db_page_space		= 128;		// use database (not temporary) page
 const ULONG TDBB_detaching				= 256;		// detach is in progress
 const ULONG TDBB_wait_cancel_disable	= 512;		// don't cancel current waiting operation
 const ULONG TDBB_cache_unwound			= 1024;		// page cache was unwound
-const ULONG TDBB_trusted_ddl			= 2048;		// skip DDL permission checks. Set after DDL permission check and clear after DDL execution
```
dyemanov: Please move the other constants' values.

src/jrd/scl.epp
```diff
@@ -295,16 +287,17 @@ void SCL_check_create_access(thread_db* tdbb, int type)
...
+	if (attachment->locksmith(tdbb, SystemPrivilege::MODIFY_ANY_OBJECT_IN_DATABASE))
 		return;
 
 	const SecurityClass::flags_t obj_mask = SCL_get_object_mask(type);
 
 	if (!(obj_mask & SCL_create))
+	{
+		const char* name = accTypeNumToStr(type);
+//		ERR_post(Arg::Gds(isc_dyn_no_create_priv) << name);
 		ERR_post(Arg::Gds(isc_dyn_no_priv));
```
dyemanov: Forgot to switch to the new constant?

dyemanov (changes requested):

src/jrd/scl.epp
romansimakov: Yep. In order to compare test results I changed it back and forgot to return.

src/dsql/Nodes.h
romansimakov: Speaking about checkPermission I agree. SCL_check_* require careful investigation. Most of them yes. But some of them return false if object isn't found. It's used while checking CREATE OR ALTER privileges.

src/include/gen/msgs.h
romansimakov: User has no CREATE @1 privilege

src/jrd/jrd.h
romansimakov: OK

src/dsql/Nodes.h
dyemanov: Then I have a question. If e.g. SCL_check_function() cannot find a function inside RDB$FUNCTIONS, why SCL_check_access() is still called (with non-existing function name and NULL security class)?

romansimakov: You're right

dyemanov (approved):
I don't have any other comments. If Alex does not have objections either, we may merge this PR.

AlexPeshkoff (approved):
No objections

PR #249, opened 2020-01-13
Correction of RefPtr::assign [B3_0]

ghost:
When RefPtr::assign returns to caller it (object) may be already destroyed. As result "return ptr" will access (read 'ptr' member) the destroyed object.

The correct code - 'return p;'

Note: usually smart pointers do not return internal/raw pointers. They return the reference to themselves.

### DIFF
--- a/src/common/classes/RefCounted.h
+++ b/src/common/classes/RefCounted.h
@@ -212,7 +212,7 @@ namespace Firebird
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


aafemt:
Object cannot be destroyed while it has reference counter bigger than zero. ptr is already assigned to value p few lines above.

PR #253, opened 2020-01-20
Generate ids.h without m4

aafemt:


mapopa:
Nice one , we get  rid of m4 when building on Posix systems , the next one would be sed :) 
Sed is kind of hard to setup on Windows 

Related discussion 

https://sourceforge.net/p/firebird/mailman/firebird-devel/thread/31fb0a5d-9f31-d00b-961a-148b9907ad6d%40ibphoenix.com/#msg36901852

AlexPeshkoff:
On 2020-01-22 11:57, Popa Adrian Marius wrote:
>
> Nice one , we get rid of m4 when building on Posix systems
>

Not completely: m4 is also used to preprocess configure.ac.
But that was not main goal - on windows appropriate step (with ids.h) 
was not executed at all and required manual changes in ids.h.

reevespaul:
On Wed, 22 Jan 2020 00:57:45 -0800
Popa Adrian Marius <notifications@github.com> wrote:

> Nice one , we get  rid of m4 when building on Posix systems , the next one
> would be sed :) Sed is kind of hard to setup on Windows 
> 
> Related discussion 
> 
> https://sourceforge.net/p/firebird/mailman/firebird-devel/thread/31fb0a5d-9f31-d00b-961a-148b9907ad6d%40ibphoenix.com/#msg36901852
> 

There is a major part missing from that discussion.

git for windows is a de facto requirement now for building firebird in
windows. It comes with a full posix tool set. So sed is already installed.
There is no need to go looking for it. You already have it. The only thing
that needs to be done is to add the appropriate directory to the path.


Paul
-- 

Paul Reeves
http://www.ibphoenix.com
Supporting users of Firebird

PR #256, opened 2020-02-06
Fixed CORE-6110: 64-bit transaction IDs are not stored properly in status vector

ilya071294:
Arg::Str is used instead of Arg::Num for isc_concurrent_transaction, isc_rec_in_limbo, isc_tra_state error messages because transaction number is 64-bit signed integer.

In the general case it doesn't matter which types of argument are used in messages. For example, when isql gets a status vector, it checks the type of argument and prints its value in the right way.

### DIFF
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -11996,11 +11996,11 @@ void AlterDatabaseNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
 
 			// msg 297: Concurrent ALTER DATABASE is not supported
 			Arg::PrivateDyn status(297);
+			string trans_num_str;
 
 			if (conflict_trans)
 			{
 				// Cannot use Arg::Num here because transaction number is 64-bit signed integer
-				string trans_num_str;
 				trans_num_str.printf("%lld", conflict_trans);
 				status << Arg::Gds(isc_concurrent_transaction) << Arg::Str(trans_num_str);
 			}
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -11998,7 +11998,12 @@ void AlterDatabaseNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
 			Arg::PrivateDyn status(297);
 
 			if (conflict_trans)
-				status << Arg::Gds(isc_concurrent_transaction) << Arg::Num(conflict_trans);
+			{
+				// Cannot use Arg::Num here because transaction number is 64-bit signed integer
+				string trans_num_str;
+				trans_num_str.printf("%lld", conflict_trans);
+				status << Arg::Gds(isc_concurrent_transaction) << Arg::Str(trans_num_str);
+			}
 
 			status_exception::raise(status);
 		}
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -12001,7 +12001,7 @@ void AlterDatabaseNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
 			if (conflict_trans)
 			{
 				// Cannot use Arg::Num here because transaction number is 64-bit signed integer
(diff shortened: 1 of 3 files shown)


vasiliy-yashkov:
Tested on Jaybird driver, starting with version 2.2 and up to 4. The driver was tested both in network mode and using fbclient 3.0.5 library. If use replacement of numeric `Arg::Num` type of argument with a string `Arg::Str`, the driver receives the transaction ID in its original form. The string type does not cause problems in the driver.
IBExpert was also tested with fbclient 3.0.5, no problems or errors occurred.
I think that other clients (that use fbclient library) will work as before.

src/dsql/DdlNodes.epp
```diff
@@ -11998,7 +11998,12 @@ void AlterDatabaseNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
 			Arg::PrivateDyn status(297);
 
 			if (conflict_trans)
-				status << Arg::Gds(isc_concurrent_transaction) << Arg::Num(conflict_trans);
+			{
+				// Cannot use Arg::Num here because transaction number is 64-bit signed integer
+				string trans_num_str;
```
AlexPeshkoff: I suppose you've checked how does your code run.
Unfortunately here it runs due to some luck. status holds direct pointer to string argument till raise() is called but trans_num_str is destroyed before it. I.e. status has a pointer to already destroyed variable when raise is called. This may work sometimes but it's definitely a bug, you should move trans_num_str out of 'if' block.

asfernandes:
Why not create a specific arg type for 64 bit numbers?

AlexPeshkoff (changes requested):

aafemt:
Backward and platform compatibility.

dyemanov:
Arg::TraNum may derive from Arg::Str but accept the TraNumber argument.

src/dsql/DdlNodes.epp
```diff
@@ -11998,7 +11998,12 @@ void AlterDatabaseNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
...
+				trans_num_str.printf("%lld", conflict_trans);
```
asfernandes: You should use something like QUADFORMAT macro.

AlexPeshkoff:
> Arg::TraNum may derive from Arg::Str but accept the TraNumber argument.

That's far non-trivial change. The big problem is how to store string argument. Remember, Arg family now stores only pointers to strings that are copied to safe place when raise() is called.

src/dsql/DdlNodes.epp
```diff
@@ -11996,9 +11996,14 @@ void AlterDatabaseNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
 
 			// msg 297: Concurrent ALTER DATABASE is not supported
 			Arg::PrivateDyn status(297);
+			string trans_num_str;
 
 			if (conflict_trans)
-				status << Arg::Gds(isc_concurrent_transaction) << Arg::Num(conflict_trans);
+			{
+				// Cannot use Arg::Num here because transaction number is 64-bit signed integer
```
AlexPeshkoff: Here and later - is it good idea to use SQUADFORMAT (which is for signed numbers) to print values of TraNumber datatype which is:
typedef FB_UINT64 TraNumber;
(i.e. unsigned)?

AlexPeshkoff (changes requested):

src/dsql/DdlNodes.epp
ilya071294: In FB3 I see this:
typedef SINT64 TraNumber;

AlexPeshkoff: Sorry, I really looked at master branch.

AlexPeshkoff (approved):

AlexPeshkoff:
please prepare PR for master in order to merge them both

src/dsql/DdlNodes.epp
aafemt: Using of macros from stdint.h would be more standard-friendly.

PR #261, opened 2020-02-26
Postfix for CORE-6110: always store strings in local StatusVector's memory

AlexPeshkoff:
This solves once and forever any problems with incorrect usage of Arg::Str and derived from it args. The price is relatively small [when no exceptions raised - sizeof(Firebird::string) per StatusVector ] memory overhead. With typical usage when StatusVector is inside a scope:
if (error)
    (Arg1 << Arg2 << ...).raise();
there is no overhead as long as error did not happen.

When error is raised overhead obviously depends upon length of arguments and complexity of operations with them.

### DIFF
--- a/src/common/StatusArg.cpp
+++ b/src/common/StatusArg.cpp
@@ -30,7 +30,6 @@
 #include "../common/StatusArg.h"
 #include "../common/utils_proto.h"
 
-#include "../common/classes/fb_string.h"
 #include "../common/classes/MetaName.h"
 #include "../common/classes/alloc.h"
 #include "fb_exception.h"
@@ -60,7 +59,8 @@ Base::Base(ISC_STATUS k, ISC_STATUS c) :
 
 StatusVector::ImplStatusVector::ImplStatusVector(const ISC_STATUS* s) throw()
 	: Base::ImplBase(0, 0),
-	  m_status_vector(*getDefaultMemoryPool())
+	  m_status_vector(*getDefaultMemoryPool()),
+	  m_strings(*getDefaultMemoryPool())
 {
 	fb_assert(s);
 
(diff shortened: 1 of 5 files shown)


src/common/StatusArg.cpp
```diff
@@ -142,6 +145,52 @@ void StatusVector::ImplStatusVector::assign(const Exception& ex) throw()
...
+
+	const char* const newEnd = &newBase[m_strings.length()];
+
+	for (ISC_STATUS* arg = m_status_vector.begin(); *arg != isc_arg_end; arg += fb_utils::nextArg(*arg))
+	{
+		if (!fb_utils::isStr(*arg))
+			continue;
+
+		const char** ptr = reinterpret_cast<const char**>(&arg[fb_utils::nextArg(*arg) - 1]);
+		if (newBase <= *ptr && *ptr < newEnd)
```
asfernandes: I think `(*ptr >= newBase && *ptr < newEnd)` would be more clear.

src/common/StatusArg.cpp
```diff
@@ -142,6 +145,52 @@ void StatusVector::ImplStatusVector::assign(const Exception& ex) throw()
...
+	}
+}
+
+void StatusVector::ImplStatusVector::setStrPointers(const char* oldBase)
+{
+	const char* const newBase = m_strings.c_str();
+	if (newBase == oldBase)
+		return;
+
+	const char* const newEnd = &newBase[m_strings.length()];
```
asfernandes: `m_strings.end()` ?

src/common/StatusArg.cpp
```diff
@@ -142,6 +145,52 @@ void StatusVector::ImplStatusVector::assign(const Exception& ex) throw()
...
+			continue;
+
+		const char** ptr = reinterpret_cast<const char**>(&arg[fb_utils::nextArg(*arg) - 1]);
+		unsigned pos = m_strings.length();
+		const char* oldBase = m_strings.c_str();
+
+		if (*arg == isc_arg_cstring)
+		{
+			m_strings.append(*ptr, arg[1]);
+			m_strings.append(1, '\0');
```
asfernandes: @AlexPeshkoff these two lines could cause to allocations in m_strings and that may makes the code fail as the first allocation can realloc m_strings to a first place and the second one may realloc it to the first place, making setStrPointers have a wrong understand of what had happened.

AlexPeshkoff: If it allocates to exactly original place - no trouble, all pointers will be correct as if there was no reallocation. But if buffer is allocated not exactly at original place but overlapping it - yes, it's really trouble. Will fix.

PR #266, opened 2020-04-27
Receive features list from a real provider

aafemt:
Starting from Firebird 3 it is not possible to know in advance what provider is used for connection to a database so EDS cannot guess features it can use. Because of this it is better to ask the provider to enum them.

It should improve cooperation of Firebird engine with Magpie project (provider for ODBC connections).

### DIFF
--- a/src/include/firebird/impl/inf_pub.h
+++ b/src/include/firebird/impl/inf_pub.h
@@ -162,6 +162,9 @@ enum db_info_types
 
 	fb_info_wire_crypt = 140,
 
+	// Return list of features supported by provider of current connection
+	fb_info_provider_features = 141,
+
 	isc_info_db_last_value   /* Leave this LAST! */
 };
 
@@ -169,6 +172,18 @@ enum db_info_crypt			/* flags set in fb_info_crypt_state */
 {
 	fb_info_crypt_encrypted = 0x01,
 	fb_info_crypt_process = 0x02
+};
+
+enum info_provider_features // response to fb_info_provider_features
+{
+	fb_feature_multi_statements 	= 0,    // Multiple prepared statements in single attachment
+	fb_feature_multi_transactions	= 1,	// Multiple concurrent transaction in single attachment
+	fb_feature_named_parameters		= 2,	// Query parameters can be named
+	fb_feature_session_reset		= 3,	// ALTER SESSION RESET is supported
+	fb_feature_read_consistency		= 4,	// Read consistency TIL is supported
+	fb_feature_statement_timeout	= 5,	// Statement timeout is supported
+
+	info_provider_features_max      // Not really a feature. Keep this last.
 };
 
 #define isc_info_version isc_info_isc_version
--- a/src/include/firebird/impl/inf_pub.h
+++ b/src/include/firebird/impl/inf_pub.h
@@ -176,13 +176,13 @@ enum db_info_crypt			/* flags set in fb_info_crypt_state */
 
 enum info_provider_features // response to fb_info_provider_features
 {
-	fb_feature_multi_statements 	= 0,    // Multiple prepared statements in single attachment
-	fb_feature_multi_transactions	= 1,	// Multiple concurrent transaction in single attachment
-	fb_feature_named_parameters		= 2,	// Query parameters can be named
-	fb_feature_session_reset		= 3,	// ALTER SESSION RESET is supported
-	fb_feature_read_consistency		= 4,	// Read consistency TIL is supported
-	fb_feature_statement_timeout	= 5,	// Statement timeout is supported
-	fb_feature_statement_long_life  = 6,    // Prepared statement can survive transaction end
+	fb_feature_multi_statements 	= 1,    // Multiple prepared statements in single attachment
+	fb_feature_multi_transactions	= 2,	// Multiple concurrent transaction in single attachment
+	fb_feature_named_parameters		= 3,	// Query parameters can be named
+	fb_feature_session_reset		= 4,	// ALTER SESSION RESET is supported
+	fb_feature_read_consistency		= 5,	// Read consistency TIL is supported
+	fb_feature_statement_timeout	= 6,	// Statement timeout is supported
+	fb_feature_statement_long_life  = 7,    // Prepared statement can survive transaction end
 
 	info_provider_features_max      // Not really a feature. Keep this last.
 };
--- a/src/include/firebird/impl/inf_pub.h
+++ b/src/include/firebird/impl/inf_pub.h
@@ -182,6 +182,7 @@ enum info_provider_features // response to fb_info_provider_features
 	fb_feature_session_reset		= 3,	// ALTER SESSION RESET is supported
 	fb_feature_read_consistency		= 4,	// Read consistency TIL is supported
 	fb_feature_statement_timeout	= 5,	// Statement timeout is supported
+	fb_feature_statement_long_life  = 6,    // Prepared statement can survive transaction end
 
 	info_provider_features_max      // Not really a feature. Keep this last.
 };
--- a/src/jrd/extds/ExtDS.h
+++ b/src/jrd/extds/ExtDS.h
@@ -197,9 +197,6 @@ protected:
 };
 
 // Provider flags
-const int prvMultyStmts		= 0x0001;	// supports many active statements per connection
-const int prvMultyTrans		= 0x0002;	// supports many active transactions per connection
-const int prvNamedParams	= 0x0004;	// supports named parameters
 const int prvTrustedAuth	= 0x0008;	// supports trusted authentication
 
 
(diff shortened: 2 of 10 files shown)


src/include/firebird/impl/inf_pub.h
```diff
@@ -171,6 +174,16 @@ enum db_info_crypt			/* flags set in fb_info_crypt_state */
...
+enum info_provider_features // Flags set in response to fb_info_provider_features
+{
+	fb_feature_multi_statements 	= 0x00001,  // Multiple prepared statements in single attachment
+	fb_feature_multi_transactions	= 0x00002,	// Multiple concurrent transaction in single attachment
+	fb_feature_named_parameters		= 0x00004,	// Query parameters can be named
+	fb_feature_session_reset		= 0x00008,	// ALTER SESSION RESET is supported
+	fb_feature_read_consistency		= 0x00010,	// Read consistency TIL is supported
+	fb_feature_statement_timeout	= 0x00020	// Statement timeout is supported
+};
+
```
dyemanov, line 186: I'm not sure I like features being described as a bitmap. What if there will be more than 32 features in FB6? I'd rather consider returning them as a variable-length byte vector in the API. Inside the extds/* modules this vector can be parsed into a int32/int64 bit mask, if you prefer to preserve the current code "as is". This becomes an implementation detail that can be changed when needed.

aafemt: I was thinking about array of bytes but failed to invent a handy and error-prone way to declare the constants so I decided to use int64 bitmask for awhile and if number of features grow over 64 another inf item (fb_info_provider_features2) can be added. If you have a good idea - I'm all ears.

AlexPeshkoff: Byte vector is IMO preferred here cause it does not have endianess issues. It also limits set of features (255) but sooner of all that's enough. What about declaration - what's bad with trivial:
enum fb_features
{
  fb_feature_1 = 1,
  fb_feature_2,
  fb_feature_3
.......
};

aafemt: Seven bits are wasted. I'd prefer to keep data exchange via network dense. Besides, could you show how declaration of ENGINE_FEATURES constant would look like in this case?

Endianess is solved by isc_portable_integer()/ClumpletReader.

AlexPeshkoff:
On 2020-04-29 16:13, Dimitry Sibiryakov wrote:
>
> *@aafemt* commented on this pull request.
>
>
> In src/include/firebird/impl/inf_pub.h 
> <https://github.com/FirebirdSQL/firebird/pull/266#discussion_r417286241>:
>
> > @@ -171,6 +174,16 @@ enum db_info_crypt			/* flags set in fb_info_crypt_state */
>   	fb_info_crypt_process = 0x02
>   };
>   
> +enum info_provider_features // Flags set in response to fb_info_provider_features
> +{
> +	fb_feature_multi_statements 	= 0x00001,  // Multiple prepared statements in single attachment
> +	fb_feature_multi_transactions	= 0x00002,	// Multiple concurrent transaction in single attachment
> +	fb_feature_named_parameters		= 0x00004,	// Query parameters can be named
> +	fb_feature_session_reset		= 0x00008,	// ALTER SESSION RESET is supported
> +	fb_feature_read_consistency		= 0x00010,	// Read consistency TIL is supported
> +	fb_feature_statement_timeout	= 0x00020	// Statement timeout is supported
> +};
> +
>
> Seven bits are wasted. I'd prefer to keep data exchange via network 
> dense.
>

With actual size of that transfer <256 bytes we may not care too much 
about sent bits too much - anyway it fits into single packet.

> Besides, could you show how declaration of ENGINE_FEATURES constant 
> would look like in this case?
>

const char ENGINE_FEATURES[] = { fb_feature1, fb_feature3, fb_feature33 };

> Endianess is solved by isc_portable_integer()/ClumpletReader.
>

Pay attention to difference 'problem missing' vs. 'problem solved'. 
Second case often means at least one well hidden bug :)

aafemt:
It would make feature check code a little more complex but ok, I don't mind to implement it this way if an agreement is established.

AlexPeshkoff:
On 2020-04-29 18:15, Dimitry Sibiryakov wrote:
>
> It would make feature check code a little more complex but ok, I don't 
> mind to implement it this way if an agreement is established.
>

Let's wait what Yemanov thinks

dyemanov:
Yes, I agree to the byte sequence instead of the mask.

aafemt:
Is it better this way?

aafemt:
Currently used statement cache in ES is incompatible with those ODBC drivers that unprepare and destroy statements on commit/rollback. Should I add check for that here or create a separated branch?

src/include/firebird/impl/inf_pub.h
```diff
@@ -171,6 +174,18 @@ enum db_info_crypt			/* flags set in fb_info_crypt_state */
...
 
+enum info_provider_features // response to fb_info_provider_features
+{
+	fb_feature_multi_statements 	= 0,    // Multiple prepared statements in single attachment
+	fb_feature_multi_transactions	= 1,	// Multiple concurrent transaction in single attachment
+	fb_feature_named_parameters		= 2,	// Query parameters can be named
+	fb_feature_session_reset		= 3,	// ALTER SESSION RESET is supported
+	fb_feature_read_consistency		= 4,	// Read consistency TIL is supported
+	fb_feature_statement_timeout	= 5,	// Statement timeout is supported
+
```
dyemanov, line 186: While we don't have any rules engraved in stone in this regard, I'd prefer the numbering started with 1.

src/jrd/extds/ExtDS.h
```diff
@@ -197,9 +197,6 @@ class Provider : public Firebird::GlobalStorage
 };
 
 // Provider flags
-const int prvMultyStmts		= 0x0001;	// supports many active statements per connection
-const int prvMultyTrans		= 0x0002;	// supports many active transactions per connection
-const int prvNamedParams	= 0x0004;	// supports named parameters
 const int prvTrustedAuth	= 0x0008;	// supports trusted authentication
```
dyemanov, line 201: prvTrustedAuth should be re-numbered as 0x0001, I suppose.

src/jrd/extds/InternalDS.cpp
```diff
@@ -188,7 +188,10 @@ void InternalConnection::attach(thread_db* tdbb)
 	m_sqlDialect = (attachment->att_database->dbb_flags & DBB_DB_SQL_dialect_3) ?
 					SQL_DIALECT_V6 : SQL_DIALECT_V5;
 
-	m_features = conFtrFB4;
+	memset(m_features, false, sizeof(m_features));
+	static const info_provider_features features[] = ENGINE_FEATURES;
+	for (int i = 0; i < sizeof(features); i++)
+		setFeature(features[i]);
 }
```
dyemanov, line 195: Wouldn't raw memcpy(m_features, ENGINE_FEATURES, sizeof(ENGINE_FEATURES)) be simpler?

src/jrd/extds/IscDS.cpp
```diff
@@ -144,17 +144,20 @@ void IscConnection::attach(thread_db* tdbb)
 		}
 	}
 
-	char buff[16];
+	char buff[64];
 	{
```
dyemanov, line 148: Maybe BUFFER_TINY instead of "magic" constants? Just a suggestion.

src/jrd/extds/IscDS.cpp
```diff
@@ -144,17 +144,20 @@ void IscConnection::attach(thread_db* tdbb)
...
+		const unsigned char info[] = {isc_info_db_sql_dialect, fb_info_provider_features, isc_info_end};
 		m_iscProvider.isc_database_info(&status, &m_handle, sizeof(info), info, sizeof(buff), buff);
 	}
 	if (status->getState() & IStatus::STATE_ERRORS) {
 		raise(&status, tdbb, "isc_database_info");
 	}
 
+	memset(m_features, false, sizeof(m_features));
+    m_sqlDialect = 1;
+
```
dyemanov, line 160: Some extra whitespace seems to exist before m_sqlDialect.

dyemanov (changes requested):

aafemt:
> Wouldn't raw memcpy(m_features, ENGINE_FEATURES, sizeof(ENGINE_FEATURES)) be simpler?

No. They are incompatible. m_features is array of bool while ENGINE_FEATURES is array of 
info_provider_features and has different size.

src/include/firebird/impl/inf_pub.h
aafemt: As a trap for bugs, I see.

src/jrd/extds/ExtDS.h
aafemt: Actually, it should be removed at all because whether remote side support trusted auth or not is a matter of config, but I'll renumber it.

src/jrd/extds/IscDS.cpp
aafemt: That's strange, I was sure that IDE is configured to have 4 spaces per tab...

hvlad:
@dyemanov: the merge was too premature, there is a serious bug in this PR, see my comment to the inf.cpp 
https://github.com/FirebirdSQL/firebird/commit/2401b66e9eeea89914273b15be3777505e3771fd#r39898043

dyemanov:
Sorry, I've missed that issue. To avoid yet another PR, I suggest Dimitry to offer a correction patch that will be applied by us.

aafemt:
I'm already preparing the postfix PR. Patch that fix all three issues spotted by Vlad would be quite big.

PR #269, opened 2020-06-03
Fixed CORE-6322 - IN predicate and quantified comparison predicates behave incorrectly with NULL

asfernandes:


src/dsql/BoolNodes.cpp
```diff
@@ -204,35 +204,25 @@ bool BinaryBoolNode::executeAnd(thread_db* tdbb, jrd_req* request) const
...
-	// otherwise, return null
-	request->req_flags |= req_null;
-	return false;
+	if (value2 == false)
+		return TriState(false);	// at least one operand was false
+	else if (value1 == true && value2 == true)
+		return TriState(true);	// both true
+	else
+		return TriState();	// otherwise, return null
 }
```
dyemanov, line 223: No "else" after "return", please.

src/dsql/BoolNodes.cpp
```diff
@@ -250,32 +240,20 @@ bool BinaryBoolNode::executeOr(thread_db* tdbb, jrd_req* request) const
...
-	// restore saved NULL state
-
-	if (flags & req_null)
-		request->req_flags |= req_null;
-
-	return false;
+	if (value2 == true)
+		return TriState(true);
+	else
+		return value1.isUnknown() || value2.isUnknown() ? TriState() : TriState(false);
```
dyemanov, line 256: Same as above (else-return).

src/dsql/BoolNodes.cpp
```diff
@@ -1438,14 +1393,10 @@ BoolExprNode* NotBoolNode::copy(thread_db* tdbb, NodeCopier& copier) const
...
+TriState NotBoolNode::execute(thread_db* tdbb, jrd_req* request) const
 {
-	bool value = arg->execute(tdbb, request);
-
-	if (request->req_flags & req_null)
-		return false;
-
-	return !value;
+	const TriState value = arg->execute(tdbb, request);
+	return value.isUnknown() ? value : TriState(!value.value);
```
dyemanov, line 1399: If TriState was a real class, I'd suggest adding "bool operator!()" and simply returning !value. But now it's a typedef and so far I don't see other good reasons to change that. Sigh.

src/dsql/ExprNodes.cpp
```diff
@@ -11264,11 +11222,9 @@ dsc* SubQueryNode::execute(thread_db* tdbb, jrd_req* request) const
...
 
 			if (*invariant_flags & VLU_null)
-				request->req_flags |= req_null;
+				return nullptr;
 			else
-				request->req_flags &= ~req_null;
-
-			return (request->req_flags & req_null) ? NULL : desc;
+				return desc;
 		}
```
dyemanov, line 11228: Again, else-return.

src/dsql/ExprNodes.cpp
```diff
@@ -12833,11 +12775,9 @@ dsc* UdfCallNode::execute(thread_db* tdbb, jrd_req* request) const
...
 		{
 			if (invariantFlags & VLU_null)
-				request->req_flags |= req_null;
+				return nullptr;
 			else
-				request->req_flags &= ~req_null;
-
-			return (request->req_flags & req_null) ? NULL : &value->vlu_desc;
+				return &value->vlu_desc;
 		}
```
dyemanov, line 12781: else-return

src/jrd/extds/ExtDS.cpp
```diff
@@ -2260,12 +2260,9 @@ void Statement::doSetInParams(thread_db* tdbb, unsigned int count, const MetaNam
...
 			if (src)
-			{
-				if (request->req_flags & req_null)
-					src->setNull();
-				else
-					src->clearNull();
-			}
+				src->clearNull();
+
+			//// Should not src->setNull() be called when !src?
```
dyemanov, line 2265: I suppose the whole if (src) {} block is not needed anymore. And below !src should be used instead of srcNull.

src/jrd/fun.epp
```diff
@@ -423,14 +425,10 @@ void FUN_evaluate(thread_db* tdbb, const Function* function, const NestValueArra
 			// doesn't carry the null flag in the descriptor. Why Borland didn't
 			// set on such flag is maybe because it only has local meaning.
 			// This closes SF Bug #728839.
-			if ((request->req_flags & req_null) && !(input && (input->dsc_flags & DSC_null)))
-			{
-			    *arg_ptr++ = NULL;
-			}
+			if (!input && !(input && (input->dsc_flags & DSC_null)))
```
dyemanov, line 428: Condition should be either (!input) if we want to return DSC_null "as is", or something like this:
if (!input || (input->dsc_flags & DSC_null)) if we want to transform DSC_null into nullptr inside *arg_ptr.

dyemanov (changes requested):

src/dsql/BoolNodes.cpp
```diff
@@ -1611,7 +1678,10 @@ DmlNode* RseBoolNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScratch*
 	if (csb->csb_currentDMLNode)
 		node->ownSavepoint = false;
 
-	return node;
+	if (blrOp == blr_ansi_all)
+		return FB_NEW_POOL(pool) NotBoolNode(pool, node);
+	else
+		return node;
 }
```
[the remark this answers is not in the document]
asfernandes: Since types are different here, it's not possible to use the ternary.

I never knew that we had a standard code style to not write return-else-return :)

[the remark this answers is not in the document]
dyemanov: I remember Claudio removing unneeded "else"s across the whole codebase. It would be inappropriate to introduce them back ;-)

PR #270, opened 2020-06-04
Update install_windows_manually.txt for 4.0

gerdus:
Added Initializing the Security Database section. 
This was in 3.0 release notes but are not present in 4.0 release notes. I encountered this as a problem when testing out 4.0 Beta 2 from zip kit.
Content is mostly lifted verbatim from the 3.0 release notes.

Changed mention of 2.1 to 4.0 and updated bat file names.

### DIFF
--- a/doc/install_windows_manually.txt
+++ b/doc/install_windows_manually.txt
@@ -1,5 +1,5 @@
 --------------------------------------
-Firebird 4.0 manual installation notes
+Firebird manual installation notes
 --------------------------------------
 
 Contents
@@ -26,11 +26,11 @@ environment by executing the following steps:
     (It is also recommended that the FIREBIRD env var also be used in
     the PATH env var, rather than hard-coding a path.)
 
-  - change the current directory to where v4.0 files are located
+  - change the current directory to where files are located
 
   - Execute install_service.bat
 
-  - optionally, you can copy fbclient.dll to the OS system directory. 
+  - optionally, you can copy fbclient.dll to the OS system directory.
     To do so, use the provided instclient.exe tool.
 
 
@@ -49,7 +49,7 @@ Firebird is run as application by executing it with the -a switch:
 Installation with a different service name
 ------------------------------------------
 
-A feature of Firebird 4.0 is the option to run multiple Firebird services
+A feature of Firebird is the option to run multiple Firebird services
 simultaneously. To do this you need to make appropriate changes to the
 firebird.conf file and then run instsvc with the -name option. For example:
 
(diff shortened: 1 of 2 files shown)


doc/install_windows_manually.txt
```diff
@@ -1,5 +1,5 @@
 --------------------------------------
-Firebird 2.1 manual installation notes
+Firebird 4.0 manual installation notes
```
AlexPeshkoff, line 2: Here and later - I do not think it's worth specifying particular version here. To avoid misunderstanding or need to change this version every new release better say just:
Firebird manual installation notes
Same for other places where version 4 is mentioned.

doc/install_windows_manually.txt
```diff
@@ -8,6 +8,7 @@ Contents
 o Installation as a service
 o Execution as an application
 o Installation with a different service name
+o Initializing the Security Database
```
AlexPeshkoff, line 11: May be better put 'Initializing the Security Database' as a separate document in doc? That will let me mention it in error message and forget this problem. Moreover that text applies not to windows-only (described method is OS independent).

doc/install_windows_manually.txt
```diff
@@ -51,27 +49,62 @@ Firebird is run as application by executing it with the -a switch:
...
+The SQL user management commands will work with any open database. Because the sample database employee.fdb 
+is present in your installation and already aliased in databases.conf, it is convenient to use
+it for the user management task.
+
+1. Stop the Firebird server. Firebird 4 caches connections to the security database aggressively. The presence
+   of server connections may prevent isql from establishing an embedded connection.
+2. In a suitable shell, start an isql interactive session, opening the employee database via its alias:
+    > isql -user sysdba employee
+3. Create the SYSDBA user:
+    SQL> create or alter user SYSDBA password 'StrongPassword';
```
AlexPeshkoff, line 91: Please mention once more that instead words StrongPassword user should invent something unique, copy-paste here can play bad joke.

AlexPeshkoff (changes requested):

doc/install_windows_manually.txt
gerdus: I have done so.

doc/install_windows_manually.txt
gerdus: Split it out to "README.security_database.txt" to fit the style of the other files.

doc/install_windows_manually.txt
gerdus: Added warning before and after the SQL snippet.

AlexPeshkoff (approved):

PR #271, opened 2020-06-09
Small meta name

AlexPeshkoff:
On my mind use of hash to access metadata names in engine is complete. This saves a lot of memory - on artificial tests like opening employee a few hundred times memory consumption becomes more than twice less. Certainly that can't be expected with real tasks (first of all with big data pages cache) but taking into an account that we still keep separate copy of metadata cache per attachment memory is really saved.

Hashing is lock-less on a main execution path (mutex is used only when we switch to new segment of memory for names, i.e. new arena). With help of Pavel Zotov we have a load-test to check metadata names (http://web.firebirdsql.org/download/rabbits/alex/meta-storage-test.rar), so I have some reasons to think there are no races in the metadata names hashing. But please take one more look at it.


src/jrd/tra.h
```diff
@@ -110,7 +110,7 @@ typedef Firebird::BePlusTree<bid, bid, MemoryPool> FetchedBlobIdTree;
 
 struct CallerName
 {
-	CallerName(int aType, const Firebird::MetaName& aName, const Firebird::MetaName& aUserName)
+	CallerName(int aType, const MetaName& aName, const Firebird::MetaString& aUserName)
```
asfernandes, line 113: It seems weird name use a type and userName use another.

src/jrd/MetaName.h
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	int compare(const Firebird::AbstractString& s) const
+	{
+		return compare(s.c_str(), s.length());
+	}
+
+	int compare(const MetaName& m) const
+	{
+		return compare(m.begin(), m.length());
```
asfernandes, line 300: When you're comparing two MetaNames, can't you just compare the word pointer? Or at least, compare it first and if they're equal the bytes does not need to be compared.

src/common/classes/ClumpletWriter.h
```diff
@@ -72,10 +72,9 @@ class ClumpletWriter : public ClumpletReader
 	void insertInt(UCHAR tag, const SLONG value);
 	void insertBigInt(UCHAR tag, const SINT64 value);
 	void insertBytes(UCHAR tag, const void* bytes, FB_SIZE_T length);
-	void insertString(UCHAR tag, const AbstractString& str);
-	void insertString(UCHAR tag, const MetaName& str);
-	void insertString(UCHAR tag, const char* str);
 	void insertString(UCHAR tag, const char* str, FB_SIZE_T length);
+	void insertString(UCHAR tag, const char* str);
+	void insertString(UCHAR tag, char* str);
```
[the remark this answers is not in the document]
AlexPeshkoff: Required to have 
    template <typename S>
    void insertString(UCHAR tag, const S& str)
Without explicit overload compiler is using template for non-const calls with obvious compile failure. Replacing typename with class does not help.

src/jrd/Database.h
```diff
@@ -521,6 +531,8 @@ class Database : public pool_alloc<type_dbb>
 	ReplicaMode dbb_replica_mode;		// replica access mode
 
 	unsigned dbb_compatibility_index;	// datatype backward compatibility level
+	Dictionary dbb_dic;					// metanames dictionary
+	Firebird::InitInstance<KeywordsMap, KeywordsMapAllocator, Firebird::TraditionalDelete> dbb_keywords_map;
```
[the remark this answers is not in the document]
AlexPeshkoff: InitInstance used to combine 2 independent features: creation on first use and destruction  according to GlobalPtr rules.
Here I need delayed (till first use) creation. Destruction should be performed according to AutoPtr rules, therefore I've added an ability to separately control creation / destruction.

src/jrd/tra.h
AlexPeshkoff: Only seems. I had to keep user/role names to be represented in old way - i.e. MetaString. They are often and actively used before dbb and tdbb are created in attachDatabase/createDatabase calls and in services where we have no tdbb at all. Use of MetaName is impossible in such places.

In this particular place one name is object name (MetaName), other - user name (i.e. MetaString). Appears logical to me.

src/jrd/MetaName.h
AlexPeshkoff: That's done in operator== and operator!=, but you are right, some optimization is possible here.

asfernandes (approved):

PR #272, opened 2020-06-13
Include CLOOP build and interfaces generation to Windows build

aafemt:


hvlad:
I'm not sure IdlFbInterfaces.h should be removed from repository.
Currently we have control on changes in this file and if cloop start to produce something wrong we have a chance to detect it sooner.
Note, all files in include\gen is present in repository.

I'm not sure IdlFbInterfaces.h should be moved into "include\gen".
But if it is moved, you should fix include\firebird\Interface.h to reflect new placement of IdlFbInterfaces.h. Instead of change a lot of project files in examples.

aafemt:
If not removed,this file will be tracked by git and marked as modified after every build. Because it is generated, it belongs to gen directory along with Firebird.pas.

Interfaces are used by Firebird itself so if this file is generated wrongly, it will immediately affect build process so we'll notice that for sure. At least POSIX build works this way and I just made Windows one to follow.

As for examples projects it would be better for them to refer to release's include directory in output_XXX instead of src one but from what I see inside of make_examples.bat it look like examples require only make_boot.bat to be run while output directory is formed by make_all.bat. That's why I added gen dir to include path instead of changing it to output_win32\include\ etc.

hvlad:
> 
> 
> If not removed,this file will be tracked by git and marked as modified after every build. 

This is wrong. File will be marked as modified only if it was really modified. Look at other files in include\gen.

> Because it is generated, it belongs to gen directory along with Firebird.pas.

Formally, yes. Actually - not really. As for Firebird.pas - it is better to move it out of include folder, I think.
 
> Interfaces are used by Firebird itself so if this file is generated wrongly, it will immediately affect build process so we'll notice that for sure. 

Probably yes, but not necessary. I see no good reason to remove it out of version control.

> At least POSIX build works this way and I just made Windows one to follow.

What do you speak about ? POSIX build regenerates IdlFbInterfaces.h, ok. but it doesn't remove it from repository nor move it into include\gen folder.
 
> As for examples projects it would be better for them to refer to release's include directory in output_XXX instead of src one but from what I see inside of make_examples.bat it look like examples require only make_boot.bat to be run while output directory is formed by make_all.bat. That's why I added gen dir to include path instead of changing it to output_win32\include\ etc.

You *broke* firebird\Interface.h. And then added not needed include directory to the lot of places as workaround. This is wrong and must be fixed. Changing INCLUDE from src\include to output\include can not fix it and have nothing common with this issue.

aafemt:
> Look at other files in include\gen.

include\gen folder is added to .gitignore list as whole. And it contains only those 
generated header that have chicken-and-egg problem i.e. are required for generating them. 
IdlInterfaces.h is not.


> it doesn't remove it from repository nor move it into include\gen folder.

Only because this file wasn't generated during Windows build so removing of it would break 
the build. And this is exactly the problem which this PR is supposed to fix.

> You broke firebird\Interface.h. And then added not needed include directory to the lot of places as workaround.

Nope. This directory is on include list of all project so I added a missing directory. But 
you are right, it was a wrong decision (see below).

> Changing INCLUDE from src\include to output\include can not fix it and have nothing common with this issue.

On contrary: examples are meant to be built by end users who has at hand only distribution 
version of headers. Building of them is supposed to check that these headers are 
sufficient and usable so it has to use output\include.

hvlad:
> 
>> Look at other files in include\gen.

> include\gen folder is added to .gitignore list as whole. 

Wrong again. Only root/gen is ignored, you may check commits history in src/include/gen.

>> it doesn't remove it from repository nor move it into include\gen folder.

> Only because this file wasn't generated during Windows build so removing of it would break the build. And this is exactly the problem which this PR is supposed to fix.

This introduce another issue - no version control over very important file. 

>> You broke firebird\Interface.h. And then added not needed include directory to the lot of places as workaround.

> Nope. This directory is on include list of all project so I added a missing directory. 

Demagogy. You **broke** existing code and "fixed" it using wrong workaround.

> But you are right, it was a wrong decision (see below).

>> Changing INCLUDE from src\include to output\include can not fix it and have nothing common with this issue.

> On contrary: examples are meant to be built by end users who has at hand only distribution version of headers. Building of them is supposed to check that these headers are sufficient and usable so it has to use output\include.

End users have NO "output" folder. 

You again argue just for the argue.
I object your PR in current state.

aafemt:
> Wrong again. Only root/gen is ignored, you may check commits history in src/include/gen.

   I checked that git doesn't react on IdlInterfaces.h staying there after build. Isn't it 
enough? Git allows to commit files in ignored directories so history proves nothing.

> This introduce another issue - no version control over very important file.

   Very important file is FirebirdInterface.idl the rest is just derived. If you want 
control over them feel free to put some static_assert's checking critical points of them 
if you know any. It works well for ods.h

> End users have NO "output" folder.

   "output" folder is exactly content of ZIP and ISS distributions so they DO have it.

hvlad:
1. IdlInterfaces.h should not be removed from git
2. If it will be moved into include\gen, files that includes it must be fixed accordingly
3. include/gen should not be added into projects
4. Feel free to fix projects in examples to include output\include instead of src\include, but remember (3) above

hvlad (changes requested):
See my prior comments

aafemt:
> include/gen should not be added into projects

It is already there from the beginning and it was not me who added it.

hvlad:
Not sure I like that DbCrypt example now can't be built when it is in Firebird code tree (it is buildable when in output only).
How do you offer to work with it to Firebrid developers ? Use copy of sources in output\examples and then copy changes back into the firebird\examples ?

Example ExtAuth now can't be built at all except of when called from make_examples - which defines "libtom_include" and "libtom_libs" environment vars (now used in project prorepties). I.e. even end-users can't just open solution in IDE and build it.

And I still insist on points 1 and 2 at https://github.com/FirebirdSQL/firebird/pull/272#issuecomment-643758495.

aafemt:
> Not sure I like that DbCrypt example now can't be built when it is in Firebird code tree (it is buildable when in output only).

I raised a topic in devel maillist asking whether these projects are supposed for distribution and the answer was "yes, why not". In this case they must be buildable in end-user environment the same way as all other examples.
If they cannot be built in end-user environment they must not be included into distribution packages.

> How do you offer to work with it to Firebrid developers ? Use copy of sources in output\examples and then copy changes back into the firebird\examples ?

It is a matter of taste. I edited them in firebird\examples and then run make_examples.bat to check the result. I doubt that there is a way to make them buildable in IDE in development tree and in end-user environment at the same time.

> Example ExtAuth now can't be built at all except of when called from make_examples - which defines "libtom_include" and "libtom_libs" environment vars (now used in project prorepties). I.e. even end-users can't just open solution in IDE and build it.

Yes and it is written in readme. Alternatively we could add tom* libraries into our distribution package but I'm not sure if it is desirable.

> And I still insist on points 1 and 2 at [#272 (comment)](https://github.com/FirebirdSQL/firebird/pull/272#issuecomment-643758495).

Sure. But I in turn still insist that generated files must not be in the tree and no additional fixes are required because support for generated headers is already done in all projects way in the past. The topic in the maillist is started, let's see others' opinions.

hvlad:
> 
> 
> > Not sure I like that DbCrypt example now can't be built when it is in Firebird code tree (it is buildable when in output only).
> 
> I raised a topic in devel maillist asking whether these projects are supposed for distribution and the answer was "yes, why not". 

But you didn't mention that projects can no longer be used by developers.

> In this case they must be buildable in end-user environment the same way as all other examples.
> If they cannot be built in end-user environment they must not be included into distribution packages.

Using your logic - these projects must not be included into Firebird tree (as it is not usable there with your patch)

> > How do you offer to work with it to Firebrid developers ? Use copy of sources in output\examples and then copy changes back into the firebird\examples ?
> 
> It is a matter of taste. I edited them in firebird\examples and then run make_examples.bat to check the result. I doubt that there is a way to make them buildable in IDE in development tree and in end-user environment at the same time.

There must be such a way. You started all this stuff - and it is your responsibility to find it and to not broke a things, isn't is ? 

> > Example ExtAuth now can't be built at all except of when called from make_examples - which defines "libtom_include" and "libtom_libs" environment vars (now used in project prorepties). I.e. even end-users can't just open solution in IDE and build it.
> 
> Yes and it is written in readme. 

Not excuses, I think.

> Alternatively we could add tom* libraries into our distribution package but I'm not sure if it is desirable.

It should be considered as an option.

> > And I still insist on points 1 and 2 at [#272 (comment)](https://github.com/FirebirdSQL/firebird/pull/272#issuecomment-643758495).
> 
> Sure. But I in turn still insist that generated files must not be in the tree and no additional fixes are required because support for generated headers is already done in all projects way in the past. The topic in the maillist is started, let's see others' opinions.

You know mine one.

aafemt:
> But you didn't mention that projects can no longer be used by developers.

They never were useful for developers except of creating of pre-built demo plugins that 
still works.

> Using your logic - these projects must not be included into Firebird tree (as it is not usable there with your patch)

Using my logic - they are not generated so they MUST be included to the tree.
But you may be right, I also thought about removing them so this option worth to be 
considered again.

aafemt:
As requested I returned back generated header and as suggested by Alex added generated Pascal file.

aafemt:
Any further issue preventing merge?

hvlad (commented):
1. makefile.nmake used to generate interface headers should have more specific name.

2. I'm not sure def_awk and isc_grep intermediate files should be put into FB_TEMP_DIR.
I think it should be at FB_TEMP_DIR\FB_CONFIG\misc (or FB_TEMP_DIR\Release\misc).

3. Why do you decide to leave files from previous build in FB_GEN_DIR subdirectories ?

aafemt:
1. Generating of headers is just one of possible tasks for this makefile. In the next PRs I planned to expand it to generate parser and preprocess ESQL files as well.
2. These files are config and platform independent. There is no point to build them for x86 and x64 separately, single build for host system is enough.
3. The same reason: not to waste time during following builds if sources were not changed. Whole build on my notebook takes more than two hours. Every optimization of this process is good.

hvlad:
>     1. Generating of headers is just one of possible tasks for this makefile. In the next PRs I planned to expand it to generate parser and preprocess ESQL files as well.

Anyway, it should not have such too generic name as you offer.
 
>     2. These files are config and platform independent. There is no point to build them for x86 and x64 separately, single build for host system is enough.

It is not about x64\win32 (which anyway builds separately). It is about debug\release and gargabe in not appropriate place.

>     3. The same reason: not to waste time during following builds if sources were not changed. Whole build on my notebook takes more than two hours. Every optimization of this process is good.

You save almost nothing by this change. File's are preprocessed anyway. 
Though I'm not sure why it was there initially and if it is really necessary.
Anyway such not related changes into PR is not good idea.
I'm not asking to undo it but in the future, please, avoid mixing few thing at one PR\commit.

So, please fix (1) and (2) and I'll accept this PR.

aafemt:
> Anyway, it should not have such too generic name as you offer.

I have no better idea. What name could you suggest?

> It is not about x64\win32 (which anyway builds separately). It is about debug\release and gargabe in not appropriate place.

These utilities have no debug/release variants. If temp directory is not an appropriate place for temp files what place is better?

hvlad:
> > Anyway, it should not have such too generic name as you offer.
> 
> I have no better idea. What name could you suggest?
 
I don't know your plans, currently it could be named `geniface.nmake` or so on.

> > It is not about x64\win32 (which anyway builds separately). It is about debug\release and gargabe in not appropriate place.
> 
> These utilities have no debug/release variants. If temp directory is not an appropriate place for temp files what place is better?

Read above: FB_TEMP_DIR\FB_CONFIG\misc or FB_TEMP_DIR\Release\misc

aafemt:
> currently it could be named `geniface.nmake` or so on.

Ok, I called it gen_helper.nmake, I hope you like it.

hvlad (approved):

extern/cloop/src/cloop/cloop.vcxproj
```diff
@@ -20,8 +20,14 @@
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <Keyword>Win32Proj</Keyword>
+    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <PlatformToolset>v141</PlatformToolset>
```
asfernandes, line 24: This change broke VS 2015 build.

hvlad: Probably small makefile to build CLOOP would be best to fix it.

aafemt: > 
> 
> This change broke VS 2015 build.

Fortunately official compiler for Firebird 4 is VS 2017 so 2015 can be simply dropped.

PR #273, opened 2020-06-15
Postfix for PR#266 problems spotted by Vlad Khorsun

aafemt:


aafemt:
I cannot understand what is wrong with Travis build. It looks like parallel build issue but I'm not sure. Can someone help me?..

PR #274, opened 2020-06-15
Fix issues spotted by Vlad Khorsun in PR#266

aafemt:


src/jrd/inf.cpp
```diff
@@ -226,8 +226,9 @@ void INF_database_info(thread_db* tdbb,
  **************************************/
  	CHECK_INPUT("INF_database_info");
 
+	const CountsBuffer::size_type bufferLength = BUFFER_SMALL;
 	CountsBuffer counts_buffer;
-	UCHAR* buffer = counts_buffer.getBuffer(BUFFER_SMALL);
+	UCHAR* buffer = counts_buffer.getBuffer(bufferLength, false);
```
aafemt, line 231: One thing worry me in this line: currently getBuffer always return inline storage because CountsBuffer is declared with initial capacity BUFFER_SMALL but nothing ensure that. If getBuffer ever return a pointer to dynamic storage and it is reallocated in process, value of buffer will become invalid.

src/jrd/extds/InternalDS.cpp
```diff
@@ -190,7 +190,7 @@ void InternalConnection::attach(thread_db* tdbb)
 
 	memset(m_features, false, sizeof(m_features));
 	static const info_features features[] = ENGINE_FEATURES;
-	for (int i = 0; i < sizeof(features); i++)
+	for (int i = 0; i < sizeof(features)/sizeof(info_features); i++)
 		setFeature(features[i]);
```
dyemanov, line 194: We have FB_NELEM macro exactly for this purpose.

src/jrd/inf.cpp
dyemanov: After every counts_buffer's resize variable buffer should be reassigned, see handling of isc_info_read_seq_count and others below.

aafemt: I don't see it in isc_info_db_id handler.

src/jrd/extds/InternalDS.cpp
aafemt: Oops, I forgot about it. Thanks.

src/jrd/inf.cpp
dyemanov: Right, it deserves fixing too.

aafemt: Here or another day?

dyemanov: Better now, if you don't mind.

aafemt: Ok.

PR #275, opened 2020-06-23
Use proper type for inf buffers

aafemt:
Also one forgotten rename.

### DIFF
--- a/src/include/firebird/impl/inf_pub.h
+++ b/src/include/firebird/impl/inf_pub.h
@@ -184,7 +184,7 @@ enum info_features // response to fb_info_features
 	fb_feature_statement_timeout	= 6,	// Statement timeout is supported
 	fb_feature_statement_long_life  = 7,    // Prepared statements are not dropped on transaction end
 
-	info_provider_features_max      // Not really a feature. Keep this last.
+	fb_feature_max      // Not really a feature. Keep this last.
 };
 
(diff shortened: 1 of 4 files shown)


src/include/firebird/impl/inf_pub.h
```diff
@@ -184,7 +184,7 @@ enum info_features // response to fb_info_features
 	fb_feature_statement_timeout	= 6,	// Statement timeout is supported
 	fb_feature_statement_long_life  = 7,    // Prepared statements are not dropped on transaction end
 
-	info_provider_features_max      // Not really a feature. Keep this last.
+	info_features_max      // Not really a feature. Keep this last.
```
asfernandes, line 187: I think all items of enums should have the same prefix.

aafemt: Yes but this one is not quite item of enum and used to be used in code like "info_features array[info_features_max]".

asfernandes: It is still an item of the enum.

AlexPeshkoff: As a compromise suggest fb_features_max

aafemt:
AppVeyor again failed to upgrade environment. WTH..?

AlexPeshkoff (approved):

PR #276, opened 2020-06-29
Fix nodDesc of NullNode being changed after its initial construction.

asfernandes:


dyemanov (commented):
I find the logic inside DsqlMapNode (setNullable / clearNull) a little bit hackery, but generally I'm OK with the PR. However, AFAIK initializers inside IF statements are C++17 feature, while we're limited to C++11. I suppose these changes should be reverted back.

asfernandes:
> I find the logic inside DsqlMapNode (setNullable / clearNull) a little bit hackery, but generally I'm OK with the PR. However, AFAIK initializers inside IF statements are C++17 feature, while we're limited to C++11. I suppose these changes should be reverted back.

What is C++17 is `if (auto x = ...; x)` syntax. I just use `if (auto n = ...)` which AFAIK is C++11.

We have already some of use of this in master.

aafemt:
> I just use `if (auto n = ...)` which AFAIK is C++11.

Nope. https://en.cppreference.com/w/cpp/language/if
Init statement is only since C++17.

asfernandes:
> > I just use `if (auto n = ...)` which AFAIK is C++11.
> 
> Nope. https://en.cppreference.com/w/cpp/language/if
> Init statement is only since C++17.

init-statement ends with semi-colon. The construct I use does not, it's not a statement.

Both clang++ and g++ agrees the syntax is valid in C++11.

When I use a init-statement they say it's C++17.

asfernandes:
Accordingly to this answer https://stackoverflow.com/questions/7836867/c-variable-declaration-in-if-expression we see that its even C++03. The declaration part is part of "condition" said in cppreference.

PR #277, opened 2020-07-14
Fix CORE-6338: Mark attachment of sweep thread with ATT_sweep_thread flag and protect it with mutex to avoid races with JRD_release_attachment

artyom-smirnov:
This is draft solution to fix CORE-6338. I've left almost all sweep thread starting logic intact but wrapped it with class to control sweeper attachment thread and added some locking to avoid racing between sweeper starting and attachment release. Please review.


AlexPeshkoff:
First question - does it help?
If yes - looks like ATT_crypt_thread & ATT_sweep_thread flags should be replaced with single ATT_internal_thread flag. All attachments, marked by it, are gathered into some container and if nothing left except them - closed. Appropriate mutex goes out of Crypto manager (and Sweeper) and placed into Database - one for all. That becomes universal solution.

artyom-smirnov:
Yes, I've ran query/shutdown script for several hours without crash/hang just with sweep and with sweep+CM threads. I will look into more universal solution.

AlexPeshkoff:
Artyom, I want to close PR cause changes discussed here (and based on ideas from PR) are in code base for a long time. Do you see any problems with it?

PR #279, opened 2020-08-18
Added new messages when duplicated trigger, domain, blob filter or pa…

Noremos:
Added new error messages:
Domain already exists
Trigger already exists
Package already exists
Filter already exists

Previous error message was like this:
-CREATE DOMAIN P failed
-violation of PRIMARY or UNIQUE KEY constraint "RDB$INDEX_2" on table "RDB$FIELDS" 
-Problematic key value is ("RDB$FIELD_NAME" = 'P')


aafemt:
If you are going to port this code to Firebird 4 it will be better to handle isc_unique_key_violation in object creation code than use select for check there.

dyemanov (changes requested):

UnknownAr:
I fixed object types and moved check try-catch block

aafemt:
Even in try-catch block this code has the same problem: attempt to create objects with the same name from parallel sessions wont be detected because you use SELECT for it. You already have unique violation code. All you need is to check violated constraint name.

PS: I still doubt that this PR should go to stable v3 branch.

AlexPeshkoff:
On 2020-08-19 12:42, Dimitry Sibiryakov wrote:
>
> Even in try-catch block this code has the same problem: attempt to 
> create objects with the same name from parallel sessions wont be 
> detected because you use SELECT for it. You already have unique 
> violation code. All you need is to check violated constraint name.
>

Even if I try to forget about parallel sessions - suggested in PR code 
is anyway overcomplication and bad solution.

> PS: I still doubt that this PR should go to stable v3 branch.
>

Agreed, should go only to master.

aafemt:
> suggested in PR code is anyway overcomplication and bad solution.

It is just a copy-paste of existing code so don't blame them.

UnknownAr:
> > IMHO using 9th value with `RDB$FUNCTION_NAME` is more stable than use 3rd value with `RDB$INDEX_45`
> 
> Nope. Searching for a substring in a string that contains user-defined values is always worse. Index names (numbers) for system constraints are fixed in idx.h, they are guaranteed to be stable.

I replaced FUNCTION_NAME by INDEX_45. Any another issues?

UnknownAr:
> There is no need to use string and find(). strcmp() is enough.

Replaced it.

aafemt:
Just in case: in "catch" block "throw" don't need parameter but in this case it doesn't hurt.

PR #282, opened 2020-08-26
Move Config to Firebird namespace

aafemt:


src/burp/mvol.cpp
```diff
@@ -203,7 +203,7 @@ static void calc_hash(Firebird::string& valid, Firebird::IDbCryptPlugin* plugin)
 //____________________________________________________________
 //
 //
-static Firebird::IKeyHolderPlugin* mvol_get_holder(BurpGlobals* tdgbl, Firebird::RefPtr<const Config>& config)
+static Firebird::IKeyHolderPlugin* mvol_get_holder(BurpGlobals* tdgbl, Firebird::RefPtr<const Firebird::Config>& config)
```
asfernandes, line 206: You should not qualify with `Firebird::` when you are already in the Firebird namespace.

aafemt: Yes but this line is not in Firebird namespace and all the rest of identifiers in it are qualified.

asfernandes: Sorry, I've meant in src/common/classes/GetPlugins.h. (but I didn't reviewed things completelly)

aafemt: >  I've meant in src/common/classes/GetPlugins.h.

Yes this is the only place where such unnecessary qualification has to be done because of https://github.com/FirebirdSQL/firebird/pull/280#discussion_r475358044 #

AlexPeshkoff:
On 2020-08-26 16:44, Dimitry Sibiryakov wrote:
>
> *@aafemt* commented on this pull request.
>
>
> In src/burp/mvol.cpp 
> <https://github.com/FirebirdSQL/firebird/pull/282#discussion_r477311210>:
>
> > @@ -203,7 +203,7 @@ static void calc_hash(Firebird::string& valid, Firebird::IDbCryptPlugin* plugin)
>   //____________________________________________________________
>   //
>   //
> -static Firebird::IKeyHolderPlugin* mvol_get_holder(BurpGlobals* tdgbl, Firebird::RefPtr<const Config>& config)
> +static Firebird::IKeyHolderPlugin* mvol_get_holder(BurpGlobals* tdgbl, Firebird::RefPtr<const Firebird::Config>& config)
>
>     I've meant in src/common/classes/GetPlugins.h.
>
> Yes this is the only place where such unnecessary qualification has to 
> be done because of #280 (comment) 
> <https://github.com/FirebirdSQL/firebird/pull/280#discussion_r475358044> #
>

I'm almost sure it's not needed in GetPlugins after subj.
When choosing between using namespace X; and namespace Y {} compiler 
should choose Y.

aafemt:
> When choosing between using namespace X; and namespace Y {} compiler should choose Y.

May be. But call of GetPlugins in Publisher.cpp is performed in anonymous namespace with both "using namespace Firebird" and "using namespace Replication" before it. I would predict "ambiguous name" error in this case.

aafemt:
> I would predict "ambiguous name" error in this case.

Let me check that...

AlexPeshkoff:
This works for me:

#include <stdio.h>

namespace X {
     class A
     {
     public:
         A() {printf("A in X\n");}
     };
}

namespace Y {
     class A
     {
     public:
         A() {printf("A in Y\n");}
     };
}

using namespace X;
using namespace Y;

namespace X {        // remove this namespace statement and get 
'ambiguous class' error
void test()
{
     A a;
}
}

int main()
{
     test();
     return 0;
}

aafemt:
Indeed it seems to work.

aafemt:
I fixed all unnecessary qualification that found. Anyone can see more...?

dyemanov (approved):
I see no problems with this PR. Unless someone has any comments/objections, I will merge it tomorrow.

PR #283, opened 2020-09-18
Fixing broken alpine build.

faserg1:
The error in alpine build was:
```/build/fb3/src/common/cvt.cpp: In function 'double CVT_get_double(const dsc*, ErrorFunction)':
/build/fb3/src/common/cvt.cpp:1271:19: error: call of overloaded 'isinf(double&)' is ambiguous
 1271 |    if (isinf(value))
      |                   ^
In file included from /usr/include/c++/9.3.0/math.h:36,
                 from /build/fb3/src/common/../common/classes/FpeControl.h:32,
                 from /build/fb3/src/common/cvt.cpp:54:
/usr/include/c++/9.3.0/cmath:861:5: note: candidate: 'typename __gnu_cxx::__enable_if<std::__is_arithmetic<_Tp>::__value, int>::__type std::isinf(_Tp) [with _Tp = double; typename __gnu_cxx::__enable_if<std::__is_arithmetic<_Tp>::__value, int>::__type = int]'
  861 |     isinf(_Tp __f)
      |     ^~~~~
In file included from /build/fb3/src/common/cvt.cpp:54:
/build/fb3/src/common/../common/classes/FpeControl.h:231:13: note: candidate: 'bool isinf(F) [with F = double]'
  231 | inline bool isinf(F x)
      |             ^~~~~
```

Because of specific of GNU implementation of standard library (including `<cmath>` in `<math.h>`), define `isinf` dissapears, but the standard function still exists by `using std::isinf;`. So, I supposed, that we must not declare own isinf function, if defined `_GLIBCXX_CMATH`.


faserg1:
> Can't you put this in main.yml? It is easy to adjust commands when they are in the same file like others builds.

Don't sure, because it run in container. It will be ok, if I move it in separate job?

asfernandes:
> > Can't you put this in main.yml? It is easy to adjust commands when they are in the same file like others builds.
> 
> Don't sure, because it run in container. It will be ok, if I move it in separate job?

I think you can still use the same job using an `include`. See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations

Then you would also need to test the container in an if expression to differentiate alpine and ubuntu.

faserg1:
Merged in one file.

.github/workflows/main.yml
```diff
@@ -25,19 +30,31 @@ jobs:
...
         (cd Firebird-[0-9]*; sudo ./install.sh -silent)
 
+    - name: Prepare (Linux, Alpine)
+      if: matrix.os == 'ubuntu-18.04' && matrix.container == 'alpine'
+      run: apk update && apk --no-cache --update add build-base libtool git autoconf automake zlib-dev icu-dev ncurses-dev libedit-dev linux-headers tar
+
+    - name: Build (Linux, Alpine)
+      if: matrix.os == 'ubuntu-18.04' && matrix.container == 'alpine'
+      run: |
+        ./autogen.sh --enable-binreloc-threads --with-builtin-tomcrypt -with-builtin-tommath --prefix=/opt/firebird
```
asfernandes, line 53: Dont this need to be "--with-builtin-tommath" (two "-")?

faserg1: Oh, sure. But it recognize it even with one `-`. But as I can see, there is no need in tomcrypt.
```
Running ./configure --enable-binreloc-threads --with-builtin-tomcrypt -with-builtin-tommath --prefix=/opt/firebird
configure: WARNING: unrecognized options: --with-builtin-tomcrypt
```
I'll remove tomcrypt from here

src/common/classes/FpeControl.h
```diff
@@ -223,7 +223,7 @@ inline bool isinf(double x)
 	return (!_finite (x) && !isnan(x));
 }
 #elif !defined(DARWIN)
-#ifndef isinf
+#if !defined (isinf) && !defined(_GLIBCXX_CMATH)
```
asfernandes, line 226: It's fine for me to merge in regard to the github action. Let's see @AlexPeshkoff about this line change.

AlexPeshkoff:
After return from vacations 5 Oct

Sent from MailDroid

src/common/classes/FpeControl.h
AlexPeshkoff: We already have over-complicated isinf() in FB3, with suggested here change it becomes even worse. IMO that's not good. Please check https://github.com/FirebirdSQL/firebird/tree/isinf - it should build on alpine w/o problems. After your confirmation I will merge that branch into main tree, and you will be able to add github action for alpine.

faserg1: @AlexPeshkoff Everything is ok

asfernandes:
@faserg1 can you prepare pull request adding github actions alpine build to master?

Difference is that in v3 branch ubuntu build was disabled (due to frequent crashes) and in master it's enabled.

faserg1:
> @faserg1 can you prepare pull request adding github actions alpine build to master?
> 
> Difference is that in v3 branch ubuntu build was disabled (due to frequent crashes) and in master it's enabled.

Sure

PR #284, opened 2020-09-21
Cascading replication option

aafemt:


aafemt:
Retargeted to master.

PR #285, opened 2020-09-27
PSQL and SQL profiler

asfernandes:


aafemt:
If start_session returns session id and multiple session can be started simultaneously (according to the table RDB$PROFILE_SESSIONS), shouldn't all other procedures accept session id as a parameter to know which session to work with?

asfernandes:
> If start_session returns session id and multiple session can be started simultaneously (according to the table RDB$PROFILE_SESSIONS), shouldn't all other procedures accept session id as a parameter to know which session to work with?

Sessions cannot be started simultaneously. Only last started session is active.

Docs says that start_session could be called with a session already started and it has the same effect as finishing the first before start the second.

aafemt:
In this case shouldn't RDB$PROFILE_SESSIONS also include finish time...?

doc/sql.extensions/README.profiler.md
```diff
@@ -0,0 +1,164 @@
...
+## Table `RDB$PROFILE_STATS`
+
+ - `RDB$PROFILE_SESSION_ID` type `BIGINT` - Profile session ID
+ - `RDB$PROFILE_REQUEST_ID` type `BIGINT` - Request ID
+ - `RDB$LINE` type `INTEGER` - Line number of the statement
+ - `RDB$COLUMN` type `INTEGER` - Column number of the statement
+ - `RDB$COUNTER` type `BIGINT` - Number of executed times of the statement
+ - `RDB$MIN_TIME` type `BIGINT` - Minimal time (in nanoseconds) of a statement execution
+ - `RDB$MAX_TIME` type `BIGINT` - Maximum time (in nanoseconds) of a statement execution
+ - `RDB$ACCUMULATED_TIME` type `BIGINT` - Accumulated execution time (in nanoseconds) of the statement
```
dyemanov, line 164: RDB$TOTAL_TIME maybe?

asfernandes:
> In this case shouldn't RDB$PROFILE_SESSIONS also include finish time...?

Maybe... And also a description/comment passed in START_SESSION.

doc/sql.extensions/README.profiler.md
asfernandes: > RDB$TOTAL_TIME maybe?

Agree.

asfernandes: Should it be MIN/MAX TIME our MINIMUM/MAXIMUM?

dyemanov: MIN/MAX is fine to me.

livius2: Can you also add AVG_TIME?

asfernandes:
I propose change this PR creating a "profiler API".

`RDB$PROFILER` package will still be an engine thing, but `RDB$PROFILE_*` tables will not.

Tables will be created (explicitly or on-demand) by the profiler plugin, and as non-system tables, will be backed-up as user tables.

`RDB$PROFILER` will call plugin methods. `RDB$PROFILER.START_SESSION` will receive list of profiler plugins and it will drive communication with them.

Engine as well will interact with profiler plugins active for the session.

I'm not yet detailing plugin API, but it will support current operations.

That will open path for real-time visual profile.

A plugin could (directly to screen in case of embedded, or via inter-process/network communication) visually display what engine is being run, instead of store things.

Plugin may also store data in different places than the database.

livius2:
HiI like the idea of "none system" tables but i dislike then RDB$ prefix.Regards,Karol Bieniaszewski
-------- Oryginalna wiadomość --------Od: Adriano dos Santos Fernandes ***@***.***> Data: 02.09.2021  19:48  (GMT+01:00) Do: FirebirdSQL/firebird ***@***.***> DW: Karol Bieniaszewski ***@***.***>, Comment ***@***.***> Temat: Re: [FirebirdSQL/firebird] PSQL and SQL profiler (#285) 
I propose change this PR creating a "profiler API".
RDB$PROFILER package will still be an engine thing, but RDB$PROFILE_* tables will not.
Tables will be created (explicitly or on-demand) by the profiler plugin, and as non-system tables, will be backed-up as user tables.
RDB$PROFILER will call plugin methods. RDB$PROFILER.START_SESSION will receive list of profiler plugins and it will drive communication with them.
Engine as well will interact with profiler plugins active for the session.
I'm not yet detailing plugin API, but it will support current operations.
That will open path for real-time visual profile.
A plugin could (directly to screen in case of embedded, or via inter-process/network communication) visually display what engine is being run, instead of store things.
Plugin may also store data in different places than the database.

asfernandes:
Replaced by #7086

PR #290, opened 2020-10-21
Don't limit by default length of logged SQL text and query arguments.

dmitry-starodubov:
It's more suitable and there is no restriction on logged record length now.


MaceWindu:
@firebirdsql, guys, sorry for pinging here, but could you please turn your tracker on. I want to report issue for DNET and tracker is down for at least two last days.
BTW, why not accepting issues through github (both firebird and dnet provider)?

asfernandes:
> @FirebirdSQL, guys, sorry for pinging here, but could you please turn your tracker on. I want to report issue for DNET and tracker is down for at least two last days.
> BTW, why not accepting issues through github (both firebird and dnet provider)?

The migration from jira to github issues is under development.

hvlad (commented):
Trace output could be very big when lot of events traced. Also, it affects performance of whole server.
Limit on logged SQL text and param values allows to reduce trace output and lower common disk and CPU load. It is especially useful for newbies. 
I see no good reason to change it.
Anyone could set value (s)he like in own trace configuration.

PR #298, opened 2020-11-27
Reserve facility 27 for r2dbc-firebird

mrotteveel:
As discussed on firebird-devel


src/msgs/facilities2.sql
```diff
@@ -21,6 +21,7 @@ set bulk_insert INSERT INTO FACILITIES (LAST_CHANGE, FACILITY, FAC_CODE, MAX_NUM
 ('2016-03-20 15:30:00', 'NBACKUP', 24, 80)
 ('2009-07-20 07:55:48', 'FBTRACEMGR', 25, 41)
 ('2015-07-27 00:00:00', 'JAYBIRD', 26, 1)
+('2020-11-27 00:00:00', 'R2DBC_FIREBIRD', 27, 1)
 stop
```
AlexPeshkoff, line 25: Why 1 - not 0 ?
BTW, in JAYBIRD it should be 0 too.

mrotteveel: The JAYBIRD entry was added by Vlad, and I followed that example. It aksi follows the convention that matches with the value of GPRE (facility 4) and INSTALL (facility 10) which also have no messages. Also note that facilities 3, 7, 8, 13, 14, 15, 18, 21, 22, 23, 24 and 25 also do not have a an error with number 0 (most start with 1, except 7, 14 and 15 which start with 2, 100 and 147 respectively).

This indicates that usually the first message number is 1, not 0, so using MAX_NUMBER = 1 when there is no message would seem to be the appropriate value. However, if you want me to, I can change this.

mrotteveel: @AlexPeshkoff Do you want me to make the change (and if so, only for R2DBC_FIREBIRD, or also for GPRE, INSTALL and JAYBIRD)? Or can this be merged?

PR #306, opened 2020-12-31
Make UID output comply RFC 4122

aafemt:


dyemanov:
Tracker ticket would be appreciated.

aafemt:
[CORE-6472](http://tracker.firebirdsql.org/browse/CORE-6472)

src/jrd/SysFunction.cpp
```diff
@@ -6264,15 +6259,16 @@ dsc* evlUuidToChar(thread_db* tdbb, const SysFunction* function, const NestValue
...
 
-	char buffer[GUID_BUFF_SIZE];
-	sprintf(buffer, GUID_NEW_FORMAT,
-		USHORT(data[0]), USHORT(data[1]), USHORT(data[2]), USHORT(data[3]), USHORT(data[4]),
-		USHORT(data[5]), USHORT(data[6]), USHORT(data[7]), USHORT(data[8]), USHORT(data[9]),
-		USHORT(data[10]), USHORT(data[11]), USHORT(data[12]), USHORT(data[13]), USHORT(data[14]),
-		USHORT(data[15]));
+	UCHAR buffer[GUID_BUFF_SIZE];
+	sprintf(reinterpret_cast<char*>(buffer),
+		"%02X%02X%02X%02X-%02X%02X-%02X%02X-%02X%02X-%02X%02X%02X%02X%02X%02X",
```
dyemanov, line 6264: Given the UCHAR inputs, why not use the "hh" size specifier here too, rather than relying on the implicit cast?

aafemt: Why not though the implicit cast is standardized and reliable making it no-op.

PR #310, opened 2021-02-07
Feature CORE-6482 - System table with keywords.

### LINKED ISSUE #6713: System table with keywords [CORE6482]
Submitted by: Maxim Kuzmin (cybermax)

Votes: 3

List of keywords \(reserved and non\-reserved\) changed in major/minor version, including release version\.
To determine whether a word is a keyword and it reserved, need to have a dictionary of such words, and update it for each version of the server\.
You should also remember that some words will become reserved in the future \(LOCALTIME, SQLSTATE, INSERTING/DELETING/UPDATING\)\.
Therefore we need a table of words with the flag "Reserved" and "Reserved in the future"\.

Commits: FirebirdSQL/firebird@dd2fcfe18a4c917398607b679b1a48e5ef740da7

asfernandes:


src/jrd/relations.h
```diff
@@ -734,3 +734,10 @@ RELATION(nam_config, rel_config, ODS_13_0, rel_virtual)
 	FIELD(f_cfg_is_set, nam_cfg_is_set, fld_cfg_is_set, 0, ODS_13_0)
 	FIELD(f_cfg_source, nam_cfg_source, fld_file_name2, 0, ODS_13_0)
 END_RELATION
+
+//// FIXME: ODS version
+// Relation 54 (RDB$KEYWORDS)
+RELATION(nam_keywords, rel_keywords, ODS_13_0, rel_virtual)
+	FIELD(f_keyword_name, nam_keyword_name, fld_keyword_name, 0, ODS_13_0)
+	FIELD(f_keyword_reserved, nam_keyword_reserved, fld_keyword_reserved, 0, ODS_13_0)
```
asfernandes, line 742: I'm not sure about the column names: `RDB$KEYWORD_NAME`, `RDB$KEYWORD_RESERVED`.

As usual, we are repetitive and maybe not well spelled in case of "KEYWORD RESERVED". I did like `RDB$CONFIG` columns where every column is prefixed with `RDB$CONFIG_`

dyemanov: I don't have any strong preference about that, as we always had RDB$RELATION_NAME inside RDB$RELATIONS and so on ;-) With just two fields in RDB$KEYWORDS I don't see it as a problem.

src/jrd/relations.h
```diff
@@ -734,3 +734,10 @@ RELATION(nam_config, rel_config, ODS_13_0, rel_virtual)
 	FIELD(f_cfg_is_set, nam_cfg_is_set, fld_cfg_is_set, 0, ODS_13_0)
 	FIELD(f_cfg_source, nam_cfg_source, fld_file_name2, 0, ODS_13_0)
 END_RELATION
+
+//// FIXME: ODS version
+// Relation 54 (RDB$KEYWORDS)
+RELATION(nam_keywords, rel_keywords, ODS_13_0, rel_virtual)
```
dyemanov, line 740: ODS number should be fixed.

dyemanov (changes requested):

src/jrd/relations.h
asfernandes: It's not clear if you mean you prefer the prefix maintained in both columns or something different.

dyemanov: Sorry for confusion, I'm OK with prefix in both columns.

dyemanov (approved):
This PR can be merged. Maybe the "squash and merge" option should be preferred to avoid unnecessary garbage in the commit log.

PR #313, opened 2021-03-04
Fix CORE-6502: remove integrity checking triggers on RDB$USER_PRIVILEGES

romansimakov:
Direct modification of the table has been forbidden since 3.0.
GRANT operations from DDL perform theirown checks.
It requires backup/restore but can work if you do not need to perform
GRANTs descrobed in the ticket.


romansimakov:
Recantly the ODS version has been bumped and I suppose now is good time for merging the PR. Isn't it?

PR #314, opened 2021-03-14
Fix for bug CORE-5515

### LINKED ISSUE #5784: When 32bit and 64bit FB 3 servers run on one Windows machine concurrently, Firebird services freeze several minutes after first disconnect. [CORE5515]
Submitted by: michalk1 (michalk1)

Attachments:
[CORE-5515.7z](https://github.com/FirebirdSQL/jira-attachments/raw/main/CORE/CORE-5515/13313_CORE-5515.7z)

Steps to reproduce:
1\) Install FB3\.0\.2 64bit with default settings \(port 3050, service name DefaultInstance\)
2\) Use FB3\.0\.2 32bit zip package to install another FB instance at the same machine, change its RemoteServicePort in firebird\.conf \(e\.g\. to 3051\), setup its security database and use instsvc to configure and start the service \(named say FB30\)
  instsvc i \-n FB30
  instsvc start \-n FB30
3\) Run isql from DefaultInstance server and make TCP connect to its employee database
  isql \-user sysdba \-password masterkey localhost/3050:employee
4\) Run isql from FB30 server instance and connect to its employee database as well
  isql \-user sysdba \-password masterkey localhost/3051:employee
5\) Close both connections and isql consoles \(ctrl\+z\)
6\) Wait at least two minutes
7\) Now, when trying to connect again to any database, isql console freezes\. Both FB services are stuck as well and must be killed from task manager\.
  
Clues:
The problem is not isql or database specific, you can use i\.e\. FlameRobin, and connect to any database to demonstrate the problem\.
What is weird is the time limit required in step 6, because when the new connection is established fast enough \(in less than about those two minutes\), everything continues to work\.
Two FB 3 64bit or two 32bit servers work together ok, it is only combination of 32bit and 64bit servers that cause hangs\.

Commits: FirebirdSQL/firebird@90270cd11a385e92355492b08ba63b1bca1e9169 FirebirdSQL/firebird@ffaf62b5ffcd5e17e26b4ce94a50813993b88c26

hvlad:
Fixed bug [CORE-5515](http://tracker.firebirdsql.org/browse/CORE-5515)

### DIFF
--- a/src/jrd/Mapping.cpp
+++ b/src/jrd/Mapping.cpp
@@ -784,12 +784,23 @@ public:
 			iscLogException("MappingIpc: Cannot initialize the shared memory region", ex);
 			throw;
 		}
-		fb_assert(sharedMemory->getHeader()->mhb_header_version == MemoryHeader::HEADER_VERSION);
-		fb_assert(sharedMemory->getHeader()->mhb_version == MAPPING_VERSION);
-
-		Guard gShared(this);
 
 		MappingHeader* sMem = sharedMemory->getHeader();
+
+		if (sMem->mhb_type != SharedMemoryBase::SRAM_MAPPING_RESET ||
+			sMem->mhb_header_version != MemoryHeader::HEADER_VERSION ||
+			sMem->mhb_version != MAPPING_VERSION)
+		{
+			string err;
+			err.printf("MappingIpc: inconsistent shared memory type/version; found %d/%d:%d, expected %d/%d:%d",
+				sMem->mhb_type, sMem->mhb_header_version, sMem->mhb_version,
+				SharedMemoryBase::SRAM_MAPPING_RESET, MemoryHeader::HEADER_VERSION, MAPPING_VERSION);
+
+			sharedMemory = NULL;
+			(Arg::Gds(isc_random) << Arg::Str(err)).raise();
+		}
+
+		Guard gShared(this);
 
 		for (process = 0; process < sMem->processes; ++process)
 		{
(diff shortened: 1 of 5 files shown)


src/jrd/Mapping.cpp
```diff
@@ -784,13 +784,24 @@ class MappingIpc FB_FINAL : public Firebird::IpcObject
...
-		fb_assert(sharedMemory->getHeader()->mhb_version == MAPPING_VERSION);
-
-		Guard gShared(this);
 
 		MappingHeader* sMem = sharedMemory->getHeader();
 
+		if (sMem->mhb_type != SharedMemoryBase::SRAM_MAPPING_RESET ||
+			sMem->mhb_header_version != MemoryHeader::HEADER_VERSION ||
+			sMem->mhb_version != MAPPING_VERSION)
+		{
```
dyemanov, line 793: Is it safe to access mhb_* members without a mutex acquired? What if shmem is being initialized by another process right at the moment?

hvlad: shmem is not accesible by other processes while it is initialized.
Also, note, on non-Windows - shmem contains mutex that should be acquired - should we access it when shmem binary layout is not known\verified ?
And, finally - former fb_assert() does the same - check before muex is acquired ;)

hvlad:
> 
> 
> @hvlad for master, can't just use `alignas`?

Yes, we can. But I consider old-style `union` way is more clear\explicit at this place.
With `union` it is explicitly clear that last member occupy 8 bytes, while with `alignas` it is far not that obvious.

PR #316, opened 2021-03-31
common: Fix varible name in utils.cpp.

Biswa96:
'lint' is defined as a macro which get confused with
LARGE_INTEGER lint variable. This fixes the error in
mingw gcc: expected unqualified-id before numeric constant

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -915,15 +915,15 @@ void get_process_times(SINT64 &userTime, SINT64 &sysTime)
 	FILETIME utime, stime, dummy;
 	if (GetProcessTimes(GetCurrentProcess(), &dummy, &dummy, &stime, &utime))
 	{
-		LARGE_INTEGER lint;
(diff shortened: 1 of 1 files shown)


AlexPeshkoff:
Defining macro lint in system headers is certainly bad style but we live in real world, let it be fixed.

aafemt:
This macro is not defined in system headers but lint utility defines it during its work the same way as, for example, resource compiler defines RC_INVOKED.

AlexPeshkoff:
Does mingw build use lint? I doubt...
Anyway name choice was not ideal.

aafemt:
There is no mingw build which I'm aware of.

Biswa96:
Currently firebird project can not be built in msys2/mingw-w64 environment. This is just a small fix.

aafemt:
In this case it is incomplete. You also must remove "-Dlint" from prefix.mingw.

PR #318, opened 2021-04-13
Handle C++ exceptions with a little more informative message in status wrappers

aafemt:


src/include/firebird/Interface.h
```diff
@@ -183,11 +184,21 @@ namespace Firebird
 			{
 				status->setErrors(e.getStatus()->getErrors());
 			}
-			catch (...)
+			catch (const std::exception& e)
```
asfernandes, line 187: Would not that add dependency on C++ runtime libraries in the client?

aafemt: At least for MSVC and GCC - no, but it looks like MacOS wants something more.

PR #6789, opened 2021-05-03
#6786, #6787, #6788 and #6790

### LINKED ISSUE #6786: Add session time zone to system context
Currently it is not possible to obtain the current session time zone other than through workarounds like `substring(current_timestamp from 26)`. 

I think we should add:
- `RDB$GET_CONTEXT('SYSTEM', 'SESSION_TIME_ZONE')` 
- `MON$TIME_ZONE` in `MON$ATTACHMENTS`

Both should return the current session time zone (either named zone or offset in {+|-}HH:MM).

This would be similar to what is provided for the session idle timeout, which is accessible from the system context using `RDB$GET_CONTEXT('SYSTEM', 'SESSION_IDLE_TIMEOUT')` and from `MON$IDLE_TIMEOUT` in `MON$ATTACHMENTS`.

Desired situation:

```sql
set time zone 'America/New_York';
select RDB$GET_CONTEXT('SYSTEM', 'SESSION_TIME_ZONE') from rdb$database;
-- returns 'America/New_York'
select MON$TIME_ZONE from MON$ATTACHMENTS where MON$ATTACHMENT_ID = CURRENT_CONNECTION;
-- returns 'America/New_York'
```
and
```sql
set time zone '+01:00';
select RDB$GET_CONTEXT('SYSTEM', 'SESSION_TIME_ZONE') from rdb$database;
-- returns '+01:00'
select MON$TIME_ZONE from MON$ATTACHMENTS where MON$ATTACHMENT_ID = CURRENT_CONNECTION;
-- returns '+01:00'
```

### LINKED ISSUE #6787: MON$ATTACHMENTS.MON$TIMESTAMP should use its session original time zone
When `isc_dpb_session_time_zone` is specified, `MON$TIMESTAMP` from `MON$ATTACHMENTS` of that session should use that time zone instead of the system configured or default time zone.

### LINKED ISSUE #6788: Extend EXTRACT to extract time zone strings
Standard SQL defines `TIMEZONE_HOUR` and `TIMEZONE_MINUTE` to `EXTRACT` as a way to extract time zone information from values.

But Firebird also supports time zone regions which current can only be extracted with `SUBSTRING`.

I suggest add `TIMEZONE_NAME` extension to `EXTRACT`.

```sql
select extract(timezone_name from timestamp '2021-05-03 10:00 America/Sao_Paulo') from rdb$database;
-- America/Sao_Paulo

select extract(timezone_name from timestamp '2021-05-03 10:00 -3:00') from rdb$database;
-- -03:00
```

`TIMEZONE_NAME` word may be changed if others has better opinions.

### LINKED ISSUE #6790: MON$ATTACHMENTS.MON$TIMESTAMP is incorrect when DefaultTimeZone is configured with time zone different from the server's default
This bug has been noted by @pavel-zotov when testing #6787.

My machine time zone is America/Sao_Paulo (-03:00).

If I set it to Europe/Moscow I get wrong value in `MON$ATTACHMENTS.MON$TIMESTAMP`:

```
SQL> select current_timestamp from rdb$database;

                                        CURRENT_TIMESTAMP 
========================================================= 
2021-05-03 20:58:10.8090 Europe/Moscow                    

SQL> select mon$timestamp from mon$attachments;

                                            MON$TIMESTAMP 
========================================================= 
2021-05-03 14:57:55.6110 Europe/Moscow                    
2021-05-03 14:57:55.6840 Europe/Moscow                    
2021-05-03 14:57:55.6850 Europe/Moscow                    
```

asfernandes:


asfernandes:
I have created this pull request to see what could be included in v4.0.

doc/sql.extensions/README.context_variables2
```diff
@@ -131,6 +131,8 @@ Usage:
    EXT_CONN_POOL_ACTIVE_COUNT   | Count of active connections, associated with pool
                                 |
    EXT_CONN_POOL_LIFETIME       | Idle connection lifetime, in seconds
+                                |
+   SESSION_TIME_ZONE            | Current session time zone.
```
dyemanov, line 135: Given we have TIMEZONE_* keywords, *TIMESTAMP and LIFETIME (all without space in the middle), I'd suggest to change the name to _SESSION_TIMEZONE_.

dyemanov:
I'd suggest to defer #6788 till 4.1 (because of a new BLR verb). Others may be merged into 4.0.

hvlad:
> I'd suggest to defer #6788 till 4.1 (because of a new BLR verb). Others may be merged into 4.0.

What problem do you see in this case ?

dyemanov (approved):
Approved all except #6788 which is suggested for postpone. While being there, for v4.1 I'd also suggest to improve the built-in BLR printer for blr_extract to pretty-print what exactly is extracted. And new fields in MON$ (to complement SESSION_TIMEZONE) also belong to v4.1.

dyemanov:
> > I'd suggest to defer #6788 till 4.1 (because of a new BLR verb). Others may be merged into 4.0.
> 
> What problem do you see in this case ?

No particular problem, just want to be as safe as possible at this stage and not introduce things that don't fix something. Otherwise we may end with RC2...

Issue #6843, opened 2021-06-04
Crash when UDR is used by system attachment, for ex. Garbage Collector


hvlad:
When UDR is used in expression index, it could be called by background garbage collector which run in special system attachment.
Note, system attachment have no assigned provider and no external handle.
When ExtEngineManager access external handle of that system attachment - it crashes.

The call stack:
```
fbclient.dll!Why::YAttachment::YAttachment(aProvider, aNext, aDbPath) Line 4980
fbclient.dll!Why::MasterImplementation::registerAttachment(provider=0x0000000000000000, attachment=0x0000000000000000) Line 3729
fbclient.dll!Firebird::IMasterBaseImpl<Why::MasterImplementation,Firebird::CheckStatusWrapper,Firebird::IVersionedImpl<Why::MasterImplementation,Firebird::CheckStatusWrapper,Firebird::Inherit<Firebird::IMaster> > >::cloopregisterAttachmentDispatcher(self=0x0000000000000000, provider=0x000007fedb9c4a79, attachment=0x0000000000000004) Line 5728
engine12.dll!Jrd::ExtEngineManager::ExternalContextImpl::ExternalContextImpl(tdbb, aEngine) Line 607
engine12.dll!Jrd::ExtEngineManager::getEngineAttachment(tdbb=0x000000000766fa40, engine=0x00000000077f0388, closing=false) Line 1598
engine12.dll!Jrd::ExtEngineManager::makeFunction(tdbb=0x000000000766fa40, csb=0x00000000052a8b40, udf=0x0000000007717150, engine={...}, entryPoint={...}, body={...}) Line 1162
engine12.dll!Jrd::Function::loadMetadata(tdbb=0x000000000766fa40, id=58688, noscan=64, flags=25408) Line 1107
engine12.dll!Jrd::Function::lookup(tdbb=0x000000000766fa40, name={...}, noscan) Line 646
engine12.dll!Jrd::UdfCallNode::parse(tdbb=0x000000000766fa40, pool={...}, csb=0x00000000052a1dc0, blrOp) Line 11057

...

engine12.dll!Jrd::Routine::parseBlr(tdbb=0x000000000766fa40, csb=0x00000000052a1dc0, blob_id=0x000000000766c1a0, blobDbg=0x000000000766c190) Line 175
engine12.dll!Jrd::Function::loadMetadata(tdbb=0x000000000766fa40, id=22720, noscan=192, flags=25408) Line 1121
[...]

PR #6844, opened 2021-06-04
Fixed bug GH-6843 : Crash when UDR is used by system attachment, for ex. Garbage Collector

hvlad:
This PR introduces ability for External Contex to correctly work with not assigned externalAttachment\externalTransaction in case of usage within system attachment.
If UDR tries to access current connection (via externalAttachment\externalTransaction), error is raised.
Note, UDR should not do anything with system attachments, else it could break too much internals.

### DIFF
--- a/src/jrd/ExtEngineManager.cpp
+++ b/src/jrd/ExtEngineManager.cpp
@@ -603,8 +603,13 @@ ExtEngineManager::ExternalContextImpl::ExternalContextImpl(thread_db* tdbb,
 
 	internalAttachment->getStable()->addRef();
 
-	externalAttachment = MasterInterfacePtr()->registerAttachment
-		(internalAttachment->getProvider(), internalAttachment->getInterface());
+	Jrd::JProvider* jProv = internalAttachment->getProvider();
+	Jrd::JAttachment* jAtt = internalAttachment->getInterface();
+
+	// System attachments (such as garbage collector) have no external interface
+	// and should not be used by UDR's
+	if (jProv && jAtt)
+		externalAttachment = MasterInterfacePtr()->registerAttachment(jProv, jAtt);
 }
 
 ExtEngineManager::ExternalContextImpl::~ExternalContextImpl()
@@ -616,6 +621,8 @@ ExtEngineManager::ExternalContextImpl::~ExternalContextImpl()
 		externalAttachment->release();
 		externalAttachment = NULL;
 	}
+	else
+		internalAttachment->getStable()->release();
 }
 
 void ExtEngineManager::ExternalContextImpl::releaseTransaction()
@@ -640,7 +647,10 @@ void ExtEngineManager::ExternalContextImpl::setTransaction(thread_db* tdbb)
 	fb_assert(!externalTransaction && !internalTransaction);
 
 	if ((internalTransaction = newTransaction))
-		externalTransaction = MasterInterfacePtr()->registerTransaction(externalAttachment, internalTransaction);
+	{
+		if (externalAttachment)
+			externalTransaction = MasterInterfacePtr()->registerTransaction(externalAttachment, internalTransaction);
+	}
 }
 
 IMaster* ExtEngineManager::ExternalContextImpl::getMaster()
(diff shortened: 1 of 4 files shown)


aafemt:
I'm afraid that in this case UDRs that has to do something with current attachment will break the internals as well. Otherwise they would not be called at at all.

Consider an UDR returning CURRENT_USER and an expression index built on it for example.

hvlad:
> 
> 
> I'm afraid that in this case UDRs that has to do something with current attachment will break the internals as well. 

They could break state of system att, but not user att. Not sure what you afraid of.

> 
> Consider an UDR returning CURRENT_USER and an expression index built on it for example.

Such UDR will be forbidden. No UDR code might run queryes in contex of system attachment.
If there will be real complains from real users, I'll think on re-implementing GC using non-system attachment, probably.

aafemt:
> Such UDR will be forbidden. No UDR code might run queryes in contex of system attachment.

Yes. That's the point. GC needs UDR to return a value to perform index cleanup, right? If the UDR has returned an error or a wrong value - at least the index won't be cleaned, at most whole GC process may be blocked.

hvlad:
You are welcome to test the fix and report any real issue found.

src/jrd/ExtEngineManager.cpp
```diff
@@ -703,6 +717,13 @@ void* ExtEngineManager::ExternalContextImpl::setInfo(int code, void* value)
 	return oldValue;
 }
 
+void ExtEngineManager::ExternalContextImpl::checkExternalAttachment()
+{
+	if (!externalAttachment)
+	{
```
AlexPeshkoff, line 723: If check for presence of external attachment is anyway needed - why not move creation of it here from constructor? I agree that overhead due to unused externalAttachment is not too big, but a lot of UDR functions (one who perform supposed activity based on passed parameters, i.e. have UDF-like functionality) do not need it at all.

src/jrd/ExtEngineManager.cpp
```diff
@@ -703,6 +717,13 @@ void* ExtEngineManager::ExternalContextImpl::setInfo(int code, void* value)
...
+		(Arg::Gds(isc_random) << Arg::Str("Use of system attachment in UDR is not allowed")).raise();
```
asfernandes, line 724: We need special error code here so UDR may know why it was not possible to get the attachment.

hvlad: Sure

asfernandes (changes requested):
We need special error code so UDR may know why it was not possible to get the attachment.

src/jrd/ExtEngineManager.cpp
hvlad: Will do, thanks

hvlad:
I'm afraid we need more complex solution. With current patch there is no more crashes (primary goal of PR) but IDX_garbage_collector ignore errors raised by UDR (there is a good reason for it) and nobody will know about this error. UDR developer might not know that UDR sometimes works in system attachment and index left not cleaned up because of this. 

I can think about additional optional declaration for UDR to explicitly state that it requires no current context. Such declaration should be verified at run-time and developer will see corresponding errors.
By default UDR's considered as they could use current context. UDR's which was not declared as not requiring current context, should be prohibited in expression indices and calculated fields (as they also could be indexed).

Better ideas ?

aafemt:
https://github.com/FirebirdSQL/firebird/pull/6844#issuecomment-854662258

asfernandes:
FB/Java may use the attachment to load Java classes stored in database. User has no control of it, if a class need to be loaded, it will call database.

> 
> Better ideas ?

Maybe your idea of not using system attachment in GC?

hvlad:
> FB/Java may use the attachment to load Java classes stored in database

Could it be implemented without direct access to the current connection (IExternalContext) ? 
I mean - using some new special method in IExternalEngine, etc ?

hvlad:
> How about force cooperative GC of expression indexes inside of user attachment/transaction?..

It is not as easy to do as it might seem. And I don't like it in general.

hvlad:
> > Better ideas ?
> 
> Maybe your idea of not using system attachment in GC?

I prefer to find better solution. If it will not be found - let it be so.

AlexPeshkoff:
> UDR could create own attachment when needed, but here we talk about the case when UDR requires access to the current attachment.
> 
UDR is often using current attachment because it's much faster than create a new one, including a case when:

> FB/Java may use the attachment to load Java classes stored in database

PR #6847, opened 2021-06-08
An attempt to fix #6830: An attempt to destroy busy rel_drop_mutex failed

### LINKED ISSUE #6830: An attempt to destroy busy rel_drop_mutex failed
When deleting relation in deferred work rel_drop_mutex is locked in the beginning at the same time setting REL_deleting flag and unlocked in the end or on error at the same time cleaning that flag. In a case of successful completion REL_deleted is set simultaneously. 

The problem happens when in some other place (when working with check_relation in metadata cache) REL_deleted is set not cleaning REL_deleting flag and not unlocking rel_drop_mutex. In case of some error in dfw MET_lookup_relation_id() does not find marked deleted relation, and can not unlock rel_drop_mutex. Therefore it remains unlocked forever (RAII holder is not used for it in DFW), and causes error on an attempt to destroy it.

AlexPeshkoff:
When working with provided core dump I've found the following - when segfault happened (during purge_attachment, in release_locks) rel_flags was 0x83324, i.e. the following flags were set:

REL_deleted
REL_check_existence
REL_sql_relation
REL_check_partners
REL_deleting
REL_temp_tran
REL_gc_lockneed

Pay attetion to both set REL_deleted|REL_deleting. And rel_drop_mutex was really locked. I failed to reproduce how to obtain such flags combination but there is one place where receiving it means keep rel_drop_mutex locked after cleanup (zero) phaze of DFW.

Suggested in this branch simple fix should help with cleanup, and I see no way for it to cause any harm.
What do you think about committing it to HEAD?

### DIFF
--- a/src/alice/aliceswi.h
+++ b/src/alice/aliceswi.h
@@ -66,6 +66,9 @@ const SINT64 sw_icu				= QUADCONST(0x0000002000000000);
 const SINT64 sw_role			= QUADCONST(0x0000004000000000);
 const SINT64 sw_replica			= QUADCONST(0x0000008000000000);
 
+// Popular combination of compatible switches
+const SINT64 sw_auth_set		= sw_user | sw_password | sw_role | sw_fetch_password | sw_trusted_auth;
+
 
 enum alice_switches
 {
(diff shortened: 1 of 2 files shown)


hvlad (commented):
Looks logical. I also see no harm by this change.

hvlad:
> Pay attetion to both set REL_deleted|REL_deleting. And rel_drop_mutex was really locked. I failed to reproduce how to obtain such flags combination

REL_deleted also could be set by `create_relation` at phase 0 and it runs before `delete_relation`. Looks like `DROP\CREATE RELATION` in the same transaction or `RECREATE RELATION` was running. The failed DDL statement could be another one in the same tx, btw.

It will be good to have reprodusible example for this condition.

AlexPeshkoff:
Create/drop table follow one another in a test, but they are in autonomous transactions:

            sql = 'create global temporary table ' || :s || ' (id int);';
            execute statement sql with autonomous transaction;
            sql = 'drop table ' || :s || ';';
            execute statement sql with autonomous transaction;

Full test here:
https://github.com/FirebirdSQL/fbt-repository/blob/master/tests/bugs/core_2879.fbt

AlexPeshkoff:
Doubt that. That loop never completes normally - when sweep is complete (test is checking sweep stability, loop we discuss is just some back load for it) it's terminated due to client's disconnect. There are 148 relations in an attachment, first 128 are system, i.e. loop iterated about 20 times.

dyemanov:
I don't foresee any problems with the patch.

PR #6864, opened 2021-06-18
Update firebird.service: tmpfile configuration

pmakowski:
use of RuntimeDirectory (https://www.freedesktop.org/software/systemd/man/systemd.exec.html)

### DIFF
--- a/builds/install/arch-specific/linux/firebird.service.in
+++ b/builds/install/arch-specific/linux/firebird.service.in
@@ -5,9 +5,10 @@ After=syslog.target network.target
 [Service]
 User=firebird
 Group=firebird
-Type=forking
 ExecStart=@FB_SBINDIR@/fbguard -daemon -forever
 StandardError=syslog
+RuntimeDirectory=firebird
+RuntimeDirectoryMode=0775
 
 [Install]
 WantedBy=multi-user.target


AlexPeshkoff:
How is removal of 
Type=forking
related with tmpfile config?

pmakowski:
Sorry, should be perhaps another merge request.
forking is not needed and if we use forking we should also set up PIDFile=
Firebird don't use forking in Fedora since https://bugzilla.redhat.com/show_bug.cgi?id=1687058 without trouble.

asfernandes:
Shouldn't also Firebird lock directory be in `/run/firebird` instead of `/tmp/firebird`?

pmakowski:
> Shouldn't also Firebird lock directory be in `/run/firebird` instead of `/tmp/firebird`?
that's the case under Fedora (--with-fbglock=%{_rundir}/%{name})

AlexPeshkoff:
Is /run accessible for embedded clients? I.e - can unprivileged user create /run/firebird himself? As far as I see - not:
drwxr-xr-x  29 root root   920 Jun 17 20:17 run

asfernandes:
> Is /run accessible for embedded clients? I.e - can unprivileged user create /run/firebird himself? As far as I see - not:
> drwxr-xr-x 29 root root 920 Jun 17 20:17 run

No.

But don't we also have a problem if service creates `/tmp/firebird` and embedded user is not in firebird group?

pmakowski:
> Is /run accessible for embedded clients?

no, only /run/firebird is if the user is a member of group firebird

AlexPeshkoff:
If service is installed and in use embedded users certainly should be in group firebird. If not - nothing currently prevents at least first of them to create lock files in default location. And if that user is in firebird group directory will become available to all group.

Telling true that's not great feature - in most cases when embedded access is used w/o firebird service databases are per-user and lock files too. But somebody may use described mode.

asfernandes:
There is `/run/lock` in Ubuntu and it's world writable.

```
drwxrwxrwt  6 root              root    160 jun 23 18:22 lock/
```

AlexPeshkoff:
On 6/24/21 12:25 AM, Adriano dos Santos Fernandes wrote:
>
> There is |/run/lock| in Ubuntu and it's world writable.
>
> |drwxrwxrwt 6 root root 160 jun 23 18:22 lock/|

On gentoo:

drwxrwxr-x 4 root uucp 80 Jun 26 11:40 /run/lock

PR #6871, opened 2021-06-24
Fix #6839 - Weird string comparison with index.

### LINKED ISSUE #6839: Weird string comparison with index
Strings containing newlines (0x0A, line feed) are handled weirdly by `>`. First, I would expect `>` to be `FALSE` instead of `TRUE`, since the string on the right is longer. Second, using `>` in a `where` clause gives different results than selecting the result of `>`. This inconsistently only happens if there is an index (or primary key) on the string column.

```
CREATE TABLE T0(C0 VARCHAR(100), PRIMARY KEY (C0));
INSERT INTO T0 (C0) VALUES ('a
b'); -- this is a 3 character string with a line feed in the middle
SELECT (('a')>(T0.C0)) FROM T0; -- {TRUE}
SELECT T0.C0 FROM T0 WHERE (('a')>(T0.C0)); -- 0 rows (although previous query returned true)

CREATE TABLE T1(C0 VARCHAR(100));
INSERT INTO T1 (C0) VALUES ('a
b');
SELECT (('a')>(T1.C0)) FROM T1; -- {TRUE}
SELECT T1.C0 FROM T1 WHERE (('a')>(T1.C0)); -- 1 row (the only difference is no primary key)
```

The example above was run with `CHARACTER SET NONE COLLATE NONE` (the default) in isql. The behavior is the same with `CHARACTER SET ASCII COLLATE ASCII` and `CHARACTER SET UTF8 COLLATE UTF8`. The behavior is also the same with JDBC.

If you use `CHARACTER SET UTF8 COLLATE UNICODE`, `CHARACTER SET UTF8 COLLATE USC_BASIC` the results are different and make more sense to me. With those settings all `SELECT` statements return `FALSE` and 0 rows.

asfernandes:


asfernandes:
This PR is so old that it must be rebased.
Nobody ever reviewed it.
Can someone review it now?

PR #6879, opened 2021-06-29
Fix #6878 - [gbak] incorrect behavior in case of using the service

### LINKED ISSUE #6878: [gbak] Incorrect behavior in case of using the service
In the case of a split backup file, gbak outputs the data:

> Done with volume #1, "/.../db.fbk"     Press return to reopen that file, or type a new name followed by return to open a different file.  Name:

But with "-se 127.0.0.1/3050:service_mgr" stdin cannot be set correctly.
The `prompt_for_name` function has `fb_assert` but it can't help avoiding the unexpected behavior

https://github.com/FirebirdSQL/firebird/blob/73a0dcdd55924d9b5bd2c3d9172f3fb5766030f0/src/burp/mvol.cpp#L1581

Noremos:
Issue: #6878


aafemt:
For master branch it would be better to remove splitting completely.

UnknownAr:
> For master branch it would be better to remove splitting completely.

In this case the backup was partitioned using the 'stlit' utility.

aafemt:
It doesn't matter. Ability for gbak working with multivolume backups (as well as multifile 
databases) is obsolete. Split utility should gone as qli did.

AlexPeshkoff:
On 6/29/21 3:44 PM, Dimitry Sibiryakov wrote:
> It doesn't matter. Ability for gbak working with multivolume backups 
> (as well as multifile
> databases) is obsolete. Split utility should gone as qli did.

+1

PR #6880, opened 2021-06-29
Skipped Warnings

oops00:
Skipped Warnings

### DIFF
--- a/src/jrd/validation.h
+++ b/src/jrd/validation.h
@@ -27,7 +27,6 @@
 #include "firebird.h"
 #include "fb_types.h"
 
-#include "../alice/alice.h"
 #include "../common/classes/array.h"
 #include "../common/SimilarToRegex.h"
 #include "../jrd/ods.h"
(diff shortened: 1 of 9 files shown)


oops00:
in the "validation.h" in the constructor of the Validation class, the skip parameter is specified, which by default leaves warnings in the log file. Change the value to true or specify this parameter when defining the class

src/jrd/validation.h
```diff
@@ -160,7 +162,7 @@ class Validation
 	void checkDPinPIP(jrd_rel *relation, ULONG page_number);
 
 public:
-	explicit Validation(thread_db*, Firebird::UtilSvc* uSvc = NULL);
+	explicit Validation(thread_db*, Firebird::UtilSvc* uSvc = NULL, bool Skipped = false);	// Edited by AIR
```
oops00, line 165: explicit Validation(thread_db*, Firebird::UtilSvc* uSvc = NULL, bool Skipped = true);	// Edited by AIR

hvlad:
The main problem of this patch is that there is no way for end user to set (or clear) "skip warns" flag.
The second problem is that code makes no respect to existing code conventions and contains few not needed formatting changes.

hvlad:
Could you, please, don't push every commit here ? 
Push *one* finall common commit when you consider patch finished.
Thanks.

PR #6892, opened 2021-07-09
Improvement #2805 [CORE2383]

TedCraft:
Add options in firebird.conf (LogsInLogDir, GfixLogFileName). Add functions in gds_proto.h (gds__log__file) and gds.cpp (gds__log__file, vgds__log__file). Rewrite gds__log function.

### DIFF
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -1165,3 +1165,19 @@
 # Type: integer
 #
 #ExtConnPoolLifeTime = 7200
+
+# ============================
+# Log files
+# ============================
+
+# Indicates whether to write logs to the logs directory
+#
+# Type: boolean
+#
+#LogsInLogDir = false
+
+# Set gfix utility logfile name
+#
+# Type: string
+#
+#GfixLogFileName = firebird.log
--- a/builds/posix/firebird.vers
+++ b/builds/posix/firebird.vers
@@ -69,6 +69,7 @@ gds__get_segment
 gds__get_slice
 gds__interprete
 gds__log
+gds__log__file
 gds__log_status
 gds__map_blobs
 gds__msg_close
--- a/builds/win32/defs/firebird.def
+++ b/builds/win32/defs/firebird.def
@@ -64,6 +64,7 @@ EXPORTS
 	gds__get_slice	@40
 	gds__interprete	@41
 	gds__log	@42
+	gds__log__file @251
 	gds__log_status	@142
 	gds__map_blobs	@240
 	gds__msg_close	@223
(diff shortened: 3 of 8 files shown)


romansimakov:
I'd prefer to have shorter parameter name. Maybe LogDirectory? I'm not sure about its meaning as well. Let it be string with path to a directory containing logs?
Originally firebird has been using the installation root as a place for log. It seems not correct. It's better to have installation root as read-only. Isn't it?
Having LogDirectory containing path to a directory when logs must be kept we allow administrators to change directory.

As for the second option with gfix log file prefix (or name) I was thinking we could just pass a file name through gfix command line. By default filebird.log will be used. However an invoker will be able to specify desired filename and use timestame if he wants. He will know what filename to find when gfix is done.

dyemanov:
I like Roman's suggestion to specify the log file name in the gfix command line (or using the corresponding Services API tag). But if it's missing, I'd prefer to write the log into a different file than firebird.log, named e.g. validation.log or recovery.log and located side by side with firebird.log.
As for moving all the logs into a separate directory, I would support that, although IMO the default location should be _$fbroot/logs_. But I'm not sure such a change is desirable for a point release (this PR is for _v4.0-release_), so maybe we could start with _LogDirectory_ pointing by default to _$fbroot_ in v4.0.1 and to _$fbroot/logs_ in v5.

hvlad:
My old opinion is that whole error logging in Firebird should be reworked. Probably using new plugin architecture.
Reworked subsystem should support flexible configuration: what to log, where to put messages, what error categories should exists and so on.
Any cosmetic changes like offered in this PR is a poor solution and not what we really need, imho. More, it will make future rework more complex as it will have to support all that cosmetic changes.
I strongly recommend to discuss this in fb-devel first. 

PS don't confuse "gfix" (utility that ask engine to do something) with "validation" (process that only engine can run).
gfix doesn't put messages into the firebird.log, engine does it.

AlexPeshkoff:
On 7/19/21 1:00 PM, Vlad Khorsun wrote:
>
> My old opinion is that whole error logging in Firebird should be 
> reworked. Probably using new plugin architecture.
> Reworked subsystem should support flexible configuration: what to log, 
> where to put messages, what error categories should exists and so on.
> Any cosmetic changes like offered in this PR is a poor solution and 
> not what we really need, imho. More, it will make future rework more 
> complex as it will have to support all that cosmetic changes.
>

+1

With one warning - logging is often specially useful when engine is in 
not completely working state - something like OOM or what else. Using 
plugins in such case may easily fail, i.e. poor by functionaiity but 
working almost always current solution to be reactivated in such case.

asfernandes:
> With one warning - logging is often specially useful when engine is in not completely working state - something like OOM or what else. Using plugins in such case may easily fail, i.e. poor by functionaiity but working almost always current solution to be reactivated in such case.

Plugin would be in y-valve, no? So I don't see this problem.

aafemt:
It would be also nice to provide log writing functionality to plugins through IMaster. I added such request into the old tracker but cannot find it after migration.

hvlad:
All, please, move discussion into fb-devel.

PR #6911, opened 2021-07-27
Added option SKIP_WARNINGS

oops00:
Example using: gfix.exe -SKIP_WARNINGS


AlexPeshkoff:
Did you check your PR carefully? When adding record only to messages2.sql you should get message that it's not good during the build.

PR #6914, opened 2021-07-31
Allow keywords in locales when creating ICU collations (Fixes issue #6903)

### LINKED ISSUE #6903: Unable to create ICU-based collation with locale keywords
Firebird does not seem to allow the creation of ICU-based collations whose locales contain keywords, such as `de@collation=phonebook` or `de-u-co-phonebk`.

For example, executing the following statement:

```
CREATE COLLATION UNICODE_TEST FOR UTF8 FROM UNICODE
  PAD SPACE
  CASE SENSITIVE
  ACCENT SENSITIVE
  'LOCALE=de-u-co-phonebk-ka-shifted';
```
fails with:

```
Message: isc_dsql_execute2 failed

SQL Message : -607
This operation is not defined for system tables.

Engine Code    : 335544351
Engine Message :
unsuccessful metadata update
CREATE COLLATION UNICODE_TEST failed
Invalid collation attributes
```

The failure originates from `UnicodeUtil::Utf16Collation::loadICU()` (file `src/common/unicode_util.cpp`). When checking the incoming locale, it is compared against the locale names obtained from ICU's enumeration. This however does not include keywords or collation subtypes, so anything other than locale base names will fail this test.

I'm attaching a patch with a suggested fix (against v4.0-release, but should work on FB3 also). The fix adds a fallback check which tries to create a collator with the locale in question. This check is only performed if the previous check failed. With this modification, locales with keywords are accepted when creating a collation, and the collation works as expected when applied to a table column.

[allow_keyword_locales.zip](https://github.com/FirebirdSQL/firebird/files/6836368/allow_keyword_locales.zip)

tkeinz:
This fix adds a fallback check when trying to create an ICU collator and the check against locale base names has failed. With this fallback check, locales with keywords are accepted when creating a collation. In my tests, the collation worked as expected when applied to a table column.

Please let me know in case this is not the right branch against which to submit the change.

### DIFF
--- a/src/common/unicode_util.cpp
+++ b/src/common/unicode_util.cpp
@@ -1718,7 +1718,7 @@ UnicodeUtil::ICU* UnicodeUtil::Utf16Collation::loadICU(
 					continue;
 
 				icu->ucolClose(testCollator);
-				if (U_FAILURE(status) || (status == U_USING_DEFAULT_WARNING))
+				if (status != U_ZERO_ERROR)
 					continue;
 			}
 		}
--- a/src/common/unicode_util.cpp
+++ b/src/common/unicode_util.cpp
@@ -1711,7 +1711,16 @@ UnicodeUtil::ICU* UnicodeUtil::Utf16Collation::loadICU(
 			}
 
 			if (avail < 0)
-				continue;
+			{
+				UErrorCode status = U_ZERO_ERROR;
+				UCollator* testCollator = icu->ucolOpen(locale.c_str(), &status);
+				if (!testCollator)
+					continue;
+
+				icu->ucolClose(testCollator);
+				if (U_FAILURE(status) || (status == U_USING_DEFAULT_WARNING))
+					continue;
+			}
 		}
 
 		char version[U_MAX_VERSION_STRING_LENGTH];


src/common/unicode_util.cpp
```diff
@@ -1711,7 +1711,16 @@ UnicodeUtil::ICU* UnicodeUtil::Utf16Collation::loadICU(
...
 			if (avail < 0)
-				continue;
+			{
+				UErrorCode status = U_ZERO_ERROR;
+				UCollator* testCollator = icu->ucolOpen(locale.c_str(), &status);
+				if (!testCollator)
+					continue;
+
+				icu->ucolClose(testCollator);
+				if (U_FAILURE(status) || (status == U_USING_DEFAULT_WARNING))
```
asfernandes, line 1721: I tried `create collation unicode_ptbr for utf8 from unicode 'LOCALE=pt_BRx'` and it worked, but should not.

Can you change that to `if (status != U_ZERO_ERROR)` to treat warnings as errors?

asfernandes (changes requested):

src/common/unicode_util.cpp
tkeinz: Oops, good catch. Thank you!

While I had tested with completely invalid locale names, I missed testing with partially valid ones. Changed as requested; now your example also fails (and the examples from my use cases continue to work).

asfernandes:
For master, I'm going to completely avoid the `icu->uloc*` functions.

pavel-zotov:
::: test details :::
See tests.bugs.gh_6903

PR #6926, opened 2021-08-23
Dependencies check fixes

ilya071294:
1. The first commit fixes this script:

set autoddl off;

create table test (id numeric);
commit;

create view v_test (id) as select id from test;
commit;

drop view v_test;
create or alter view v_test (id) as select id from test;
drop table test;
commit;

The last commit fails and the transaction should be rolled back but the test table is already broken. It happens because the phase 4 of delete_relation DFW cannot be rolled back. It's already completed by the moment of the exception. With the fix the exception will occur at the phase 3.

2. The second commit fixes this:

set autoddl off;

create table test (id numeric);
commit;

create table test1 (id numeric);
commit;

set term ^;

create procedure NEW_PROCEDURE
returns (
    OUT integer)
as
begin
  select first 1 id from test into out;
  suspend;
end^
commit^

alter procedure NEW_PROCEDURE
returns (
    OUT integer)
as
begin
  select first 1 id from test1 into out;
  suspend;
end^

drop table test1^
commit^

set term ;^

rollback;

drop table test1;
commit;

drop procedure NEW_PROCEDURE;
commit;

drop table test;
commit;

The last commit should work without any errors but it fails because the test table still depends on NEW_PROCEDURE even though the procedure is deleted. The test1 table is also broken here (like in the first script).

I would also suggest to add a commit where the phase 4 of delete_relation DFW is moved to the phase 6 (modifyRoutine DFW has the phase 5, and exceptions may occur there as well). Or maybe it should be moved to the post commit stage but it seems wrong.

### DIFF
--- a/src/jrd/RecordSourceNodes.cpp
+++ b/src/jrd/RecordSourceNodes.cpp
@@ -479,7 +479,7 @@ RelationSourceNode* RelationSourceNode::parse(thread_db* tdbb, CompilerScratch* 
 				PAR_name(csb, *aliasString);
 			}
 
-			if (!(node->relation = MET_lookup_relation_id(tdbb, id, false)))
+			if (!(node->relation = MET_lookup_relation_id(tdbb, id, false, false)))
 				name.printf("id %d", id);
 
 			break;
@@ -496,7 +496,7 @@ RelationSourceNode* RelationSourceNode::parse(thread_db* tdbb, CompilerScratch* 
 				PAR_name(csb, *aliasString);
 			}
 
-			node->relation = MET_lookup_relation(tdbb, name);
+			node->relation = MET_lookup_relation(tdbb, name, false);
 			break;
 		}
 
--- a/src/jrd/dfw.epp
+++ b/src/jrd/dfw.epp
@@ -722,6 +722,7 @@ namespace
 						return false;
 
 					routine->flags &= ~Routine::FLAG_BEING_ALTERED;
+					routine->flags |= Routine::FLAG_OBSOLETE;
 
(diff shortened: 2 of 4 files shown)


aafemt:
IMHO, both scripts should fail on "drop table test", long before commit.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -479,7 +479,7 @@ RelationSourceNode* RelationSourceNode::parse(thread_db* tdbb, CompilerScratch*
 				PAR_name(csb, *aliasString);
 			}
 
-			if (!(node->relation = MET_lookup_relation_id(tdbb, id, false)))
+			if (!(node->relation = MET_lookup_relation_id(tdbb, id, false, false)))
```
hvlad, line 482: Why not simply check for REL_deleting here ?

src/jrd/dfw.epp
```diff
@@ -721,6 +721,8 @@ namespace
 					if (!routine)
 						return false;
 
+					routine->flags &= ~Routine::FLAG_BEING_ALTERED;
+
```
hvlad, line 725: Should it be marked with FLAG_OBSOLETE also ?

hvlad:
> 
> 
> IMHO, both scripts should fail on "drop table test", long before commit.

Sound logical, at first look.

ilya071294:
> IMHO, both scripts should fail on "drop table test", long before commit.

I agree but to achieve this we need to compile a procedure/view, update and check dependencies at dsql stage.

PR #6927, opened 2021-08-23
Separate list of events for each trace session

aafemt:


hvlad:
This patch will change nothing in trace behaviour. 
Why do you need it ?

aafemt:
I could make a mistake in the patch of course but it was supposed to fix the problem when trace plugin receives events it didn't subscribe to. I asked Pavel to test it.

hvlad:
The patch makes no change to the trace plugin and it still subscribes to the all known events.
If you want to do something really useful, you should introduce ITracePlugin::trace_needs(), 
fill it after reading trace configuration and use it in the TraceManager instead of useless ITraceFactory::trace_needs(),
which could be declared as obsolete and unused.
If your custom trace plugin fills ITraceFactory::trace_needs() based on trace configuration it is a misuse of Factory pattern.
And I can't accept this patch as it makes no benefit for most Firebird users.

aafemt:
Surely I will do it for master branch, but I want (even if provisory) solution for v4 as well. Changing of interfaces is not an option here.

hvlad:
I see no problem for such interface change in 4.x, we already did it earlier.

PR #6958, opened 2021-09-10
Rework messages processing, simplifying the build and avoiding store generated files in git.

asfernandes:


AlexPeshkoff (approved):
I did not check commit in details but approach is OK for me.

romansimakov:
I know you sent the proposal to devel-firebird but that was for discussion. Could you please add short description of the made changes?

asfernandes:
> I know you sent the proposal to devel-firebird but that was for discussion. Could you please add short description of the made changes?

Almost everything as proposed there.

However preprocessor is not called directly to generate any files.

First, preprocessor can not be used to generate `#define`.

Second, it was easy to change build_msg to generate it. And codes (who was generating that files before) was merged in build_msg.

Only C and Pascal files are generated now and some files directly in the output tree, so they are not versioned.

One Pascal files is left versioned (including with error messages). I think that together with the C++ files generated by cloop, we can think in remove them from the tree in another patch.

C++ files works directly with the error files, they do not need the generated files.

PR #6961, opened 2021-09-12
rebuild codes with configure --enable-regen-codes (new option)

real-dam:
This started as a Debian-specific patch to satisfy the need that all generated files are re-made during package build. Rebuilding makes the resulting files reproducible and guarantees that they are produced from sources and do not contain stray modifications.

Perhaps other distributions would find it useful, therefore including it in the upstream source seems the right thing to do.


asfernandes:
Looks like that would not be necessary after https://github.com/FirebirdSQL/firebird/pull/6958

real-dam:
Yes indeed. Good catch!
I saw the messages on firebird-devel, but didn't look at the MR.

Feel free to reject this one once #6958 is merged.

PR #6972, opened 2021-09-15
add --with-system-boost configure option for using system-wide boost headers

real-dam:
This is the last one of the "use system $lib" series.


src/include/firebird/Message.h
```diff
@@ -25,7 +25,15 @@
 
 #include "ibase.h"
 #include "./Interface.h"
+#if defined USE_SYSTEM_BOOST
```
asfernandes, line 28: This is public header file. So I think name must be better `FB_USE_SYSTEM_BOOST`.

aafemt:
Taking into account that C++11 features are fully available, do we still need boost?

real-dam:
> Taking into account that C++11 features are fully available, do we still need boost?

I like the idea of replacing an external dependency with a compiler feature. Hopefully someone more fluent in C++ than me would try it.

asfernandes:
> Taking into account that C++11 features are fully available, do we still need boost?

We didn't discussed C++11 in public headers yet. I support it.

configure.ac
```diff
@@ -560,6 +560,12 @@ AC_ARG_WITH(system-re2,
   [RE2_BUILD=N])
 AC_SUBST(RE2_BUILD)
 
+LOCAL_BOOST=Y
+AC_ARG_WITH(system-boost,
+  [  --with-system-bools     use system-wide boost library instead of embedded copy],
```
asfernandes, line 565: This is misspelled (bools).

asfernandes:
How that will work for public headers?

Will you put a dependency for boost in firebird-client-dev (or the name you use) package?

Will you deploy fb-boost headers in that package?

Will the user also need to use the define?

real-dam:
> How that will work for public headers?
> 
> Will you put a dependency for boost in firebird-client-dev (or the name you use) package?
> 
> Will you deploy fb-boost headers in that package?
> 
> Will the user also need to use the define?

Good question. Requiring that all consumers of `firebird/Message.h` define `USE_SYSTEM_BOOST` would be bad.

Perhaps I should reverse the define and rename it to `USE_FB_BOOST`? This way consumers using the Debian package would only need to have boost around, maybe pulled by a dependency.

That dependency, while technically needed, is not present in firebird3.0 packages which have a similar patch (crudely using system boost and replacing `FB_BOOST_xxx` with `BOOST_xxx` in `Message.h`) and nobody complained for the last 5+ years. I guess some softer form of dependency would be enough.

Will update the PR with `USE_FB_BOOST` and reverse meaning.

aafemt:
> nobody complained for the last 5+ years.

It just means that nobody uses Messages.h.

real-dam:
> Will update the PR with `USE_FB_BOOST` and reverse meaning.
Oh, but that would mean that any `Message.h` consumers that have it from non-packaged Firebird would need the define. Not good either.

I think the approach used in fb3 Debian packages is the least evil.

The cleanest way from consumer side would be to have `Messages.h` with either `include <...boost...>` (plus a dependency) or `include "...boost..."`, both hard-coded and not controlled by a define.

fb3 debs took one approach at this, patching the installed files.

An approach that controls the output via `--with-system-boost` would require similar sort of patching, for example via sed. Shouldn't be a problem since this affects only posix builds. Will be kind of ugly, but should work.

real-dam:
This became as good as patching during package build with the added benefit that including it upstream will make the functionality available for other downstream distributions.

Issue #6886, opened 2021-07-05
Differerent interfaces behaviour depending upon source of interface


AlexPeshkoff:
ResultSet::close function behaves differently depending upon provider of interface - in YValve it releases result set on success (according to docs), in Engine/Remote - does not release result set. Same difference in bahaviour is present in all functions, closing access to the object (Transaction::commit, Attachment::detach, etc.).

That causes memory leaks when attachments from engine are used directly (like in Mappong.cpp).

aafemt:
Do you suggest YValve to stop release interface breaking every existing application or make plugins to release objects breaking every existing third-party plugin?

aafemt:
IMHO, this behavior must be simply documented and every API user to be aware that they must release interfaces by hand if they were gotten from plugin itself rather than Y-Valve.

AlexPeshkoff:
That's a question I was waiting for. To know - do some 3d-party 
providers exist.

What about your suggestion. Yes, documented bug becomes feature, but I 
prefer to fix it. There are ways to avoid failures with old providers.

aafemt:
Yes, Magpie (provider for ODBC) exists and I raised thread in firebird-devel about exactly this issue during its development. But provider-type plugin is not the only one which have interfaces with terminating methods.

aafemt:
Perhaps the best thing you can do is to introduce a new set of Y-interfaces without terminal methods and declare old set deprecated. Also it is a good occasion to divorce client API for and plugin API.

AlexPeshkoff:
This fixed approx.80% of memory leaks in engine when running TCS. Number of leaks in yValve and network server also got smaller. But I reopen it in order to wait for full QA runs and after it backport to FB 4 and 3.

PR #6974, opened 2021-09-16
Raised API version because of changes related to #6886

aafemt:


AlexPeshkoff:
Dmitry & Vlad!
I'm not sure - is it correct reason to change FB_API_VER ?

hvlad (commented):
Since I see no usage of FB_API_VER in Firebird code, I'd say we need to define a rules first.
Every interface supports own versioning and FB_API_VER gives nothing to this, IMHO.
Perhaps FB_API_VER should be changed after massive changes in interfaces. And #6886 could be such a case. Lets define rules first. 
What is main purpose of FB_API_VER ? 
How it should be used by end-users ?

aafemt:
I would use it like this:
```
void ODBCAttachment::detach(Firebird::CheckStatusWrapper* status)
{
    internalDetach(status);
#if FB_API_VER > 40
   release();
#endif
}
```
```
#if FB_API_VER > 40
void ODBCAttachment::deprecatedDetach(Firebird::CheckStatusWrapper* status)
{
    internalDetach(status);
}
#endif
```

asfernandes:
We can also make cloop generate per-interface `#define` version, instead of only C++ const.

AlexPeshkoff:
Generating preprocessor constant together with C++ one seems logical. At least much better than artificial FB_API_VER.

PR #7010, opened 2021-10-19
Attempt to fix #7009

aafemt:
The new parameter of rollforwardSavepoint() has no default value to make sure that no call left not reviewed.


aafemt:
> There is no comments in the code and method name give me nothing to understand 
> its purpose.

   The comment is at tra.cpp:4025 as for any other routine.

hvlad:
> 
> There is no explanations when which of two methods should be used.
> If you read further there is "to optimize following rollback."

I don't know how else to explain you that there is NO info about what method should be used and when.
Until it arrives, the PR should not me merged.

aafemt:
Mentioned comment clearly say "use mergeSavepoint if rollbackSavepoint will 
surely follow". If unsure - use releaseSavepoint or rollbackSavepoint as usual.

I have no idea what can be added to this. If you have a good suggestion - I'm 
all ears.

PR #7049, opened 2021-11-22
Make savepoint statements pure DSQL, without being backed by BLR

### LINKED ISSUE #7048: Release of user savepoint releases too much
Following script fails on Firebird 4:

SQL> savepoint a;
SQL> savepoint b;
SQL> release savepoint b;
SQL> release savepoint a;
Statement failed, SQLSTATE = 3B000
Unable to find savepoint with name A in transaction context 

Firebird 3 works well.

dyemanov:
Initially this work was started as a bugfix for #7048 which looked like a one-liner (change a condition inside execute_looper() from `== savNumber` to `>= savNumber`). But I always hated the old tricks with reusing the request savepoint for `SET SAVEPOINT` and remembering that fact inside execute_looper(), it does not look robust at all. So it seemed logical to avoid creating the automatic savepoint frame around savepoint statements at all and I did that using a new request flag (see [the patch](https://github.com/FirebirdSQL/firebird/files/7582017/save.txt)). But then I realized that savepoint statements don't need to be executed through the looper at all, as they're not available in PSQL. Hence this PR is suggested (at least for master).


AlexPeshkoff (approved):
I did not review in details but an idea to skip BLR for user savepoints is definitely right.

dyemanov:
I'm mostly in doubt whether it's OK to treat savepoint statements as belonging to the "transaction management" group (TransactionNode) or better keep them as regular statements just without BLR (DsqlOnlyStmtNode).

aafemt:
Isn't it sometimes considered to make all SQL statements "blr-less"?..

AlexPeshkoff:
When answering from the blank list POV this is definitely transaction management. Would it break something in dozens of client software (like Delphi components) - hard to say for sure, but sooner of all yes, will break something somewhere.

dyemanov:
It's still described as isc_info_sql_stmt_savepoint, so nobody would notice any difference. It's just a matter of our internal consistency.

PR #7055, opened 2021-11-30
Make gfix handle io errors and other changes

romansimakov:
1) Fix page number data type to ULONG.
2) Fix subtraction backversion pages from chain pages.
3) Fix call release_page according to fetch_page.
4) CCH_fetch handles I/O errors and returns NULL in buffer if read_shadow is not set.
- Check all calls of fetch_page to handle returned value.
- Add several corrupt messages.


hvlad:
> What's about this PR?

Will look at it in a next few days.

PR #7081, opened 2021-12-21
ResultSet::getInfo() implementation

dyemanov:
At the moment only record count is supported. I've decided to return a special value -1 for uni-directional cursors, because it appears we never return isc_infona (what a pity) and only in one case return isc_infinap, so I'm not sure applications and libraries are prepared to get an error code for a properly valid info tag.

Now it's good time to consider other possible information requests and I will add them before the merge. Possible options:

- whether prefetch is enabled or not
- whether cursor is updatable
- cursor name
- mirror opcodes for isc_info_sql_records / isc_info_sql_get_plan / isc_info_sql_explain_plan (useful if result set is created via Attachment::openCursor())

Any other suggestions?

### DIFF
--- a/src/jrd/build_no.h
+++ b/src/jrd/build_no.h
@@ -3,16 +3,16 @@
                *** DO NOT EDIT ***
   TO CHANGE ANY INFORMATION IN HERE PLEASE
   EDIT src/misc/writeBuildNum.sh
-  FORMAL BUILD NUMBER:347
+  FORMAL BUILD NUMBER:351
 */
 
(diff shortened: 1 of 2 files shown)


aafemt:
Currently isc_infunk is heavily returned and thus handled by libraries. I expect no problem with error returned to applications.

AlexPeshkoff:
What about additional info requests - I do not think we should pay too much attention to inventing them right now. As soon as one gets really needed it may be easily added.
For me existing list is OK.

PR #7083, opened 2021-12-21
ResultSet::getInfo() implementation

dyemanov:
At the moment only record count is supported. I've decided to return a special value -1 for uni-directional cursors, because it appears we never return isc_infona (what a pity) and only in one case return isc_infinap, so I'm not sure applications and libraries are prepared to get an error code for a properly valid info tag.

Now it's good time to consider other possible information requests and I will add them before the merge. Possible options:

- whether prefetch is enabled or not
- whether cursor is updatable
- cursor name
- mirror opcodes for isc_info_sql_records / isc_info_sql_get_plan / isc_info_sql_explain_plan (useful if result set is created via Attachment::openCursor())

Any other suggestions?


aafemt:
Cursor name is never generated by server but always set by client application. I see no harm in it but no necessity as well.

PR #7086, opened 2021-12-23
PSQL and SQL profiler

asfernandes:


asfernandes:
Are there any objections to integrate this in master, or anything yet to discuss?

romansimakov:
In firebird-devel we talked about an ability to profile other connections. Has it been added or are you going to add it later?

asfernandes:
> In firebird-devel we talked about an ability to profile other connections. Has it been added or are you going to add it later?

I think this change would not invalidate the current design, so I think it does not make much sense to continue adding features before an initial integration and lack of feedback on what we have now.

romansimakov:
> > In firebird-devel we talked about an ability to profile other connections. Has it been added or are you going to add it later?
> 
> I think this change would not invalidate the current design, so I think it does not make much sense to continue adding features before an initial integration and lack of feedback on what we have now.

I don't mind if so

dyemanov:
I'd prefer the default plugin to be used without naming in `START_SESSION` (for example, declare NULL as a default value). It could be done via `DefaultProfilerPlugin` option in firebird.conf, for example. Alternatively, we may have `ProfilerPlugin` option in firebird.conf (similar to `TracePlugin = fbtrace`) that substitutes the missing plugin name in `START_SESSION` to any specified plugin (not only the default one). One may still specify the plugin name explicitly in `START_SESSION`, if necessary.

dyemanov:
Once we support profiling other connections (I agree this is a top priority feature), should we extend `START_SESSION` with a new parameter (`ATTACHMENT_ID DEFAULT CURRENT_ATTACHMENT`) or some new function will be introduced?

dyemanov:
While I don't mind `CREATE_SESSION` and `FINISH_SESSION` routine names, I'd suggest to rename `SESSION_ID` columns inside tables to `PROFILE_ID`. Having `SESSION_ID` and `ATTACHMENT_ID` side to side looks somewhat ambiguous, considering we have `SESSION_ID` and `USER_SESSION` meaning attachments in `RDB$GET_CONTEXT`, and also considering the SQL specification having `SESSION_USER` where _session_ also means an attachment.

dyemanov:
Why `FBPROF$` prefix for tables/views instead of just `PROF$`? We're Firebird anyway, so what's the point.

dyemanov:
Given the default profiler plugin is supposedly to be really usable (not just a draft example of how things should be coded), why its tables/views are created dynamically rather than being part of ODS? Are they expected to change significantly with time?

dyemanov:
I really miss timings (count/min/max/total) for statements as a whole. It would be inconvenient to measure them with some external tools and then look into PROF$ tables/views for details, it would be more handy to have everything in a single place. Especially if we speak about multiple executions of a single prepared/cached statement.

dyemanov:
I've attempted to collect execution times in the past, but wanted to split them into _total/cpu/wait_ parts, with the _wait_ part being also detailed (I/O, lock, latch, pause, etc). Given that extra measurements are not always dirty cheap, I had a doubt they should be presented via MON$ tables unconditionally. Now it looks like it could be integrated with your profiler design after it's committed and thus measured on demand. But the question is how deep we need to dive into the CPU time. Is it OK to calculate it as `(total - wait)` (or leave this calculation up to users), or do we want to see real CPU time and possibly deal with some delta _(total - wait - cpu)_ that remains unknown. And if we need real CPU time, do we want _user/kernel_ times? All these things complicate the implementation and make it somewhat platform-dependent.

dyemanov:
I feel the profiler package also needs a routine `SET_FLUSH_INTERVAL` that does flushing automagically, based on timer. Zero (default) means manual flushing, as designed currently.

aafemt:
> It could be done via `DefaultProfilerPlugin` option in firebird.conf, for example. Alternatively, we may have `ProfilerPlugin` option in firebird.conf (similar to `TracePlugin = fbtrace`)

Talking about configuration: because profiling is entirely engine-related, I'd say that this setting should belong to Engine14.conf, not firebird.conf.

asfernandes:
> Given the default profiler plugin is supposedly to be really usable (not just a draft example of how things should be coded), why its tables/views are created dynamically rather than being part of ODS? Are they expected to change significantly with time?

1. Then it's not really plugin design
2. It would require many code to make backup and restore of them work
3. It would not support primary and foreign keys

asfernandes:
> Why `FBPROF$` prefix for tables/views instead of just `PROF$`? We're Firebird anyway, so what's the point.

My idea was to have a FB namespace to differentiate things from users objects, as they can also use the dollar sign.

But since it's new name convention, I would not have a problem to change it.

We also have `PLG$*` tables, so maybe we should name them `PLG$PROF*`?

asfernandes:
> I really miss timings (count/min/max/total) for statements as a whole. It would be inconvenient to measure them with some external tools and then look into PROF$ tables/views for details, it would be more handy to have everything in a single place. Especially if we speak about multiple executions of a single prepared/cached statement.

How these stored timings would be different than aggregate the request-based timings per `STATEMENT_ID`?

asfernandes:
> I feel the profiler package also needs a routine `SET_FLUSH_INTERVAL` that does flushing automagically, based on timer. Zero (default) means manual flushing, as designed currently.

Data (even when flushed) is stored as part of user transaction, that may be rolled back later.

I do not see a way that automatic flush would work with this or would be less confusing than manually flush the data before read.

But this could be useful in the case of profiling others connections.

dyemanov:
> We also have `PLG$*` tables, so maybe we should name them `PLG$PROF*`?

This looks good to me.

dyemanov:
> Data (even when flushed) is stored as part of user transaction, that may be rolled back later.

Automatic flushing could behave like it's executed in autonomous transaction. Rollback will surely not be possible, but one may always delete the rows manually. This should be well-documented, of course.

dyemanov:
> How these stored timings would be different than aggregate the request-based timings per `STATEMENT_ID`?

They cannot be aggregated from `RECORD_SOURCE_STATS`, as time may also be spent between cursor operations. Perhaps it could be aggregated from `PSQL_STATS`, but only for PSQL routines. Imagine I execute `INSERT FROM SELECT` and 90% of time is spent inside VIO_store(), how could I get the total execution time for my statement? What about non-PSQL procedures/functions?

AlexPeshkoff:
On 5/1/22 19:02, Adriano dos Santos Fernandes wrote:
>
>     Given the default profiler plugin is supposedly to be really
>     usable (not just a draft example of how things should be coded),
>     why its tables/views are created dynamically rather than being
>     part of ODS? Are they expected to change significantly with time?
>
>  1. Then it's not really plugin design
>

I want to explicitly agree with Adriano here.
I see no reasons for one particular plugin to have required for it 
tables in ODS.

dyemanov:
> I want to explicitly agree with Adriano here. I see no reasons for one particular plugin to have required for it tables in ODS.

I don't mind.

AlexPeshkoff:
On 5/1/22 17:26, Dimitry Sibiryakov wrote:
>
>     It could be done via |DefaultProfilerPlugin| option in
>     firebird.conf, for example. Alternatively, we may have
>     |ProfilerPlugin| option in firebird.conf (similar to |TracePlugin
>     = fbtrace|)
>

One more possible way to go is to have a list of plugins in 
|ProfilerPlugin| option. By default first is used, users may select from 
listed in this option. As an addition this makes it possible to tune the 
set of profiler plugins for particular databases in databases.conf.

> Talking about configuration: because profiling is entirely 
> engine-related, I'd say that this setting should belong to 
> Engine14.conf, not firebird.conf.
>

More than half of firebird.conf 'is entirely engine-related'. Do you 
suggest to move all that settings to engineXX.conf? Take into an account 
that most of that settings may be tuned in databases.conf too.

hvlad:
> > Data (even when flushed) is stored as part of user transaction, that may be rolled back later.
> 
> Automatic flushing could behave like it's executed in autonomous transaction. Rollback will surely not be possible, but one may always delete the rows manually. This should be well-documented, of course.

I think that profiler data should be saved independent of final state of user transaction.
I.e. it should be available for user analysis even after rollback.

hvlad:
> > We also have `PLG$*` tables, so maybe we should name them `PLG$PROF*`?
> 
> This looks good to me.

What does it means ? PLG == ???

AlexPeshkoff:
On 5/4/22 11:38, Vlad Khorsun wrote:
>
>         Data (even when flushed) is stored as part of user
>         transaction, that may be rolled back later.
>
>     Automatic flushing could behave like it's executed in autonomous
>     transaction. Rollback will surely not be possible, but one may
>     always delete the rows manually. This should be well-documented,
>     of course.
>
> I think that profiler data should be saved independent of final state 
> of user transaction.
> I.e. it should be available for user analysis even after rollback.
>

May be let user specify separate transaction for profiling? If that 
specific transaction is rolled back certainly profiler data is lost, 
moreover, if user wants it may be same transaction as one used in other 
DB operations. But when needed profiler data is easily done available 
after main transaction rollback.

aafemt:
> More than half of firebird.conf 'is entirely engine-related'. Do you
> suggest to move all that settings to engineXX.conf?

   Yes. AFAIU they are kept in firebird.conf for historical reason only and it 
makes coexistence of several engines harder.

> Take into an account
> that most of that settings may be tuned in databases.conf too.

   That's the reason for https://github.com/FirebirdSQL/firebird/issues/4802

hvlad:
> May be let user specify separate transaction for profiling?

Sorry, I see no sense in dropping profile data when user tx is rolled back. 
Should we drop run-time stats in such a case ? (of course - not) 
How to profile on-rollback trigger, for example ?

AlexPeshkoff:
On 5/4/22 12:29, Vlad Khorsun wrote:
>
>     May be let user specify separate transaction for profiling?
>
> Sorry, I see no sense in dropping profile data when user tx is rolled 
> back.
> Should we drop run-time stats in such a case ? (of course - not)
>

Certainly, they are initially put out of transaction control. BTW, this 
has big price - accessing them from SQL is impossible, i.e. one can't 
join run-time stats with a table or use in PL/SQL procedure.

> How to profile on-rollback trigger, for example ?
>

With separate profiled and profiler transactions it's not a problem. 
Profiled transaction performs rollback, in profiler transaction results 
are visible.

AlexPeshkoff:
On 5/4/22 12:26, Dimitry Sibiryakov wrote:
> > More than half of firebird.conf 'is entirely engine-related'. Do you
> > suggest to move all that settings to engineXX.conf?
>
> Yes. AFAIU they are kept in firebird.conf for historical reason only 
> and it
> makes coexistence of several engines harder.
>
> > Take into an account
> > that most of that settings may be tuned in databases.conf too.
>
> That's the reason for https://github.com/FirebirdSQL/firebird/issues/4802

If we talk about all settings - yes, that certainly makes sense. But 
should be well thought in many aspects. Anyway - it's offtopic in 
profiler discussion.

hvlad:
>>>May be let user specify separate transaction for profiling? 

>> Sorry, I see no sense in dropping profile data when user tx is rolled back. 
Should we drop run-time stats in such a case ? (of course - not)

> Certainly, they are initially put out of transaction control. 

Not sure I got you - certainly yes (drop run-time stats on rollback) or not ?

>BTW, this has big price - accessing them from SQL is impossible, i.e. one can't join run-time stats with a table or use in PL/SQL procedure.

It is not about price, it is sooner about way to represent non-SQL data. 
Nobody force us to save profiler data into persistent tables.
If one want SQL interface, it could be any kind of tables - persistent, virtual (like RDB$CONFIG) and so on.
It also could be not a tables but stored procedures.
Also, I can imagine XML or JSON -like interface to query all profile data at one call.

>> How to profile on-rollback trigger, for example ?

> With separate profiled and profiler transactions it's not a problem. Profiled transaction performs rollback, in profiler transaction results are visible.

I see no reason to make any relation between user transaction and collecting (storing) of profile data.
Any use of transactions by profiler is implementation details and it should not affect user session as much as possible.

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1695,3 +1696,49 @@ interface ReplicatedSession : PluginBase
...
+		ISC_TIMESTAMP_TZ timestamp);
+
+	void flush(Status status, Transaction transaction);
+}
+
+interface ProfilerSession : Disposable
+{
+	const uint FLAG_BEFORE_EVENTS = 0x1;
+	const uint FLAG_AFTER_EVENTS = 0x2;
+
```
hvlad, line 1715: What is the purpose of these flags ?

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1695,3 +1696,49 @@ interface ReplicatedSession : PluginBase
...
+	int64 getId();
+	uint getFlags();
+
```
hvlad, line 1718: What is the purpose of this function ?

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1695,3 +1696,49 @@ interface ReplicatedSession : PluginBase
...
 }
 
+// Profiler interfaces
+
+interface ProfilerPlugin : PluginBase
+{
+	void init(Status status, Attachment attachment, Transaction transaction);
+
+	ProfilerSession startSession(Status status, Transaction transaction, const string description,
+		ISC_TIMESTAMP_TZ timestamp);
```
hvlad, line 1706: Why is requires `transaction` ? 
What transaction will be passed into this routine - user tx or separate one ?
How this transaction could be used by plugin ?

Could it be so that plugin creates own metadata using this transaction ? 
Could plugin save own data using this transaction ?

AlexPeshkoff:
On 5/4/22 13:08, Vlad Khorsun wrote:
>
>             May be let user specify separate transaction for profiling?
>
>         Sorry, I see no sense in dropping profile data when user tx is
>         rolled back.
>         Should we drop run-time stats in such a case ? (of course - not)
>
>     Certainly, they are initially put out of transaction control.
>
> Not sure I got you - certainly yes (drop run-time stats on rollback) 
> or not ?
>

Currently of course - not.

>     BTW, this has big price - accessing them from SQL is impossible,
>     i.e. one can't join run-time stats with a table or use in PL/SQL
>     procedure.
>
> It is not about price, it is sooner about way to represent non-SQL data.
> Nobody force us to save profiler data into persistent tables.
> If one want SQL interface, it could be any kind of tables - 
> persistent, virtual (like RDB$CONFIG) and so on.
> It also could be not a tables but stored procedures.
>

This can also be temp tables - all advantages of accessing like 
persistent one but not poison database.

> Also, I can imagine XML or JSON -like interface to query all profile 
> data at one call.
>

We need such a way to access data but discussing that related to 
profiling only is definitely wrong.

>         How to profile on-rollback trigger, for example ?
>
>     With separate profiled and profiler transactions it's not a
>     problem. Profiled transaction performs rollback, in profiler
>     transaction results are visible.
>
> I see no reason to make any relation between user transaction and 
> collecting (storing) of profile data.
> Any use of transactions by profiler is implementation details and it 
> should not affect user session as much as possible.
>

That's also possible approach - but on my mind _abiity_ to control that 
may be useful.
Even to quickly ropllback that data sometimes (sheet, created wrong 
request which returned useless data...).
I.e. one can pass null transaction and in that case everything works 
like you describe. But when needed user transaction may take place too.

asfernandes:
Let me describe relation of profiling and transactions.

A profile when started profiles the whole actions happening in an attachment.

When the plugin receives the start event it receives a transaction (the user one calling RDB$PROFILER.START_SESSION), that it may only use during that call, for example, to create metadata in the database. It can't hold that transaction for later use.

Then the plugin receive events and aggregate then in memory. These events do not receive a transaction.

Later when user ask for flush, plugin receives a transaction (the user one calling RDB$PROFILER.FLUSH or other RDB$PROFILER function allowing flush) and stores the data.

That works very well for cases that user opens ISQL (or other tool), start profiler session, run commands, flush and query data.

User also is able to flush data in autonomous transaction and query this data in another tool.

Flush is the more costing action. In this case it does not happen together with the events the user is profiling.

And it also allows the user to just see profiling data while testing their software, for example, with an snapshot transaction.

This is a real use case for the profiler, so it must be possible to do it in this way.

There are may be others use cases when profiling others attachments, so we should have some extension that works well in this context too.

src/include/firebird/FirebirdInterface.idl
asfernandes: It's user transaction and plugin could use it during the call. It should not hold it for later use.

That should be documented.

src/include/firebird/FirebirdInterface.idl
asfernandes: The plugin is able to receive "before" or "after" events, but it may not be interested in both.

The default plugin only deals with "after" events.

"Before" events may be used by some plugin that tracks what code is running and visually show it to the user.

asfernandes:
> Sorry, I see no sense in dropping profile data when user tx is rolled back.

That is not dropped as it is stored in memory.

Data (of whole profile session) is only dropped if session is never flushed or finished.

hvlad:
> Let me describe relation of profiling and transactions.
> 
> A profile when started profiles the whole actions happening in an attachment.
> 
> When the plugin receives the start event it receives a transaction (the user one calling RDB$PROFILER.START_SESSION), that it may only use during that call, for example, to create metadata in the database. It can't hold that transaction for later use.
> 
> Then the plugin receive events and aggregate then in memory. These events do not receive a transaction.
> 
> Later when user ask for flush, plugin receives a transaction (the user one calling RDB$PROFILER.FLUSH or other RDB$PROFILER function allowing flush) and stores the data.

So, if user call RDB$PROFILER.START_SESSION and RDB$PROFILER.FLUSH in the same (own) transaction, profiler will create metadata and attempt to use it (to INSERT something) in the same transaction ?

hvlad:
> > Sorry, I see no sense in dropping profile data when user tx is rolled back.
> 
> Data is not dropped as it is stored in memory.
> 
> Data (of whole profile session) is only dropped if session is never flushed or finished.

What happens when user flushes the data and then rolled back transaction ?

src/include/firebird/FirebirdInterface.idl
hvlad: Can I get the answers on two last questions too, pls ?

asfernandes: I described things in others answers of what default plugin does with it and here what plugin could not do (hold transaction for later use).

Of course, any prohibited usage thing by Firebird should not be done by plugin - example: create metadata and use in same transaction.

asfernandes:
> So, if user call RDB$PROFILER.START_SESSION and RDB$PROFILER.FLUSH in the same (own) transaction, profiler will create metadata and attempt to use it (to INSERT something) in the same transaction ?

No, and actually things happens a bit different.

The default plugin uses the transaction from startSession to query a sequence to make the session id.

There is the init method. Init receives a attachment and transaction.

The default plugin uses the attachment to start a new transaction to create the metadata.

Init is called by engine once per database/plugin when a profiler session is about to be created.

src/include/firebird/FirebirdInterface.idl
hvlad: Sorry, this is not the answer I could expect.
My point is that good design should nor allow nor provoke users to do wrong things.
So far I see no strong reason to expose user transaction in the profiler plugin interface.

hvlad:
> > So, if user call RDB$PROFILER.START_SESSION and RDB$PROFILER.FLUSH in the same (own) transaction, profiler will create metadata and attempt to use it (to INSERT something) in the same transaction ?
> 
> No, and actually things happens a bit different.
> 
> The default plugin uses the transaction from startSession to query a sequence to make the session id.
> 
> There is the init method. Init receives a attachment and transaction.
> 
> The default plugin uses the attachment to start a new transaction to create the metadata.
> 
> Init is called by engine once per database/plugin when a profiler session is about to be created.

  Thanks. I still see no reason to expose user transaction to the profiler plugin:
- `Init` is not uses user transaction,
- `startSession` could use user attachment to make the session id using separate (non-user) transaction, or
- engine could generate and provide profile session id for profiler plugin.

aafemt:
Besides plugin can generate session id by any other external mean, including GUID. It is plugin's internal thing, right?

hvlad:
> Besides plugin can generate session id by any other external mean, including GUID. It is plugin's internal thing, right?

Good question, it should be considered also for multi-connection profile sessions and a multi-process environment (CS, embedded, etc).

src/include/firebird/FirebirdInterface.idl
asfernandes: The passed transactions could be (and are) used to query if metadata is already created in init and get new sequence in startSession.

Otherwise plugin would need to start transaction to do that. So IMO avoid the transaction does not simplify things nor make things faster.

It's a matter of document how not to use transaction. It's up to plugin to use it in valid situations or not use it.

hvlad: One the one side we have optimization to avoid new transaction when profile session starts.

One the other side we have potential security issue when plugin run own actions in context of user transaction, 
also we have design issue when public API expose too much for external code.

I see no problem having to start and commit new transaction once per profiler session - it is not often and not requires to be highly optimized event.

BTW, I'm not sure profiler metadata should be owned by random user who happens to be first to use profiler. It is better to use separate account for it and make all necessary grants for public (or less wide) usage, IMHO.

AlexPeshkoff: I agree that metadata creation must not run under random user because in that case we can't correctly control access to it (BTW, that user may simply have no appropriate rights). I think that for this particular operation new SYSDBA attachment should be created which must create appropriate tables, role (something like PLG$PROFILE) if it does not exist yet and grant access on that tables to that role.

What about transaction in which profiler checks presence & access to that table - both approaches (use users transaction vs. start new one) have some advantages, i.e. I do not think it's critical API issue.

hvlad: > I agree that metadata creation must not run under random user because in that case we can't correctly control access to it (BTW, that user may simply have no appropriate rights). I think that for this particular operation new SYSDBA attachment should be created which must create appropriate tables, role (something like PLG$PROFILE) if it does not exist yet and grant access on that tables to that role.

Yes, I have similar idea.

> What about transaction in which profiler checks presence & access to that table - both approaches (use users transaction vs. start new one) have some advantages, i.e. I do not think it's critical API issue.

Expose user transaction to the unknown (untrusted ?) external plugin... hmm... may be not critical but definitely not safe, IMHO. More, now I see that allow plugin to use user attachment is also very far from safe.

Formally, profiler should be allowed to collect some stats and save it somewhere for further usage, correct ? 
I see no requirement here to provide profiler with a way to change user context or user data.
Ideally we should limit profiler access so that it is only allowed to create metadata and run DML on it.

Public API should not allow external plugin to do something that it should not do.

aafemt: Plugins are working inside of server and they receive master interface that allows them to connect any database including security one in embedded mode bypassing any security checks. There is no point to close the stable door.

asfernandes: > BTW, I'm not sure profiler metadata should be owned by random user who happens to be first to use profiler.

I also do not like it.

> It is better to use separate account for it and make all necessary grants for public (or less wide) usage, IMHO.

Is it 100% safe to get database name with `rdb$get_context('SYSTEM', 'DB_NAME')` in server and reconnect to it with `SYSDBA` user and no password in the server?

asfernandes: > Plugins are working inside of server and they receive master interface that allows them to connect any database including security one in embedded mode bypassing any security checks. There is no point to close the stable door.

Yes. Plugins must be trusted by server admin. They can also crash the server.

AlexPeshkoff:
On 5/5/22 22:15, Adriano dos Santos Fernandes wrote:
>
> Is it 100% safe to get database name with |rdb$get_context('SYSTEM', 
> 'DB_NAME')| in server and reconnect to it with |SYSDBA| user and no 
> password in the server?
>

For second part I'm 100% sure - it's safe. What about first part - also 
appears to be OK but be aware that result depends upon DatabaseAccess in 
firebird.conf and may be something else - do not remember right now.

aafemt:
Even with NONE you'll get the alias useful for connection.

AlexPeshkoff:
On 5/6/22 15:27, Dimitry Sibiryakov wrote:
>
> Even with NONE you'll get the alias useful for connection.
>

Yes. But this does not work if in the middle someone changed 
databases.conf. To be precise in a case of such change & 
databaseAccess=NONE you are never able to attach to required DB.

src/include/firebird/FirebirdInterface.idl
hvlad: > Plugins are working inside of server and they receive master interface that allows them to connect any database including security one in embedded mode bypassing any security checks. There is no point to close the stable door.

Engine could pass special impl of master interface that not allows to make connections at all.

AlexPeshkoff:
On 5/6/22 15:36, Vlad Khorsun wrote:
>
>
>
> In src/include/firebird/FirebirdInterface.idl 
> <https://github.com/FirebirdSQL/firebird/pull/7086#discussion_r866783562>:
>
> > @@ -1695,3 +1696,49 @@ interface ReplicatedSession : PluginBase
>   	void setSequence(Status status, const string name, int64 value);
>   }
>   
> +// Profiler interfaces
> +
> +interface ProfilerPlugin : PluginBase
> +{
> +	void init(Status status, Attachment attachment, Transaction transaction);
> +
> +	ProfilerSession startSession(Status status, Transaction transaction, const string description,
> +		ISC_TIMESTAMP_TZ timestamp);
>
>     Plugins are working inside of server and they receive master
>     interface that allows them to connect any database including
>     security one in embedded mode bypassing any security checks. There
>     is no point to close the stable door.
>
> Engine could pass special impl of master interface that not allows to 
> make connections at all.
>

IMaster* fullMaster = fb_get_master_interface();

hvlad:
>> Engine could pass special impl of master interface that not allows to make connections at all.

> IMaster* fullMaster = fb_get_master_interface();

This is not a problem as engine know it calls plugin code and ready to handle callbacks. 
Just decide if we want/need to hardening security in such a way.

AlexPeshkoff:
On 5/6/22 15:45, Vlad Khorsun wrote:
>
>         Engine could pass special impl of master interface that not
>         allows to make connections at all.
>
>     IMaster* fullMaster = fb_get_master_interface();
>
> This is not a problem as engine know it calls plugin code and ready to 
> handle callbacks.
>

Including other threads? :)

> Just decide if we want/need to hardening security in such a way.
>

Do you think we can limit access of part of a code running inside a 
process from the rest of that process?

hvlad:
> And main problem - what a sense to try to limit in something plugin's code which anyway works in same process with engine? With little hackery it can just find list of databases in engine and add own attachment to one of them.

Ok, we give up to make external code safer. At least until it run in the engine process.
In any case I still see no good reason to expose user transaction to the profiler.

Honestly, I see no good reason to make new plugin interface at all - offered functionality is hard to extend by different impl and it could be just a part of engine (as part of existing runtime stats).

aafemt:
Yes, neither user transaction nor user attachment make sense in a profiling plugin. If the profiler indeed need to be a plugin - it can receive ITrace* objects that provides enough information about them.

AlexPeshkoff:
On 5/6/22 15:52, Vlad Khorsun wrote:
>
> Honestly, I see no good reason to make new plugin interface at all - 
> offered functionality is hard to extend by different impl and it could 
> be just a part of engine (as part of existing runtime stats).
>

That's what we should begin with.
May be Adriano has ideas why profiler is implemented as plugin?

asfernandes:
> On 5/6/22 15:52, Vlad Khorsun wrote: Honestly, I see no good reason to make new plugin interface at all - offered functionality is hard to extend by different impl and it could be just a part of engine (as part of existing runtime stats).
> That's what we should begin with.

That is what we *had* begun with. Almost two years ago in this another PR https://github.com/FirebirdSQL/firebird/pull/285!

> May be Adriano has ideas why profiler is implemented as plugin?

Also commented on changes on initial design here: https://github.com/FirebirdSQL/firebird/pull/285#issuecomment-911919810

We should better have feedbacks sooner than later when one asks for a merge.

And I still insist this design with plugin is better than everything in the engine.

asfernandes:
> > > Engine could pass special impl of master interface that not allows to make connections at all.
> 
> > IMaster* fullMaster = fb_get_master_interface();
> 
> This is not a problem as engine know it calls plugin code and ready to handle callbacks. Just decide if we want/need to hardening security in such a way.

Security of server plugins are very offtopic here. They were "insecure" by design in C++ code.

Security could only be handled with others layers like containers or things like apparmor/selinux.

It's funny that in FB/Java I made the Java code run in Java sandbox configurable security and many were against it saying if a code is in server, someone should had trusted it.

asfernandes:
> But please agree that n that both cases transaction is useless.

It's required in flush method. It allows to save data in same transaction to meet use case that I previously described.

In init method it's used for initial metadata inspection.

It avoids more code in the plugin and more work for the engine and database, not introducing any real problem.

If it hurts someone, I see no sense in fight for a so minor thing.

asfernandes:
> Once we support profiling other connections (I agree this is a top priority feature), should we extend `START_SESSION` with a new parameter (`ATTACHMENT_ID DEFAULT CURRENT_ATTACHMENT`) or some new function will be introduced?

The former options seems good for me.

asfernandes:
I've implemented profiling of others attachments with this commit set. The semantics of it is documented in the readme.

To make things not confusing, user's transaction is not passed to plugins anymore and flush always starts its own transaction.

asfernandes:
> > How these stored timings would be different than aggregate the request-based timings per `STATEMENT_ID`?
> 
> They cannot be aggregated from `RECORD_SOURCE_STATS`, as time may also be spent between cursor operations. Perhaps it could be aggregated from `PSQL_STATS`, but only for PSQL routines. Imagine I execute `INSERT FROM SELECT` and 90% of time is spent inside VIO_store(), how could I get the total execution time for my statement? What about non-PSQL procedures/functions?

I think we need TOTAL time of requests (as stored in `PLG$PROF_REQUESTS`) which is valid for SQL and PSQL.

Then it's possible to calculate MIN/MAX per STATEMENT (as stored in `PLG$PROF_STATEMENTS`).

asfernandes:
> I've attempted to collect execution times in the past, but wanted to split them into _total/cpu/wait_ parts, with the _wait_ part being also detailed (I/O, lock, latch, pause, etc). Given that extra measurements are not always dirty cheap, I had a doubt they should be presented via MON$ tables unconditionally. Now it looks like it could be integrated with your profiler design after it's committed and thus measured on demand. But the question is how deep we need to dive into the CPU time. Is it OK to calculate it as `(total - wait)` (or leave this calculation up to users), or do we want to see real CPU time and possibly deal with some delta _(total - wait - cpu)_ that remains unknown. And if we need real CPU time, do we want _user/kernel_ times? All these things complicate the implementation and make it somewhat platform-dependent.

AFAIK there is no way to directly get the "wait time" from the OS. It's the elapsed time minus the thread's CPU time (user + kernel).

I think we can pass the elapsed and total thread's CPU time. The helper views could also calculate the total wait time.

We should decide if the APIs that currently receives only `uint64 runTime` would get a new `uint64 cpuTime` parameter, hence will not be extensible.

Or if we would need to add another interface which would be extensible and plugin writers would call its methods to get timings. The downside of this is that it will be slower.

dyemanov:
Yes, no way to get wait time from the system. But we can measure "logical waits" ourselves in most points where we checkout from the engine. Some of them won't be honest (IO wait time will actually be IO CPU time when reads are performed from filesystem cache, for example), but this is OK as they're expected to be short and unlikely to be noticeable as "top" waits. If the CPU time is also measured, then we could calculate "wait" as "total time - CPU time" and it should be more or less real, but I doubt it gonna be useful. So I'm somewhat skeptical in whether we really need CPU time...

asfernandes:
Anyone see a blocking point to have this merged in master?

AlexPeshkoff:
On 5/4/22 11:39, Vlad Khorsun wrote:
>
>         We also have |PLG$*| tables, so maybe we should name them
>         |PLG$PROF*|?
>
>     This looks good to me.
>
> What does it means ? PLG == ???
>

That's just prefix like RDB$, see PLG$SRP_USERS in sec db.

asfernandes:
On 04/05/2022 07:49, Vlad Khorsun wrote:
> 
> So, if user call RDB$PROFILER.START_SESSION and RDB$PROFILER.FLUSH in
> the same (own) transaction, profiler will create metadata and attempt to
> use it (to INSERT something) in the same transaction ?
> 

No, and actually things happens a bit different.

The default plugin uses the transaction from startSession to query a
sequence to make the session id.

There is the init method. Init receives a attachment and transaction.

The default plugin uses the attachment to start a new transaction to
create the metadata.

Init is called by engine once per database/plugin when a profiler
session is about to be created.


Adriano

pavel-zotov:
@@@ QA issue @@@
Test verifies only example from doc/sql.extensions/README.profiler.md
More complex checks will be implementer later.

PR #7088, opened 2021-12-24
Fix MacOS UDR and Legacy_UserManager plugins not working due to not exported entry point.

asfernandes:
Add FB_DLL_EXPORT to public headers.

Use default visibility also for Linux/Unix (in addition to MacOS) as
user application can also be compiled with -fvisibility=hidden.

### DIFF
--- a/src/include/ibase.h
+++ b/src/include/ibase.h
@@ -66,8 +66,12 @@
 
 #if defined(WIN32) || defined(_WIN32) || defined(__WIN32__)
 #define FB_DLL_EXPORT __declspec(dllexport)
-#elif defined(__APPLE__) || defined(__linux__) || defined(unix) || defined(__unix__) || defined(__unix)
+#elif defined __has_attribute
+#if __has_attribute (visibility)
 #define FB_DLL_EXPORT __attribute__ ((visibility("default")))
+#else
+#define FB_DLL_EXPORT
+#endif
 #else
 #define FB_DLL_EXPORT
 #endif
--- a/src/include/ibase.h
+++ b/src/include/ibase.h
@@ -64,6 +64,14 @@
 #define FB_API_DEPRECATED
 #endif
 
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__)
+#define FB_DLL_EXPORT __declspec(dllexport)
+#elif defined(__APPLE__) || defined(__linux__) || defined(unix) || defined(__unix__) || defined(__unix)
+#define FB_DLL_EXPORT __attribute__ ((visibility("default")))
+#else
+#define FB_DLL_EXPORT
+#endif
+
 #include "./firebird/impl/types_pub.h"
 
 /***********************/
--- a/doc/Using_OO_API.html
+++ b/doc/Using_OO_API.html
@@ -1673,7 +1673,7 @@ only for some specific OS you can make this place a bit simpler. In
 minimum case the function should register module and all factories in
 plugin manager:</font></p>
 <p style="margin-bottom: 0cm"><font size="4" style="font-size: 14pt">extern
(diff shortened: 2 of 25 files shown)


src/include/ibase.h
```diff
@@ -64,6 +64,14 @@
 #define FB_API_DEPRECATED
 #endif
 
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__)
+#define FB_DLL_EXPORT __declspec(dllexport)
+#elif defined(__APPLE__) || defined(__linux__) || defined(unix) || defined(__unix__) || defined(__unix)
+#define FB_DLL_EXPORT __attribute__ ((visibility("default")))
```
AlexPeshkoff, line 70: I'm not sure - does \_\_attribute\_\_ ((visibility("default"))) work in non-GCC case? I.e. is not check for unix too wide, may be something like check for defined(\_\_GNUC\_\_) should be added?

aafemt: Another problem here is that __declspec() must be used BEFORE function name while __attribute__() - AFTER it. I.e. declaration `void FB_DLL_EXPORT foo()` won't work with GCC.

AlexPeshkoff:
On 12/24/21 19:39, Dimitry Sibiryakov wrote:
>
>
>
> In src/include/ibase.h 
> <https://github.com/FirebirdSQL/firebird/pull/7088#discussion_r775052721>:
>
> > @@ -64,6 +64,14 @@
>   #define FB_API_DEPRECATED
>   #endif
>   
> +#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__)
> +#define FB_DLL_EXPORT __declspec(dllexport)
> +#elif defined(__APPLE__) || defined(__linux__) || defined(unix) || defined(__unix__) || defined(__unix)
> +#define FB_DLL_EXPORT __attribute__ ((visibility("default")))
>
> Another problem here is that __declspec() must be used BEFORE function 
> name while *attribute*() - AFTER it. I.e. declaration |void 
> FB_DLL_EXPORT foo()| won't work with GCC.
>

According to https://gcc.gnu.org/wiki/Visibility it works.

aafemt:
It looks like it can work in form `FB_DLL_EXPORT void foo()` though I remember to have a problem with it on Linux where GCC issued warning about attribute have no effect.

src/include/ibase.h
asfernandes: > I'm not sure - does __attribute__ ((visibility("default"))) work in non-GCC case? I.e. is not check for unix too wide, may be something like check for defined(__GNUC__) should be added?

Or maybe this?

https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html

AlexPeshkoff:
On 12/24/21 20:00, Adriano dos Santos Fernandes wrote:
>
>
>
> In src/include/ibase.h 
> <https://github.com/FirebirdSQL/firebird/pull/7088#discussion_r775055414>:
>
> > @@ -64,6 +64,14 @@
>   #define FB_API_DEPRECATED
>   #endif
>   
> +#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__)
> +#define FB_DLL_EXPORT __declspec(dllexport)
> +#elif defined(__APPLE__) || defined(__linux__) || defined(unix) || defined(__unix__) || defined(__unix)
> +#define FB_DLL_EXPORT __attribute__ ((visibility("default")))
>
>     I'm not sure - does *attribute* ((visibility("default"))) work in
>     non-GCC case? I.e. is not check for unix too wide, may be
>     something like check for defined(*GNUC*) should be added?
>
> Or maybe this?
>
> https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html
>

Certainly, it's ok.

AlexPeshkoff (approved):

PR #7125, opened 2022-02-06
Make DDL object type constants stable and extendable

romansimakov:
Rework DDL access checks.
Remove unneeded constants, duplicating code.
Rename some functions to suit their purpose.
Provide a way to extent object type constants in future not changing existing ones.


src/jrd/obj.h
```diff
@@ -24,56 +24,89 @@
...
-const int obj_filters			= obj_last_non_ddl + 13;
-
-const int obj_type_MAX			= obj_last_non_ddl + 14;	// keep this last!
-
-// used in the parser only / no relation with obj_type_MAX (should be greater)
-const int obj_user_or_role		= 100;
-const int obj_schema			= 101;
-const int obj_parameter			= 102;
-
-inline const char* get_object_name(int object_type)
```
AlexPeshkoff, line 76: I doubt this change of object list to enum is needed. On the one hand it's correct change in good direction but on the other when that data is stored in database and used in for loops what we get is a lot of casts here & there, poisoning code. I like use of ObjectType but may be it should be not enum, just typedef? Or (if you want thin control on what's happening) create class ObjectType with appropriate auto casts, with checks in DEV_BUILD. Though must say I do not remember problems when illegal value of integer was used as object type...

AlexPeshkoff (changes requested):

src/jrd/obj.h
romansimakov: Done

src/jrd/dfw.epp
```diff
@@ -2650,7 +2650,7 @@ static bool grant_privileges(thread_db* tdbb, SSHORT phase, DeferredWork* work,
 		return true;
 
 	case 3:
-		GRANT_privileges(tdbb, work->dfw_name, work->dfw_id, transaction);
+		GRANT_privileges(tdbb, work->dfw_name, (ObjectType)work->dfw_id, transaction);
```
AlexPeshkoff, line 2653: is C-cast really needed here?

romansimakov: not really

AlexPeshkoff (approved):

PR #7130, opened 2022-02-15
Assertion failure when tracing is enabled

Noremos:
In the debug build I got the following error:
`Assertion (false) failure: ..\..\..\src\jrd\trace\TraceObjects.cpp 239`
when I tried to create a database via isql with the following trace configuration:
```
database
{
	enabled = true
	log_statement_start = true
	log_statement_finish = true
	time_threshold = 0
}
```
This started happening after a recent refactoring. So I tried to fix the problem.

### DIFF
--- a/src/dsql/DsqlRequests.h
+++ b/src/dsql/DsqlRequests.h
@@ -241,6 +241,11 @@ public:
 		Firebird::IMessageMetadata* outMetadata, UCHAR* outMsg,
 		bool singleton) override;
 
+	bool isDml() const override
+	{
+		return true;
+	}
+
 private:
 	NestConst<TransactionNode> node;
 };


asfernandes:
I tried to put this content in `fbtrace.conf` and had no problem creating a database. How to reproduce?

Your fix is not correct. `isDml()` means cast to `DsqlDmlRequest*` should be possible, and that is not the case for `DsqlTransactionRequest`.

asfernandes:
I reproduced it. Please test the code I committed to master now.

Noremos:
> Your fix is not correct. `isDml()` means cast to `DsqlDmlRequest*` should be possible, and that is not the case for `DsqlTransactionRequest`.

I took a closer look at the code and now I see that you are right. I will correct my patch

asfernandes:
> I took a closer look at the code and now I see that you are right. I will correct my patch

Please take a look if it's already corrected in master.

Noremos:
> Please test the code I committed to master now.

Now it works brilliantly. Thanks

PR #7131, opened 2022-02-16
Improve performance and fix problem related to character contractions in the end of indexed fields

### LINKED ISSUE #7093: Improve indexed lookup speed of strings when the last keys characters are part of collated contractions
This relates to https://github.com/FirebirdSQL/firebird/issues/6915 but without the need to set `DISABLE-COMPRESSIONS=1`.

Test cases:

```
create collation UNICODE_CSCZ_CI
   for UTF8  
   from UNICODE  
   case insensitive  
   'LOCALE=cs_CZ'
;
   
create collation UNICODE_CSCZ_CS
   for UTF8  
   from UNICODE  
   case sensitive  
   'LOCALE=cs_CZ'
;

CREATE TABLE TEST1M (
  ANSI_CZ VARCHAR(10)  CHARACTER SET WIN1250 COLLATE PXW_CSY,
  UNICODE_CS_CZ VARCHAR(10)  CHARACTER SET UTF8 COLLATE UNICODE_CSCZ_CS,
  UNICODE_CI_CZ VARCHAR(10) CHARACTER SET UTF8 COLLATE UNICODE_CSCZ_CI,
  UNICODE_CS VARCHAR(10) CHARACTER SET UTF8,
  UNICODE_CI VARCHAR(10) CHARACTER SET UTF8 COLLATE UNICODE_CI
);

CREATE OR ALTER PROCEDURE GetStr(AORDERID BIGINT)
RETURNS (AResult CHAR(10)) AS
declare variable Base36Chars CHAR(36);
declare variable mResult VARCHAR(10);
declare variable ID BIGINT;
declare variable I INT;
BEGIN
    Base36Chars = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ';
    mResult = '';
    AResult = mResult;
    ID = AORDERID;
    WHILE (ID > 0) DO
    BEGIN
      I = MOD(ID, 36);
      ID = ID / 36;
      mResult = mResult || SubString(Base36Chars from I + 1 for 1);
    END
    AResult = LEFT(mResult || '0000000', 7);
  Suspend;
END;

-- Generate test string data
-- 000000, 100000...900000...A00000...Z00000,
-- 010000, 110000...910000...A10000...Z10000,
-- ...

EXECUTE BLOCK
AS
  DECLARE ROWSCOUNT INT = 1000000;
  DECLARE I INT = 0;
  DECLARE C INT = 0;
  DECLARE Str VARCHAR(10);
BEGIN
  WHILE (C < ROWSCOUNT) DO
  BEGIN
    SELECT AResult from GetStr(:I) into :Str;
    -- Skip Y, Z
    IF ((LEFT(Str, 1) <> 'Y') AND (LEFT(Str, 1) <> 'Z')) THEN BEGIN
      INSERT INTO TEST1M(ANSI_CZ, UNICODE_CS_CZ, UNICODE_CI_CZ, UNICODE_CS, UNICODE_CI) VALUES (:Str, :Str, :Str, :Str, :Str);
      C = C + 1;
    END
    I = I + 1;
  END
END;

CREATE INDEX TEST1M_ANSI_CZ ON TEST1M (ANSI_CZ);
CREATE INDEX TEST1M_UNICODE_CS_CZ ON TEST1M (UNICODE_CS_CZ);
CREATE INDEX TEST1M_UNICODE_CI_CZ ON TEST1M (UNICODE_CI_CZ);
CREATE INDEX TEST1M_UNICODE_CS ON TEST1M (UNICODE_CS);
CREATE INDEX TEST1M_UNICODE_CI ON TEST1M (UNICODE_CI);

SELECT COUNT(*) FROM TEST1M;
  -- Time 269ms                           <<<<<<< OK >>>>>>>
  -- Result 1000000

--######################################### Scenario use WHERE >= #########################################
-- Problem only in CZECH collate case insensitive
   
  SELECT ANSI_CZ FROM TEST1M
  WHERE ANSI_CZ >= 'Z'
  ORDER BY ANSI_CZ;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing                     
 
  SELECT UNICODE_CS FROM TEST1M
  WHERE UNICODE_CS >= 'Z'
  ORDER BY UNICODE_CS;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing                     
 
  SELECT UNICODE_CI FROM TEST1M
  WHERE UNICODE_CI >= 'Z'
  ORDER BY UNICODE_CI;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing

  SELECT UNICODE_CS_CZ FROM TEST1M
  WHERE UNICODE_CS_CZ >= 'Z'
  ORDER BY UNICODE_CS_CZ;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing

  SELECT UNICODE_CI_CZ FROM TEST1M        
  WHERE UNICODE_CI_CZ >= 'Z'
  ORDER BY UNICODE_CI_CZ;
    -- Time: 4,294s                       <<<<<<< COLLATE CASE INSENSITIVE HEAR IS PERFORMANCE PROBLEM FOR VALUE "Z" 4,294s WHY? VALUE "Y" IS OK 0ms >>>>>>>
    -- Result nothing

  SELECT UNICODE_CI_CZ FROM TEST1M
  WHERE UNICODE_CI_CZ >= 'Y'
  ORDER BY UNICODE_CI_CZ;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing
 
  SELECT FIRST 1 UNICODE_CI_CZ FROM TEST1M        
  WHERE UNICODE_CI_CZ >= 'C'
  ORDER BY UNICODE_CI_CZ;
    -- Time: 1,531s                       <<<<<<< collate case INSENSITIVE NEXT PERFOMANCE PROBLEM HAVE VALUE "C" 1,531s WHY? VALUE "D" IS OK 0ms >>>>>>>
    -- Result C000000   

  SELECT FIRST 1 UNICODE_CI_CZ FROM TEST1M        
  WHERE UNICODE_CI_CZ >= 'D'
  ORDER BY UNICODE_CI_CZ;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result D000000      
 
--######################################### Scenario use WHERE like #########################################   
-- Problem in CZECH collate case sensitive and insensitive    
  SELECT ANSI_CZ FROM TEST1M
  WHERE ANSI_CZ LIKE 'Z%'
  ORDER BY ANSI_CZ;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing      
                   
  SELECT UNICODE_CS FROM TEST1M
  WHERE UNICODE_CS LIKE 'Z%'
  ORDER BY UNICODE_CS;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing                     

  SELECT UNICODE_CI FROM TEST1M
  WHERE UNICODE_CI LIKE 'Z%'
  ORDER BY UNICODE_CI;
    -- Time: 0ms                          <<<<<<< OK >>>>>>>
    -- Result nothing

  SELECT UNICODE_CS_CZ FROM TEST1M
  WHERE UNICODE_CS_CZ LIKE 'Z%'
  ORDER BY UNICODE_CS_CZ;
    -- Time: 4,247s                       <<<<<<< collate case SENSITIVE HEAR IS PERFOMANCE PROBLEM FOR VALUE "Z" 4,247s WHY? VALUE "Y" IS OK 0ms >>>>>>>
    -- Result nothing

  SELECT UNICODE_CS_CZ FROM TEST1M
  WHERE UNICODE_CS_CZ LIKE 'Y%'
  ORDER BY UNICODE_CS_CZ;
    -- Time: 0ms                          <<<<<<< FOR letter "Y" IS NOT PROBLEM :) WHY? >>>>>>>
    -- Result nothing
 
  SELECT UNICODE_CI_CZ FROM TEST1M        
  WHERE UNICODE_CI_CZ LIKE 'Z%'
  ORDER BY UNICODE_CI_CZ;
    -- Time: 4,52s                        <<<<<<< HEAR IS PERFORMANCE PROBLEM WHY? >>>>>>>
    -- Result nothing

  SELECT UNICODE_CI_CZ FROM TEST1M
  WHERE UNICODE_CI_CZ LIKE 'Y%'
  ORDER BY UNICODE_CI_CZ;
    -- Time: 0ms                          <<<<<<< FOR letter "Y" IS NOT PROBLEM :) WHY? >>>>>>>
    -- Result nothing
```

### LINKED ISSUE #7094: Incorrect indexed lookup of strings when the last keys characters are part of collated contractions and there is multi-segment insensitive descending index
Test case:

```
create collation unicode_ci_cz for utf8 from unicode case insensitive 'LOCALE=cs_CZ';

create table t1 (
    c1 varchar(10) character set utf8 collate unicode_ci_cz,
    c2 varchar(10) character set utf8 collate unicode_ci_cz
);

insert into t1 (c1) values ('a');
insert into t1 (c1) values ('b');
insert into t1 (c1) values ('c');
insert into t1 (c1) values ('ch');
insert into t1 (c1) values ('d');
insert into t1 (c1) values ('e');
insert into t1 (c1) values ('f');
update t1 set c2 = c1;
commit;

create desc index t1_c1_c2_desc on t1 (c1, c2);


select * from t1 where c1 > 'c' order by c1, c2;

-- No data


select * from t1 where c1 > 'c' plan (t1 natural) order by c1, c2;

/*
C1         C2         
========== ========== 
d          d          
e          e          
f          f          
ch         ch         
*/
```

This relates to https://github.com/FirebirdSQL/firebird/issues/7093

### LINKED ISSUE #7140: Wrong select result in case of special sort character.
Originally discovered with Polish collation but same problem for Czech. Tested with FB3, FB4 and ICU4.2, ICU 5.2, ICU 6.3

To get wrong result you must meet all of the following conditions:
 - test column that has a sort assigned with a special sort for some characters
 - use the less "<" operator
 - use a character with a special order, eg in Czech it is an "c" alphabetical order a, b, c, č, d
 - combine it with statement "or value is null"

create collation test_cz for UTF8 FROM UNICODE CASE INSENSITIVE ACCENT SENSITIVE 'LOCALE=cs_CZ';
CREATE TABLE TBL_TEST 
   (	   C1 VARCHAR(50) collate test_cz
   );
CREATE INDEX IDX_c1 ON TBL_TEST  (C1);
CREATE DESCENDING INDEX IDX_c1_d ON TBL_TEST (C1);

INSERT INTO TBL_TEST (C1) VALUES ('aaa');
INSERT INTO TBL_TEST (C1) VALUES ('bbb');
INSERT INTO TBL_TEST (C1) VALUES ('ccc');
INSERT INTO TBL_TEST (C1) VALUES ('ddd');
INSERT INTO TBL_TEST (C1) VALUES (NULL);

select * from TBL_TEST where c1 < 'b' or c1 is null order by c1 desc
result:
1 aaa
2 [null]

select * from TBL_TEST where c1 < 'c' or c1 is null order by c1 desc
result:
1 [null]

select * from TBL_TEST where c1 < 'd' or c1 is null order by c1 desc
result:
1 ccc
2 bbb
3 aaa
4 [null]

asfernandes:


asfernandes:
The solution was tested and approved by customer.

Once it's reviewed and integrated in v3, I will frontport it to v4 and master branches.

asfernandes:
The changes is like discussed in fb-devel:
- Operators (`>`, `>=`, `<`, `<=`) are fixed to lookup for sort keys without remove contractions
- `STARTING WITH` generates multiple prefixes (not more than 3) starting keys

src/jrd/Optimizer.cpp
```diff
@@ -1003,26 +1006,35 @@ void OptimizerRetrieval::getInversionCandidates(InversionCandidateList* inversio
 				if (segment->scope == scope)
 					scratch.scopeCandidate = true;
 
-				if (segment->scanType != segmentScanMissing && !(scratch.idx->idx_flags & idx_unique))
+				const USHORT iType = scratch.idx->idx_rpt[j].idx_itype;
+
+				if (iType >= idx_first_intl_string)
 				{
 					const USHORT iType = scratch.idx->idx_rpt[j].idx_itype;
```
dyemanov, line 1013: I believe the second definition of `iType` must be removed.

src/jrd/Optimizer.cpp
```diff
@@ -1003,26 +1006,35 @@ void OptimizerRetrieval::getInversionCandidates(InversionCandidateList* inversio
...
 						}
 					}
+
+					if (segment->scanType == segmentScanStarting)
+					{
+						if (textType->getFlags() & TEXTTYPE_MULTI_STARTING_KEY)
+							scratch.useMultiStartingKeys = true;	// use INTL_KEY_MULTI_STARTING
+						else
+							scratch.usePartialKey = true;
+					}
```
dyemanov, line 1033: Maybe a better condition sequence would be:
```
if (segment->scanType == segmentScanStarting)
...
else if (segment->scanType != segmentScanMissing && !(scratch.idx->idx_flags & idx_unique))
...
```
?

Currently I see that `scratch.usePartialKey` may be assigned twice.

src/jrd/Optimizer.cpp
```diff
@@ -1003,26 +1006,35 @@ void OptimizerRetrieval::getInversionCandidates(InversionCandidateList* inversio
...
 				}
 
 				// Check if this is the last usable segment
-				if (!scratch.fuzzy &&
+				if (!scratch.usePartialKey &&
```
dyemanov, line 1037: I believe this check should be extended:
`if (!scratch.usePartialKey && !scratch.useMultiStartingKeys &&`
as I don't see how for index {A, B} segment B can be used if segment A is looked up using irb_multi_starting.

src/jrd/btr.cpp
```diff
@@ -679,152 +682,161 @@ void BTR_evaluate(thread_db* tdbb, const IndexRetrieval* retrieval, RecordBitmap
...
-			// Check if pointer is still valid
-			if (pointer > endPointer) {
-				BUGCHECK(204);	// msg 204 index inconsistent
 			}
 		}
-	}
 
-	CCH_RELEASE(tdbb, &window);
+		CCH_RELEASE(tdbb, &window);
+	} while ((lower = lower->key_next.get()) && (upper = upper->key_next.get()));
```
dyemanov, line 839: I'm not sure putting the whole body into the loop is a good thing. First of all, the parts with `retrieval->irb_lower_count == 0` and `retrieval->irb_upper_count == 0` will never be executed more than once, as `irb_multi_starting` implies that both bounds are defined. Second, it's worth testing that locating the next lower bound via `BTR_find_page()` (i.e. from the tree root) is really faster than iterating through `find_node_start_point()` (i.e. using jump nodes and sibling pointers) (*). I expect contractions to have a common prefix (e.g. 'C' and 'CH'), so they should be located quite close to each other on leaf pages.

I don't request these changes, just want you to consider them, at least for master.

(*) I mean something like this:

```c++
if (retrieval->irb_upper_count)
{
  while (true)
  {
    while (scan(tdbb, pointer, bitmap, bitmap_and, &idx, retrieval, prefix, upper,
                       skipLowerKey, *lower))
    {
      page = (btree_page*) CCH_HANDOFF(tdbb, &window, page->btr_sibling, LCK_read, pag_index);
      pointer = page->btr_nodes + page->btr_jump_size;
      prefix = 0;
    }

    if ((!(lower = lower->key_next.get()) && (upper = upper->key_next.get())))
      break;

    while (!(pointer = find_node_start_point(page, lower, 0, &prefix,
              idx.idx_flags & idx_descending, (retrieval->irb_generic & (irb_starting | irb_partial)))))
    {
      page = (btree_page*) CCH_HANDOFF(tdbb, &window, page->btr_sibling, LCK_read, pag_index);
    }

    prefix = IndexNode::computePrefix(upper->key_data, upper->key_length,
                                                            lower->key_data, lower->key_length);
  }
}
```

src/jrd/btr.cpp
```diff
@@ -1500,13 +1519,11 @@ idx_e BTR_make_key(thread_db* tdbb,
 	key->key_flags = 0;
 	key->key_nulls = 0;
 
+	const bool fuzzy = keyType == INTL_KEY_PARTIAL || keyType == INTL_KEY_MULTI_STARTING;
```
dyemanov, line 1522: Parenthesis around the boolean for better readability, please.

src/jrd/btr.cpp
```diff
@@ -2433,78 +2455,164 @@ static void compress(thread_db* tdbb,
...
+				{
+					DSC to;
+
+					// convert to an international byte array
+					to.dsc_dtype = dtype_text;
+					to.dsc_flags = 0;
+					to.dsc_sub_type = 0;
+					to.dsc_scale = 0;
+					to.dsc_ttype() = ttype_sort_key;
+					to.dsc_length = MAX_KEY * 4;
```
dyemanov, line 2509: In FB4, `MAX_KEY * 4 == 32768` while `MAX_COLUMN_SIZE == 32767`. Are you sure `dsc_length > MAX_COLUMN_SIZE` is safe? At least `MOV_make_string2` could throw as a result.

src/jrd/recsrc/IndexTableScan.cpp
```diff
@@ -492,33 +547,32 @@ UCHAR* IndexTableScan::getPosition(thread_db* tdbb, Impure* impure, win* window)
...
 		// be false equality hits.
-		impure->irsb_nav_upper_length = MIN(m_length + 1, upper.key_length);
-		memcpy(impure->irsb_nav_data + m_length, upper.key_data, impure->irsb_nav_upper_length);
+		impure->irsb_nav_upper_length = MIN(m_length + 1, upper->key_length);
+		memcpy(impure->irsb_nav_data + m_length, upper->key_data, impure->irsb_nav_upper_length);
 	}
 
 	if (retrieval->irb_lower_count)
-		limit_ptr = &lower;
+		limit_ptr = lower;
```
dyemanov, line 576: Similarly to btr.cpp, jumping through keys via BTR_find_leaf() would make the patch against IndexScanNode.cpp 10x smaller and supposedly not slower. Something like this in `getRecord()`:

```c++
if (retrieval->irb_upper_count &&
  compareKeys(idx, key.key_data, key.key_length, &upper, flags) > 0)
// here the new code starts
{
  if (! get new lower/upper key)
    break;

  key = upper = new key;

  while (!(nextPointer = BTR_find_leaf(page, &upper, impure->irsb_nav_data, NULL,
            (idx->idx_flags & idx_descending), (retrieval->irb_generic & (irb_starting | irb_partial)))))
  {
    page = (Ods::btree_page*) CCH_HANDOFF(tdbb, &window, page->btr_sibling, LCK_read, pag_index);
  }

  continue;
}
```

dyemanov (commented):
I've commented about the things that got my attention. Suggestions for btr.cpp / IndexScanNode.cpp are based on my own multi-key lookup work that's yet to be presented via PR. I cannot comment about the INTL part of your changes, as I don't feel confident enough in that area.

dyemanov:
And I should note that amount of "core" changes is quite big, this makes me worried about unexpected side effects. I'd feel safer if this is committed into master only, with the interested customer provided with custom v3 / v4 builds by IBPhoenix.

asfernandes:
> And I should note that amount of "core" changes is quite big, this makes me worried about unexpected side effects. I'd feel safer if this is committed into master only, with the interested customer provided with custom v3 / v4 builds by IBPhoenix.

@dyemanov @pcisar what if a special parameter is put in v3/4 with default value as disabled?

pcisar:
> And I should note that amount of "core" changes is quite big, this makes me worried about unexpected side effects. I'd feel safer if this is committed into master only, with the interested customer provided with custom v3 / v4 builds by IBPhoenix.

This change was developed in 3.0.8, it passed the standard project QA and was thoroughly tested by sponsor in their test lab and in real deployment (their customer). The same was done on custom build based on 3.0.9. We're highly confident that this patch works and does not break anything.

src/jrd/Optimizer.cpp
asfernandes: It may be assigned twice, but due to the internal `if's`, the only other way would be to use a single giant `if` with lots of condtions (`(x1 && y1 && z1) || (x2 && y2)`). I think it's going to be more confusing.

src/common/intlobj_new.h
```diff
@@ -57,9 +57,15 @@ typedef USHORT (*pfn_INTL_keylength) (texttype* tt, USHORT len);
 
 /* Types of the keys which may be returned by str2key routine */
 
-#define INTL_KEY_SORT    0 /* Full sort key */
-#define INTL_KEY_PARTIAL 1 /* Starting portion of sort key for equality class */
-#define INTL_KEY_UNIQUE  2 /* Full key for the equality class of the string */
+#define INTL_KEY_SORT           0 /* Full sort key */
+#define INTL_KEY_PARTIAL        1 /* Starting portion of sort key for equality class */
+#define INTL_KEY_UNIQUE         2 /* Full key for the equality class of the string */
+#define INTL_KEY_MULTI_STARTING 3 /* Multiple starting keys */
```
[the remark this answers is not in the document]
asfernandes: INTL_KEY_PARTIAL is part of public collation API. Did you want to rename it?

dyemanov:
@pcisar We already had cases when "thoroughly tested" and "QA passed" features caused regressions. Simply because one customer is never enough for a wide testing. That said, I'm not going to stop the progress, especially recalling how long it was discussed before being implemented. I'd just like to see my review questions answered before anything is committed. Another concern I have is mostly related to IBP, so I've sent it to Paul B. privately.

src/jrd/btr.cpp
asfernandes: > I expect contractions to have a common prefix (e.g. 'C' and 'CH'), so they should be located quite close to each other on leaf pages.

No. CH sorts after H while C sorts after B.

src/jrd/btr.cpp
```diff
@@ -1659,7 +1704,7 @@ void BTR_make_null_key(thread_db* tdbb, const index_desc* idx, temporary_key* ke
 	// If the index is a single segment index, don't sweat the compound stuff
 	if ((idx->idx_count == 1) || (idx->idx_flags & idx_expressn))
 	{
-		compress(tdbb, &null_desc, key, tail->idx_itype, true, descending, false);
+		compress(tdbb, &null_desc, key, tail->idx_itype, true, descending, INTL_KEY_SORT);
```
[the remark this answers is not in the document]
asfernandes: I treated it like we would do with an enum, so I explicitly passed the "default" constant and let the routine deal with the implementation details (not use keyType when isNull == true).

src/common/intlobj_new.h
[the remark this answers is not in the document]
dyemanov: While I don't think anyone will be broken, let's better avoid renaming something in the API ;-) As for `usePartialKey`, I don't have any preference whether its name should correspond to `INTL_KEY_PARTIAL` or to `irb_starting`. Both are surely acceptable. Thus no serious need to change anything, sorry.

src/jrd/btr.cpp
dyemanov: That's sad ;-)

src/jrd/btr.cpp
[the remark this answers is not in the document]
dyemanov: OK

src/common/intlobj_new.h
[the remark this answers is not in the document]
aafemt: Which part of API it is exactly? This header is not included in distribution packages and I cannot remember any sort of API function related to collations.

src/jrd/Optimizer.cpp
dyemanov: I seem to be wrong here, sorry. `usePartialKey` belongs to `segmentScanStarting` and should never appear for equalities.

src/jrd/Optimizer.cpp
```diff
@@ -1296,9 +1308,12 @@ InversionNode* OptimizerRetrieval::makeIndexScanNode(IndexScratch* indexScratch)
 			retrieval->irb_generic |= irb_exclude_upper;
 	}
 
-	if (indexScratch->fuzzy)
+	if (indexScratch->usePartialKey)
 		retrieval->irb_generic |= irb_starting;	// Flag the need to use INTL_KEY_PARTIAL in btr.
 
+	if (indexScratch->useMultiStartingKeys)
+		retrieval->irb_generic |= irb_multi_starting; // Flag the need to use INTL_KEY_MULTI_STARTING in btr.
+
```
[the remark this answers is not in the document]
asfernandes: It's not fully mutually exclusive now. Some places checking for irb_starting is also being used (and is necessary) when irb_multi_starting is set.

The code had already some confusion in relation of irb_partial and irb_starting.

I currently have not a good answer of what would be the best way (maintain it, verify irb_multi_starting in more places, always set irb_starting when irb_multi_starting is used).

src/jrd/btr.cpp
asfernandes: Could your approach have degraded performance due to this fact (distance of keys)?

src/jrd/Optimizer.cpp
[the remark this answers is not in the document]
dyemanov: I was just looking at this:

```c++
const USHORT keyType =
			(retrieval->irb_generic & irb_multi_starting) ? INTL_KEY_MULTI_STARTING :
			(retrieval->irb_generic & irb_starting) ? INTL_KEY_PARTIAL :
			(retrieval->irb_desc.idx_flags & idx_unique) ? INTL_KEY_UNIQUE :
			INTL_KEY_SORT;
```
where `irb_multi_starting` has a preference over `irb_starting`, but I agree that `irb_starting` may be checked in other places independently. So far I think that `irb_multi_starting` should imply `irb_starting` (your last option), but it should be better investigated and I'm not going to insist on that for v3/v4 anyway.

src/jrd/btr.cpp
dyemanov: Yes, but it's mostly targeted at many keys being scanned at once, as the more keys are present the more a range scan (via sibling pointers) wins over multiple lookups from the tree root. For the case of few keys I just fallback to the old algorithm (independent lookups OR'ed into a single bitmap), as I deal with equalities only. I believe I will be able to utilize your solution for this case.

src/jrd/recsrc/IndexTableScan.cpp
asfernandes: I suppose this and the find_node_start_point comments should not be changed then due to probable distance of keys, correct?

dyemanov: Yes, my suggestion is not going to suit you well.

asfernandes:
Is there anything pending from me that I could do to make possible the merge of this?

dyemanov:
I see no issues that remain unresolved, so go ahead merging.

Issue #7147, opened 2022-03-07
Problems with use of big timeout (or no timeout at all) in trace service


AlexPeshkoff:
Was originally reported by Pavel Zotov privately.

Bug was found when working with new python testing tool. That tool has no timeout when using trace service.
When working with CS python tool hangs on exit to OS and a number of instances of CS remain running on the server.

It was found that all processes hang in same place - they wait indefinitely for more trace data, and there is no way to stop them. Bug has gone unnoticed when using fbsvcmgr / fbtracemgr cause both have 1 sec timeout. That <1 sec delay when closing utility appears to be not a problem for interactive user and was successfully ignored by all test suites.

It's trivial to reproduce without python slightly modifying out own utility. It was noticed that client also has serious problem - there is no way to interrupt Service::query(). fb_shutdown() also does not help.

AlexPeshkoff:
Applied minimalist fix to FB3.

Noremos:
> Applied minimalist fix to FB3.

Hello Alex.
In classic mode, there seems to be a new issue after your fix. The following messages have begun to appear in my firebird.log:
"Shutting down the server with 1 active connection(s) to 1 database(s), 0 active service(s)"

In the case of the OO API, this happens when not all attachments are detached.
I did a little investigation. Putting sleep(3000ms) before the connection counting function makes the message disappear. looks like a race condition.
I have a simple example to reproduce this.

```
#include <iostream>

#include <stdio.h>
#include <vector>
#include <unistd.h>

#include <ibase.h>
#include <firebird/Interface.h>

using namespace Firebird;
using namespace std;

static IMaster* master = fb_get_master_interface();
static IProvider* prov = master->getDispatcher();

void create_database()
{
	IStatus* st = master->getStatus();
	IUtil* utl = master->getUtilInterface();
	IAttachment* att = NULL;
	IXpbBuilder* dpb = NULL;

	try
	{
		ThrowStatusWrapper status(st);
		dpb = utl->getXpbBuilder(&status, IXpbBuilder::DPB, NULL, 0);
		dpb->insertInt(&status, isc_dpb_page_size, 4 * 1024);

		// create empty database
		att = prov->createDatabase(&status, "/tmp/fbtests.fdb", dpb->getBufferLength(&status),
								   dpb->getBuffer(&status));
		printf("Database /tmp/fbtests.fdb created\n");

		// detach from database
		att->detach(&status);
		att = NULL;
	}
	catch (const FbException& error)
[...]

PR #7148, opened 2022-03-07
Fix for #7147: Problems with use of big timeout (or no timeout at all) in trace service

AlexPeshkoff:


hvlad (commented):
Does you consider to use already existing PortsCleanup ?

src/remote/client/interface.cpp
```diff
@@ -995,6 +998,18 @@ class RProvider : public StdPlugin<IProviderImpl<RProvider, CheckStatusWrapper>
 void RProvider::shutdown(CheckStatusWrapper* status, unsigned int /*timeout*/, const int /*reason*/)
 {
 	status->init();
+
+	// make OS abort all outgoing connections
+	MutexLockGuard g(outPortsMutex, FB_FUNCTION);
+	for (auto p = outPorts->begin(); p != outPorts->end(); ++p)
```
asfernandes, line 1004: Why not `for (auto port : outPorts)`?

AlexPeshkoff: It does not work with GlobalPtr<> variable.

hvlad:
@AlexPeshkoff : could you explain what exact bug you going to fix and how ? I'm not sure I got it reading #7147 description.

AlexPeshkoff:
The problem may happen with any service, querying server w/o timeout, as long as server itself never finishes connection with client. Most of our services do not behave this way - as soon as requested job is complete they return no data in Service::query() and finish this API call. But trace service behaves in another manner - when there is no data to return to the client it waits for timeout expiration or indefinitely when no timeout. I.e. in order to break connection with server client must be able to at least disconnect from it gracefully. The problem is that remote provider does nothing on an attempt to fb_shutdown() it. Therefore application which started Service::query() w/o timeout can't close connection calling fb_shutdown() and after using it waits for completion of Service::query() call. As long as server has enough data to return buffer full of data.

When working with trace service we were almost always using fbtracemgr & fbsvcmgr. Both tools use 1 second timeout when getting data from service. That was hiding mention effects for a long time. Looks like users were also always using timeout - may be looking at our code in fb*mgrs...

AlexPeshkoff:
@hvlad Cancel happened to be really simple task - most of time took API modification. What do you think - do we want to support it remotely? Currently there is a stub in interface.cpp.

hvlad:
> @hvlad Cancel happened to be really simple task - most of time took API modification. What do you think - do we want to support it remotely? Currently there is a stub in interface.cpp.

It is good to have, but not required for this ticket, imho.
If it could be implemented - why not ?

PR #7153, opened 2022-03-16
Fixes for #7042: ON DISCONNECT triggers are not executed during forced attachment shutdown

ilya071294:


AlexPeshkoff (commented):
Would it make sense to make timeout for ON DISCONNECT triggers related with -FORCE parameter of gifx? If one specifies '-force 0' but due to problems in trigger has to wait 180 sec (default timeout) that appears not good.

ilya071294:
> Would it make sense to make timeout for ON DISCONNECT triggers related with -FORCE parameter of gifx? If one specifies '-force 0' but due to problems in trigger has to wait 180 sec (default timeout) that appears not good.

I think the timeout should be defined only by the config parameter to get a more predictable and strict behavior. '-force 1' may give strange results. For example, some attachments will complete the trigger in 1 second, and some of them won't.
I found another problem. When I specify '-tran 0', ON DISCONNECT triggers are not able to run because they need a transaction, and new transactions are not allowed (DBB_shut_tran flag is set). But when I specify '-force 0', new transactions are allowed. It seems not right. We can either allow transactions for ON DISCONNECT triggers, or do not execute triggers at all during DB shutdown. Any opinions?

AlexPeshkoff:
OK, let's leave timeouts as is for now.

What about DB shutdown - I think we should not run ON DISCONNECT triggers only when engine is shutting down cause in this case we must complete our job as soon as possible. Or kernel may just kill our process with -9. In all other cases ON DISCONNECT triggers should be executed. And yes - starting transaction(s) for them should be enabled. BTW, if autonomous transaction feature is used inside it should be startable too.

src/jrd/exe.cpp
```diff
@@ -1200,7 +1200,19 @@ void EXE_execute_triggers(thread_db* tdbb,
...
+				if (trigger_action == TRIGGER_DISCONNECT)
+				{
+					if (!trigger->req_timer)
+						trigger->req_timer = FB_NEW_POOL(*tdbb->getAttachment()->att_pool) TimeoutTimer();
+
+					const unsigned int timeOut = tdbb->getDatabase()->dbb_config->getOnDisconnectTrigTimeout() * 1000;
+					trigger->req_timer->setup(timeOut, isc_cfg_stmt_timeout);
+					trigger->req_timer->start();
+					thread_db::TimerGuard timerGuard(tdbb, trigger->req_timer, true);
+					EXE_start(tdbb, trigger, transaction);
```
asfernandes, line 1212: A comment here would be good, like `// under timerGuard scope` as the line is identical to the "else" case and some wrong refactor may put both outside of the "if".

ilya071294:
When I run 'gfix -shut full -force 30', I am still able to create attachments and start new transactions during 30 seconds. Is this the correct behavior?

AlexPeshkoff:
> When I run 'gfix -shut full -force 30', I am still able to create attachments and start new transactions during 30 seconds. Is this the correct behavior?

Never used to work with something except '-force 0', therefore hard to answer how this should be. Looking at the code - yes, it's normal behavior. It may be also useful to check how does something true old (like 1.5) behaves in that case.

dyemanov:
From the InterBase Operations Guide:
```
FORCE SHUTDOWN AFTER TIMEOUT
<...> This option does not deny new transactions or connections during the timeout period. <...>
```

ilya071294:
> Never used to work with something except '-force 0', therefore hard to answer how this should be. Looking at the code - yes, it's normal behavior. It may be also useful to check how does something true old (like 1.5) behaves in that case.

Anyway, I think this problem is not related with this PR, and a new issue should be created for it.
Is the PR okay now?

AlexPeshkoff (approved):

AlexPeshkoff:
We need the same for master branch

ilya071294:
> We need the same for master branch

Done. Should I port at least some commits to FB3?

AlexPeshkoff:
Not sure. Usually we backport to old branch only bugfixes. What particular commits do you want to backport?

ilya071294:
> What particular commits do you want to backport?

If we are not going to change ON DISCONNECT triggers in FB3, then only https://github.com/FirebirdSQL/firebird/pull/7153/commits/a86592803a3f32c3cf0e10b9d870e32062f5b88c.

AlexPeshkoff:
> > What particular commits do you want to backport?
> 
> If we are not going to change ON DISCONNECT triggers in FB3, then only [a865928](https://github.com/FirebirdSQL/firebird/commit/a86592803a3f32c3cf0e10b9d870e32062f5b88c).

@ilya071294 Please prepare pull request

PR #7187, opened 2022-05-19
Fix for #7178: Add PUBLIC user to sql script condition which selects privileges from granted user roles.

### LINKED ISSUE #7178: DEFAULTed grants to PUBLIC must act as DEFAULTed to every user
A role can be default granted to PUBLIC.

create role r1;
grant default r1 to public;

In this case I think the role should be act as a defaulted role to every user.

But it's not working.

The role privileges only works if user uses it explicitly.

AlexeyMochalov:
Fixes #7178: Add PUBLIC user to sql script condition which selects privileges from granted user roles.

### DIFF
--- a/src/jrd/scl.epp
+++ b/src/jrd/scl.epp
@@ -1065,7 +1065,9 @@ void UserId::findGrantedRoles(thread_db* tdbb) const
 		string usr_get_priv;
 		sql << "with recursive role_tree as ( "
 			<< "   select rdb$relation_name as nm, 0 as ur from rdb$user_privileges "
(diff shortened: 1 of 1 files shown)


AlexeyMochalov:
Thanks for your reply. It's possible I misunderstood documentation.
It says:
"Unless the role is granted as a default role, a user that is granted a role must explicitly specify that role, either with their login credentials or activating it using SET ROLE, in order to exercise the associated privileges. **Any other privileges granted to the user or received through default roles are not affected by explicitly specifying a role**."
in https://firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref40/fblangref40-security-granting.html
Could you please clarify?
If user set role explicitly it should not receive any privileges except from specified role? Is that right?

asfernandes:
You seem to be correctly, sorry.

PR #7201, opened 2022-06-07
Corrected determining of relation's security class for checking acces…

dmitry-starodubov:
…s rights to BLOB.


AlexPeshkoff:
Please same for HEAD!

PR #7213, opened 2022-06-17
Use Windows private namespace for kernel objects used in server-to-server IPC.

hvlad:
It allows to synchronize processes across different windows sessions.

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -617,7 +617,7 @@ private:
 
 	void init()
 	{
-		char sid[SECURITY_MAX_SID_SIZE];
+		alignas(SID) char sid[SECURITY_MAX_SID_SIZE];
 		DWORD cbSid = sizeof(sid);
 
 		// For now use EVERYONE, could be changed later
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -625,7 +625,7 @@ private:
 		if (!CreateWellKnownSid(WinWorldSid, NULL, &sid, &cbSid))
 			raiseError("CreateWellKnownSid");
 
-		// Create security desctriptor which allows generic access to the just created SID
+		// Create security descriptor which allows generic access to the just created SID
 
 		SECURITY_ATTRIBUTES sa;
 		RtlSecureZeroMemory(&sa, sizeof(sa));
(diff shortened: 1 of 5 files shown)


src/common/utils.cpp
```diff
@@ -555,6 +558,147 @@ bool isGlobalKernelPrefix()
...
+	{
+		char sid[SECURITY_MAX_SID_SIZE];
+		DWORD cbSid = sizeof(sid);
+
+		// For now use EVERYONE, could be changed later
+		cbSid = sizeof(sid);
+		if (!CreateWellKnownSid(WinWorldSid, NULL, &sid, &cbSid))
+			raiseError("CreateWellKnownSid");
+
+		// Create security desctriptor which allows generic access to the just created SID
```
asfernandes, line 628: typo here: desctriptor -> descriptor

src/common/utils.cpp
```diff
@@ -555,6 +558,147 @@ bool isGlobalKernelPrefix()
...
+	const char* sBoundaryName = "FirebirdCommonBoundary";
+
+	void raiseError(const char* apiRoutine)
+	{
+		(Firebird::Arg::Gds(isc_sys_request) << apiRoutine << Firebird::Arg::OsError()).raise();
+	}
+
+	void init()
+	{
+		char sid[SECURITY_MAX_SID_SIZE];
```
asfernandes, line 620: Can this be in the stack in unaligned location? Maybe some alignas should be used?

Here https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-sid I understand that AllocateAndInitializeSid would be needed.

hvlad: Here used [CreateWellKnownSid()](https://docs.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-createwellknownsid) and I see no requirements for sid alignment.
Also, looking for [usage samples](https://cpp.hotexamples.com/examples/-/-/CreateWellKnownSid/cpp-createwellknownsid-function-examples.html) one can see that its uses either [LocalAlloc()](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-localalloc) or stack allocation. 
Also, I see no notes about alignment of allocated by LocalAlloc() memory.

Anyway, I'll add `alignas(SID)` to be sure.

PR #7221, opened 2022-06-28
Add json-related blr, constants, messages and DSC changes

Noremos:
This is an initial commit of the JSON functions implementation (see ISO/IEC TR 19075-6 SQL support for JavaScript Object Notation (JSON))


asfernandes:
I don't think partial changes should be integrated.

asfernandes:
It would also be good to have what and how this is going to be implemented before the code is implemented.

Noremos:
> I don't think partial changes should be integrated.

This PR was created because of blr. Reserved blr_josn will help avoid unnecessary conflicts in the future.


> It would also be good to have what and how this is going to be implemented before the code is implemented.

JSON_VALUE, JSON_QUERY, JSON_EXISTS, IS JSON, JSON_ARRAY, JSON_ARRAYAGG and JSON_OBJECTAGG have already been implemented. The full PR will be soon. Just need some time to cleanup the code.

dyemanov:
I see no problems reserving `blr_json` (or maybe it should better be named `blr_json_func`?), either explicitly or in the commented form. However, the rest of the changes should better be combined in a single pull request. As for `DSC_json`, I need more details on how it gonna be used before accepting it.

So far I'm willing to commit the BLR change myself and suggest the author to re-target this PR for Firebird v6 and continue adding the other JSON-related commits here.

Noremos:
> I see no problems reserving `blr_json` (or maybe it should better be named `blr_json_func`?), either explicitly or in the commented form. However, the rest of the changes should better be combined in a single pull request.

I agree with your remarks. As for the blr name, in best case there need 8 blr codes for 8 functions, but it's pretty greedy. So i desided to use a simple factory with sub codes (like agg nodes with blr_agg_function). So yes, it's better to use blr_json_func or blr_json_function.

About DSC_json... It needs to make FORMAT clause more clear and user friendly. The clauses uses to escape an input string (or don't escape if it is a JSON)

PR #7225, opened 2022-06-30
Fixed ORDER BY for big (>34 digits) int128 values with indexed access

AlexPeshkoff:
Adds new type of index for ods 13.1


src/common/Int128.h
```diff
@@ -303,6 +310,16 @@ class Int128 //: public Decimal128Base
 		return (UCHAR*)(&v); 
 	}
 
+	static const unsigned Bias = 128;
```
asfernandes, line 313: Didn't we use ALL_UPPER_CASE in most (*) class's static constants?

(*) Sure not all, but I didn't remember PascalCase being used.

PR #7244, opened 2022-07-28
Add support for FreeBSD/powerpc64le

pkubaj:


AlexPeshkoff:
Main question - was the build actually tested?

pkubaj:
Yes, I have started a Firebird server and created a database.

AlexPeshkoff:
Can you also try one more test?
Start server on PPC & from *intel / amd* box (i.e. x64) connect to it:

# ./isql -z -user sysdba -pas masterkey ppc64le:employee
Server version:
LI-T5.0.0.591-dev Firebird 5.0 Initial
LI-T5.0.0.591-dev Firebird 5.0 Initial/tcp (fbs3)/P18:C
<Here you should get ppc64le server desription>
Database: localhost:employee, User: SYSDBA
SQL> set list;
SQL> select 1.5, cast(1.5 as decimal(30,1)), cast(1.5 as double 
precision), cast(1.5 as decimal(15,1)), cast(1.5 as decfloat) from 
rdb$database;
CONSTANT                        1.5
CAST 1.5
CAST                            1.500000000000000
CAST                            1.5
CAST 1.5
SQL>

I.e. all values should be (as expected) 1.5. If something goes wrong 
that means network layer (converting messages to/from network format) 
works wrong on PPC - bad byte order. I ask cause this platform was 
actually untested for a long time, and checking for correct byte order 
is generally useful thing for new port.

Alex.

pkubaj:
OK, I will test that, but I don't think anything will be wrong with byte order since powerpc64le is actually little-endian, same as amd64.

pkubaj:
I can confirm that it works as you wrote.

builds/posix/prefix.freebsd_powerpc64el
```diff
@@ -0,0 +1,31 @@
...
+
+COMMON_FLAGS=-DFREEBSD -DPPC64EL -pipe -MMD -fPIC -fsigned-char
+PROD_FLAGS=$(COMMON_FLAGS) -O3 -g -fno-builtin -Wno-deprecated
+DEV_FLAGS=$(COMMON_FLAGS) -ggdb -p -Wall -Wno-non-virtual-dtor
+
+# These files are generated incorrectly (e.g. array.epp => array.cpp)
+%/array.o %/blob.o %/alice_meta.o %/restore.o %/backup.o: COMMON_FLAGS += -Wno-narrowing
+%/OdsDetection.o %/dba.o: COMMON_FLAGS += -Wno-narrowing
+
+CXXFLAGS := $(CXXFLAGS) -std=c++11
```
AlexPeshkoff, line 31: What a reason to change default for master branch c++17 to c++11? I hardly understand how does it build at all cause in alloc.h we use c++14 specific std::enable_if_t, see https://en.cppreference.com/w/cpp/types/enable_if

AlexPeshkoff (changes requested):

builds/posix/prefix.freebsd_powerpc64el
pkubaj: Ah, I based this patch on the Firebird 4 that we have in our ports tree and some other prefix files used std=c++11.

AlexPeshkoff (approved):

PR #7251, opened 2022-08-08
Windows build (installer, dev experience, speed, GH actions) and GH refactor

asfernandes:
- Fix errors when paths has spaces and different (or localized) utilities versions

The zip/installer steps were not working correctly with localized tools or installed in paths with spaces.

- Improve Windows build process making it generating public files directly in the output directory, separated by debug/release configuration

The Windows development experience was always very below the Linux experience.
Public binaries were generated by VS/msbuild in `temp` directory and copied by `make_all.bat` to the `output` directory.
But others files (`.conf`) were only copied to `output`.
So to debug in VS one would need to manually copy files to `temp`.
It was also wasting time and disk space having files in both places.

- Create and save Windows installer in GitHub actions

We were lacking the Windows installer generation and upload.

- Upload zip generated in builds/install_images in GitHub actions

The zip kits uploaded to GH Actions were the result of the `output` directory.
With this change it gets the zip generated in `builds/install_images`

- Improve Windows build speed not building both release and debug external libraries

Another source of slow build, external libraries were being built for *both* release and debug build.

- Refactor GitHub Actions using independent jobs instead of multiple conditional steps

While working with GitHub Actions, the conditional (per OS / build type) steps in the same job matrix was refactored to use multiple jobs.


hvlad:
>     * Improve Windows build process making it generating public files directly in the output directory
> 
> 
> The Windows development experience was always very below the Linux experience. Public binaries were generated by VS/msbuild in `temp` directory and copied by `make_all.bat` to the `output` directory. But others files (`.conf`) were only copied to `output`. So to debug in VS one would need to manually copy files to `temp`. It was also wasting time and disk space having files in both places.

  AFAIU, there is no separate DEBUG and RELEASE output folders and to quickly switch between DEBUG and RELEASE one need to rebuild (relink, at least) necessary binaries ? In case of partial build (not all targets) it will create mess of DEBUG and RELEASE  binaries in output folder.

AlexPeshkoff:
On 8/8/22 09:23, Vlad Khorsun wrote:
>
>     |* Improve Windows build process making it generating public files
>     directly in the output directory |
>
>     The Windows development experience was always very below the Linux
>     experience. Public binaries were generated by VS/msbuild in |temp|
>     directory and copied by |make_all.bat| to the |output| directory.
>     But others files (|.conf|) were only copied to |output|. So to
>     debug in VS one would need to manually copy files to |temp|. It
>     was also wasting time and disk space having files in both places.
>
> AFAIU, there is no separate DEBUG and RELEASE output folders and to 
> quickly switch between DEBUG and RELEASE one need to rebuild (relink, 
> at least) necessary binaries ? In case of partial build (not all 
> targets) it will create mess of DEBUG and RELEASE binaries in output 
> folder.
>

Probably the best way to go is to have such folders in output? LIke 
posix does.
.conf-files may be also copied to both.

asfernandes:
> Probably the best way to go is to have such folders in output?

I'm doing it.

PR #7252, opened 2022-08-10
Do not allow the AST thread to release EX lock on a relation while it's in the deleting state

ilya071294:
The problem is rarely reproduced by bugs.core_2879 test. Without the fix the phase 4 of delete_relation DFW may fail because MET_lookup_relation_id cannot find a relation.

### DIFF
--- a/src/jrd/dfw.epp
+++ b/src/jrd/dfw.epp
@@ -4955,6 +4955,7 @@ static bool delete_relation(thread_db* tdbb, SSHORT phase, DeferredWork* work, j
 	case 4:
 		relation = MET_lookup_relation_id(tdbb, work->dfw_id, true);
 		if (!relation) {
+			fb_assert(false);
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


ilya071294:
I also have a question about jrd_rel::rel_drop_mutex. Looks like it does not make sense anymore since relation vector was moved from dbb to att. Did I miss something?

dyemanov (approved):

dyemanov:
> I also have a question about jrd_rel::rel_drop_mutex. Looks like it does not make sense anymore since relation vector was moved from dbb to att. Did I miss something?

Implementation of the shared metadata cache is in progress, so the mutex could become useful again.

dyemanov:
> Should I prepare the PR for FB3?

Not necessary, I'll cherry-pick your commit to another branches myself.

ilya071294:
> I'll cherry-pick your commit to another branches myself.

OK, it would be good.

Issue #3750, opened 2011-03-14
Partial Indexes [CORE3384]


firebird-issue-importer:
Submitted by: Vladimir Arkhipov (arkinform)

Votes: 4

Example:

CREATE UNIQUE INDEX orders\_number\_idx ON orders \(number\)
    WHERE number is not null

Description from <http://www.postgresql.org/docs/8.0/static/indexes-partial.html>

"A partial index is an index built over a subset of a table; the subset is defined by a conditional expression \(called the predicate of the partial index\)\. The index contains entries for only those table rows that satisfy the predicate\."

"A major motivation for partial indexes is to avoid indexing common values\. Since a query searching for a common value \(one that accounts for more than a few percent of all the table rows\) will not use the index anyway, there is no point in keeping those rows in the index at all\. This reduces the size of the index, which will speed up queries that do use the index\. It will also speed up many table update operations because the index does not need to be updated in all cases\."

Very very useful \!\!\!

justin-fyfe:
+1 if this feature can be done. Not for performance but for unique enforcement. For example, consider the table:

```
CREATE TABLE FOO_TBL (
     ID    UUID NOT NULL,
     USER_ID UUID NOT NULL,
     UNIQUE_VALUE VARCHAR(256) NOT NULL, -- THE UNIQUE VALUE
     START_TS TIMESTAMP NOT NULL, -- THE TIME THE VALUE BECAME ACTIVE
     STOP_TS TIMESTAMP, -- THE TIME THAT THE VALUE WAS NO LONGER ACTIVE
     CONSTRAINT PK_FOO PRIMARY KEY (ID),
     CONSTRAINT FK_FOO_USER FOREIGN KEY (USER_ID) REFERENCES USER_TBL(USER_ID)
);
```

Basically a table which needs to track changes over time of a value (can be anything like an medical record number, etc.) but basically the business requirement is: **value must be unique if it is active**.

In PostgreSQL, Oracle, and even SQLite this is done via:

```
CREATE UNIQUE INDEX FOO_UQ_IDX ON FOO_TBL(UNIQUE_VALUE) WHERE (STOP_TS IS NULL);
```

i.e. Enforce the unique constraint on `UNIQUE_VALUE` if it has not been indicated as no longer valid.

[...]

PR #7257, opened 2022-08-12
Support for partial indices

dyemanov:
As requested in #3750. Also supported by MS SQL, PostgreSQL, SQLite, MongoDB. Brief description can be found in [Wikipedia](https://en.wikipedia.org/wiki/Partial_index).


livius2:
Thank you for implementing this. Next step partial `Foreign Key` like in my comment in https://github.com/FirebirdSQL/firebird/issues/3750 :)

src/burp/backup.epp
```diff
@@ -1617,68 +1617,152 @@ void put_index( burp_rel* relation)
 	// requests--this requires more code but it is well worth it
 	// for the performance benefits, especially remotely--deej
 
-	FOR (REQUEST_HANDLE tdgbl->handles_put_index_req_handle1)
-		X IN RDB$INDICES WITH
-		X.RDB$RELATION_NAME EQ relation->rel_name
+	if (tdgbl->runtimeODS >= DB_VERSION_DDL13_1)
```
asfernandes, line 1620: Diff comparation is not easy here.

Can't we use unique code where non-existent fields are read as NULL?

Or are you preventing problem of new gbak being connected to very old engine that treat non-existent fields as error?

src/burp/restore.epp
```diff
@@ -6616,6 +6649,40 @@ bool get_index(BurpGlobals* tdgbl, const burp_rel* relation)
 		general_on_error ();
 	END_ERROR;
 
+	if (hasConditionSrc || hasConditionBlr)
```
asfernandes, line 6652: Same here. Write to non-existent fields is possible since many versions ago.

asfernandes:
I suppose that if v4 opens an ODS 13.1 database with partial index, it may not work properly, trying to find entries in index without considering its condition, correct?

Do we consider these cases as valid for minor ODS upgrade?

dyemanov:
> I suppose that if v4 opens an ODS 13.1 database with partial index, it may not work properly, trying to find entries in index without considering its condition, correct?
> 
> Do we consider these cases as valid for minor ODS upgrade?

v4 cannot attach an ODS 13.1 database.

src/burp/backup.epp
dyemanov: Yes, this is to please old engine versions.

pavel-zotov:
Following script causes FB crash (checked on 5.0.0.938; problem in line marked as "[ ! ]"):
```
set bail on;
shell del c:\temp\tmp4test.fdb 2>nul;
create database 'localhost:c:\temp\tmp4test.fdb';
show version;
set list on;
set echo on;
recreate table test(
    id int primary key
    ,f01 int
    ,f02 int
);
create index test_f01 on test computed by (f01)
where f01 = 1 --------------------------------- [ ! ]
;
```

dyemanov:
@pavel-zotov Should be fixed now.

pavel-zotov:
@@@ QA issue @@@
Currently this test contains only trivial cases for check.
More complex examples, including test of backup/restore, validation, using of misc datatypes (non-ascii, decfloat and int128), will be added later.

PR #7265, opened 2022-08-16
Make Windows build scripts reuse previous configuration

asfernandes:


hvlad (commented):
Looks ok. While I'm not sure why FB_CONFIG was replaced by FBBUILD_BUILDTYPE and why so long name ? We have all env vars prefixed with  FB_, thus FB_BUILD_TYPE looks better for me.

asfernandes:
We had both and I didn't know why.
As Dmitry cited FBBUILD_BUILDTYPE, I just maintained it.
I have nothing against rename it.

hvlad:
> We had both and I didn't know why. As Dmitry cited FBBUILD_BUILDTYPE, I just maintained it. I have nothing against rename it.

Aha, I found it - FBBUILD_BUILDTYPE was used in run_all.bat only (as all other FBBUILD_XXX vars).
run_all.bat was introduced especially to build installer.

I never use it by myself and not plan to use it. I never need it as developer.

So, please, return back FB_CONFIG to avoid confusion and don't use FBBUILD_BUILDTYPE outside of run_all.bat

reevespaul:
On Tue, 16 Aug 2022 05:15:54 -0700
Adriano dos Santos Fernandes ***@***.***> wrote:

> I see no good reason to use both names.
> Maybe @reevespaul can say something?
> 

I think the fundamental problem is that we have never developed a fully
integrated build environment. Vlad says (quite correctly from his point of
view) that he never uses run_all.bat and never will. That's normal. My
problem was that I needed a way to reliably package the windows builds, so I
wrote run_all.bat, and BuildExecutableInstall.bat (the packaging script) is
tightly integrated with it. Run_all.bat does work with/around the make_*.bat
files but from the start I did not want to touch them. The aim was always to
take the standard, well debugged output from the make_*.bat files and package
it.

Releases have always been done manually by me. But now that software
development everywhere is moving to automated CI/CD pipelines we need to
accept the reality that run_all.bat (or something similar) is the Boss! It
should still be possible to run individual batch files as part of the
code/compile/test/debug process but overall we should accept that one master
script can control the entire process such that fully working release
packages will come out just from executing run_all.bat.

When I reworked the build process recently I tried to centralise as much as
possible around setenvvars.bat. We had never really dealt with the fact that
not only do we need to support different versions of MSVC, we also need to
account for the fact that MS is releasing new minor versions of Visual Studio
every eight weeks or so and major new releases arrive every two or three
years. And we then need to add in the fact the github seems to be quite
prompt to deprecate pipelines based on O/S or compilers that are no longer
supported.

So we have two main objectives - create a build system that is responsive to
a constantly changing external environment while maintaining a usable
development environment. (We could minimise the changes in the external
environment if we had our own dedicated CI/CD pipelines instead of using
github actions/appveyor/whatever but that doesn't help much as we still need
to adapt to the changing environment.)

This means we need to clearly define the layers and not mix them up.
Originally, run_all.bat was written to define its own variables and reset
them each time it is run (in theory, anyway.) The make_*.bat files were
never dependant upon run_all.bat. I don't think that works any more.

I think we need to start from the point of view that there is a master
script, and that the make_*.bat files must also be able to be run
individually. Setenvvars.bat would be the logical place to co-ordinate
between the requirements of the end-to-end build/packaging process and the
development/build process itself.

I'll make time tomorrow to look into this a bit more deeply. When I studied
this a few weeks ago I was worried about some of the environment variable
naming. In the end I decided to do nothing ( I was working on v4.0.2 at the
time). I think the names of environment vars could do with a lot of cleaning
up but at the same time there is a high risk of breaking things. If we are
going to do this the pre-Alpha release is the time to do it.


Paul
-- 

Paul Reeves
http://www.ibphoenix.com
Supporting users of Firebird

asfernandes:
Currently (with this PR) all .bat files can be called individually and with an optional configuration (RELEASE / DEBUG) argument.

When a batch is called for the first time (setenvvars.bat not called before), the default configuration is RELEASE.

If setenvvars.bat was already been called, then the default configuration is the one set before (currently set in FBBUILD_BUILDTYPE).

All scripts can be called intermixed, including the one to build the installer.

```
:: build release
run_all.bat

:: preprocess release
preprocess.bat

:: build debug
run_all.bat DEBUG

:: preprocess debug
preprocess.bat

:: boot build release
make_boot.bat RELEASE

:: make ICU debug
make_icu.bat DEBUG

:: boot build debug
make_boot.bat
```

This seems to me to cover CI and dev/debug workflows.

hvlad:
Just avoid usage of env vars from "installer" scripts at "compile\build" scripts. 
I.e. return back usage of FB_CONFIG at make_XXX that was replaced by FBBUILD_BUILDTYPE

asfernandes:
The installer, after your request to have Debug/Release build in different directories, now uses the configuration to also know the path of the files.

I see no good reason to use two environment variables for same thing, one for the build and another for the installer. It makes no sense.

The installer script now acts as all others scripts, accepting the configuration as parameter, default to RELEASE on first run and default to last used in subsequent calls.

Have a unified handling of this is certainly better than have different procedures for different scripts.

dyemanov (commented):
While I don't like two macros serving the same goal, I think Paul's point about layering/isolation makes sense. If he created a bunch of variables (including FBBUILD_BUILDTYPE) for the "master" script only (mainly intended for the installer) exactly with the purpose to keep make_* scripts independent, then we should use the same approach **in this PR**, i.e. revert back to FB_CONFIG inside the make_* scripts. Then the run_all stuff should be revisited (by Paul or someone else) and maybe better integrated with the build scripts, including macro unification (all macros, not only FBBUILD_BUILDTYPE!). I don't see much value in fixing/improving FBBUILD_BUILDTYPE now and leaving everything else as is.

dyemanov (approved):

PR #7274, opened 2022-08-19
Fixed  #7255: READ COMMITTED READ CONSISTENCY mode is broken in Classic / SuperClassic on Linux

### LINKED ISSUE #7255: READ COMMITTED READ CONSISTENCY mode is broken in Classic / SuperClassic on Linux (the newly created users are not visible to execute statement)
The script below works in 3.0 (all architectures) and in 4+ (Supersever only). But 4+ Classic and Superclassic on Linux fail with the error "Your user name and password are not defined. Ask your database administrator to set up a Firebird login.".
Although it will work fine if we wait 10 seconds before execute block.
```
create user test password 'test';
commit;

set term ^;
create or alter procedure sp_main(a_usr varchar(31), a_pwd varchar(31))
  returns(x int) as
begin
  for
          execute statement
          'select 1 from rdb$database'
          as user a_usr password a_pwd into x
  do
          suspend;
end
^
set term ;^
commit;

grant execute on procedure sp_main to test;
commit;

-- wait 10 seconds and it will work

set term ^;
execute block returns(x int) as
begin
  for execute statement ('select * from sp_main(:u, :p)') ( u := 'test', p := 'test' )
  into x
  do suspend;
end
^
set term ;^
commit;

drop user test;
commit;
```

AlexPeshkoff:
TPC-cache related shared files are always (no matter are there other DBBs using that file or not) deleted by TipCache::finalizeTpc(). That works fine on windows cause 'safe delete' technique is used in class SharedMemory. That is not a problem for SS cause there is always single DBB in it. But in Classic / SuperClassic non-windows it's severe problem breaking FB operation when READ CONSISTENCY is used in READ COMMITTED transactions (default mode for that transactions).

Sample in #7255 with newly created user is bright illustration for the bug.

I've decided to make deletion of shared files on linux safe (unlink only when last descriptor is closed), suppose it will be useful in other places where shared memory is used.

### DIFF
--- a/src/common/isc_sync.cpp
+++ b/src/common/isc_sync.cpp
@@ -286,7 +286,7 @@ typedef GenericMap<Pair<NonPooled<DevNode, Firebird::CountedFd*> > > FdNodes;
 static GlobalPtr<Mutex> fdNodesMutex;
 static GlobalPtr<FdNodes> fdNodes;
 
-FileLock::FileLock(const char* fileName, InitFunction* init)
+FileLock::FileLock(const char* fileName)
 	: level(LCK_NONE), oFile(NULL),
 #ifdef USE_FCNTL
 	  lStart(0),
@@ -313,11 +313,6 @@ FileLock::FileLock(const char* fileName, InitFunction* init)
 		CountedFd** put = fdNodes->put(getNode(fd));
 		fb_assert(put);
 		*put = oFile;
-
-		if (init)
-		{
-			init(fd);
-		}
 	}
 
 	rwcl = getRw();
@@ -1168,14 +1163,14 @@ void SharedMemoryBase::removeMapFile()
 
 	if (!sh_mem_header->isDeleted())
 	{
-#ifndef WIN_NT
-		unlinkFile();
-#else
+#ifdef WIN_NT
 		fb_assert(!sh_mem_unlink);
 		sh_mem_unlink = true;
-#endif // WIN_NT
 
 		sh_mem_header->markAsDeleted();
+#else
+		unlinkFile();
+#endif // WIN_NT
 	}
 }
 
@@ -1200,7 +1195,26 @@ void SharedMemoryBase::unlinkFile()
 	if (hFile != INVALID_HANDLE_VALUE)
 		CloseHandle(hFile);
 #else
-	unlink(expanded_filename);
+	if (sh_mem_unlink_called)
+		return;
+
+	// time to release count of attaches to file
+	FileLockHolder finiLock(initFile);
+
+	if (sh_mem_unlink_called)
+		return;
+
+	if (sh_mem_header->mhb_flags < 2)
+		fatal_exception::raiseFmt("Lost connect to shared memory file %s", expanded_filename);
+
+	sh_mem_header->mhb_flags -= 2;
+	if (sh_mem_header->mhb_flags < 2)
+	{
+		sh_mem_header->markAsDeleted();
+		unlink(expanded_filename);
+	}
+
+	sh_mem_unlink_called = true;
 #endif // WIN_NT
 }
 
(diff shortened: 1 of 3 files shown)


src/jrd/tpc.cpp
```diff
@@ -130,11 +130,11 @@ void TipCache::finalizeTpc(thread_db* tdbb)
...
 	// be called during database shutdown when AST delivery is already disabled
 
-	// wait for all initializing processes (PR)
+	// wait for other processes
 	Lock lock(tdbb, 0, LCK_tpc_init);
 
-	if (!LCK_lock(tdbb, &lock, LCK_SW, LCK_WAIT))
-		ERR_bugcheck_msg("Unable to obtain TPC lock (SW)");
+	if (!LCK_lock(tdbb, &lock, LCK_EX, LCK_WAIT))
+		ERR_bugcheck_msg("Unable to obtain TPC lock (EX)");
```
hvlad, line 138: Are we need serialization here ? Any better ways ?
Same in initializeTpc() few lines below

src/common/isc_sync.cpp
```diff
@@ -1460,7 +1474,12 @@ SharedMemoryBase::SharedMemoryBase(const TEXT* filename, ULONG length, IpcObject
 		}
 	}
 
+	// time to count our attach to file
+	const USHORT UPPER_BOUND = MAX_USHORT - 1;
+	if (sh_mem_header->mhb_flags >= UPPER_BOUND)
+		(Arg::Gds(isc_random) << "Too many connects to single shared memory file").raise();
```
hvlad, line 1480: Taking into account that mhb_flags is USHORT and incremented by 2, I'm not sure this code could prevent overflow.
Last valid value is 65534 or MAX_USHORT - 2, for this value condition `mhb_flags >= UPPER_BOUND` is `false` and at next line we will assign 0 to the mhb_flags. Am I wrong ?

src/jrd/tpc.cpp
AlexPeshkoff: Probably it can be replaced with atomic access to mhb_flags. But before doing that - are we sure such serialization is a problem?

src/common/isc_sync.cpp
AlexPeshkoff: You are wrong. 
MAX_USHORT is 65535 (0xFFFF), not 65536. I.e. MAX_USHORT - 1 is 65534 and I see no problems this place.

hvlad: Yes, indeed.

src/jrd/tpc.cpp
hvlad: I think - yes, it will be a problem when attach\dettach calls rate is high. 
Probably we need some perf testing.

hvlad:
> I've decided to make deletion of shared files on linux safe (unlink only when last descriptor is closed), suppose it will be useful in other places where shared memory is used.

BTW, what about processes that was terminated forcibly, without releasing own mappings ?

AlexPeshkoff:
On 8/19/22 20:17, Vlad Khorsun wrote:
>
>     I've decided to make deletion of shared files on linux safe
>     (unlink only when last descriptor is closed), suppose it will be
>     useful in other places where shared memory is used.
>
> BTW, what about processes that was terminated forcibly, without 
> releasing own mappings ?
>

For SS/SC nothing can be done - single process died. For CS current code 
can't help - though in theory something can be done.

Such support is present in Mapping.cpp. There is process'es table, and 
everyone scanning it (sush scan is done also when it's probably time to 
remove file) checks - is another process in a table alive? If not that 
process is deleted from the table, file is unmapped when last process is 
removed from the table. In theory we can have something like this in SM 
header in FB5. But we need a solution for variable size table.

dyemanov:
Out of curiosity, is there any reason why we cannot use reference counting inside the TPC cache like it's done for all other shmem objects?

PR #7277, opened 2022-08-23
Fix using LM locks, SharedMemory left almost untouched

AlexPeshkoff:


AlexPeshkoff:
Another fix for #7255

src/jrd/tpc.cpp
```diff
@@ -146,24 +149,38 @@ void TipCache::finalizeTpc(thread_db* tdbb)
...
+			if (nmSnap.hasData())
+				SharedMemoryBase::unlinkFile(nmSnap.c_str());
+			if (nmHdr.hasData())
+				SharedMemoryBase::unlinkFile(nmHdr.c_str());
+		}
+		else
+			tdbb->tdbb_status_vector->init();
+	}
+
+	LCK_release(tdbb, m_lock);
```
hvlad, line 183: If LCK_lock() in line 172 failed, here we'll get bugcheck in LM

hvlad: Actually, not a bugcheck (due to additional check in LCK), but anyway - this is not fully correct

src/jrd/tpc.cpp
```diff
@@ -375,14 +405,30 @@ TipCache::StatusBlockData::~StatusBlockData()
...
+	if (fName.hasData())
+	{
+		if (LCK_lock(tdbb, &existenceLock, LCK_EX, LCK_NO_WAIT))
+		{
+			SharedMemoryBase::unlinkFile(fName.c_str());
+			LCK_release(tdbb, &existenceLock);
+		}
+		else
+			tdbb->tdbb_status_vector->init();
+	}
```
hvlad, line 431: Here same optimization as in `TipCache::finalizeTpc()` could be applied - don't call `LCK_release` before `LCK_lock`.
I'm not insist on it, though.

PR #7286, opened 2022-08-31
Make Android build work with modern NDK

asfernandes:
Modern NDK does not recommend usage of `make_standalone_toolchain`.


asfernandes:
I'm now deriving NDK_TOOLCHAIN by NDK, when NDK_TOOLCHAIN does not exist.

I also fixed some bugs in x86 and x86_64 builds.

AlexPeshkoff (approved):

PR #7287, opened 2022-09-02
The ability to change the initial value of generator/sequence was added.

DmEye:
You need to write "ALTER GENERATOR gen_name START WITH some_integer;" if you want to change namly the initial value of generator/sequence.


asfernandes:
We have `create or alter sequence <name> start with <n>` already acting very different than `alter sequence <name> start with <n>` that you are creating, i.e., when sequence exists is works like `alter sequence <name> restart with <n>`.

This is not feeling right. I think more discussion is needed for this command.

PR #7293, opened 2022-09-06
Create Android packages with all necessary files in all architectures (x86, x64, arm32, arm64)

asfernandes:
Add GitHub action to run second phase in MacOS machine, where Android emulator can run.

Currently this is done for ARM 32/64 builds.


AlexPeshkoff:
Am I right that w/o emulator build can not be completed?

asfernandes:
> Am I right that w/o emulator build can not be completed?

Without emulator first phase can be more or less like before, except that I removed stripped build there.

I added scripts with setup for emulator with local build, in this case both phases can run in Linux.

PR #7320, opened 2022-09-28
Fix wrong calculation of the header page end

ilya071294:


src/utilities/gstat/ppg.cpp
```diff
@@ -231,7 +231,7 @@ void PPG_print_header(const header_page* header, ULONG page,
 	TEXT temp[257];
 
 	const UCHAR* p = header->hdr_data;
-	for (const UCHAR* const end = p + header->hdr_page_size; p < end && *p != HDR_end; p += 2 + p[1])
+	for (const UCHAR* const end = (UCHAR*)header + header->hdr_page_size; p < end && *p != HDR_end; p += 2 + p[1])
```
dyemanov, line 234: Please either add a space after `(UCHAR*)` or use `reinterpret_cast<UCHAR*>(header)`. I'll merge this PR then.

PR #7339, opened 2022-10-08
Hash table based on lock-free list for page cache.

hvlad:
Replaced hash table guarded with the rw-lock by lock-free implementation.
It removes one of the main bottlenecks in the shared page cache.

The new hash table uses lock-free linked list and hazard pointers safe memory reclamation implemented in [CDS library](https://github.com/khizmax/libcds).

Many thanks to @romansimakov who intensively tested this code, found few bugs and added some comments and assertions.


src/jrd/cch.cpp
```diff
@@ -3539,164 +3571,194 @@ static bool expand_buffers(thread_db* tdbb, ULONG number)
...
+		bcb->bcb_flags |= BCB_free_pending;
+		if (!(bcb->bcb_flags & BCB_writer_active))
+			bcb->bcb_writer_sem.release();
+
+		if (walk)
 		{
-			//bdb->bdb_flags &= ~(BDB_faked | BDB_prefetch);
-			return lsOk;
+			bdb->release(tdbb, true);
+//			if (!--walk)
```
asfernandes, line 3706: Why maintain commented code (and using wrong indentation)?

src/jrd/cch.cpp
```diff
@@ -4316,60 +4232,74 @@ static ULONG memory_init(thread_db* tdbb, BufferControl* bcb, SLONG number)
...
 		if (memory + page_size > memory_end)
-			memory = 0;
+		{
+			const auto blk = bcb->bcb_bdbBlocks.end() - 1;
+			const BufferDesc* bdb = blk->m_bdbs;
+
+			if (!(bcb->bcb_flags & BCB_exclusive))
+			{
+				// first lock block is after last BufferDesc
+				fb_assert((char*)bdb->bdb_lock >= (char*)tail);
```
asfernandes, line 4290: Please add space after casts.

src/jrd/cch.cpp
```diff
@@ -5312,3 +5247,269 @@ void BufferDesc::unLockIO(thread_db* tdbb)
...
+			if (bNew)
+				val.second = bdb;
+			else
+				while (!(bdb2 = val.second))
+					cds::backoff::pause();
+		},
+		true);
+*/
+
+	auto ret = list.update(page, bdb, [&bdb2](bool bNew, BdbList::value_type& val)
```
asfernandes, line 5385: How well maintained is this library?

Your issue from Jul 25, 2021 is not answered?

We depend on custom version, so it's not possible to use system version of library in Linux like we use for others libraries?

src/jrd/jrd.cpp
```diff
@@ -384,6 +384,9 @@ static void threadDetach()
 {
 	ThreadSync* thd = ThreadSync::findThread();
 	delete thd;
+
+	if (cds::threading::Manager::isThreadAttached())
```
asfernandes, line 388: Please encapsulate usage of this library somewhere, like we do with others external libraries.

src/jrd/InitCDSLib.h
```diff
@@ -0,0 +1,70 @@
...
+{
+
+class InitCDS
+{
+public:
+
+	explicit InitCDS(MemoryPool&);
+	~InitCDS();
+
+	// Creates memory pool that will not be deleted until cds finish its work.
```
asfernandes, line 48: How is this relevant for current master, which have the external pool?
External pool and special pool destroy logic was added just for these things and it works very weel with boost.test.
Did you tested the code against recent codebase without this workaround?

asfernandes:
Also, Alex has been implementing hazard ptrs in metadata cache branch. Can't his solution be used here too?

AlexPeshkoff:
On 10/9/22 00:40, Adriano dos Santos Fernandes wrote:
>
> Also, Alex has been implementing hazard ptrs in metadata cache branch. 
> Can't his solution be used here too?
>

I prefer to switch to cds.

src/jrd/cch.cpp
```diff
@@ -1977,6 +2082,7 @@ void CCH_release(thread_db* tdbb, WIN* window, const bool release_tail)
 					// Reassert blocking AST after write failure with dummy lock convert
 					// to same level. This will re-enable blocking AST notification.
 
+					if (!(bcb->bcb_flags & BCB_exclusive))
 					{ // scope
```
dyemanov, line 2086: `// scope` comment is not longer needed

src/jrd/cch.cpp
```diff
@@ -3539,164 +3571,194 @@ static bool expand_buffers(thread_db* tdbb, ULONG number)
...
+		bcb->bcb_hashTable->resize(number);
 
-	const bcb_repeat* const old_end = bcb->bcb_rpt + bcb->bcb_count;
+	SyncLockGuard syncEmpty(&bcb->bcb_syncEmpty, SYNC_EXCLUSIVE, FB_FUNCTION);
+	ULONG allocated = memory_init(tdbb, bcb, number - bcb->bcb_count);
 
-	bcb_repeat* const new_rpt = FB_NEW_POOL(*bcb->bcb_bufferpool) bcb_repeat[number];
-	bcb_repeat* const old_rpt = bcb->bcb_rpt;
+	bcb->bcb_count += allocated;
+	bcb->bcb_free_minimum = (SSHORT)MIN(bcb->bcb_count / 4, 128);	/* 25% clean page reserve */
```
dyemanov, line 3587: Space after cast, please

src/jrd/cch.cpp
```diff
@@ -4316,60 +4232,74 @@ static ULONG memory_init(thread_db* tdbb, BufferControl* bcb, SLONG number)
...
-
 			bcb->bcb_memory.push(memory);
-			memory_end = memory + memory_size;
+
+			tail = (BufferDesc*) FB_ALIGN(memory, alignof(BufferDesc));
+
+			BufferControl::BDBBlock blk;
+			blk.m_bdbs = tail;
+			blk.m_count = to_alloc;
+			bcb->bcb_bdbBlocks.push(blk);
```
dyemanov, line 4240: I suspect some memory overhead is possible here, given that our Array always doubles its storage size when expanding. Maybe bcb_bdbBlocks should be pre-allocated with the known maximum size, e.g. BufferControl::create(dbb, number) where number is passed to BufferControl::ctor and then into bcb_bdbBlocks(p, number)?

src/jrd/cch.cpp
hvlad: Fixed

src/jrd/cch.cpp
hvlad: Fixed

src/jrd/cch.cpp
hvlad: Library was maintained very well until last year when decision to use was made.
It have good reputation and stability, a lot of tests provided.
I see no problem re (not) usage of system version, we already does it and nobody died so far.

src/jrd/InitCDSLib.h
hvlad: I'll look at it. 
The code was written on the v4 code base initially. And I going to maintain it in my custom tree at least for some time.
What problem do you see with this code ?

src/jrd/cch.cpp
hvlad: Wow! Fixed :)

src/jrd/cch.cpp
hvlad: Fixed

src/jrd/jrd.cpp
hvlad: Could you be more specific here, what exactly do you mean ?

src/jrd/InitCDSLib.h
asfernandes: > What problem do you see with this code ?

As said, a not more needed workaround since the solution for the problem was already implemented.

src/jrd/jrd.cpp
asfernandes: We have encapsulation of extern code: ICU (unicode_utils), tommath (BigInteger), tomcrypt (TomCryptHash). absl (Int128).

Why use this library here directly in `jrd.cpp` instead of following this good practice?

hvlad: I still not undertand what you want. Callbacks from engine (?) to attach\detach threads ?
I see no big sence in implementing it, at least until we have just one library that requires it.
Something else ? Explain, pls

BTW, why some methods of library is OK to call from engine and others - not OK ?

src/jrd/InitCDSLib.h
hvlad: The external pool give no benefits for CDS, AFAIU.

CDS allows full control on memory usage and code uses it.
Current code uses two pools:
- one allocated in InitCDS::InitCDS() and used for internal needs of CDS, and
- one allocated in InitPool::InitPool() (at cch.cpp) and used for list nodes (in hash table) only 
Note, there could be allocated very big number of list nodes and distinct pool allows to exactly control its memory usage and avoid fragmentation of pool memory.
~InitCDS() guarantees that pools destroyed after CDS terminated its operations.
Thus there is no need to allocate pools for CDS from "never freed" external pool.
While it is possible, it is not necessary.
And it will not avoid usage of InitCDS().
And I don't want to make all CDS allocations of single common (external) pool, if you offers exactly this.

dyemanov (approved):

PR #7350, opened 2022-10-15
SKIP LOCKED clause for SELECT WITH LOCK, UPDATE and DELETE

asfernandes:


src/jrd/vio.cpp
```diff
@@ -6294,6 +6299,9 @@ static int prepare_update(	thread_db*		tdbb,
 				{
 					tdbb->bumpRelStats(RuntimeStatistics::RECORD_CONFLICTS, relation->rel_id);
 
+					if (writeLockSkipLocked == true)
+						return PREPARE_LOCKERR;
```
dyemanov, line 6303: I'm not sure I like `PREPARE_LOCKERR` meaning somewhat different things depending on the `writeLockSkipLocked` parameter. Maybe we need a different result code e.g. `PREPARE_LOCK` for this particular task?

src/jrd/vio.cpp
```diff
@@ -4556,12 +4557,15 @@ bool VIO_writelock(thread_db* tdbb, record_param* org_rpb, jrd_tra* transaction)
 				}
 			}
 			org_rpb->rpb_runtime_flags |= RPB_refetch;
-			return false;
+			return WriteLockResult::CONFLICTED;
```
dyemanov, line 4560: I'm wondering whether we could return `WriteLockResult::SKIPPED` for _ReadCommitted_ transactions for the `PREPARE_DELETE` case, thus avoiding a refetch inside `LockedStream::LockRecord()` (or request restart in the _ReadConsistency_ mode). @hvlad, what do you think? Do I miss anything?

dyemanov (changes requested):
I have no other objections except comments already posted.

src/jrd/vio.cpp
hvlad: I need some time to review, will report later.

doc/sql.extensions/README.skip_locked.md
```diff
@@ -0,0 +1,95 @@
...
+  [RETURNING ...]
+```
+
+## Notes
+
+When `SELECT ... WITH LOCK` or `DELETE`/`UPDATE` statements are used without `SKIP LOCKED`
+but with `FIRST`/`SKIP`/`ROWS`/`OFFSET`/`FETCH` clauses (say, "skip/limit" clauses), lock happens
+in the last step (i.e. after the "skip/limit" is done).
+
+When `SKIP LOCKED` is used with "skip/limit", "lock / skip locked" happens before "skip/limit".
```
hvlad, line 53: Coud you explain: does it means that `SELECT SKIP 100 ... FROM ... WITH LOCK SKIP LOCKED` will lock 100 skipped records ?

src/jrd/vio.cpp
```diff
@@ -156,13 +156,17 @@ static void list_staying_fast(thread_db*, record_param*, RecordStack&, record_pa
...
 
-const int PREPARE_OK		= 0;
-const int PREPARE_CONFLICT	= 1;
-const int PREPARE_DELETE	= 2;
-const int PREPARE_LOCKERR	= 3;
+enum class PrepareResult
+{
+	SUCCESS,
+	CONFLICT,
+	DELETE,
```
hvlad, line 163: This break Windows build as `DELETE` is already defined in winnt.h as access type constant.

doc/sql.extensions/README.skip_locked.md
dyemanov: Vlas has raised a good point - only rows returned from this query must be locked, this is by design of the WITH LOCK feature. So SKIP clause should skip records before locking. Maybe correct order should be SKIP->LOCK->FIRST? AFAIU, it should not break anything for WITH LOCK, as I don't see how swapping FIRST/LOCK would cause any difference. But for WITH LOCK SKIP LOCKED this would work more correctly than now. The only remaining question is SINGULAR, as it should be the last but it always fetches one more row, thus locking it. But it will throw an error is the second record exists, which is likely to cause a rollback and release the redundant lock.

AlexPeshkoff:
On 10/25/22 19:12, Vlad Khorsun wrote:
>
>
>
> In src/jrd/vio.cpp 
> <https://github.com/FirebirdSQL/firebird/pull/7350#discussion_r1004691232>:
>
> > @@ -156,13 +156,17 @@ static void list_staying_fast(thread_db*, record_param*, RecordStack&, record_pa
>   static void notify_garbage_collector(thread_db* tdbb, record_param* rpb,
>   	TraNumber tranid = MAX_TRA_NUMBER);
>   
> -const int PREPARE_OK		= 0;
> -const int PREPARE_CONFLICT	= 1;
> -const int PREPARE_DELETE	= 2;
> -const int PREPARE_LOCKERR	= 3;
> +enum class PrepareResult
> +{
> +	SUCCESS,
> +	CONFLICT,
> +	DELETE,
>
> This break Windows build as |DELETE| is already defined in winnt.h as 
> access type constant.
>

This is not first time when use of CAPITAL letters constants break build 
for some particular OS due to cpp level conflict.
May be we should use some other formatting for that constants?

doc/sql.extensions/README.skip_locked.md
asfernandes: I set this behavior based on tests of Oracle and PgSQL but saved no tests.
I will do it again and post here.

asfernandes: # PostgreSQL:

It locks skipped records.

## Prepare:
```
create table t(n integer);

insert into t values (0);
insert into t values (1);
insert into t values (2);
insert into t values (3);
insert into t values (4);
insert into t values (5);
insert into t values (6);
insert into t values (7);
insert into t values (8);
insert into t values (9);
```

## Session 1:
```
postgres=# begin transaction;
BEGIN
postgres=*# 
postgres=*# select * from t order by n offset 3 limit 2 for update;
 n 
---
 3
 4
(2 rows)

```

## Session 2:
```
postgres=# select * from t order by n offset 2 limit 2 for update skip locked;
 n 
---
 7
 8
(2 rows)
```

asfernandes: I set this behavior because with `WITH LOCK` (without `SKIP LOCKED`) it "skips, limit and lock" - in this order.

This behavior is incompatible with `SKIP LOCKED` - it does not make sense to limit before lock/skip.

Also made it because PostgreSQL work in this way.

But when changing this it created different behavior for skip with `WITH LOCK` and `WITH LOCK SKIP LOCKED`. This is not good.

I see no good case for SKIP with `WITH LOCKED SKIP LOCKED` (as well with `WITH LOCK`).

So I may revert this behavior.

What I insist is to LOCK/SKIP before limit (FIRST / ROWS) when `SKIP LOCKED` is used.

We may even prohibit SKIP with `WITH LOCK SKIP LOCKED`.

Now for the case of "skip, limit and lock" vs "skip, lock, limit" order, the later should be problematic without `SKIP LOCKED`, so there should be this difference anyway.

asfernandes: > Maybe correct order should be SKIP->LOCK->FIRST? AFAIU, it should not break anything for WITH LOCK, as I don't see how swapping FIRST/LOCK would cause any difference.

AFAIK it would work wrongly, locking one more row than ROWS clause wants.

asfernandes: > > Maybe correct order should be SKIP->LOCK->FIRST? AFAIU, it should not break anything for WITH LOCK, as I don't see how swapping FIRST/LOCK would cause any difference.
> 
> AFAIK it would work wrongly, locking one more row than ROWS clause wants.

Looks like you are correct, it should work as FirstRowsStream verify it before call the next stream.

asfernandes:
> This is not first time when use of CAPITAL letters constants break build for some particular OS due to cpp level conflict. May be we should use some other formatting for that constants?

It's sad because `enum class` is much more elegant and less error prone, but I think we should choose others names variations instead of changing case convention.

doc/sql.extensions/README.skip_locked.md
asfernandes: > The only remaining question is SINGULAR

AFAIU `WITH LOCK` is not allowed in this context.

dyemanov: As for Postgres, this is explicitly documented:
> If a LIMIT is used, locking stops once enough rows have been returned to satisfy the limit (but note that **rows skipped over by OFFSET will get locked**)
However, this behaviour really looks weird. It's like `UPDATE ... ROWS 5 TO 10` updating ten records instead of five. But PG does not have any LIMIT-like clause for `UPDATE/DELETE`, so this may be not so counter-intuitive for them as for us.
MariaDB does not document their behaviour for co-existing `OFFSET` and `FOR UPDATE`. Their `UPDATE` supports only `LIMIT` clause (same as our `FIRST`), without the `OFFSET` part. So again this does not shows them any inconsistencies even if working like PG.
Personally, I don't see any practical sense in `SKIP/OFFSET` used together with `SKIP LOCKED`, be it inside `SELECT` or `UPDATE/DELETE`. So I won't be objecting to prohibit this. But if we're able to make it working consistently (skip -> lock/skip locked -> first), this is IMHO also acceptable.

dyemanov: > > The only remaining question is SINGULAR
> 
> AFAIU `WITH LOCK` is not allowed in this context.

Yes, it looks so. Good ;-)

src/jrd/tra.cpp
```diff
@@ -1948,7 +1948,7 @@ void TRA_sweep(thread_db* tdbb)
 }
 
 
-int TRA_wait(thread_db* tdbb, jrd_tra* trans, TraNumber number, jrd_tra::wait_t wait)
+int TRA_wait(thread_db* tdbb, jrd_tra* trans, TraNumber number, jrd_tra::wait_t wait, Nullable<SSHORT> waitTimeout)
```
hvlad, line 1951: This looks like redundant change: the only case when `waitTimeout` is passed here is `VIO\wait()` and the passed value is 0. It might be better to change `VIO\wait()` to pass `tra_no_wait` instead and not change `TRA_wait()` at all. 
As for me, there must be more strong reason to change `TRA` interface.

src/jrd/vio.cpp
hvlad: Looks like you're correct here

src/include/firebird/impl/blr.h
```diff
@@ -456,4 +459,6 @@
 #define blr_outer_map_message		(unsigned char) 1
 #define blr_outer_map_variable			(unsigned char) 2
 
+#define blr_skip_locked		(unsigned char) 222
```
hvlad, line 462: BLR codes are limited and we close to the limit. 
Could `blr_skip_locked` be implemented as sub-code of `blr_writelock` ?

src/jrd/vio.cpp
asfernandes: Do you mean that or something different?

```diff
diff --git a/src/jrd/vio.cpp b/src/jrd/vio.cpp
index 9b7cf4679c..d86336ae43 100644
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -4545,8 +4545,12 @@ WriteLockResult VIO_writelock(thread_db* tdbb, record_param* org_rpb, jrd_tra* t
        switch (prepare_update(tdbb, transaction, org_rpb->rpb_transaction_nr, org_rpb, &temp, &new_rpb,
                                                   stack, TriState(skipLocked)))
        {
-               case PrepareResult::CONFLICT:
                case PrepareResult::DELETED:
+                       if (skipLocked && (transaction->tra_flags & TRA_read_committed))
+                               return WriteLockResult::SKIPPED;
+                       // fall thru
+
+               case PrepareResult::CONFLICT:
                        if ((transaction->tra_flags & TRA_read_consistency))
                        {
                                Request* top_request = tdbb->getRequest()->req_snapshot.m_owner;
```

src/include/firebird/impl/blr.h
asfernandes: > Could `blr_skip_locked` be implemented as sub-code of `blr_writelock` ?

I don't think so.

dyemanov: As SKIP LOCKED was suggested also for UPDATE/DELETE, this is not going to be possible. Perhaps it could be done using blr_marks, but personally I don't like it much.
We're not that close to the limit yet, as we have also gaps in numbering to be reused. And we may reserve the last BLR verb as a prefix for the second level of numbering (and use USHORT for BLR verbs internally).

src/jrd/vio.cpp
dyemanov: Yes, something like this.

src/jrd/tra.cpp
asfernandes: What seems to work is `tra_probe`.

dyemanov (approved):

src/jrd/tra.cpp
hvlad: Exactly!

src/include/firebird/impl/blr.h
hvlad: Agree.

hvlad (approved):

pavel-zotov:
I've decided to check outcome for different Tx isolation levels and lock resolution.
Several questions appeared.

**isql-1:**
```
recreate table test(id int primary key, f01 int);
commit;
insert into test(id, f01) select row_number()over(), 0 from rdb$types rows 10;
commit;
update test set f01 = 1 where id in (1,5,9);
SQL> 
```

**isql-2:**

**=== 1 ===** Check all possible TIL with lock resolution = 'NO WAIT'
```
SQL> rollback; set transaction read committed read consistency NO wait; select id from test rows 2 with lock skip locked;
ID                              2
ID                              3
-- OK, expected
```

```
SQL> rollback; set transaction snapshot NO wait; select id from test rows 2 with lock skip locked;
ID                              2
ID                              3
-- OK, expected
```


```
SQL> rollback; set transaction snapshot table stability NO wait; select id from test rows 2 with lock skip locked;
Statement failed, SQLSTATE = 40001
lock conflict on no wait transaction
-Acquire lock for relation (TEST) failed
-- YES, also expected (because TIL requires the whole table will be unchanged
```

**Unexpected:**
```
SQL> rollback; set transaction read committed record_version NO wait; select id from test rows 2 with lock skip locked;
Statement failed, SQLSTATE = 40001
deadlock
-read conflicts with concurrent update
-concurrent transaction number is 107
-- Why error here ?
```

**Unexpected:**
```
SQL> rollback; set transaction read committed no record_version NO wait; select id from test rows 2 with lock skip locked;
Statement failed, SQLSTATE = 40001
deadlock
-read conflicts with concurrent update
-concurrent transaction number is 107
-- Why error here ?
```


**=== 2 ===** Check all possible TIL with lock resolution = 'lock timeout <N>'
```
SQL> rollback; set transaction read committed read consistency lock timeout 3; select id from test rows 2 with lock skip locked;
ID                              2
ID                              3
-- OK, expected
```

```
SQL> rollback; set transaction snapshot lock timeout 3; select id from test rows 2 with lock skip locked;
ID                              2
ID                              3
-- OK, expected
```

```
SQL> rollback; set transaction snapshot table stability lock timeout 3; select id from test rows 2 with lock skip locked;
Statement failed, SQLSTATE = 40001
lock time-out on wait transaction
-Acquire lock for relation (TEST) failed
-- YES, also expected (because TIL requires the whole table will be unchanged
```

**Unexpected:**
```
SQL> rollback; set transaction read committed no record_version lock timeout 3; select id from test rows 2 with lock skip locked;
Statement failed, SQLSTATE = 40001
deadlock
-read conflicts with concurrent update
-concurrent transaction number is 107
-- Why error here ?
```

**Unexpected:**
```
SQL> rollback; set transaction read committed record_version lock timeout 3; select id from test rows 2 with lock skip locked;
Statement failed, SQLSTATE = 40001
deadlock
-read conflicts with concurrent update
-concurrent transaction number is 107
-- Why error here ?
```

=== 3 === Check all possible TIL with lock resolution = 'WAIT'
```
SQL> rollback; set transaction read committed read consistency wait; select id from test rows 2 with lock skip locked;
ID                              2
ID                              3
-- OK, expected
```

```
SQL> rollback; set transaction snapshot wait; select id from test rows 2 with lock skip locked;
ID                              2
ID                              3
-- OK, expected
```

```
SQL> rollback; set transaction snapshot table stability wait; select id from test rows 2 with lock skip locked;
--- ++ EXPECTEDLY HANGS +++ after rollback in ISQL-1 output will be:
ID                              1
ID                              2
But the same will be if we COMMIT in isql-1 (why ?)
```

**Unexpected:**
```
SQL> rollback; set transaction read committed no record_version wait; select id from test rows 2 with lock skip locked;
--- @@@ UNEXPECTEDLY HANGS @@@ after commit or rollback in ISQL-1 output will be:
ID                              1
ID                              2
```

**Unexpected:**
```
SQL> rollback; set transaction read committed record_version wait; select id from test rows 2 with lock skip locked;
--- @@@ UNEXPECTEDLY HANGS @@@ after commit or rollback in ISQL-1 output will be:
ID                              1
ID                              2
```

pavel-zotov:
PS. Above mentioned issues raise only when config parameter ReadConsistency is changed to 0.

asfernandes:
> PS. Above mentioned issues raise only when config parameter ReadConsistency is changed to 0.

I'm verifying. It's a interaction of ReadConsistency = 0 and MaxStatementCacheSize > 0.

asfernandes:
> I'm verifying. It's a interaction of ReadConsistency = 0 and MaxStatementCacheSize > 0.

But may happen with MaxStatementCacheSize = 0 too and a stored routine.

We need new issue if beta-1 tag is not going to be moved again.

asfernandes:
The unexpected errors when `no record_version` is used is expected IMO.
As these selects do not work even without `with lock`.

asfernandes:
@pavel-zotov, I created https://github.com/FirebirdSQL/firebird/issues/7486 and committed a fix.
Please comment there if you think there are more problems.

pavel-zotov:
> Please comment there if you think there are more problems.

Intermediate snapshot with timestamp 26-feb-2023 08:00 seems work OK.

> The unexpected errors when no record_version is used is expected IMO.
> As these selects do not work even without with lock.

This I can't understand. If we want to get some records for lock *and* tell engine to skip locked then it is somewhat like: "give us records that have no versions now", isn't ?
And if yes then why NO record_version must issue errors in that case ?
Exception raises because engine sees that records are locked but clause 'SKIP locked' just is not taken in account in that case.
Is it right ?

asfernandes:
`WITH LOCK [SKIP LOCKED]` needs a record read before, but this locked records cannot be read with `NO RECORD VERSION`.
Considering that this transaction mode is replaced by default I would only document it as in fact I don't think there are anything we could do.

PR #7375, opened 2022-11-02
Use hard links instead of soft links

aafemt:
I've found that they works even without elevated privileges


aafemt:
Perhaps it may be a solution as well. But what is wrong with hard links for those who are not aware of Developers Mode?

hvlad:
I hardly see someone not in Dev mode trying to build Firebird (or anything).
As for "problem" with hard links - for cleanup it requires to delete both file and a link, AFAIU.
This could add some unexpected confusion for non-linux people.

PS IMHO, the real root of this issue is that some our .epp files historycally uses not original database names but symlinks to it.

aafemt:
Softlinks on Windows are deleted automatically if original file is deleted? That's a shocking surprise.

hvlad:
Softlink doesn't preserve original file.
Thanks for patience ;)

aafemt:
Ah, this way. I see.

Issue #7388, opened 2022-11-14
Non-equal execution when using view or CTE


chrpos:
Hi,

I am using Firebird Version: 4.0.2 (issue is there also with Firebird 3.0.7)
Due to performance issues I was facing with an sql query on a database, I generated a simple reproducible example, based on the "employee" sample database of firebird.

The issue in short: The query in question uses a view. When I replace the view, that is referenced in the query, with a common table expression (CTE), using the same SQL for the CTE as for the view, the query executes in fact very differently. See the comparison of the performance analysis. Interestingly enough, the execution plans of both queries are the same, though execute differently.

In the following example, Query A references the view "EMP_DEPT_TEAM". Query B uses the content of the view's query as CTE. Query A and B produce the same execution plans. I would expect query A and B performing just the same, but query A performs less good than query B.

**How to reproduce** (use the database "employee.fdb"):
Create a view:
```
create view EMP_DEPT_TEAM
as
select
    EMP_NO,
    DEPARTMENT.DEPT_NO,
    iif(PROJECT.TEAM_LEADER is not null, 'yes', 'no') as IS_TEAMLEADER
from EMPLOYEE
    inner join DEPARTMENT on EMPLOYEE.DEPT_NO = DEPARTMENT.DEPT_NO
    left join PROJECT on EMPLOYEE.EMP_NO = PROJECT.TEAM_LEADER;
```

**Query A (using the view):**
```
select
    EMP_NO,
    DEPT_NO,
    IS_TEAMLEADER
from EMP_DEPT_TEAM
    where EMP_DEPT_TEAM.EMP_NO = (select max(EMP_NO) from EMP_DEPT_TEAM);
```

**Query B (not using the view, but a CTE):**
```
with EMP_DEPT_TEAM_ as (select
    EMP_NO,
    DEPARTMENT.DEPT_NO,
    iif(PROJECT.TEAM_LEADER is not null, 'yes', 'no') as IS_TEAMLEADER
from EMPLOYEE
    inner join DEPARTMENT on EMPLOYEE.DEPT_NO = DEPARTMENT.DEPT_NO
    left join PROJECT on EMPLOYEE.EMP_NO = PROJECT.TEAM_LEADER
)
select
    EMP_NO,
    DEPT_NO,
    IS_TEAMLEADER
from EMP_DEPT_TEAM_
    where EMP_NO = (select max(EMP_NO) from EMP_DEPT_TEAM_);
```

[...]

PR #7390, opened 2022-11-16
Fix invariants optimization involving views (#7388)

dyemanov:
The idea is that we should not compare view field's stream with streams from other RSEs until the field has been expanded to the proper expression. Once it's done, `sub->pass1()` will recursively take care of the invariant checking. The test case from #7388 now shows good performance in both cases, but I'd like to have a second opinion.

### DIFF
--- a/src/dsql/ExprNodes.cpp
+++ b/src/dsql/ExprNodes.cpp
@@ -6539,8 +6539,6 @@ ValueExprNode* FieldNode::pass1(thread_db* tdbb, CompilerScratch* csb)
 {
 	StreamType stream = fieldStream;
 
-	markVariant(csb, stream);
-
 	CompilerScratch::csb_repeat* tail = &csb->csb_rpt[stream];
 	jrd_rel* relation = tail->csb_relation;
 	jrd_fld* field;
@@ -6551,6 +6549,7 @@ ValueExprNode* FieldNode::pass1(thread_db* tdbb, CompilerScratch* csb)
 		if (relation && (relation->rel_flags & REL_being_scanned))
 			csb->csb_g_flags |= csb_reload;
 
+		markVariant(csb, stream);
 		return ValueExprNode::pass1(tdbb, csb);
 	}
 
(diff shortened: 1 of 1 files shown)


livius2:
Is it possible to see difference of such execution in explain plan? 
As now both plans are equal. Do you consider some extension to it?

dyemanov:
I agree it could be useful. Will take look at possible implementation.

dyemanov:
This is the suggested plan output:

```
Sub-query (invariant)
    -> Singularity Check
        -> Aggregate
            -> Nested Loop Join (outer)
                -> Filter
                    -> Hash Join (inner)
                        -> Table "RDB$RELATION_FIELDS" as "SUB RF" Full Scan
                        -> Record Buffer (record length: 529)
                            -> Table "RDB$RELATIONS" as "SUB R" Full Scan
                -> Filter
                    -> Table "RDB$SECURITY_CLASSES" as "SUB SC" Access By ID
                        -> Bitmap
                            -> Index "RDB$INDEX_7" Unique Scan
Select Expression
    -> Filter
        -> Nested Loop Join (outer)
            -> Nested Loop Join (inner)
                -> Filter
                    -> Table "RDB$RELATIONS" as "SUB R" Access By ID
                        -> Bitmap
                            -> Index "RDB$INDEX_1" Range Scan (full match)
                -> Filter
                    -> Table "RDB$RELATION_FIELDS" as "SUB RF" Access By ID
                        -> Bitmap
                            -> Index "RDB$INDEX_4" Range Scan (full match)
            -> Filter
                -> Table "RDB$SECURITY_CLASSES" as "SUB SC" Access By ID
                    -> Bitmap
                        -> Index "RDB$INDEX_7" Unique Scan

```

livius2:
Will be good to see both plans in new version. 
Or it will be now like this?

```
Sub-query (invariant)
    -> Singularity Check
```

and second version

```
Select Expression
    -> Singularity Check
```

or it will look differently?


```
Sub-query (invariant)
    -> Singularity Check
```

and second version

```
Sub-query (variant)
    -> Singularity Check
```

dyemanov:
`"Select Expression"` is the main query, it does not change. For sub-query now we see also `"Select Expression"`, after the patch it will look as either `"Sub-query"` or `"Sub-query (invariant)"`.

dyemanov:
Most sub-queries are correlated in practice, so reporting them as "variant" IMO is about to cause more questions than keeping is simple.

livius2:
Yes, may concern was only if for variant there will be `Select Expression` or also `Sub-query` :)

dyemanov:
Plan explain improvements will be applied separately.

PR #7397, opened 2022-11-25
Inline minor ODS upgrade

dyemanov:
This feature was proposed at the last Firebird conference in Berlin. It allows to upgrade the existing database to the latest minor ODS version, without a need in backup/restore.

1. Upgrade must be done manually, using `gfix -upgrade`
2. It requires exclusive access to the database, error is thrown otherwise
3. `USE_GFIX_UTILITY` system privilege is required
4. Upgrade is transactional, all changes are reverted if any error happens


AlexPeshkoff:
I see no problems with this patch

asfernandes (approved):

PR #7405, opened 2022-12-01
Surface internal optimization modes (all rows vs first rows) at the SQL and configuration levels

dyemanov:
The idea is to force the optimizer to consider another (hopefully better) plan if only a subset or rows if fetched initially. For example, this is common for Delphi applications to load only the first records to the grid and continue fetching the following ones on demand (or in background). This feature may improve the response time in such applications. This was internally supported since v3 (used under the hood for FIRST <n> queries, and also for EXISTS subqueries), now these hints may be used for user queries as well.

Syntax `OPTIMIZE FOR {ALL | FIRST} ROWS` is taken from DB2 which also supports this feature. In fact DB2 allows to specify how many first rows should be taken into account by the optimizer, but I think it complicates its usage, so this is not supported by this patch. Although we could also add it if it's considered useful.

I'm not really happy with new conflicts in the parser, but making `OPTIMIZE` a reserved keyword is IMO not a good thing either.


livius2:
Hi. 
Shouldn't it be settings by RDB$GET_CONTEXT? We can then turn it On before some statements and Off before another.

dyemanov:
It's kinda hint, but a very high-level one. It does not suggest the optimizer what exactly to do and you don't need to understand the plan under the hood. It only asks to use a different cost algorithm. As for RDB$CONTEXT, we never used them to turn something on or off. However, a session-level control statement may be useful for this feature to define per-attachment behaviour.

src/dsql/StmtNodes.h
```diff
@@ -1344,6 +1344,7 @@ class SelectNode final : public TypedNode<StmtNode, StmtNode::TYPE_SELECT>
 	bool dsqlForUpdate = false;
 	bool dsqlWithLock = false;
 	bool dsqlSkipLocked = false;
+	TriState dsqlOptimize;
```
asfernandes, line 1347: It would look for informative some name like `dsqlOptimizeForFirst` as a bool here has no meaning.

PR #7412, opened 2022-12-06
Fix bug when uncommitted records may become visible after the database sweep

ilya071294:
When the server crash happens, the database might have pointer pages with flags which are not consistent with flags on data pages. If ppg_dp_swept is set for a data page which does not actually have dpg_swept set, then records on this page won't be garbage-collected, and they will become visible after OIT is advanced at the end of the sweep. Checking for dead transactions between OIT and OAT helps to determine if the server has been crashing recently. If it has, we cannot trust ppg_dp_swept flag.

### DIFF
--- a/src/jrd/tra.cpp
+++ b/src/jrd/tra.cpp
@@ -1864,6 +1864,25 @@ void TRA_sweep(thread_db* tdbb)
 		TraNumber transaction_oldest_active = transaction->tra_oldest_active;
 		tdbb->setTransaction(transaction);
 
+		// Check if we have dead transactions between OIT and OAT
+		// because it helps to determine if the server has been
+		// crashing recently. If it has, then the database might have
+		// pointer pages with flags which are not consistent with
+		// flags on data pages. If ppg_dp_swept is set for a data page
+		// which does not actually have dpg_swept set, then records on
+		// this page won't be garbage-collected, and they will become
+		// visible after OIT is advanced at the end of the sweep.
+		// Therefore, we cannot trust ppg_dp_swept in this case.
+		// Note that TRA_cleanup has already been called, and dead
+		// transactions are marked as tra_dead.
+		int oldest_state = 0;
+		const TraNumber oldest_dead =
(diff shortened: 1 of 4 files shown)


src/jrd/tra.cpp
```diff
@@ -1864,6 +1864,25 @@ void TRA_sweep(thread_db* tdbb)
...
+		// visible after OIT is advanced at the end of the sweep.
+		// Therefore, we cannot trust ppg_dp_swept in this case.
+		// Note that TRA_cleanup has already been called, and dead
+		// transactions are marked as tra_dead.
+		int oldest_state = 0;
+		const TraNumber oldest_dead =
+			TPC_find_states(tdbb, transaction->tra_oldest, transaction_oldest_active,
+				1 << tra_dead, oldest_state);
+
+		if (oldest_dead)
```
dyemanov, line 1883: I'd also consider adding `fb_assert(oldest_state == tra_dead);` inside this `if`.

hvlad:
> Checking for dead transactions between OIT and OAT helps to determine if the server has been crashing recently. If it has, we cannot trust ppg_dp_swept flag.

I'm afraid with such approach almost every sweep run will not trust `swept` flag.
We need better way to detect database in 'suspected' state, IMO
.

aafemt:
Can careful writes dependency stack force writing of data page before pointer page for setting the flag?

ilya071294:
> Can careful writes dependency stack force writing of data page before pointer page for setting the flag?

It already works this way if we are talking about setting the flag. The problem happens when the flag is cleared (for example, during inserts). Currently the data page will still be written before the pointer page. But in this case we need the dependency to be opposite. It is possible to do so but it should be done carefully because it can ruin the performance.

I'm going to close the PR because it has problems:
1. Dead transaction can appear as a result of "hard" rollback, and if it happens often, the sweep will almost always be slow.
2. Ignoring ppg_dp_swept flag when executing the sweep may not be enough because ppg_dp_empty flag has a similar problem. If the flag is set on the pointer page, and the data page actually has garbage, the sweep will skip this data page as well.

PR #7431, opened 2022-12-21
Crypto manager key correctness check (FB5 port of PR #7423)

AlexeyMochalov:
Related to #7415 issue
This should add key correctness check in case of resetting encryption key on Classic server with multiple attachments, and fix hanging Super server in case of accidental key damaging/deleting


AlexPeshkoff:
Let's wait for regular tests results before merging into stable branches.

PR #7441, opened 2023-01-02
More cursor-related details in the plan output

dyemanov:
1) Main queries are still reported as "select expression", while its sub-queries are reported as "sub-query", and declared cursors are reported as "cursor <name>".
2) Invariant sub-queries and scrollable cursors are reported as such (see also #7390).
3) If the plan is requested for EXECUTE BLOCK or some cached PSQL object (procedure/function/trigger), every select expression inside that PSQL module is prefixed with its position (line/column numbers).

Examples:
```sql
-- line 23, column 2
PLAN (DISTRICT INDEX (DISTRICT_PK))
-- line 28, column 2
PLAN JOIN (CUSTOMER INDEX (CUSTOMER_PK), WAREHOUSE INDEX(WAREHOUSE_PK))
```

```sql
Select Expression (line 23, column 2)
    -> Singularity Check
        -> Filter
            -> Table "DISTRICT" Access By ID
                -> Bitmap
                    -> Index "DISTRICT_PK" Unique Scan
Select Expression (line 28, column 2)
    -> Singularity Check
        -> Nested Loop Join (inner)
            -> Filter
                -> Table "CUSTOMER" Access By ID
                    -> Bitmap
                        -> Index "CUSTOMER_PK" Unique Scan
            -> Filter
                -> Table "WAREHOUSE" Access By ID
                    -> Bitmap
                        -> Index "WAREHOUSE_PK" Unique Scan
```

### DIFF
--- a/src/jrd/recsrc/Cursor.cpp
+++ b/src/jrd/recsrc/Cursor.cpp
@@ -46,34 +46,82 @@ namespace
 
 		return true;
 	}
+}
 
-	// Initialize dependent invariants
-	void initializeInvariants(Request* request, const VarInvariantArray* invariants)
+// ---------------------
+// Select implementation
+// ---------------------
+
+void Select::initializeInvariants(Request* request) const
+{
+	// Initialize dependent invariants, if any
+
+	if (m_rse->rse_invariants)
 	{
-		if (invariants)
+		for (const auto offset : *m_rse->rse_invariants)
 		{
-			for (const ULONG* iter = invariants->begin(); iter < invariants->end(); ++iter)
-			{
-				impure_value* const invariantImpure = request->getImpure<impure_value>(*iter);
-				invariantImpure->vlu_flags = 0;
-			}
+			const auto invariantImpure = request->getImpure<impure_value>(offset);
+			invariantImpure->vlu_flags = 0;
 		}
 	}
+}
+
+void Select::printPlan(thread_db* tdbb, string& plan, bool detailed) const
+{
+	if (detailed)
+	{
+		if (m_rse->flags & RseNode::FLAG_SUB_QUERY)
+		{
+			plan += "\nSub-query";
+
+			if (!(m_rse->flags & RseNode::FLAG_VARIANT))
+				plan += " (invariant)";
+		}
+		else if (m_cursorName.hasData())
+		{
+			plan += "\nCursor \"" + string(m_cursorName) + "\"";
+
+			if (m_rse->flags & RseNode::FLAG_SCROLLABLE)
+				plan += " (scrollable)";
+		}
+		else
+			plan += "\nSelect Expression";
+
+		if (m_line || m_column)
+		{
+			string pos;
+			pos.printf(" (line %u, column %u)", m_line, m_column);
+			plan += pos;
+		}
+	}
+	else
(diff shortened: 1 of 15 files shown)


src/jrd/recsrc/Cursor.h
```diff
@@ -31,26 +31,64 @@ namespace Jrd
...
+		MetaName m_cursorName;	// optional name for explicit PSQL cursors
+
+	private:
+		ULONG m_line = 0;
+		ULONG m_column = 0;
+	};
+
+	// SubQuery class (simplified forward-only cursor)
+
+	class SubQuery : public Select
```
asfernandes, line 78: Looks like `SubQuery` and `Cursor` may be declared as `final` classes.

src/jrd/recsrc/Cursor.cpp
```diff
@@ -130,7 +179,7 @@ void Cursor::close(thread_db* tdbb) const
 
 bool Cursor::fetchNext(thread_db* tdbb) const
 {
-	if (m_scrollable)
+	if (m_rse->flags & RseNode::FLAG_SCROLLABLE)
```
asfernandes, line 182: An inline `isScrollable()` method would make code more clear.

asfernandes:
Is this going to beta 1?

I already started to add same information to the profiler, with means another table for cursor data (it was under a TODO, but now that data should be available) and would like to avoid add migrations after beta 1.

sim1984:
It does not interfere.

dyemanov:
This patch is not critical for Beta (which is late enough), so I planned to merge it later.
IMHO, PR #7397 is more important for Beta, but nobody reacted on it yet :-(

asfernandes:
This comment became not true as now the name of `FOR ... AS CURSOR <name>` became visible.
This name is not being displayed in the plan.

```
// ASF: We cannot define the name of the cursor here, but this is not a problem,
// as implicit cursors are always positioned in a valid record, and the name is
// only used to raise isc_cursor_not_positioned.
```

dyemanov:
I print cursor names only for `DECLARE CURSOR`. I didn't intend to distinguish between `FOR SELECT` and `FOR SELECT AS CURSOR C`. Perhaps it would be handy, but I don't see how this name can be retrieved at the JRD level.

asfernandes:
> Perhaps it would be handy, but I don't see how this name can be retrieved at the JRD level.

We can't do it with `fb_dbg_map_curname` but we can do using something similar to `fb_dbg_map_src2blr` where we can associate position of `blr_for` with the name.

I'm implementing it for #7442.

asfernandes:
> We can't do it with `fb_dbg_map_curname` but we can do using something similar to `fb_dbg_map_src2blr` where we can associate position of `blr_for` with the name.
> 
> I'm implementing it for #7442.

I committed it (https://github.com/FirebirdSQL/firebird/commit/74a18d925f441982f7bdedb499f71f1c51c670e1).

It's much more related to this branch, so we can cherry-pick it here without conflicts if you want.

pavel-zotov:
QA note. 
Test not needed: issues are covered by (at least) bugs/gh_7466_plans_tracking_test.py and bugs/gh_7853_test.py

PR #7442, opened 2023-01-04
Add PLG$PROF_CURSORS and some others fixes

asfernandes:


asfernandes:
This PR is based on yet un-merged https://github.com/FirebirdSQL/firebird/pull/7441.

If it's not going to be merged for beta1, then I'd postpone it only for next major version, as there is no mechanism in place for upgrade of `PLG$PROF_*` objects (they are just normal objects).

But then some fixes need to be collected from this branch to master.

dyemanov (approved):

dyemanov:
I will move the tag after this PR is merged.

PR #7479, opened 2023-02-18
Fix #7476 - Improvement on procedures/triggers to recognize changes in a domain used in COALESCE.

### LINKED ISSUE #7476: Improvement on procedures/triggers to recognize changes in a domain used in COALESCE
Looks like if I have a field based on a custom domain (ie: varchar(18)) and I change the domain to varchar(30), and have procedures/triggers referencing fields using this domain, as in:

insert into table1 (afield) select coalesce(afield,'') from table2;

The change in the domain is not automatically detected by the procedure, producing errors like:

ISC ERROR CODE:335544321 ISC ERROR MESSAGE: arithmetic exception, numeric overflow, or string truncation string right truncation expected length 18, actual 30 At procedure 'XXXXXXXXX line: YY, col: 3.

Vlad sais that the problem is probably due to the use of coalesce creating an implicit cast based on the field definition when the procedure was created.

Tested with Firebird 3.

asfernandes:


asfernandes:
I do not remember why this was not merged, maybe because beta 1 was on the way.
Should we wait for 5.0.1 or merge now?

PR #7502, opened 2023-03-12
fix macOS (Apple Silicon) build

th1722:
Change the specified minimum version of macOS (Apple Silicon)  from 12 (Monterey) to 11 (Big sur).


asfernandes:
Why downgrade this version would make the build work for a newer hardware? What was the error?

Issue #7514, opened 2023-03-24
Segfault when detaching after deleting shadow on Classic


Noremos:
The easiest way to reproduce it is fbtest

```Python
{
'id': 'bug',
'qmid': None,
'title': 'Deleting a shadow',
'description': '',
'min_versions': '2.6',
'versions': [
{
 'firebird_version': '3.0',
 'platform': 'All',
 'page_size': '4096',
 'test_type': 'python',
 'test_script': '''
import os

Shadow1Path = os.path.join(context['temp_directory'], 'DDL_Shadow.shd')

script="""
CREATE SHADOW 1 MANUAL '%s';
commit;

DROP SHADOW 1;
""" %(Shadow1Path)
runProgram('isql', [dsn, '-user', user_name, '-password', user_password, '-q'], script)

 ''',
 'expected_stdout': """ 
 """,
 'expected_stderr': """ 
 """
}
]
}
```

The stack:
```CPP
Jrd::Attachment::createPool(Jrd::Attachment * const this) (src\jrd\Attachment.cpp:150)
CMP_compile(Jrd::thread_db * tdbb, const UCHAR * blr, ULONG blrLength, bool internalFlag, ULONG dbginfoLength, const UCHAR * dbginfo) (src\jrd\cmp.cpp:155)
CMP_compile_request(Jrd::thread_db * tdbb, const UCHAR * blr, ULONG blrLength, bool internalFlag) (src\jrd\cmp.cpp:216)
Jrd::AutoRequest::compile(Jrd::AutoRequest * const this, Jrd::thread_db * tdbb, const UCHAR * blr, ULONG blrLength) (src\jrd\exe_proto.h:171)
MET_get_shadow_files(Jrd::thread_db * tdbb, bool delete_files) (..\Debug\jrd\met.cpp:6539)
SDW_get_shadows(Jrd::thread_db * tdbb) (src\jrd\sdw.cpp:584)
CCH_fetch_lock(Jrd::thread_db * tdbb, Jrd::WIN * window, int lock_type, int wait, SCHAR page_type) (src\jrd\cch.cpp:869)
CCH_fetch(Jrd::thread_db * tdbb, Jrd::WIN * window, int lock_type, SCHAR page_type, int wait, const bool read_shadow) (src\jrd\cch.cpp:802)
CCH_FETCH(Jrd::thread_db * tdbb, Jrd::win * window, USHORT lock_type, SCHAR page_type) (src\jrd\cch_proto.h:84)
[...]

PR #7524, opened 2023-03-29
Suggested fix for #7514: Segfault when detaching after deleting shadow on Classic

AlexPeshkoff:
The reason is as obvious as bug is hard to reproduce. 
When process A deletes shadow it sends appropriate AST to others. Each time before dealing with shadows list appropriate flag is checked which normally causes reload of shadows info. But when database is closing (JRD_shutdown_database) no attachments are left and that does not prevent receiving shadow's AST, i.e. it may be needed to reload shadows info when writing transaction counters to header.

The only visible solution is to skip shadows reload at this moment.

### DIFF
--- a/src/jrd/sdw.cpp
+++ b/src/jrd/sdw.cpp
@@ -581,7 +581,11 @@ void SDW_get_shadows(thread_db* tdbb)
 	// to prevent missing any new ones later on, although it does not
 	// matter for the purposes of the current page being written
 
-	MET_get_shadow_files(tdbb, false);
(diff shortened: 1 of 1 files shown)


src/jrd/sdw.cpp
hvlad, line 1: Since `MET_get_shadow_files()` going to execute request, it expect to have current attachment (and transaction). 
From this POV the patch is very logical.

Just one question - does you consider to move call of `TRA_update_counters()` after setting flag `DBB_no_ast` ?

hvlad (approved):

PR #7549, opened 2023-04-16
Fix #7548 - SET BIND OF TIMESTAMP WITH TIME ZONE TO CHAR is not working with UTF8 connection charset.

### LINKED ISSUE #7548: SET BIND OF TIMESTAMP WITH TIME ZONE TO CHAR is not working with UTF8 connection charset
FB 4.0.2 windows x64

SQL Script:
CREATE DATABASE 'xnet://d:\fbdata\test.fdb' USER 'SYSDBA' PASSWORD 'masterkey';

RECREATE TABLE a(
  t TIMESTAMP WITH TIME ZONE
);

INSERT INTO a VALUES(CURRENT_TIMESTAMP);

SELECT * FROM a;

SET TERM ^;
EXECUTE BLOCK AS
BEGIN
  SET BIND OF TIMESTAMP WITH TIME ZONE TO CHAR;
END
^
SET TERM ;^

SELECT * FROM a;
COMMIT;

WIN1250 works fine:
c:\fb\isql.exe -ch WIN1250 -i set_bind_error.sql

UTF8 are failing:
c:\fb\isql.exe -ch UTF8 -i set_bind_error.sql

T
==============
Statement failed, SQLSTATE = 22001
Dynamic SQL Error
-SQL error code = -303
-arithmetic exception, numeric overflow, or string truncation
-string right truncation
-expected length 14, actual 41

asfernandes:


AlexPeshkoff:
Can we have non-ascii characters in TZ name? If not may be better force use of ASCII charset to coerced fields.

asfernandes:
> Can we have non-ascii characters in TZ name? If not may be better force use of ASCII charset to coerced fields.

I thought on it but it looks very invasive to change the charset, specially thinking on the fix for v4.

Even it being ascii-only, our last pass generally is to convert things for the connection charset and that would not be done in this case.

AlexPeshkoff (approved):

PR #7557, opened 2023-04-27
Named arguments, DEFAULT in argument list and CALL statement

### LINKED ISSUE #7566: Allow DEFAULT keyword in argument list
When routine has many parameters and you want to pass non-default values for the last ones, without DEFAULT you need to pass every parameter up to the last informed.

SQL standard allows to use `DEFAULT` to help in this case:

```
func(1, 2, default, default, 3)
```

SQL spec:
```
<SQL argument> ::=
    <value expression>
    ...
    | <contextually typed value specification>
    | <named argument specification>

<named argument specification> ::=
    <SQL parameter name> <named argument assignment token>

<named argument SQL argument> ::=
    <value expression>
    | <target specification>
    | <contextually typed value specification>

<contextually typed value specification> ::=
    ...
    | <default specification>

<default specification> ::=
    DEFAULT
```

This feature is about functions and procedures (in execute and select).

### LINKED ISSUE #7586: Named arguments for function call, EXECUTE PROCEDURE and procedure record source
Named arguments allows you to specify function and procedure arguments by their names, rather than only
by their positions.

It is especially useful when the routine have a lot of parameters and you want to specify them in arbitrary
order or not specify some of those who have default values.

As the positional syntax, all arguments without default values are required to be present in the call.

A call can use positional, named or mixed arguments. In mixed syntax, positional arguments must appear before
named arguments.

## Syntax

```
<function call> ::=
    [<package name> .] <function_name>( [<arguments>] )

<procedure selection> ::=
    [<package name> .] <procedure_name> [( <arguments> )]

<execute procedure> ::=
    EXECUTE PROCEDURE [<package name> .] <procedure name>
    [{ (<arguments>) | <arguments> }]
    [RETURNING_VALUES ...]

<arguments> ::=
    <positional arguments> |
    [ {<positional arguments>,} ] <named arguments>

<positional arguments> ::=
    <value> [ {, <value>}... ]

<named arguments> ::=
    <named argument> [ {, <named argument>}... ]

<named argument> ::=
    <argument name> => <value>
```

## Examples

```
select function_name(parameter2 => 'Two', parameter1 => 1)
  from rdb$database
```

```
select function_name(1, parameter2 => 'Two')
  from rdb$database
```

```
execute procedure insert_customer(
  last_name => 'SCHUMACHER',
  first_name => 'MICHAEL')
```

```
select *
  from get_customers(city_id => 10, last_name => 'SCHUMACHER')
```

### LINKED ISSUE #7587: CALL statement
`CALL` statement is similar to `EXECUTE PROCEDURE`, but allow the caller to get specific output parameters, or none.

When using the positional or mixed parameter passing, output parameters follows the input ones.

When passing `NULL` to output parameters, they are ignored, and in the case of DSQL, not even returned.

In DSQL output parameters are specified using `?`, and in PSQL using the target variables or parameters.

## Syntax

```
<call statement> ::=
    CALL [<package name> .] <procedure name> (<arguments>)

<arguments> ::=
    <positional arguments> |
    [ {<positional arguments>,} ] <named arguments>

<positional arguments> ::=
    <value or default> [ {, <value or default>}... ]

<named arguments> ::=
    <named argument> [ {, <named argument>}... ]

<named argument> ::=
    <argument name> => <value or default>

<value or default> ::=
    <value> |
    DEFAULT
```

## Examples

```
create or alter procedure insert_customer (
    last_name varchar(30),
    first_name varchar(30)
) returns (
    id integer,
    full_name varchar(62)
)
as
begin
    insert into customers (last_name, first_name)
        values (:last_name, :first_name)
        returning id, last_name || ', ' || first_name
        into :id, :full_name;
end
```

```
-- Not all output parameters are necessary.
call insert_customer(
    'LECLERC',
    'CHARLES',
    ?)
```

```
-- Ignore first output parameter (using NULL) and get the second.
call insert_customer(
    'LECLERC',
    'CHARLES',
    null,
    ?)
```

```
-- Ignore ID output parameter.
call insert_customer(
    'LECLERC',
    'CHARLES',
    full_name => ?)
```

```
-- Pass inputs and get outputs using named arguments.
call insert_customer(
    last_name => 'LECLERC',
    first_name => 'CHARLES',
    last_name => ?,
    id => ?)
```

```
create or alter procedure do_something_and_insert_customer returns (
    out_id integer,
    out_full_name varchar(62)
)
as
    declare last_name varchar(30);
    declare first_name varchar(30);
begin
    call insert_customer(
        last_name,
        first_name,
        out_id,
        full_name => out_full_name);
end
```

asfernandes:
- https://github.com/FirebirdSQL/firebird/issues/7586
- https://github.com/FirebirdSQL/firebird/issues/7566
- https://github.com/FirebirdSQL/firebird/issues/7587


aafemt:
Why names are not resolved in DSQL stage and require creation of new BLR verbs?

asfernandes:
> Why names are not resolved in DSQL stage and require creation of new BLR verbs?

Routines can be changed later, hence existing calls should not pass parameters in wrong order.

aafemt:
Dependency tracking is supposed to ensure metadata consistence and prevent changing signature of routines that are used in existing calls.

asfernandes:
> Dependency tracking is supposed to ensure metadata consistency and prevent changing of signature of routines that are used in existing calls.

Then we create a nice feature (named arguments, with allows routine changes switching parameter orders) for no real use because you do not want the new BLRs. Makes 0 sense.

aafemt:
This feature is for changing of parameters order in call operator. Changing of parameters order of called routine is a completely different thing. And you are right, I see no real use for this feature except Oracle syntax compatibility in either case (unless you are going to prohibit positional parameters completely).

doc/sql.extensions/README.named_arguments.md
```diff
@@ -0,0 +1,52 @@
...
+
+<arguments> ::=
+    <named arguments>
+    <positional arguments>
+
+<named arguments> ::=
+    <named argument> [ {, <named argument>}... ]
+
+<named argument> ::=
+    <argument name> => <value>
```
hvlad, line 33: For me it looks a bit confusing, like assigning name to the value (name => value).
Does you consider to reuse existing `TOK_BIND_PARAM` ( `:=` ) instead ?
I understand that `=>` used in Oracle, but it not means we should copy it :)

aafemt: I see no reason not to use both.

asfernandes: In addition to be used by Oracle and PgSQL, `=>` is SQL standard, so it's not confusing.
I also remember to suggest it when you created `:=`, but in fact as things are not exactly identical, I see no problem to use each of them in their respective places.

asfernandes:
As far as I understood the SQL standard, one must exclusively choose between positional and named arguments in a given usage. So it's what I did here.

But both Oracle and PgSQL allows positional arguments followed by named ones. This syntax is also available in C# language. It's useful to override specific defaults letting others with their defaults.

The BLR syntax would allow it and we may implement it too.

dyemanov:
Looking at this:
```

<SQL argument list> ::=
    <left paren> [ <SQL argument> [ { <comma> <SQL argument> }... ] ] <right paren>

<SQL argument> ::=
    <value expression>
  | <generalized expression>
  | <target specification>
  | <contextually typed value specification>
  | <named argument specification>

<generalized expression> ::=
    <value expression> AS <path-resolved user-defined type name>

<named argument specification> ::=
    <SQL parameter name> <named argument assignment token> <named argument SQL argument>

<named argument SQL argument> ::=
    <value expression>
  | <target specification>
```
it looks like positioned and named arguments are allowed to be intermixed.

asfernandes:
> it looks like positioned and named arguments are allowed to be intermixed.

I read this in SQL 2013:

```
If the <SQL argument list> contains any <SQL argument> that is a <named argument specification>, then
all <SQL argument>s contained in that <SQL argument list> shall be <named argument specification>s;
no two <SQL parameter name>s contained in the <SQL argument list> shall be equivalent.
```

src/dsql/pass1.cpp
```diff
@@ -538,25 +538,68 @@ dsql_ctx* PASS1_make_context(DsqlCompilerScratch* dsqlScratch, RecordSourceNode*
...
+				fb_assert(procNode->dsqlInputArgNames->getCount() == inputList->items.getCount());
+
+				LeftPooledMap<MetaName, const dsql_fld*> argsByName;
+
+				for (const auto* field = procedure->prc_inputs; field; field = field->fld_next)
+					argsByName.put(field->fld_name, field);
+
+				bool mismatchError = false;
+				Arg::StatusVector mismatchStatus;
+				mismatchStatus << Arg::Gds(isc_prcmismat) << procNode->dsqlName.toString();
```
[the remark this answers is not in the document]
asfernandes: How many milliseconds we are losting per thousand of statement prepare and what percentage of it in the whole prepare timings?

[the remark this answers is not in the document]
aafemt: Many a mickle makes a muckle.

Issue #7576, opened 2023-05-08
Allow nested parenthesized joined table


mrotteveel:
The current syntax allows only a single level of parentheses around joined tables, but the SQL:2016 syntax is:

```
<table primary> ::=
     [...]
  | <parenthesized joined table>

<parenthesized joined table> ::=
    <left paren> <parenthesized joined table> <right paren>
  | <left paren> <joined table> <right paren>

[...]
```

A simple syntax change allows this (I'll push a pull request shortly).

asfernandes:
Please give a example of a query that was not working.

mrotteveel:
The SQL standard syntax allows arbitrary depth of nesting in parentheses.

An example of a query that doesn't work without this syntax change:
```sql
select c.customer, ctr.currency
from ((
  customer c inner join country ctr 
    on c.country = ctr.country 
))
```

Obviously, the error can be solved by removing a level of parentheses, but I have seen tools generate similar queries, expecting arbitrary nesting to work.

PR #7577, opened 2023-05-08
Allow nested parenthesized joined table

mrotteveel:
See #7576


asfernandes (approved):

mrotteveel:
@asfernandes Thanks for reviewing. Do you think it is OK to merge now, or should I wait for opinions on firebird-devel?

dyemanov:
I'd vote for merging it now.

Issue #7599, opened 2023-05-23
Conversion of text with '\0' to DECFLOAT without errors.


dmitry-lipetsk:
```
SQL> show version;
ISQL Version: WI-V4.0.3.2937 Firebird 4.0
Server version:
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-V4.0.3.2937 Firebird 4.0"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-V4.0.3.2937 Firebird 4.0/tcp (HOME4)/P17:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-V4.0.3.2937 Firebird 4.0/tcp (HOME4)/P17:C"
on disk structure version 13.0
SQL> select cast('123'||x'00'||'321' as integer) from rdb$database;

        CAST
============
Statement failed, SQLSTATE = 22018
conversion error from string "123"
SQL> select cast('123'||x'00'||'321' as DECFLOAT(34)) from rdb$database;

                                      CAST
==========================================
                                       123

```

AlexPeshkoff:
BTW diags in case of integer is also rather far from ideal

dmitry-lipetsk:
Alex, I rewrote decFloatFromString to work with a range of characters, and I seem to have found a few pointer issues.

If you haven't started dealing with this issue yet, you can wait while I finish and provide more information.

AlexPeshkoff:
Have not. It will be great. 
BTW, of there are issues with decFloatFromString, that probably should be reported to IBM.

dmitry-lipetsk:
About a problem with pointers. I'll try to explian a problem as simple as possible :)

Sample query:

```sql
select cast('-0000000000000000000000000000000000000000000000000000000000000000000000000000000.E+7000' as DECFLOAT(34))
from rdb$database
```

[decFinalize](https://github.com/FirebirdSQL/firebird/blob/5ed9d76d2951501e0e5b344ef45c064e0da0131e/extern/decNumber/decCommon.c#L230) function receives 'num' with this data:
![image](https://github.com/FirebirdSQL/firebird/assets/4594554/a7dffba5-de65-4764-817e-09acf2afdc09)

As you can see, lsd<msd and both members are pointing to uninitialized memory.

**msd** is pointing on MOST significant digit
**lsd** is pointing LEAST significant digit

decFinalize expectes that num contains at least one digit - it means that **msd<=lsd**

Look at this code of decFinalize:

[...]

PR #7609, opened 2023-05-30
Proposal for fixing #7599

dmitry-lipetsk:
Using range-based parser (decFloatFromString_fb) instead null-terminated algorithm (decFloatFromString).

New code was added into decCommon_fb.c file.

Cross test of the old and new code was done (logs are attached).

Original code of new parser was created on C++ and has lots of debug checks those were removed.

---
Please, check the new code in cvt.cpp file

[20230530_201557_1--new.log](https://github.com/FirebirdSQL/firebird/files/11602744/20230530_201557_1--new.log)
[20230530_201142_1--old.log](https://github.com/FirebirdSQL/firebird/files/11602745/20230530_201142_1--old.log)


dmitry-lipetsk:
It was needed to put new code into decCommon.c :)

Correct me, please, if I am wrong.

As I undestand, it is need to correct these files:

CMakeLists.txt
 - [add] [line 374] list(FILTER decNumber_src EXCLUDE REGEX ".*decCommon_fb.c")

extern\decNumber\Makefile
 - [change] [line 5] 	$(CC) $(CROSS_FLAGS) -c -O3 -fPIC $(subst decCommon_fb.c,,$(subst decCommon.c,,$(subst decBasic.c,,$(wildcard *.c))))

dmitry-lipetsk:
As I understand, this patch has not already required.

Issue #7580, opened 2023-05-10
Write firebird.log entry for unauthorized / failed database attachment attemps


pcisar:
In Firebird 2.5, it was possible to detect unauthorized db attachment attempts via trace. However, this does not work anymore since v3, as unauthorized attachments are handled before engine code that handles appropriate trace event. As this feature could be used to detect attempts to hack the database from outside, it would be good to at least write a firebird.log entry.

As it's very small change, I hope it would be added to all versions since v3.

dyemanov:
Tracing of failed connect attempts is already implemented by RedSoft. We'll prepare PR for Alex's or Vlad's review.

Noremos:
The PR is here: https://github.com/FirebirdSQL/firebird/pull/7620

PR #7620, opened 2023-06-05
Trace failed attaches #7580

Noremos:
I created a JRD-free ServerTraceManager to use it at Remote. It enables catching failed attachments. The new class is set as a parent class to TraceManager to reduce code duplication. I decided not to use virtual methods to maintain performance


src/utilities/strace/ServerTraceManager.h
```diff
@@ -0,0 +1,174 @@
...
+ *  License. You may obtain a copy of the License at
+ *  http://www.ibphoenix.com/main.nfs?a=ibphoenix&page=ibp_idpl.
+ *
+ *  Software distributed under the License is distributed AS IS,
+ *  WITHOUT WARRANTY OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing rights
+ *  and limitations under the License.
+ *
+ *  The Original Code was created by Artyom Abakumov
+ *  <artyom.abakumov (at) red-soft.ru> for Red Soft Corporation.
```
hvlad, line 18: Really ???

src/jrd/trace/TraceManager.h
```diff
@@ -53,7 +44,7 @@ class jrd_tra;
 class DsqlRequest;
 class Service;
 
-class TraceManager
+class TraceManager : public ServerTraceManager
```
hvlad, line 47: It is very confused.
I suppose the base class should be a TraceManager, and derived classes should be a JrdTraceManager,  (Net)ServerTraceManager, etc

src/remote/server/server.cpp
```diff
@@ -406,6 +413,151 @@ class FailedLogins : private SortedObjectsArray<FailedLogin,
 	}
 };
 
+
+// For Trace
+
+class TraceFailedAuthImpl :
```
hvlad, line 419: Is it necessary to pollute server.cpp with such aux classes ? Could it be moved into separate file ?

hvlad:
> > Also, new folder `src/utilities/strace` adds confusion with already existing `src/utilities/ntrace`. I suggest to use more distinct name, such as `src/utilities/TraceManager`
> 
> To be precise - TraceManager is not utility. Placing it into utilities folder is rather confusing. 

It was my first thought too

> BTW, keeping in jrd is also bad idea - that's not part of DB engine any more, that's common code for engine and netserver. I think we should add new folder for it. Things like auth & replication (also have common parts for engine/netserver) should also go into it.

Agree. 

> 
> I.e.
> 
> ```
> src.......
>          common
>          jrd
>          newfolder.......
>                       auth
>                       replication
>                       trace

Please, TraceManager (or likewise) - to not confuse with other trace-related code.

>          remote
> ```
> 
> Sorry, no idea how to call it.

I can live with 'utilities'. But if you consider there should be something different... I can offer: aux, assist, supplement, support, helper

aafemt:
> I can live with 'utilities'. But if you consider there should be something different... I can offer: aux, assist, supplement, support, helper

So far such "helper" things end up in common/classes folder.

src/utilities/strace/ServerTraceManager.h
Noremos: Sorry, I didn't mean something like that

src/remote/server/server.cpp
AlexPeshkoff: I see no reasons to move that class out of server.cpp. Or we should review a great lot of our sources - for example, TraceFailedConnection (cousin of TraceFailedAuthImpl) successfully lives in jrd.cpp.

builds/win32/msvc15/fbserver.vcxproj
```diff
@@ -213,6 +213,10 @@
     <ClCompile Include="..\..\..\src\remote\server\ReplServer.cpp" />
     <ClCompile Include="..\..\..\src\jrd\replication\Config.cpp" />
     <ClCompile Include="..\..\..\src\jrd\replication\Utils.cpp" />
+    <ClCompile Include="..\..\..\src\utilities\ntrace\compress.cpp" />
```
AlexPeshkoff, line 216: What does compress.cpp do here? Looks like causes windows build fail.

AlexPeshkoff (approved):

hvlad:
The moving of TraceManager core out of Jrd requires much more refactoring that it is done in this patch.
I don't like to accept it in current state, sorry.
I can finish it by myself, if needed, but I can't (and don't want to) commit into external (Red Soft) repository.
Perhaps, this patch could be commited into separate branch at Firebird repository to continue to improve it there.

dyemanov:
Moved the patch into the Firebird repo: https://github.com/FirebirdSQL/firebird/tree/trace-failed-attach.
@hvlad , the ball is in your court now.

PR #7622, opened 2023-06-06
Fixed #7579: Cannot nbackup a firebird 3.0 database in firebird 4.0 service with engine12 setup in Providers

### LINKED ISSUE #7579: Cannot nbackup a firebird 3.0 database in firebird 4.0 service with engine12 setup in Providers
In firebird 4.0, I configure `firebird.conf` to support both `engine12` and `engine13` plugins:

    Providers = Remote,Engine13,Engine12,Loopback

The firebird service may access both Firebird 3.0 and 4.0 database.

Next, I try to nbackup a Firebird 4.0 employee database and it works:

    fbsvcmgr service_mgr action_nbak dbname c:\4.0\employee.fdb nbk_file r:\fb40.nb0 nbk_level 0

Finally, I try to use nbackup a Firebird 3.0 employee database:

    fbsvcmgr service_mgr action_nbak dbname c:\3.0\employee.fdb nbk_file r:\fb30.nb0 nbk_level 0

and it fail with message:

    unsupported on-disk structure for file c:\3.0\employee.fdb; found 12.0, support 13.0

AlexPeshkoff:
In my environment the following commit in B3_0_Release is needed to make engine12 use ICU already loaded before it by engine13: https://github.com/FirebirdSQL/firebird/commit/2052fde1f2138859eb03da6d782edbfa50a879dd
It has nothing to do with provider selection, just not backported earlier fix for mod_loader.

For me after applying this PR commands:
```
fbsvcmgr localhost: user sysdba password masterkey action_nbak dbname e12 nbk_file e12copy nbk_level 0
fbsvcmgr localhost: user sysdba password masterkey action_db_stats sts_hdr_pages dbname e12
```
work fine with FB5 server and the following config:
firebird.conf:
`Providers = Remote,Engine13,Engine12,Loopback`
databases.conf:
`e12 = $(dir_sampleDb)/e12.fdb`
where e12.fdb has ODS major == 12.

### DIFF
--- a/src/include/firebird/FirebirdInterface.idl
+++ b/src/include/firebird/FirebirdInterface.idl
@@ -734,6 +734,10 @@ version:	// 3.0.9 => 3.0.10, 4.0.1 => 4.0.2
 
 interface Provider : PluginBase
 {
+	const uint PTYPE_ANY		= 0;
+	const uint PTYPE_LOCAL		= 1;
+	const uint PTYPE_NETWORK	= 2;
+
 	Attachment attachDatabase(Status status, const string fileName,
 		uint dpbLength, const uchar* dpb);
 	Attachment createDatabase(Status status, const string fileName,
@@ -742,6 +746,9 @@ interface Provider : PluginBase
 		 uint spbLength, const uchar* spb);
 	void shutdown(Status status, uint timeout, const int reason);
 	void setDbCryptCallback(Status status, CryptKeyCallback cryptCallback);
+
+version:	// 3.0.10 => 3.0.11, 4.0.2 => 4.0.3
+	uint getType(Status status);
 }
 
 // Helper to start transaction over >1 attachments (former TEB)
--- a/src/include/firebird/FirebirdInterface.idl
+++ b/src/include/firebird/FirebirdInterface.idl
@@ -734,10 +734,6 @@ version:	// 3.0.9 => 3.0.10, 4.0.1 => 4.0.2
 
 interface Provider : PluginBase
 {
-	const uint PTYPE_ANY		= 0;
-	const uint PTYPE_LOCAL		= 1;
-	const uint PTYPE_NETWORK	= 2;
-
 	Attachment attachDatabase(Status status, const string fileName,
 		uint dpbLength, const uchar* dpb);
 	Attachment createDatabase(Status status, const string fileName,
@@ -746,9 +742,6 @@ interface Provider : PluginBase
 		 uint spbLength, const uchar* spb);
 	void shutdown(Status status, uint timeout, const int reason);
 	void setDbCryptCallback(Status status, CryptKeyCallback cryptCallback);
-
-version:	// 3.0.10 => 3.0.11, 4.0.2 => 4.0.3
-	uint getType(Status status);
 }
 
 // Helper to start transaction over >1 attachments (former TEB)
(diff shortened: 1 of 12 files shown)


src/yvalve/why.cpp
```diff
@@ -6320,6 +6321,43 @@ void YService::start(CheckStatusWrapper* status, unsigned int spbLength, const u
...
+				{
+					alternativeHandle->start(&st, spb.getBufferLength(), spb.getBuffer());
+					if (st.isSuccess())
+					{
+						status->init();
+						return;
+					}
+				}
+
+				// we are not going to attach network providers, therefore const svcName is OK
```
aafemt, line 6343: Actually - why? What's wrong with remote providers running service on different (local) server if engine cannot handle this database? I can easily imagine configuration "Providers=Engine13,loopback" and Firebird 3 installed as well.

src/include/firebird/FirebirdInterface.idl
```diff
@@ -734,6 +734,10 @@ version:	// 3.0.9 => 3.0.10, 4.0.1 => 4.0.2
 
 interface Provider : PluginBase
 {
+	const uint PTYPE_ANY		= 0;
```
asfernandes, line 737: Method is `getType`, so why `PTYPE_` instead of `TYPE_`?
The names are already at scope `Provider`.
We don't use the extra specifications in the others constants.

src/yvalve/why.cpp
```diff
@@ -6515,83 +6553,122 @@ YService* Dispatcher::attachServiceManager(CheckStatusWrapper* status, const cha
 			IntlSpb().toUtf8(spbWriter);
 		}
 
-		// Build correct config
-		RefPtr<const Config> config(Config::getDefaultConfig());
-		if (spbWriter.find(isc_spb_config))
+		IProvider* p;
+		service = internalServiceAttach(status, svcName, spbWriter,
+			[](CheckStatusWrapper*, IService*){ }, Dispatcher::PTYPE_ANY, &p);
```
asfernandes, line 6558: Can't this empty lambda be just `{}`?

src/yvalve/why.cpp
AlexPeshkoff: I suppose you talk about client (i.e. embedded usage)? Or also about FB4 server on non-default port (XNET disabled) and FB3 on default port (XNET enabled)? 

First case formally works (provided one is using services for engine13 embedded access which actually makes no sense). 

Second will work only when Redirection=1 in firebird.conf. But for single server with default network setup that will cause infinite recursion, even when no additional providers present, Redirection=1 is enough. Something not good.

Looks like good alternative to avoid recursion is to cleanup Redirection at all. That will avoid need in getType(), enable first case (at client) and reliably protect from infinite recursion.

src/yvalve/why.cpp
AlexPeshkoff: Can - but I see no use in it, in that case additional null check in internalServiceAttach() needed.

src/include/firebird/FirebirdInterface.idl
AlexPeshkoff: I will fix it - provided this method does not go away at all after "Redirection" cleanup.

src/yvalve/why.cpp
AlexPeshkoff: See devel (Redirection parameter in firebird.conf) - why additionally pollute code supporting generally useless feature?

PR #7630, opened 2023-06-15
Firebird.conf DefaultDbCachePages better description

livius2:
Hi. As many times on Firebird group is question about DefaultDbCachePages, i have written probably better description. Users ask why my server use so low memory even if i setup big value in config. They do not know how this cache work and that firebird.conf setting can be overrided by different one.

Please fill free to fix it as i am not native English

### DIFF
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -253,7 +253,7 @@
 # Priority of setings is: direcly in database, databases.conf, firebird.conf.
 # The SuperServer default value is 2048 pages for each database and the classic
 # value is 256 pages per client connection per database.
-# As this value means maximum number of pages cached, at start server use minimal cache.
+# As this value means maximum number of pages cached, at start server use minimal RAM.
 # Cache is filled over time as server access different pages by issuing any command e.g. SELECT.
 #
 # Per-database configurable.
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -244,13 +244,13 @@
 # ----------------------------
 # Number of cached database pages
 #
-# This setings is used to cache in memory database pages. 
-# It specify maximum number of pages which can be cached.
+# This seting is used to cache in memory database pages. 
+# It specifies maximum number of pages which can be cached.
 # In SuperServer it is per database, in Classic per client connection.
 # Per-database configurable. This value can be overrided by databases.conf or 
 # directly in database file by issuing command:
 # gfix -b[uffers] BUFFERS
-# Priority of setings is: direcly in database, databases.conf, firebird.conf.
+# Priority of settings is: directly in database, databases.conf, firebird.conf.
 # The SuperServer default value is 2048 pages for each database and the classic
 # value is 256 pages per client connection per database.
 # As this value means maximum number of pages cached, at start server use minimal RAM.
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -244,7 +244,7 @@
 # ----------------------------
 # Number of cached database pages
 #
-# This seting is used to cache in memory database pages. 
+# This setting is used to cache in memory database pages. 
 # It specifies maximum number of pages which can be cached.
 # In SuperServer it is per database, in Classic per client connection.
 # Per-database configurable. This value can be overrided by databases.conf or 
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -244,11 +244,17 @@
 # ----------------------------
 # Number of cached database pages
 #
-# This sets the number of pages from any one database that can be held
-# in cache at once. If you increase this value, the engine will
-# allocate more pages to the cache for every database. By default, the
-# SuperServer allocates 2048 pages for each database and the classic
-# allocates 256 pages per client connection per database.
+# This setings is used to cache in memory database pages. 
+# It specify maximum number of pages which can be cached.
+# In SuperServer it is per database, in Classic per client connection.
+# Per-database configurable. This value can be overrided by databases.conf or 
+# directly in database file by issuing command:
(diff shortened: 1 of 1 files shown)


builds/install/misc/firebird.conf
```diff
@@ -244,11 +244,17 @@
 # ----------------------------
 # Number of cached database pages
 #
-# This sets the number of pages from any one database that can be held
-# in cache at once. If you increase this value, the engine will
-# allocate more pages to the cache for every database. By default, the
-# SuperServer allocates 2048 pages for each database and the classic
-# allocates 256 pages per client connection per database.
+# This setings is used to cache in memory database pages.
```
omachtandras, line 247: setting

builds/install/misc/firebird.conf
```diff
@@ -244,11 +244,17 @@
 # ----------------------------
 # Number of cached database pages
 #
-# This sets the number of pages from any one database that can be held
-# in cache at once. If you increase this value, the engine will
-# allocate more pages to the cache for every database. By default, the
-# SuperServer allocates 2048 pages for each database and the classic
-# allocates 256 pages per client connection per database.
+# This setings is used to cache in memory database pages. 
+# It specify maximum number of pages which can be cached.
```
omachtandras, line 248: specifies

builds/install/misc/firebird.conf
```diff
@@ -244,11 +244,17 @@
...
+# In SuperServer it is per database, in Classic per client connection.
+# Per-database configurable. This value can be overrided by databases.conf or 
+# directly in database file by issuing command:
+# gfix -b[uffers] BUFFERS
+# Priority of setings is: direcly in database, databases.conf, firebird.conf.
```
omachtandras, line 253: settings

omachtandras (changes requested):

builds/install/misc/firebird.conf
livius2: I see here one typo `direcly` should be `directly`

builds/install/misc/firebird.conf
```diff
@@ -244,11 +244,17 @@
...
+# This setting is used to cache in memory database pages. 
+# It specifies maximum number of pages which can be cached.
+# In SuperServer it is per database, in Classic per client connection.
+# Per-database configurable. This value can be overrided by databases.conf or 
+# directly in database file by issuing command:
+# gfix -b[uffers] BUFFERS
+# Priority of settings is: directly in database, databases.conf, firebird.conf.
+# The SuperServer default value is 2048 pages for each database and the classic
+# value is 256 pages per client connection per database.
+# As this value means maximum number of pages cached, at start server use minimal RAM.
```
romansimakov, line 256: Engine allocates all buffers at the attachment. It's true both for SuperServer and for ClassicServer. Therefore, SuperServer allocates all buffers at the first connection. ClassicServer allocates buffers for every attachment and this number depends on number of attachments.
I think suggested description is not complete as well and doesn't sole original question.

livius2: @romansimakov 
I did not observe such thing like SuperServer allocation of all buffers. I start server, connect to database and server still use low RAM. Only if i select from multiple tables it increase RAM usage.

romansimakov: > @romansimakov I did not observe such thing like SuperServer allocation of all buffers. I start server, connect to database and server still use low RAM. Only if i select from multiple tables it increase RAM usage. But i still use FB3. Maybe was some changes in recent version of FB?

What's your OS and how do you see? I'm afraid OS just doesn't allocate asked memory phisically and leaves it as virtually allocated. When engine addresses a memory page maps on phisical memory and from that time OS account it as used by the process. 

For example, on my Ubuntu I can see the following for 2048 cache on SuperServer:
1) Just after server start
```
roman:bin$ grep Vm /proc/`pidof rdbserver`/status 
VmPeak:	  470764 kB
VmSize:	  405680 kB
VmHWM:	   42504 kB
VmRSS:	   42504 kB
VmData:	   58240 kB
```
2) After connect to a big (80GB) database:
roman:bin$ grep Vm /proc/`pidof rdbserver`/status 
```
VmPeak:	 1384600 kB
VmSize:	 1319452 kB
VmHWM:	   72432 kB
VmRSS:	   72432 kB
VmData:	  632884 kB
```
3) After select from a big table:
roman:bin$ grep Vm /proc/`pidof rdbserver`/status 
```
VmPeak:	 1384600 kB
VmSize:	 1319068 kB
VmHWM:	  103860 kB
VmRSS:	  103860 kB
VmData:	  632500 kB
```
4) After disconnect
roman:bin$ grep Vm /proc/`pidof rdbserver`/status 
```
VmPeak:	 1384600 kB
VmSize:	  764724 kB
VmHWM:	  103860 kB
VmRSS:	   40868 kB
VmData:	   90628 kB
```
Note how VmRSS and VmSize change.

romansimakov: I.e. it's controlled not by the Engine but by OS

livius2: I am working only on Windows, and i simply looked into TaskManager->Resource monitor
At start FB (SuperServer) show really low RAM usage, only when i do select from multiple tables it increase:
image

dyemanov: Get a difference between virtual memory (working set) and physical (committed) memory. When you allocate memory, virtual memory counter is increased. When you access that allocated memory, physical memory counter is increased. So Firebird allocates (read: reserves in virtual memory) the whole page cache at startup, but its usage (physical/committed memory) will grow only when pages are read into the cache.

livius2: I suppose that talking about Phisical Memory usage is more interesting for people which read config values description and what they observe after FB start. But my text was only proposition of description, fill free to extend that info of Virtual Memory preservation and any other info which help to understand that parameter.

PR #7635, opened 2023-06-16
Improvement #7634 : Include Performance Cores only in default affinity mask

### LINKED ISSUE #7634: Include Performance Cores only in default affinity mask
Some modern CPU's could contain two set of cores - Efficient (E) and Performance (P).
By default Firebird runs on all available cores. Many times I saw significant performance drop in this (default) case. 
Setting affinity mask to exclude E-cores restore performance to the expected level. More, I see the same problem not 
only with Firebird, for example Firebird build runs near 3 times slower when E-cores are used on my i7-12700 CPU (8P + 4E). 
So far this problem was not confirmed on non-Windows OS. Note, CpuAffinityMask is implemented on Windows only.

I offer to exclude E-cores when affinity mask is not set, i.e. by default.

hvlad:


src/common/os/win32/os_utils.cpp
```diff
@@ -560,6 +561,65 @@ void getUniqueFileId(HANDLE fd, UCharBuffer& id)
...
+
+	DWORD len = 0;
+
+	if (!GetLogicalProcessorInformationEx(RelationProcessorCore, nullptr, &len))
+	{
+		if (GetLastError() != ERROR_INSUFFICIENT_BUFFER)
+			return;
+	}
+
+ 	HalfStaticArray<UCHAR, 1024> buff;
```
asfernandes, line 585: Don't you need this variable aligned as `PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX`? (using third parameter of `HalfStaticArray`)?

aafemt: It is an usual requirement for WinAPI and providing misaligned pointer can end up in ERROR_NOACCESS returned. As an example GetTokenInformation() does that though it is not documented anywhere and for ReadDirectoryChangesEx() it is written explicitly.

hvlad: Not documented == not usual. 
Anyway, I check with 1, 2, 4 and 8 bytes aligned pointers and found that 1 and 2 indeed returns `ERROR_NOACCESS`. 
Note, `alignof(SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX) == 8`.
Also, `InlineStorage<>` by default returns 8-byte aligned pointer.

hvlad: Fixed.

aafemt: In this case function's parameter is declared as a pointer to struct so "not documented" == "compiler default ABI". If the parameter had been declared as PVOID that would indeed might vary.

asfernandes: > Also, `InlineStorage<>` by default returns 8-byte aligned pointer.

It does not.

PR #7641, opened 2023-06-18
Fix wrong profiler measurements due to overflow.

asfernandes:


src/plugins/profiler/Profiler.cpp
```diff
@@ -240,9 +248,10 @@ class ProfilerPlugin final : public StdPlugin<IProfilerPluginImpl<ProfilerPlugin
 
 //--------------------------------------
 
-void ProfilerPlugin::init(ThrowStatusExceptionWrapper* status, IAttachment* attachment)
+void ProfilerPlugin::init(ThrowStatusExceptionWrapper* status, IAttachment* attachment, FB_UINT64 ticksFrequency)
 {
 	userAttachment = attachment;
+	::ticksFrequency = (SINT64) ticksFrequency;
```
dyemanov, line 254: Maybe just nitpicking, but I'd prefer to rename the variable (ticksFreq or frequency) to avoid the :: scope specifier.

PR #7654, opened 2023-06-29
Update allowed C++ version

asfernandes:


doc/README.modern_cpp.md
```diff
@@ -1,11 +1,12 @@
 # Modern C++ features
 
-We define here as "modern C++" features the ones introduced since C++17. These features should be used in a "controlled" form in Firebird code, as not all compilers fully support them.
+It's allowed to use in Firebird internal code C++ features up to and including C++17.
 
-Only ones mentioned in this document could be used, but as necessities appears, discussion should be started in the devel list or pull requests.
+Public files (as API headers) is limited to C++11.
```
asfernandes, line 5: Note that I also want to change version of public code to C++11.

asfernandes:
If we agree on C++17, I will also update build files.

doc/README.modern_cpp.md
hvlad: How this limitation can be enforced on practice ?

hvlad:
> If we agree on C++17, I will also update build files.

No objections.

doc/README.modern_cpp.md
asfernandes: We may build the examples with this option.

asfernandes:
> If we agree on C++17, I will also update build files.

As we had previously defined C++14 with C++17 extensions, build files are already defined to C++17.

doc/README.modern_cpp.md
aafemt: So now API headers can handle std::exception?

asfernandes: `std::exception` is pre-c++11 feature, so explain what you want in fb-devel, as it seems to have nothing related with this discussion.

hvlad:
Looks like some incompatibility (with commit b31d37b, probalby) after merge: MacOS build of examples can't find path for libstdc++ headers

PR #7655, opened 2023-06-30
Fix crash on assert or memory corruption when too long a username is …

dmitry-starodubov:
…used for authentication

### DIFF
--- a/src/auth/SecureRemotePassword/server/SrpServer.cpp
+++ b/src/auth/SecureRemotePassword/server/SrpServer.cpp
@@ -274,7 +274,7 @@ int SrpServer::authenticate(CheckStatusWrapper* status, IServerBlock* sb, IWrite
 
 			account = sb->getLogin();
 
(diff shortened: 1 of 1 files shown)


src/auth/SecureRemotePassword/server/SrpServer.cpp
```diff
@@ -274,6 +274,10 @@ int SrpServer::authenticate(CheckStatusWrapper* status, IServerBlock* sb, IWrite
 
 			account = sb->getLogin();
 
+			const size_t len = strlen(account.c_str());
```
AlexPeshkoff, line 277: Why not
const size_t len = account.length();

AlexPeshkoff (changes requested):

src/auth/SecureRemotePassword/server/SrpServer.cpp
dmitry-starodubov: Yes, that would be better

PR #7662, opened 2023-07-06
Fix performance issues in prepare_update()

ilya071294:
Commits in the PR fix similar cases where the temporary record is inserted twice in prepare_update().

### DIFF
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -574,8 +574,6 @@ namespace
 	{
 		if (rpb->rpb_flags & rpb_not_packed)
 		{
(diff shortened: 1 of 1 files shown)


hvlad (approved):

dyemanov (approved):

ilya071294:
The first commit is only for FB5 but the second one is actual for FB3/4. Should I backport it?

dyemanov:
I don't mind about backporting.

hvlad:
> I don't mind about backporting. @hvlad, what do you think ?

Agree

PR #7663, opened 2023-07-06
Print warnings occurred during commit

dmitry-starodubov:


src/isql/isql.epp
```diff
@@ -244,6 +244,7 @@ namespace IcuUtil
 	}
 }
 
+void ISQL_warning(Firebird::CheckStatusWrapper* st);
```
aafemt, line 247: It would be better to fix declaration in isql_proto.h.

dmitry-starodubov: Indeed. Didn't see it there.

AlexPeshkoff: @dmitry-starodubov Will you apply changes, suggested by DS, or I can do it myself?

dmitry-starodubov: Done

PR #7671, opened 2023-07-09
Cache sort blocks by the sort owners to avoid contention in shared cache of sorts blocks.

hvlad:
If request performs many sorts while executing, it often uses shared cache of sort blocks (maintained at dbb-level).
When such requests run concurrently, its creates a big load on shared cache and wait for each other significant time, reducing common performance.

This problem present at SS architecture only.

The offered solution is to avoid return of used sort block into shared common cache until requiest finished its execution.

### DIFF
--- a/src/jrd/sort.cpp
+++ b/src/jrd/sort.cpp
@@ -2195,7 +2195,7 @@ void SortOwner::unlinkAll()
 
 	if (buffers.hasData())
 	{
-		// Here we cache blocks to be reused later
+		// Move cached buffers to the database level cache to be reused later by other attachments
 
 		const size_t MAX_CACHED_SORT_BUFFERS = 8; // 1MB
 
--- a/src/jrd/sort.cpp
+++ b/src/jrd/sort.cpp
@@ -68,12 +68,6 @@ const USHORT MAX_MERGE_LEVEL	= 2;
 using namespace Jrd;
 using namespace Firebird;
 
-void SortOwner::unlinkAll()
-{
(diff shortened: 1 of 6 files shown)


src/jrd/sort.h
```diff
@@ -426,9 +426,14 @@ class SortOwner
 		return pool;
 	}
 
+	UCHAR* allocateBuffer();
+	void releaseBuffer(UCHAR*);
+
 private:
 	MemoryPool& pool;
+	Database* dbb;
```
dyemanov, line 434: `Database* const dbb`, please.

src/jrd/sort.cpp
```diff
@@ -2181,6 +2163,53 @@ void Sort::sortRunsBySeek(int n)
...
+
+
+void SortOwner::unlinkAll()
+{
+	while (sorts.getCount())
+		delete sorts.pop();
+
+	if (buffers.hasData())
+	{
+		// Here we cache blocks to be reused later
```
dyemanov, line 2198: Maybe the comment should be changed to something like: "Here we move the locally cached blocks to the database level cache, to be reused by other connections"

dyemanov (approved):
Approved with a couple of minor comments.

src/jrd/sort.h
hvlad: Done

src/jrd/sort.cpp
hvlad: Done

PR #7673, opened 2023-07-11
Make async replication reliable on Linux CS (replica side)

### LINKED ISSUE #7387: Unreliable replication behaviour in Linux Classic
Replication always stops on the same segment.
The only error in the log is that the process has stopped.
Every restart ends the same.
OS: CentOS 8 Stream.
FB mode: classic.
```
SRV-01-SI-4-REP (replica) Wed Nov  2 10:56:02 2022
        Database: /db-zpue/db/si.fdb
        VERBOSE: Segment 743 (34832 bytes) is replicated in 0 ms, preserving the file due to 1 active transaction(s) (oldest: 204980332 in segment 731)

SRV-01-SI-4-REP (replica) Wed Nov  2 10:56:02 2022
        Database: /db-zpue/db/si.fdb
        VERBOSE: Segment 744 (47351 bytes) is replicated in 0 ms, preserving the file due to 1 active transaction(s) (oldest: 204980332 in segment 731)

SRV-01-SI-4-REP (replica) Wed Nov  2 10:56:02 2022
        Database: /db-zpue/db/si.fdb
        VERBOSE: Segment 745 (53729 bytes) is replicated in 0 ms, preserving the file due to 1 active transaction(s) (oldest: 204980332 in segment 731)

SRV-01-SI-4-REP (replica) Wed Nov  2 10:56:24 2022
        Database: /db-zpue/db/si.fdb
        ERROR: connection shutdown
        Engine is shutdown.
        At segment 746, offset 22884

SRV-01-SI-4-REP (replica) Wed Nov  2 10:56:24 2022
        Database: /db-zpue/db/si.fdb
        VERBOSE: Suspending for 60 seconds

SRV-01-SI-4-REP (replica) Wed Nov  2 10:57:01 2022
        Database: /db-zpue/db/si.fdb
        VERBOSE: Started replication thread

SRV-01-SI-4-REP (replica) Wed Nov  2 10:57:01 2022
        Database: /db-zpue/db/si.fdb
        VERBOSE: Added 8188 segment(s) to the processing queue

```

dyemanov:
The initial implementation was buggy as it allowed to replicate journals by worker processes which appear/disappear sporadically. Replication should rather be performed by a dedicated process. This fixes #7387.

### DIFF
--- a/src/remote/server/ReplServer.cpp
+++ b/src/remote/server/ReplServer.cpp
@@ -75,8 +75,24 @@ namespace
 	const USHORT CTL_VERSION1 = 1;
 	const USHORT CTL_CURRENT_VERSION = CTL_VERSION1;
 
-	volatile bool* shutdownPtr = NULL;
+	volatile bool shutdownFlag = false;
 	AtomicCounter activeThreads;
+	Semaphore shutdownSemaphore;
+
+	int shutdownHandler(const int, const int, void*)
+	{
+		if (!shutdownFlag && activeThreads.value())
+		{
+			shutdownFlag = true;
+			shutdownSemaphore.release(activeThreads.value() + 1);
+
+			do {
+				Thread::sleep(10);
+			} while (activeThreads.value());
+		}
+
+		return 0;
+	}
 
 	struct ActiveTransaction
 	{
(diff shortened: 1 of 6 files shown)


src/remote/server/ReplServer.cpp
```diff
@@ -75,8 +75,24 @@ namespace
 	const USHORT CTL_VERSION1 = 1;
 	const USHORT CTL_CURRENT_VERSION = CTL_VERSION1;
 
-	volatile bool* shutdownPtr = NULL;
+	volatile bool shutdownFlag = false;
```
asfernandes, line 78: Not the single problematic place we have, but actually this is not correct.
Solution with `std::call_once` would be safe.

AlexPeshkoff: @asfernandes  In this particular case - no danger. shutdownHandler() may be called only from fb_shutdown(), which is reliably synchronized internally, i.e. 2 calls in parallel never happens.

AlexPeshkoff (approved):

PR #7679, opened 2023-07-16
Fix invalid conversion in src/common/unicode_util.cpp

th1722:
I found this error when I built Firebird as part of LibreOffice.

[...]
> workdir/UnpackedTarball/firebird/src/common/unicode_util.cpp:1381:48:
> error: invalid conversion from 'UChar* {aka char16_t*}' to
> 'const short unsigned int*' [-fpermissive]
>      const Array<USHORT> str(strChars, prefixLen);
>                                                 ^
[...]

Use the reinterpret_cast operator to avoid this error.

### DIFF
--- a/src/common/unicode_util.cpp
+++ b/src/common/unicode_util.cpp
@@ -1373,7 +1373,7 @@ UnicodeUtil::Utf16Collation* UnicodeUtil::Utf16Collation::create(
 
 			for (int prefixLen = 1; prefixLen < len; ++prefixLen)
 			{
-				const Array<USHORT> str(strChars, prefixLen);
+				const Array<USHORT> str(reinterpret_cast<USHORT*>(strChars), prefixLen);
 				SortKeyArray* keySet = obj->contractionsPrefix.get(str);
 
 				if (!keySet)


aafemt:
UChar declared as char16_t? It is not what Firebird sources could expect.

th1722:
> UChar declared as char16_t? It is not what Firebird sources could expect.

No, UChar is declared as unsigned char in LibreOffice.

aafemt:
Quoted error text has different opinion.

AlexPeshkoff:
On 7/17/23 05:55, th1722 wrote:
>
>     UChar declared as char16_t? It is not what Firebird sources could
>     expect.
>
> No, UChar is declared as unsigned char in LibreOffice.
>

Do not know for LibreOffice, but thi particular UChar is ICU-specific. 
***@***.*** - do not mix with UCHAR in firebird)
And appears it's expected to be 16-bit entity (something like unsigned 
short). No idea how is char16_t defined in your case, may be you've 
tried to change ICU version for the build?

PS.Sorry for late answer, was at vacations.

PR #7693, opened 2023-08-02
Improvement #7692 : Make trace config parser resolve symlinks in database file path in trace configuration

### LINKED ISSUE #7692: Make trace config parser resolve symlinks in database file path in trace configuration
Hello.
Currently, if a database file is available in one location on Linux and this location is available via the real path and via Linux symbolic link, fbtrace is unable to get the data when using link. I get only "Trace session started" and that's all.
Main engine can connect without problems in both ways.
Best regards,
Tomasz.

hvlad:


AlexPeshkoff (approved):

src/utilities/ntrace/TraceConfiguration.cpp
```diff
@@ -129,8 +129,8 @@ void TraceCfgReader::readConfig()
 			PathName expandedName;
 
 			if (m_databaseName == noQuotePattern ||
-				expandDatabaseName(noQuotePattern, expandedName, nullptr) &&
-				m_databaseName == expandedName)
+				(expandDatabaseName(noQuotePattern, expandedName, nullptr),
+				m_databaseName == expandedName) )
```
aafemt, line 133: Why would you like to compare m_databaseName with garbage in expandedName if expandDatabaseName() failed?

AlexPeshkoff: "false" returned by expandDatabaseName does not mean that we got garbage in expandedName. It just reports that alias was not found in databases.conf, rarely that info is (or 'was' - not sure for today) needed to the caller.

AlexPeshkoff: expandedName always contains the most correct name of DB file

PR #7694, opened 2023-08-03
Fix false positives of "missing entries for record X" error during index validation when a deleted record version is committed and has a backversion

ilya071294:
The existence of an index entry is not required for such version chain because it is all garbage since a transaction is committed. To reproduce the false positive, the server should be stopped just after the chain is garbage-collected and an index page is written to disk.

### DIFF
--- a/src/jrd/validation.cpp
+++ b/src/jrd/validation.cpp
@@ -1790,17 +1790,22 @@ Validation::RTN Validation::walk_data_page(jrd_rel* relation, ULONG page_number,
 				// is a backversion then at least one of the record versions is
 				// committed. If there's no backversion then check transaction
 				// state of the lone primary record version. Unless it is already deleted.
+				// Note, if the primary record version is deleted and committed, the
+				// existence of an index entry is not required for such version chain
+				// because it is all garbage.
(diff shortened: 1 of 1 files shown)


hvlad (approved):

src/jrd/validation.cpp
```diff
@@ -1790,17 +1790,22 @@ Validation::RTN Validation::walk_data_page(jrd_rel* relation, ULONG page_number,
...
+				// Note, if the primary record version is deleted and committed, the
+				// existence of an index entry is not required for such version chain
+				// because it is all garbage.
+				const bool deleted_flag = header->rhd_flags & rhd_deleted;
 
-				if (header->rhd_b_page)
+				if (header->rhd_b_page && !deleted_flag)
 					RBM_SET(pool, &vdr_rel_records, number.getValue());
-				else if ((header->rhd_flags & rhd_deleted) == 0)
+				else if (header->rhd_b_page || !deleted_flag)
```
aafemt, line 1800: Both ifs contain condition `header->rhd_b_page`. Cannot they be replaced with single `if`?

ilya071294: I think I tried it before and it led to unnecessary code duplication.

PR #7695, opened 2023-08-03
Fix false positives of "missing entries for record X" error during index validation

ilya071294:
Cases of false positives:
1. When a deleted record version is committed and has a backversion. The existence of an index entry is not required for such version chain because it is all garbage since a transaction is committed.
2. When a deleted record version is not committed and has a backversion which is also not committed. This can happen when a record is inserted and deleted in the same transaction.


aafemt:
Case 2 should not ever happen. update_in_place() and delete_in_place() make sure that there is no uncommitted backversions.

PR #7701, opened 2023-08-06
Fix #7700 - SKIP LOCKED returns conflict error or wait when config ReadConsistency = 0

### LINKED ISSUE #7700: SKIP LOCKED returns conflict error or wait when config ReadConsistency = 0
Test case:

**isql-1:**
```
recreate table test(id int primary key, f01 int);
insert into test(id, f01) select row_number()over(), 0 from rdb$types rows 10;
commit;

update test set f01 = 1 where id in (1,5,9);
SQL> 
```

**isql-2:**

```
SQL> rollback; set transaction read committed record_version NO wait; select id from test with lock skip locked;

          ID 
============ 
           2 
           3 
           4 
Statement failed, SQLSTATE = 40001
deadlock
-read conflicts with concurrent update
-concurrent transaction number is 136
-- Why error here ?
```

```
SQL> rollback; set transaction read committed record_version wait; select id from test with lock skip locked;

          ID 
============ 
           2 
           3 
           4 
-- Why wait here?
```

asfernandes:


src/jrd/vio_proto.h
```diff
@@ -44,6 +44,13 @@ namespace Jrd
 		DPM_next_pointer_page	// data pages from one pointer page
 	};
 
+	enum class RecordLock
+	{
+		DONT,
```
dyemanov, line 49: A bit of nitpicking, but perhaps `NONE` would be better?

dyemanov (approved):

PR #7702, opened 2023-08-08
Improvement #5577 : embedded: after fb_shutdown some files stay locked [CORE5299] 

hvlad:


asfernandes:
This would be problematic for applications that uses ICU and FB-embedded.
Firebird may call cleanup of a library it didn't loaded.
ICU is used by many applications.
I think this should not be done.

hvlad:
> This would be problematic for applications that uses ICU and FB-embedded. Firebird may call cleanup of a library it didn't loaded. 

Are you means the case when app already loaded ICU before engine does ?
Here I agree, `u_cleanup()` could be dangerous for the app.

We might "just unload" ICU libraries, not calling `u_cleanup()`. 
This will unload ICU libraries if there are no more references only and should be safe.
But all data files loaded by ICU (.dat and *.res) will remains in memory.

Maybe we could introduce some way to control this (to call `u_cleanup()` or not) ?

hvlad:
One more question - how to handle the case when ICU is loaded multiply times by Firebird itself.
It could be loaded by the engine, the trace and even by the fbclient, AFAIU.

asfernandes:
> Are you means the case when app already loaded ICU before engine does ?

Yes, but the inverse may also be possible.

I see no safe way to have `u_cleanup()` called.

And if you unload without calling it, then there will be multiple memory leaks each time this is done.

Issue #4915, opened 2014-11-09
Improve the index scan to match multiple values at once [CORE4600]


firebird-automations:
Submitted by: @dyemanov

Votes: 1

The existing equality index scan deals with a single value and returns a bitmap matching that value\. It could be extended to match several values with one index scan and return the combined bitmap for all those values\. This allows to replace multiple index scans with a single scan in some cases\. I can see such an improvement to be used e\.g\. by the block nested loops or the IN \(<list\>\) predicate\.

I suppose this should require the list of values to be sorted, but this is up to the implementation whether the storting will be performed by the caller or by the index scan code itself\.

dyemanov:
Implemented in Firebird 5.0, currently used by IN <list> predicate only.

PR #7707, opened 2023-08-10
Better processing and optimization if IN <list> predicates

dyemanov:
This PR implements native IN <list> processing rather than converting it into a binary tree of ORed predicates. Details are described below.

1) Processing is now linear rather than recursive, thus no runtime stack limitations. The artificial limit of 1500 items is gone. The current implementation has a hard-limit of 64K items just because BLR needs some limit and because this value looks OK from the sanity POV.

2) Lists that are known to be constant are pre-evaluated as invariants and cached as a binary search tree, thus making comparisons faster if condition needs to be checked for many rows or if the value list is long enough.

3) If the list is very long or if the IN predicate is not selective, the index scan supports searching groups using the sibling pointer (i.e. horizontally) rather than searching every group from root (i.e. vertically). This corresponds to #4915.

Performance results in TPC-R queries containing IN <list> predicates:

Q12: ~ 2x improvement (list scan is used instead of full-scan with bitmap)
Q16: no difference (the plan is the same)
Q19: ~ 10x improvement (list scan is used instead of many range scans)
Q22: ~ 2x improvement (the plan is the same, but expression evaluation is more optimal)

The patch (in its original incarnation, based on v3 and v4 codebases) was tested by our customer in production since 2022 and looked stable, although they didn't see such a noticeable performance boost. But they needed longish lists (> 1500 items) more than a better performance ;-) When porting this into master I improved some things I didn't like originally.

### DIFF
--- a/src/dsql/BoolNodes.h
+++ b/src/dsql/BoolNodes.h
@@ -166,13 +166,13 @@ public:
 	virtual bool dsqlMatch(DsqlCompilerScratch* dsqlScratch, const ExprNode* other, bool ignoreMapCast) const;
 	virtual bool sameAs(const ExprNode* other, bool ignoreStreams) const;
 	virtual BoolExprNode* pass1(thread_db* tdbb, CompilerScratch* csb);
-	virtual void pass2Boolean1(thread_db* tdbb, CompilerScratch* csb);
-	virtual void pass2Boolean2(thread_db* tdbb, CompilerScratch* csb);
+	void pass2Boolean(thread_db* tdbb, CompilerScratch* csb, std::function<void ()> process);
 	virtual bool execute(thread_db* tdbb, Request* request) const;
 
 public:
 	NestConst<ValueExprNode> arg;
 	NestConst<ValueListNode> list;
+	NestConst<LookupValueList> lookup;
 };
 
 
--- a/src/dsql/BoolNodes.h
+++ b/src/dsql/BoolNodes.h
@@ -150,7 +150,7 @@ public:
 
 	static DmlNode* parse(thread_db* tdbb, MemoryPool& pool, CompilerScratch* csb, const UCHAR blrOp);
 
-	virtual void getChildren(NodeRefsHolder& holder, bool dsql) const
+	void getChildren(NodeRefsHolder& holder, bool dsql) const override
 	{
 		BoolExprNode::getChildren(holder, dsql);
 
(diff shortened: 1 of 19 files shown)


src/dsql/BoolNodes.h
```diff
@@ -138,6 +141,41 @@ class ComparativeBoolNode final : public TypedNode<BoolExprNode, ExprNode::TYPE_
 };
 
 
+class InListBoolNode : public TypedNode<BoolExprNode, ExprNode::TYPE_IN_LIST_BOOL>
+{
+	const static UCHAR blrOp = blr_in_list;
```
asfernandes, line 146: Why make a constant instead of use the blr directly in gen function?

src/dsql/BoolNodes.h
```diff
@@ -138,6 +141,41 @@ class ComparativeBoolNode final : public TypedNode<BoolExprNode, ExprNode::TYPE_
...
+
+public:
+	InListBoolNode(MemoryPool& pool, ValueExprNode* aArg = nullptr, ValueListNode* aList = nullptr);
+
+	static DmlNode* parse(thread_db* tdbb, MemoryPool& pool, CompilerScratch* csb, const UCHAR blrOp);
+
+	virtual void getChildren(NodeRefsHolder& holder, bool dsql) const
```
asfernandes, line 153: As it is new code, worth change the `virtual` redeclarations to use `override` instead.

src/dsql/ExprNodes.cpp
```diff
@@ -194,6 +192,55 @@ static void setParameterInfo(dsql_par* parameter, const dsql_ctx* context);
...
+}
+
+TriState LookupValueList::find(thread_db* tdbb, Request* request, const ValueExprNode* value, const dsc* desc) const
+{
+	const auto impure = request->getImpure<impure_value>(m_impureOffset);
+	auto sortedList = impure->vlu_misc.vlu_sortedList;
+
+	if (!(impure->vlu_flags & VLU_computed))
+	{
+		delete impure->vlu_misc.vlu_sortedList;
```
asfernandes, line 214: It is good to initialize the field to `nullptr` just after the delete.
Because if the new constructor fail, it could be left with garbage.
This is done after `delete impure->vlu_misc.vlu_invariant;` in `SubstringSimilarNode::execute`.

src/dsql/ExprNodes.cpp
```diff
@@ -194,6 +192,55 @@ static void setParameterInfo(dsql_par* parameter, const dsql_ctx* context);
 //--------------------
 
 
+int SortValueItem::compare(const dsc* desc1, const dsc* desc2)
+{
+	return MOV_compare(JRD_get_thread_data(), desc1, desc2);
+}
+
+LookupValueList::LookupValueList(MemoryPool& pool, const ValueListNode* values, ULONG impure)
```
asfernandes, line 200: Wouldn't be possible to avoid `const_casts` replacing some constructors to not use pointer to consts?

src/dsql/BoolNodes.h
dyemanov: It's used in two methods (also in `internalPrint`) and I don't like duplicating things ;-)

src/dsql/BoolNodes.h
dyemanov: Agreed.

src/dsql/ExprNodes.cpp
dyemanov: Agreed.

src/dsql/ExprNodes.cpp
dyemanov: I will try.

PR #7718, opened 2023-08-23
Allow create database with a different owner

aafemt:


src/include/gen/Firebird.pas
```diff
@@ -15196,6 +15197,7 @@ function ITraceProcedureImpl_getPerfDispatcher(this: ITraceProcedure): Performan
 
 function ITraceProcedureImpl_getStmtIDDispatcher(this: ITraceProcedure): Int64; cdecl;
 begin
+	Result := 0;
```
AlexPeshkoff, line 15200: Interesting and sooner of all useful fix but:
- unrelated with particular PR
- should be accomplished with changes in cloop sources

src/jrd/jrd.cpp
```diff
@@ -1342,7 +1343,7 @@ static void		start_transaction(thread_db* tdbb, bool transliterate, jrd_tra** tr
 static void		rollback(thread_db*, jrd_tra*, const bool);
 static void		purge_attachment(thread_db* tdbb, StableAttachmentPart* sAtt, unsigned flags = 0);
 static void		getUserInfo(UserId&, const DatabaseOptions&, const char*,
-	const RefPtr<const Config>*, bool, Mapping& mapping, bool);
```
AlexPeshkoff, line 1345: Also not related with this PR

aafemt: It is. The check for privilege was moved into createDatabase() to allow impersonalization making the parameter obsolete and unused.

AlexPeshkoff (changes requested):
Feature is for sure useful, and I finally agree with removing parameter from getUserInfo() as related with this PR - but please remove unrelated changes from .pas file.

hvlad:
It is always good idea to explain why you consider proposed feature useful and how it could/should be used.

aafemt:
> why you consider proposed feature useful and how it could/should be used.

Among other purposes it is planned to be used as a base for other feature: gbak preserving DB owner on restore. It would bring consistency between nbackup preserving owner and gbak rewriting it.

AlexPeshkoff:
Agreed, this change is related.

AlexPeshkoff:
> Ok, but should I really edit autogenerated file by hand after every build?

Certainly not. I will fix file in master branch and after it pls fix PR.

aafemt:
That's why I insisted that this file must not be in the repository years ago...

AlexPeshkoff:
@aafemt Look like you've created PR against rather old copy of this file. If you make fresh clone of master branch and run something like:
git log -L 14927,14935:Firebird.pas
you will see why I think so. 
I do not pretend to understand how exactly you've achieved this :)

AlexPeshkoff:
Last fix was done by Vlad and I doubt he used linux build for that :-)
Well , I suggest to treat this incident as resolved :)

What about PR as a whole - let's wait for DE cause list of features to be added to FB5 was already discussed and currently is closed. Personally I see no problems adding it - your code appears quite clear here, but let's wait what others say.

aafemt:
> Personally I see no problems adding it - your code appears quite clear here, but let's wait what others say.

I don't expect this PR to be accepted into v5. V6 is fine. I just need to reserve DPB item number.

sim1984:
I did not see support for specifying the owner of the database in the create database statement. It is clear that this pseudo-statement in reality calls the api to create a database, but it also needs to be done. If it is, then why is it not reflected in the documentation?

AlexPeshkoff:
On 8/23/23 18:15, Dimitry Sibiryakov wrote:
>
>     Last fix was done by Vlad and I doubt he used linux build for that :-)
>
> But you are looking at different line. Here is link to current master 
> version: 
> https://github.com/FirebirdSQL/firebird/blob/5432699f284dad21443eff5ed86b9f717614fa6f/src/include/gen/Firebird.pas#L15197
>

Wrong link. Coorect one is 
https://github.com/FirebirdSQL/firebird/blob/master/src/include/gen/Firebird.pas#L14927
Your link is not to master. Instead 
5432699f284dad21443eff5ed86b9f717614fa6f 
<https://github.com/FirebirdSQL/firebird/blob/5432699f284dad21443eff5ed86b9f717614fa6f/src/include/gen/Firebird.pas#L15197> 
is used.

aafemt:
> Wrong link. Coorect one is https://github.com/FirebirdSQL/firebird/blob/master/src/include/gen/Firebird.pas#L14927 Your link is not to master.

It is to the last commit of master but ok, here is the really right link:
https://github.com/FirebirdSQL/firebird/blob/master/src/include/gen/Firebird.pas#L15197

AlexPeshkoff:
On 8/23/23 18:49, Dimitry Sibiryakov wrote:
>
>     Wrong link. Coorect one is
>     https://github.com/FirebirdSQL/firebird/blob/master/src/include/gen/Firebird.pas#L14927
>     Your link is not to master.
>
> It is to the last commit of master but ok, here is the really right link:
> https://github.com/FirebirdSQL/firebird/blob/master/src/include/gen/Firebird.pas#L15197
>

You are right. A bug though not in cloop but in posix makefile. Will fix.

src/jrd/DbCreators.h
```diff
@@ -36,7 +36,14 @@
 
 namespace Jrd {
 
-bool checkCreateDatabaseGrant(const Firebird::MetaString& userName, const Firebird::MetaString& trustedRole,
+enum class CreateGrant
+{
+	none,		// This user cannot create databases
```
asfernandes, line 41: Our `enum class` members uses ALL_UPPER_CASE convention.

src/jrd/jrd.cpp
```diff
@@ -2852,7 +2853,44 @@ JAttachment* JProvider::createDatabase(CheckStatusWrapper* user_status, const ch
...
+			// Check user's power level
+			CreateGrant powerLevel = CreateGrant::assumed; // By default it is a boot build or embedded mode where everything is allowed
+			if (options.dpb_auth_block.hasData())
+			{
+				powerLevel = checkCreateDatabaseGrant(userId.getUserName(), userId.getTrustedRole(), userId.getSqlRole(), config->getSecurityDatabase());
+			}
+			switch(powerLevel)
+			{
+			case CreateGrant::none:
+				{
```
asfernandes, line 2867: Do not use braces in cases if you are not declaring variables.

src/jrd/jrd.cpp
```diff
@@ -2852,7 +2853,44 @@ JAttachment* JProvider::createDatabase(CheckStatusWrapper* user_status, const ch
...
-			getUserInfo(userId, options, filename, &config, true, mapping, false);
+			getUserInfo(userId, options, filename, &config, mapping, false);
+
+			// Check user's power level
+			CreateGrant powerLevel = CreateGrant::assumed; // By default it is a boot build or embedded mode where everything is allowed
+			if (options.dpb_auth_block.hasData())
+			{
+				powerLevel = checkCreateDatabaseGrant(userId.getUserName(), userId.getTrustedRole(), userId.getSqlRole(), config->getSecurityDatabase());
+			}
+			switch(powerLevel)
```
asfernandes, line 2864:
```suggestion

			switch (powerLevel)
```

asfernandes (changes requested):

src/jrd/DbCreators.h
aafemt: Isn't it only for old code which was converted from macros? All-uppercase identifiers are ugly.

AlexPeshkoff:
On 8/24/23 14:45, Dimitry Sibiryakov wrote:
>
>
>
> In src/jrd/DbCreators.h 
> <https://github.com/FirebirdSQL/firebird/pull/7718#discussion_r1304203448>:
>
> > @@ -36,7 +36,14 @@
>   
>   namespace Jrd {
>   
> -bool checkCreateDatabaseGrant(const Firebird::MetaString& userName, const Firebird::MetaString& trustedRole,
> +enum class CreateGrant
> +{
> +	none,		// This user cannot create databases
>
> Isn't it only for old code which was converted from macros?
>

No.

> All-uppercase identifiers are ugly.
>

When they represent constants - may be not too ugly.

aafemt:
> When they represent constants - may be not too ugly.

Ok, though it is a scoped enum and its members cannot be mistaken for anything else.

aafemt:
> Wouldn't it be better to make the options available `grant DATABASE_OWNER to USER_NAME`?

Changing of existing database is problematic. Set owner during creation is simpler.

dyemanov (approved):

pavel-zotov:
Is it possible to implement similar for ALTER database ? I mean:
```
alter database set owner to <some_other_user>;
```

sim1984:
Implementing changing the owner of just the database will not solve the problems. By introducing this feature, you will also have to provide the ability to change owners for other database objects. And also think about recursively changing the owner of all database objects.

```
ALTER <object_type> <object_name> SET OWNER TO <user>

<object_type> ::=
  TABLE | VIEW | PROCEDURE | FUNCTION | TRIGGER | PACKAGE |
  DOMAIN | EXCEPTION | FILTER | ROLE
```

And special case for database:

```
ALTER DATABASE SET OWNER TO <user> [WITH RECURSIVE]
```

pavel-zotov:
> And also think about recursively changing the owner of all database objects.

Yes, number of these objects can be huge. But not infinite. 
And not all of objects must be affected but only those who were created by previous owner.
So, several minutes usually must be enough to complete this job :-)

PS. Of course, i speak about 'service time' interval when SYSDBA has exclusive access to DB

aafemt:
> Is it possible to implement similar for ALTER database ?

In theory - yes. On practice it is much harder task than impersonation during single call.

PR #7737, opened 2023-09-05
Fix cases where the precedence relationship between a record page and a blob page is not set

ilya071294:
The precedence is not set when:
1. A blob is deleted during the intermediate garbage collection (FB4+).
2. A blob is created for a modified record.

Both cases can lead to a situation where a record points to a blob that doesn't exist or that belongs to another record, which is quite dangerous.

### DIFF
--- a/src/jrd/blb.cpp
+++ b/src/jrd/blb.cpp
@@ -448,6 +448,7 @@ void BLB_garbage_collect(thread_db* tdbb,
  **************************************/
 	SET_TDBB(tdbb);
 
+	fb_assert(prior_page > 0);
 	RecordBitmap bmGoing;
 	ULONG cntGoing = 0;
 
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -2485,25 +2485,26 @@ static void delete_version_chain(thread_db* tdbb, record_param* rpb, bool delete
 
 	ULONG prior_page = 0;
 
-	if (!delete_head)
+	// Note that the page number of the oldest version in the chain should
+	// be stored in rpb->rpb_page before exiting this function because
+	// VIO_intermediate_gc will use it as a prior page number.
+	while (rpb->rpb_b_page != 0 || delete_head)
 	{
-		prior_page = rpb->rpb_page;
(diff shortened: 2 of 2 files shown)


hvlad (approved):

pavel-zotov:
> The precedence is not set when:
...
> 2. A blob is created for a modified record.

Can it be reproduced in ISQL ? Is it reproducible when work in the single attachment ?
Perhaps, I did not understand properly doing following:

```
C:\FB\50sc>C:\FB\50sc\isql /:employee -z
ISQL Version: WI-T5.0.0.1182 Firebird 5.0 Beta 2
Server version:
WI-T5.0.0.1182 Firebird 5.0 Beta 2
WI-T5.0.0.1182 Firebird 5.0 Beta 2/tcp (Image-PC1)/P18:C
WI-T5.0.0.1182 Firebird 5.0 Beta 2/tcp (Image-PC1)/P18:C
Database: /:employee, User: SYSDBA
SQL> recreate table test(id int primary key, x int, b blob);
SQL> insert into test(id,x) values(1, 100);
SQL> commit;
SQL> update test set x = -x;
SQL> update test set b = (select list(uuid_to_char(gen_uuid())) from rdb$types rows 3);
SQL> commit;
SQL> set blob all;
SQL> set list on;
SQL> select * from test;

ID                              1
X                               -100
B                               8b:1e0
88DBFFFC-A5DC-422A-B1AB-1BEFD27A7529,7A67BB30-ECF2-4E8E-99CF-F6C7158F6B80,03C444C2-211A
-66242F88D5D1,98EADB17-F94D-4401-A587-57AC4467AC49,C77C21FE-6D45-498E-9167-22532239CAEA
DD6D-46E9-89CB-91E383F125EA,A50A1E36-1BF5-4127-B229-FC1C14AD8246,705D53F9-C0AF-4AB7-B0A
. . .
```

ilya071294:
> Can it be reproduced in ISQL ?

No. It's a bit hard to reproduce because the server should be abnormally terminated before a page with a new blob is written to the disk while a page with a modified record version is already written.

PR #7742, opened 2023-09-07
Fix #7715 - Alternative String Literals and multibyte (UTF8) alternatives

### LINKED ISSUE #7715: Alternative String Literals and multibyte (UTF8) alternatives
FB3 and FB4 don't support multibyte alternatives

Connection charset: **UTF8**

```sql
select  _win1251 q'Жмама, мама, что я буду делать?Ж' from rdb$database

select  _win1251 q'🙃мама, мама, что я буду делать?🙃' from rdb$database
```

---
Correct me, if I am not right:

Server expects that
1. **"мама, мама, что я буду делать?"** is the text with WIN1251 charset

2. **Ж** and 🙃 are the text with UTF8 charset

asfernandes:


dmitry-lipetsk:
Hello,

I tested this patch with FB4 (after the minimal correction of Parser constructor).

Works fine.

Please apply this patch to FB4 too.

Thanks.

PR #7748, opened 2023-09-13
Allow collation to be a part of data type

aafemt:
As discussed in https://groups.google.com/d/msgid/firebird-devel/0803eb29-46ef-41ff-a1c7-768ea69e7988%40lawinegevaar.nl


AlexPeshkoff:
I see nothing bad that we get closer to SQL standard.

AlexPeshkoff (approved):

dyemanov:
A few comments after looking at parse.y:

1) By SQL spec, COLLATE can be specified not only for specific character set, but also for `national_character_type`.
2) I see `charset_clause` also in `udf_data_type` and `array_type`, they're not covered in this patch.

Also, accordingly to the SQL spec, COLLATE cannot be specified for BINARY. It corresponds to your new grammar, but what about CHARACTER SET OCTETS being specified explicitly with some COLLATE? Is it going to be rejected due to collation mismatch after INTL tables lookup or maybe we should prohibit it with a dedicated error message?

doc/sql.extensions/README.ddl.txt
```diff
@@ -603,3 +603,8 @@ ALTER TABLE <name> ... [ {ENABLE | DISABLE} PUBLICATION ]
 
 Defines whether replication is enabled for the specified table.
 If not specified in the CREATE TABLE statement, the database-level default behaviour is applied.
+
+24) COLLATE clause can be used as a part of character data type as per SQL standard.
+(Dmitry Sibiryakov)
+
+If is used twice, an error is returned.
```
dyemanov, line 610: This does not correspond to changes below. For parameters you seem to allow both and use the latest one. What about local variables?

aafemt: This PR is intended for table fields only, I'm not sure if parameters and variables are covered. Should they?

aafemt:
array_type is not covered on purpose for two reasons:
1) arrays are not much supported in general;
2) it adds another parser conflict that is not good.

doc/sql.extensions/README.ddl.txt
dyemanov: You have added the code to `ParameterClause::ctor` with a comment // Explicit collation has precedence over data type collation

And parameters/variables are declared using:
`column_domain_or_non_array_type collate_clause ...`
where `column_domain_or_non_array_type` now includes a `collate_clause` itself after your changes...
`domain_or_non_array_type -> domain_or_non_array_type_name -> non_array_type -> simple_type`

dyemanov (changes requested):
Basically, I agree with the PR, but the questioned issues must be addressed before merging.

dyemanov:
> array_type is not covered on purpose for two reasons:
> 
>     1. arrays are not much supported in general;
>     2. it adds another parser conflict that is not good.

I don't mind.

doc/sql.extensions/README.ddl.txt
aafemt: This comment applies to cases when data type is derived from domain or column including collation. Old code silently reset collation to default if no explicit COLLATE was provided so I've added the condition and the comment.

aafemt:
> but what about CHARACTER SET OCTETS being specified explicitly with some COLLATE? Is it going to be rejected due to collation mismatch after INTL tables lookup or maybe we should prohibit it with a dedicated error message?

Charset OCTETS has only one collation so unless somebody created a custom collation for it (is it possible at all?) you'll get error "collation not found" trying to set collation to anything but default "OCTETS". I don't think that using of COLLATE should be prohibited for particular charset(s). At least for the sake of backward compatibility.

doc/sql.extensions/README.ddl.txt
aafemt: It is pity that setClause() is not available outside of parser. Can they be made static?

dyemanov:
> > but what about CHARACTER SET OCTETS being specified explicitly with some COLLATE? Is it going to be rejected due to collation mismatch after INTL tables lookup or maybe we should prohibit it with a dedicated error message?
> 
> Charset OCTETS has only one collation so unless somebody created a custom collation for it (is it possible at all?) you'll get error "collation not found" trying to set collation to anything but default "OCTETS". I don't think that using of COLLATE should be prohibited for particular charset(s). At least for the sake of backward compatibility.

OK, fine with me.

doc/sql.extensions/README.ddl.txt
dyemanov: > It is pity that setClause() is not available outside of parser. Can they be made static?

IMHO, they (along with `setClauseFlag` / `checkDuplicateClause` / `isDuplicateClause`) should be `static`, as they don't access the parser state. @asfernandes , do you agree?

dyemanov: I would question the need in `collate_clause` after `domain_or_non_array_type` and `column_domain_or_non_array_type`, as now it's also included inside these parser rules. The only regression would be that `NOT NULL` must be written after `COLLATE`, not before as now. Is it an acceptable breakage, given the low probability that both co-exist in input/output parameters / return values in real-world scripts? And there will be no need to check for duplicated definitions then.

src/dsql/parse.y
```diff
@@ -4737,14 +4737,16 @@ array_range
 %type <legacyField> simple_type
 simple_type
 	: non_charset_simple_type
-	| character_type charset_clause
+	| character_type charset_clause collate_clause
 		{
```
dyemanov, line 4767: I'm not sure whether it works in our codebase, but accordingly to the SQL specification `collate_clause` must also be supported for text blobs.

doc/sql.extensions/README.ddl.txt
asfernandes: If nodes are going to call `Parser::setClause`, I don't agree. It's layer violation.
In this case I would move them to `Node` class.

src/dsql/parse.y
aafemt: Yes, but Firebird has no standard character blobs (CLOB or CHARACTER LARGE OBJECT), only binary blobs so this part is not for this PR.

aafemt:
> 1. By SQL spec, `COLLATE` can be specified not only for specific character set, but also for `national_character_type`.

Added.

doc/sql.extensions/README.ddl.txt
aafemt: Collation can be removed from ParameterClause constructor parameters and assigned later. Is it ok?

dyemanov (approved):

asfernandes:
@aafemt this change introduced a problem making this query work:
```
select cast('x' as varchar(10) character set utf8 collate wrong) from rdb$database;
```

aafemt:
If something that never worked started to work I wouldn't call it "a problem".
If `cast` ignores collation in target data type definition it is the problem with `cast` and not data type definition. Thus I extract your comment to a separate issue.

asfernandes:
> If something that never worked started to work I wouldn't call it "a problem".

Not true. It worked perfectly well, giving an error in a not supported contruction.

> If `cast` ignores collation in target data type definition it is the problem with `cast` and not data type definition. Thus I extract your comment to a separate issue.

No. It's a problem introduced in this PR.

asfernandes:
> Now the construction is supported and the query doesn't produce "not supported" error. It is logical and usual behavior of any new functionality.

I've no idea of what are you talking, really. Makes no sense for me.

Is it a way to say that you're not going to fix regression #8249 introduced by this your change?

aafemt:
> Is it a way to say that you're not going to fix regression #8249 introduced by this your change?

It is a way to tell you "do not call this regression". Regression is a case when working documented functionality doesn't work anymore.

#8249 is a separate bug that should be fixed by me or anyone else like any other bug. You assigned it to me - fine. Now there is nothing to do but to sit and wait.

PR #7758, opened 2023-09-22
Added the ability to change deterministic option

### LINKED ISSUE #7427: Alter function DETERMINISTIC
Please add extension to current syntax of `Alter Function` to be able to change `DETERMINISTIC` only without specification of whole function body as e.g.:

`ALTER FUNCTION funcname {SET | DROP} DETERMINISTIC `

Zhdanov0:
Added the ability to change deterministic opt without specifiying the entire body of the function. Extended "ALTER FUNCTION" syntax:
```
ALTER FUNCTION name DETERMINISTIC;
ALTER FUNCTION name NOT DETERMINISTIC;
```

### DIFF
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -1726,7 +1726,7 @@ void CreateAlterFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsql
 	AutoSavePoint savePoint(tdbb, transaction);
 	bool altered = false;
 
-	bool alterIndividualParameters = !(body || external);
+	const bool alterIndividualParameters = !(body || external);
 
 	// first pass
 	if (alter)
@@ -1738,19 +1738,16 @@ void CreateAlterFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsql
 			else
 				status_exception::raise(Arg::Gds(isc_dyn_func_not_found) << Arg::Str(name));
 		}
+		else if (executeAlter(tdbb, dsqlScratch, transaction, false, true))
+		{
+			altered = true;
+		}
 		else
 		{
-			if (executeAlter(tdbb, dsqlScratch, transaction, false, true))
-			{
-				altered = true;
-			}
+			if (create)	// create or alter
+				executeCreate(tdbb, dsqlScratch, transaction);
 			else
-			{
-				if (create)	// create or alter
-					executeCreate(tdbb, dsqlScratch, transaction);
-				else
-					status_exception::raise(Arg::Gds(isc_dyn_func_not_found) << Arg::Str(name));
-			}
+				status_exception::raise(Arg::Gds(isc_dyn_func_not_found) << Arg::Str(name));
 		}
 	}
 	else
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -1726,19 +1726,31 @@ void CreateAlterFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsql
 	AutoSavePoint savePoint(tdbb, transaction);
 	bool altered = false;
 
+	bool alterIndividualParameters = !(body || external);
+
 	// first pass
 	if (alter)
 	{
-		if (executeAlter(tdbb, dsqlScratch, transaction, false, true))
+		if (alterIndividualParameters)
 		{
-			altered = true;
+			if (executeAlterIndividualParameters(tdbb, dsqlScratch, transaction, false, true))
+				altered = true;
+			else
+				status_exception::raise(Arg::Gds(isc_dyn_func_not_found) << Arg::Str(name));
 		}
 		else
 		{
-			if (create)	// create or alter
-				executeCreate(tdbb, dsqlScratch, transaction);
+			if (executeAlter(tdbb, dsqlScratch, transaction, false, true))
+			{
+				altered = true;
+			}
 			else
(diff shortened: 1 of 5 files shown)


src/dsql/DdlNodes.h
```diff
@@ -466,6 +469,7 @@ class CreateAlterFunctionNode : public DdlNode
 	MetaName name;
 	bool create;
 	bool alter;
+	bool alterIndividualParameters;
```
aafemt, line 472: This variable is not needed, the node can determine full or partial update looking whether body is set or not.

src/dsql/parse.y
```diff
@@ -2903,6 +2925,13 @@ alter_function_clause
 			$$->alter = true;
 			$$->create = false;
 		}
+	| change_opt_function_clause
```
aafemt, line 2928: Instead of separate branch you could make whole AS part optional. As a bonus security attributes would become uneatable as well.

src/dsql/DdlNodes.h
Zhdanov0: okay

Zhdanov0: > This variable is not needed, the node can determine full or partial update looking whether body is set or not.

okay

src/dsql/DdlNodes.epp
```diff
@@ -2062,6 +2077,42 @@ bool CreateAlterFunctionNode::executeAlter(thread_db* tdbb, DsqlCompilerScratch*
...
+		}
+
+		if (!secondPass && runTriggers && package.isEmpty())
+		{
+			executeDdlTrigger(tdbb, dsqlScratch, transaction, DTW_BEFORE,
+				DDL_TRIGGER_ALTER_FUNCTION, name, NULL);
+		}
+
+		MODIFY FUN
+			FUN.RDB$DETERMINISTIC_FLAG = deterministic ? TRUE : FALSE;
```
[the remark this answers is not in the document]
Zhdanov0: > NULL flag should be reset as well. Just in case.

okay

src/dsql/parse.y
Zhdanov0: > Instead of separate branch you could make whole AS part optional. As a bonus security attributes would become uneatable as well.

You mean make psql-body and external-body optional, right? Then you will have to write:
alter function funcname returns type deterministic / not deterministic; 
Accordingly, the "return <type>" part should also be made optional?

aafemt: Why not? It would be standard deviation but quite acceptable, IMHO. But let's wait for others' opinion. It was just an idea, not request from me.

dyemanov: `function_clause` is shared between `CREATE` and `ALTER`, so without a separate branch it would be possible to write `CREATE FUNCTION <funcname> DETERMINISTIC`. I'm not sure this is what we want, even if handled internally with some dedicated error.

src/dsql/DdlNodes.h
dyemanov: Just check not only for `body` but also for `external`.

Zhdanov0: I removed the variable and added a flag calculation based on the `body` and `external` parameters

src/dsql/DdlNodes.epp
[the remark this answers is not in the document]
Zhdanov0: added NULL flag reset

sim1984:
In my opinion, the following syntax would be more correct:

```
ALTER FUNCTION name SET DETERMINISTIC;
ALTER FUNCTION name DROP DETERMINISTIC;
```

Thus, if the NAME field does not have `SET | DROP`, then the full `ALTER FUNCTION` syntax is required.

There is another similar task. Change `SQL SECURITY {INVOKER | DEFINER}`.

```
ALTER FUNCTION name SET SQL SECURITY {INVOKER|DEFINER};
ALTER FUNCTION name DROP SQL SECURITY;
```

dyemanov:
Yes, SET / DROP was suggested initially, but I didn't like it.

src/dsql/DdlNodes.epp
```diff
@@ -1726,27 +1726,44 @@ void CreateAlterFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsql
 	AutoSavePoint savePoint(tdbb, transaction);
 	bool altered = false;
 
+	bool alterIndividualParameters = !(body || external);
```
dyemanov, line 1729: `const bool`, please

src/dsql/DdlNodes.epp
```diff
@@ -1726,27 +1726,44 @@ void CreateAlterFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsql
...
+				altered = true;
+			}
 			else
-				status_exception::raise(Arg::Gds(isc_dyn_func_not_found) << Arg::Str(name));
+			{
+				if (create)	// create or alter
+					executeCreate(tdbb, dsqlScratch, transaction);
+				else
+					status_exception::raise(Arg::Gds(isc_dyn_func_not_found) << Arg::Str(name));
+			}
```
dyemanov, line 1753: I believe this should finally look as `if ... else if ... else` rather than having a nested `if ... else`.

dyemanov (approved):

asfernandes:
It's missing some doc in `doc/sql.extensions/README.ddl.txt`.

livius2:
I suppose that https://github.com/FirebirdSQL/firebird/issues/7427 can be linked here and closed as implemented.

PR #7763, opened 2023-09-26
This should fix QA tests core_4418 and core_6044

aafemt:
...but may break any other that check isql output for procedure and function parameters.

### DIFF
--- a/src/isql/extract.epp
+++ b/src/isql/extract.epp
@@ -459,8 +459,7 @@ int EXTRACT_list_table(const SCHAR* relation_name,
 
 				if (!FLD.RDB$CHARACTER_SET_ID.NULL)
 					char_set_id = FLD.RDB$CHARACTER_SET_ID;
-				if (char_set_id != default_char_set_id)
-					ISQL_get_character_sets(char_set_id, 0, true, false, false, true, char_sets);
+				ISQL_get_character_sets(char_set_id, collation, char_set_id != default_char_set_id, false, false, true, char_sets);
 				if (char_sets[0])
 					isqlGlob.prints(char_sets);
 				// CVC: Someone deleted the code that checks intchar when handling collations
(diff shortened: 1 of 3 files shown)


aafemt:
Field B. In your output it is as expected by the tests. In my comment it is shorter (though semantically the same).

AlexPeshkoff:
I.e. p/r restores legacy behavior with output of DB default charset - right? In that case what test can be broken by THIS p/r?

sim1984:
Maybe the tests should still be rewritten for the correct output, rather than trying to restore the passing of these tests?

dyemanov:
> Maybe the tests should still be rewritten for the correct output, rather than trying to restore the passing of these tests?

Maybe, we just need to agree what is a correct output ;-)

aafemt:
> I.e. p/r restores legacy behavior with output of DB default charset - right? In that case what test can be broken by THIS p/r?

It depends on how deeply you count "legacy behavior". This PR probably restores behavior of v4 but I'm not sure.

What I said that may be broken is following script:
`create procedure p(A varchar(20) collate UTF8) as begin end; show procedure p;`
In contrast to the table above it will show collation without charset for the parameter. It can be fixed too but not so simple.

dyemanov:
Personally, I prefer `isql -x` to generate as simplified script as possible. So omitting the character set which matches the database default character set looks like a good idea. But I'm not so sure about the `SHOW` command, where IMHO it makes sense to print the detailed definition, thus avoiding the need to run `SHOW DATABASE` to understand the real charset of the shown object(s).

AlexPeshkoff:
Agree with @dyemanov, just want to add that in a case we have to choose between simplified script and detailed SHOW information - I prefer the second, script getting a little longer will never hurt, not understood by user SHOW result is much worse.

aafemt:
I see the point. In this case you can simply merge this PR and see if any other QA test fail. If there will be any - I'll do more adjustments.

PR #7807, opened 2023-10-16
Added the ability to change sql security option without specifying the entire body of the function/procedure/package

### LINKED ISSUE #7744: Provide ability to run "ALTER SQL SECURITY DEFINER / INVOKER" without specifying further part of proc/func/package
Currently we can run:
```
alter trigger sql security <invoker | definer>
```
-- and not specify body of affected trigger.
But it is not so for other kind of PSQL units (SP, functions and package headers).
It will be useful if we have such ability also, i.e. without need to put into ALTER statement the further part of unit being changed (input/output parameters for SP/func; inner functions and procedures for packages).

Zhdanov0:
Added the ability to change sql security opt without specifiying the entire body of the function/procedure/package.
1) ALTER FUNCTION <name> [ {DETERMINISTIC | NOT DETERMINISTIC} ] [ SQL SECURITY {DEFINER | INVOKER} | DROP SQL SECURITY ]
Examples:
`ALTER FUNCTION name DETERMINISTIC SQL SECURITY DEFINER`
`ALTER FUNCTION name SQL SECURITY INVOKER`
`ALTER FUNCTION name NOT DETERMINISTIC DROP SQL SECURITY`
2) ALTER PROCEDURE <name> SQL SECURITY {DEFINER | INVOKER} | DROP SQL SECURITY
3) ALTER PACKAGE <name> SQL SECURITY {DEFINER | INVOKER} | DROP SQL SECURITY


Zhdanov0:
Added the ability to change sql security (definer / invoker / drop) to the existing syntax ALTER FUNCTION <name> { DETERMINISTIC | NOT DETERMINISTIC }.
So that you can write `ALTER FUNCTION funcname DETERMINISTIC SQL SECURITY DEFINER`

I did it by analogy with "alter_domain_ops"

The deterministic field type was changed from bool to TriState.
The third "null" state appears if you write: `ALTER FUNCTION funcname SQL SECURITY DEFINER`. Then, with a partial update, deterministic_flag is not set to false.
If you do not change it to TriState, then `ALTER FUNCTION funcname SQL SECURITY DEFINER`
will set deterministic_flag to false, because in createAlterFunctionNode it defaults to false.

The sql security field type was changed from TriState to std::optional<SqlSecurity> because the SS_DROP state was added.
Added sql security clauses:
states for full update: definer / invoker / nothing
for partial update: definer / invoker / drop
In full update prohibited "drop sql security"
in partial update prohibited "nothing" because it will be infinite recursion.

sim1984:
See also https://github.com/FirebirdSQL/firebird/issues/7744

dyemanov:
I'm fine with this implementation, please go ahead for stored procedures.

src/dsql/PackageNodes.epp
```diff
@@ -215,6 +215,9 @@ string CreateAlterPackageNode::internalPrint(NodePrinter& printer) const
 
 DdlNode* CreateAlterPackageNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
 {
+	if(alter && !items)
```
asfernandes, line 218:
```suggestion
	if (alter && !items)
```

src/dsql/PackageNodes.epp
```diff
@@ -319,9 +322,18 @@ void CreateAlterPackageNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlS
 	// run all statements under savepoint control
 	AutoSavePoint savePoint(tdbb, transaction);
 
+	const bool alterIndividualParameters = (alter && !items);
+
 	if (alter)
 	{
-		if (!executeAlter(tdbb, dsqlScratch, transaction))
+		if(alterIndividualParameters)
```
asfernandes, line 329:
```suggestion
		if (alterIndividualParameters)
```

src/dsql/PackageNodes.epp
```diff
@@ -319,9 +322,18 @@ void CreateAlterPackageNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlS
...
+		{
+			if(!executeAlterIndividualParameters(tdbb, dsqlScratch, transaction))
```
asfernandes, line 331:
```suggestion
			if (!executeAlterIndividualParameters(tdbb, dsqlScratch, transaction))
```

src/dsql/PackageNodes.epp
```diff
@@ -475,6 +487,44 @@ bool CreateAlterPackageNode::executeAlter(thread_db* tdbb, DsqlCompilerScratch*
...
+	{
+		modified = true;
+
+		executeDdlTrigger(tdbb, dsqlScratch, transaction, DTW_BEFORE,
+			DDL_TRIGGER_ALTER_PACKAGE, name, NULL);
+
+		MODIFY PKG
+			if (ssDefiner.has_value())
+			{
+				if(ssDefiner.value() != SqlSecurity::SS_DROP)
```
asfernandes, line 507:
```suggestion
				if (ssDefiner.value() != SqlSecurity::SS_DROP)
```

src/dsql/parse.y
```diff
@@ -2707,9 +2707,13 @@ procedure_clause
 	| external_procedure_clause
 	;
 
+%type <createAlterProcedureNode> change_opt_procedure_clause
+change_opt_procedure_clause
```
asfernandes, line 2711: What is this empty rule?

src/dsql/parse.y
```diff
@@ -2966,6 +2987,16 @@ package_clause
 		}
 	;
 
+%type <createAlterPackageNode> change_opt_package_clause
+change_opt_package_clause
```
asfernandes, line 2991: Maybe `change_opt_package_clause` and `change_opt_procedure_clause` could be put directly in the alter rules?
I think these rules has not so great names to be put standalone and they are not reused.

src/dsql/DdlNodes.epp
```diff
@@ -2742,11 +2742,22 @@ void CreateAlterProcedureNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsq
...
+			altered = true;
+		else
+			status_exception::raise(Arg::Gds(isc_dyn_proc_not_found) << Arg::Str(name));
+	}
+	else if (alter)
 	{
 		if (executeAlter(tdbb, dsqlScratch, transaction, false, true))
+		{
 			altered = true;
+		}
```
asfernandes, line 2760:
```suggestion
			altered = true;
```

src/dsql/parse.y
Zhdanov0: I made rules `change_opt_package_clause` and `change_opt_procedure_clause` to separate full alter and partial alter. I can change the names to `partial_alter_package_clause` and `partial_alter_procedure_clause`. This way it will be clearer and will separate full alter and partial alter.

src/dsql/parse.y
Zhdanov0: It's not empty, the definition is below.

asfernandes: And why you "declared it" here?
```
%type <createAlterProcedureNode> change_opt_procedure_clause
change_opt_procedure_clause
	;
```

src/dsql/parse.y
asfernandes: You call it partial and I call it just another form of alter. Every alter is partial, they change what was specified. So my suggestion to just use another branch of the original rules directly.

Zhdanov0: "Every alter is partial, they change what was specified." - is not correct for functions, procedures and packages. 
Functions/procedures/packages change all parameters, even those that were not specified. For example: 
```
ALTER FUNCTION func1 (str varchar(5))
RETURNS INT
AS
DECLARE result INTEGER;
BEGIN
result = POSITION('a', str);
RETURN result;
END;
```
The deterministic and sql security parameters were not specified, but are still being modified
(reset).
I will call this a full alter (the `executeAlter` method is used).

Now partial alter (`executeAlterIndividualParameters` method is used):
Here parameters that are not specified will not be modified.  Only the individual specified parameters will be changed.
For example:
```
ALTER FUNCTION name SQL SECURITY INVOKER
```
The deterministic parameter is not specified here and it is not modified.

My opinion:
This is different behavior and it should be in different branches.
But it is better to change the name to a more clarifying one: `partial_alter_package_clause` for example.

asfernandes: So we have a big inconsistency with triggers:

```
SQL> create table t1 (n1 integer)!
SQL> create trigger t1_ins inactive before insert position 4 on t1 SQL SECURITY DEFINER as begin end!

SQL> show trigger t1_ins!
Triggers on Table T1:
T1_INS, Sequence: 4, Type: BEFORE INSERT, Inactive, SQL SECURITY DEFINER
Trigger text:
=============================================================================
as begin end
=============================================================================

SQL> alter trigger t1_ins as begin end!

SQL> show trigger t1_ins!
Triggers on Table T1:
T1_INS, Sequence: 4, Type: BEFORE INSERT, Inactive, SQL SECURITY DEFINER
Trigger text:
=============================================================================
as begin end
=============================================================================
```

dyemanov:
Unless there will be other issues posted, I'll merge this PR later this week.

dyemanov:
@Zhdanov0, please resolve the existing conflicts.

pavel-zotov:
QA is covered by test for #7744

PR #7811, opened 2023-10-20
Improve SKIP LOCKED implementation

### LINKED ISSUE #7810: Improve SKIP LOCKED implementation
Make SKIP LOCKED works in all transaction isolation levels, including RC NRV
Avoid preliminary lock for UPDATE and DELETE statements 

More details at thread "SKIP LOCKED problems" in fb-devel
https://groups.google.com/g/firebird-devel/c/BIbAz_YJMFw

hvlad:
See #7810

### DIFF
--- a/src/dsql/StmtNodes.cpp
+++ b/src/dsql/StmtNodes.cpp
@@ -152,6 +152,40 @@ namespace
 		{}
 	};
 
+	// Combined conditional savepoint and its change marker.
+	class CondSavepointAndMarker
+	{
+	public:
+		CondSavepointAndMarker(thread_db* tdbb, jrd_tra* trans, bool cond) :
+			m_savepoint(tdbb, trans, cond),
+			m_marker(cond ? trans->tra_save_point : nullptr)
+		{}
+
+		~CondSavepointAndMarker()
+		{
+			rollback();
+		}
+
+		void release()
+		{
+			m_marker.done();
+			m_savepoint.release();
+		}
+
+		void rollback()
+		{
+			m_marker.done();
+			m_savepoint.rollback();
+		}
+
+	private:
+		// Prohibit unwanted creation/copying
+		CondSavepointAndMarker(const CondSavepointAndMarker&) = delete;
+		CondSavepointAndMarker& operator=(const CondSavepointAndMarker&) = delete;
+
+		AutoSavePoint m_savepoint;
+		Savepoint::ChangeMarker m_marker;
+	};
 }	// namespace
 
 
@@ -2314,7 +2348,7 @@ StmtNode* EraseNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
 			PASS1_limit(dsqlScratch, dsqlRows->length, dsqlRows->skip, rse);
 
 		if (dsqlSkipLocked)
-			rse->flags |= RseNode::FLAG_WRITELOCK | RseNode::FLAG_SKIP_LOCKED;
+			rse->flags |= RseNode::FLAG_SKIP_LOCKED;
 	}
 
 	if (dsqlReturning && dsqlScratch->isPsql())
@@ -2348,6 +2382,7 @@ string EraseNode::internalPrint(NodePrinter& printer) const
 	NODE_PRINT(printer, dsqlReturning);
 	NODE_PRINT(printer, dsqlRse);
 	NODE_PRINT(printer, dsqlContext);
+	NODE_PRINT(printer, dsqlSkipLocked);
 	NODE_PRINT(printer, statement);
 	NODE_PRINT(printer, subStatement);
 	NODE_PRINT(printer, stream);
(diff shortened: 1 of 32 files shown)


asfernandes:
My own suite of tests is showing various errors after these changes. Here is the tests.
[skip-locked-tests.tar.gz](https://github.com/FirebirdSQL/firebird/files/13053365/skip-locked-tests.tar.gz)

src/dsql/StmtNodes.cpp
```diff
@@ -152,6 +152,40 @@ namespace
...
+
+		void rollback()
+		{
+			m_marker.done();
+			m_savepoint.rollback();
+		}
+
+	private:
+		// Prohibit unwanted creation/copying
+		CondSavepointAndMarker(const CondSavepointAndMarker&) = delete;
```
asfernandes, line 183: It's better to put deleted constructor/operator in public section, so error message will be about this fact. If they are in private section, then error is about accessibility.

hvlad:
> My own suite of tests is showing various errors after these changes. Here is the tests. [skip-locked-tests.tar.gz](https://github.com/FirebirdSQL/firebird/files/13053365/skip-locked-tests.tar.gz)

Thanks for the info, but I have no idea how to use this.
Tests with isql is preferred.

asfernandes:
> > My own suite of tests is showing various errors after these changes. Here is the tests. [skip-locked-tests.tar.gz](https://github.com/FirebirdSQL/firebird/files/13053365/skip-locked-tests.tar.gz)
> 
> Thanks for the info, but I have no idea how to use this. Tests with isql is preferred.

It's easy to run, but unfortunately it's a closed project (temporary opened at that time) as per my message "Another test framework?" sent in fb-admins a year ago, and closed again due to not be useful for others (no response).

Anyway, it's very easy to understand and extract tests and run them in other ways. ;)

asfernandes:
> Then it shouldn't be hard for you to run it and show expected and actual results.

The file I attached here has exactly this: test source code, expected run and erroneous run. It doesn't only have the test framework.

If you or @pavel-zotov find it useful and want my help so this can be put in the official test suite, just ask.

dyemanov:
I believe your tests should be copied into the official fbtest framework to be widely useful, so I'd appreciate your help with this. AFAIK, Pavel is on vacation now, but should be back one week later.

hvlad:
I've fixed DELETE RETURNING with simplest possible way. Perhaps more complex solution - like in ModifyNode, would be better.
The our QA tests is passed (considering required changes in RC NRV mode).
I can't run tests by Adriano but it should fail in places where "skip" logic was changed: now engine doesn't take into account skipped locked records for the ROWS clause purposes.

src/dsql/StmtNodes.cpp
hvlad: Will be fixed in next commit.

src/dsql/StmtNodes.cpp
```diff
@@ -5632,6 +5689,16 @@ StmtNode* ForNode::pass2(thread_db* tdbb, CompilerScratch* csb)
...
+	if (rse->hasSkipLocked())
+	{
+		SortedStreamList streamList;
+		rse->collectStreams(streamList);
+		for (auto stream : streamList)
+		{
+			csb->csb_rpt[stream].csb_flags |= csb_skip_locked;
+		}
+	}
+
```
dyemanov, line 5701: IIRC, `rse->collectStreams()` collects all the streams found inside the RSE, including those buried inside subqueries inside different expressions in FIRST/SKIP/WHERE/etc.

I'd move this code to `Optimizer::compile()` instead, it already has `compileStreams` array which contains only streams inside the `FROM` clause. Maybe even share the existing loop between `WITH LOCK` and `SKIP LOCKED` features.

dyemanov (changes requested):

dyemanov (approved):

asfernandes:
As I previously said, this implementation changes behavior of what was delivered (and is documented) in v5.

https://github.com/FirebirdSQL/firebird/issues/7810 does not mention behavior changes.
This PR does not changes documentation.

For me, this implementation has bug in relation of how FIRST/SKIP/ROWS/etc is documented to interact with SKIP LOCKED.

asfernandes:
It also changed generated BLR for DELETE ... RETURNING (without SKIP LOCKED clause).

src/dsql/StmtNodes.cpp
hvlad: Done

sim1984:
I think the behavior after the patch: skip everything blocked, and only then apply the FIRSK/SKIP/ROWS/FETCH/OFFSET limiters, is much more logical and understandable to the end user. And in my mind it should be moved to v5 before it becomes a release. If this is done later, it will create additional difficulties when migrating to 6.0, for those who have time to use this feature in 5.0.

hvlad:
I have the same opinion, thanks.

EPluribusUnum:
Where is the new SELECT syntax? Does SKIP LOCKED will be allowed in CTE/subselect/derived table? 
It should be allowed to support composability. Also as it affects the resultset and effectively works as a WHERE condition it should be accepted everywhere where WHERE is accepted also.

hvlad:
You ask in a wrong place, the ticket with feature itself is #7350

Anyway, current description is
https://github.com/FirebirdSQL/firebird/blob/master/doc/sql.extensions/README.skip_locked.md

and updated, but not merged yet is
https://github.com/FirebirdSQL/firebird/blob/work/gh-7810-2/doc/sql.extensions/README.skip_locked.md

the difference is in 'Notes' section only

EPluribusUnum:
WITH LOCK section is not part of the composable select and its applied on the final resultset. I think in this case also SKIP LOCKED should apply on the final resultset, not between the skip and limit.

PR #7824, opened 2023-11-03
Add sw_64 support

haomi123:
Add sw_64 support

### DIFF
--- a/builds/posix/prefix.linux_sw_64
+++ b/builds/posix/prefix.linux_sw_64
@@ -1,0 +1,28 @@
+# The contents of this file are subject to the Interbase Public
+# License Version 1.0 (the "License"); you may not use this file
+# except in compliance with the License. You may obtain a copy
+# of the License at http://www.Inprise.com/IPL.html
+#
+# Software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


AlexPeshkoff:
Did you try to run tests in extern/libcds? That's HW-dependent library, and I want to make sure does it work correctly on sw-64. Malfunction of it may cause unstable FB operation.

mariuz:
@haomi123 here is example for LoongArch patch for libcds in the firebird tree

https://github.com/FirebirdSQL/firebird/commit/fcc471e420525dbd4a1dda59ed51487d5ead5921#diff-31135add04bb8d8584529be04bae88c6a088d3a90c73a6c8c7716bd8bec815db

also upstream pull request 
https://github.com/khizmax/libcds/pull/183

ps : @AlexPeshkoff upstream project seems to be unmaintained 
https://github.com/khizmax/libcds/pulls

AlexPeshkoff:
@mariuz I know that libcds is unmaintained. Not good, but not too big trouble - we already have unmaintained btyacc and it did not cause too big troubles for >10 years. And IMO it's not a reason to skip it's tests when porting to new hardware. For S390x & mentioned here LoongArch tests were reported to pass successfully.

Issue #1708, opened 2007-05-30
Avoid data retrieval if the WHERE clause always evaluates to FALSE [CORE1287]


firebird-issue-importer:
Submitted by: Alexander Tyutik (tut)

Is duplicated by [CORE6394](https://github.com/FirebirdSQL/firebird/issues?q=CORE6394+in%3Atitle)
Is duplicated by [CORE5394](https://github.com/FirebirdSQL/firebird/issues?q=CORE5394+in%3Atitle)

Votes: 14

In query like this

`SELECT * FROM SOME_TABLE WHERE 1 = 0`

It is obviously that condition in WHERE clause is always FALSE, but FB will read all records from SOME_TABLE without any benefits.

Now real example where this improvement can be useful. I need to write recursive CTE and specify level deep limit. I wrote such test query:

```
EXECUTE BLOCK
RETURNS (
  ID INTEGER,
  LEV SMALLINT)
AS
  DECLARE MAX_LEV SMALLINT = 0;
BEGIN
  FOR WITH RECURSIVE TR AS (
      SELECT T.*, 1 AS LEV FROM "Tree" T WHERE PARENT_ID IS NULL
      UNION ALL
      SELECT T.*, TR.LEV + 1 AS LEV FROM "Tree" T, TR WHERE PARENT_ID = TR.ID AND TR.LEV < :MAX_LEV)
    SELECT ID, LEV FROM TR INTO :ID, :LEV AS CURSOR CUR DO
  BEGIN
    SUSPEND;
  END
END
```

and in this query part

`SELECT T.*, TR.LEV + 1 AS LEV FROM "Tree" T, TR WHERE PARENT_ID = TR.ID AND TR.LEV < :MAX_LEV`

will be make unnecessary work for each leaf and performance will be worse.

PR #7853, opened 2023-11-19
Do not consider non-deterministic expressions as invariants in pre-filters

dyemanov:
Related to my prior commit for #1708. Even if the boolean does not access table fields it may still be non-deterministic, in this case it should not be treated as invariant and evaluated just once.

Example:

`select count(*) from employee where rand() > 0.5`

The condition is not correlated with any data but still produces a different result per every invocation. Prior to v5, it returned (approximately) half rows of the result set. In v5, it returns either zero rows or a half of them, depending on what was the result of `RAND()` during its first invocation.

Tested with constant expressions, `RAND()` and SQL functions defined as deterministic and non-deterministic. This patch does not detect determinism too deeply (e.g. whether a field references stream from the current scope or the outer one), only to the level currently needed for the optimizer.


src/jrd/SysFunction.cpp
```diff
@@ -6962,3 +6962,16 @@ void SysFunction::checkArgsMismatch(int count) const
 		status_exception::raise(Arg::Gds(isc_funmismat) << Arg::Str(name));
 	}
 }
+
+
+bool SysFunction::deterministic() const
+{
+	const MetaName funcName(name);
+
+	return (funcName != "GEN_UUID" &&
```
asfernandes, line 6971: I think this should be defined in table at `SysFunction::functions`.

pavel-zotov:
Only 5.x is checked now.
FB 6.x has not yet improved, test will fail on it.
Waiting for front-port.

dyemanov:
Frontported to v6.

pavel-zotov:
QA note: removed upper bound for version after this feature was front-ported to 6.x. Checked on 6.0.0.532.

PR #7857, opened 2023-11-20
Fix possible BUGCHECK "cannot find tip page" during validation with -full or -mend option

ilya071294:
If a primary record version has a transaction number which is greater than NT and seems corrupted, its state will be treated as tra_active to avoid an attempt to fetch a non-existing TIP page.


hvlad (approved):
I have no objection, just one question - what consequences you expect with such change, if it is `Next` on disk is wrong (less than actual value) ? 
Perhaps such condition could be detected in `walk_tip()` ?

ilya071294:
> if it is `Next` on disk is wrong (less than actual value)

The actual value - do you mean it as the number of the last non-active transaction on the TIP?

hvlad:
> > if it is `Next` on disk is wrong (less than actual value)
> 
> The actual value - do you mean it as the number of the last non-active transaction on the TIP?

Yes, exactly.

ilya071294:
> Yes, exactly.

If such corruption is detected, is it possible to fix it safely? Which page can we trust more - the header or the TIP?

hvlad:
> If such corruption is detected, is it possible to fix it safely? Which page can we trust more - the header or the TIP?

If `Next` points to the last TIP page and if TIP chain is valid, then both header and last TIP could be trusted in the same degree, IMHO.

dyemanov:
Do we expect any more commits inside this PR? Or can it be merged now?

ilya071294:
Conflicts are solved. Please review.

Issue #7854, opened 2023-11-19
Performance degradation with global temporary table


shavluk:
I've identified a performance issue related to the execution speed when working with a global temporary table in 5.0.0.1271-RC2 Below is a simplified example to reproduce the problem:

```
create global temporary table gtt_test (
    id  integer not null,
    t   timestamp default current_timestamp
) on commit preserve rows;

alter table gtt_test add constraint pk_gtt_test primary key (id);
```

The problem occurs during the insertion of a large number of rows, and the performance has significantly decreased compared to Firebird 3.0.10. The insertion speed has dropped by approximately 2 times.

Here is an example of inserting a large number of rows:
```
execute block as
declare variable id_session int;
declare variable k int = 0;
begin
  while (k < 1000000) do
  begin
    insert into gtt_test(id) values (:k);
    k = k + 1;
  end
end
```

sim1984:
Starting with Firebird 4.0, current_timestamp returns a timestamp with time zone type value. Therefore, the compare is not quite correct. Replace with default localtimestap and compare again.

hvlad:
Could you provide runtime statistics for both servers also ?

shavluk:
I have the server running in compatibility mode with DataTypeCompatibility = 3.0
The localtimestamp and current_timestamp show the same times

When using localtimestamp the problem does go away. 

10000000 record(s) was(were) inserted into GTT_TEST
firebird 3.0
------ Performance info ------
Prepare time = 0ms
Execute time = 48s 625ms
Current memory = 442 306 032
Max memory = 446 333 296
Memory buffers = 50 000
Reads from disk to cache = 0
Writes from cache to disk = 33 317
Fetches from cache = 59 352 468


firebird 5.0 (localtimestamp)
------ Performance info ------
Prepare time = 0ms
Execute time = 42s 969ms
Current memory = 432 436 544
Max memory = 436 109 328
Memory buffers = 50 000
Reads from disk to cache = 0
Writes from cache to disk = 7 487
Fetches from cache = 59 391 477

firebird 5.0 (current_timestamp)
------ Performance info ------
Prepare time = 0ms
Execute time = 1m 54s 344ms
Current memory = 432 425 632
Max memory = 436 098 416
Memory buffers = 50 000
Reads from disk to cache = 0
Writes from cache to disk = 7 487
Fetches from cache = 59 391 477

sim1984:
[...]

PR #7858, opened 2023-11-20
Fixed performance issue #7854

hvlad:


aafemt:
I wonder why all this atomic juggles instead of obtaining UCalendar in constructor? TimeZoneDesc itself is created on demand.

src/common/TimeZoneUtil.cpp
```diff
@@ -70,9 +83,26 @@ namespace
 			return unicodeName.begin();
 		}
 
+		UCalendar* getCalendar(const Jrd::UnicodeUtil::ConversionICU& icuLib, UErrorCode* err) const
+		{
+			UCalendar* val = calendar.load();
+			if (!val)
+			{
+				val = icuLib.ucalOpen(getUnicodeName(), -1, NULL, UCAL_GREGORIAN, err);
```
asfernandes, line 91: I'd get icuLib here instead of receive by argument.

src/common/TimeZoneUtil.cpp
```diff
@@ -70,9 +83,26 @@ namespace
...
+				if (!val)
+					return val;
+
+				UCalendar* old = calendar.exchange(val);
+				if (old)
+					return old;
+			}
+			return icuLib.ucalClone(val, err);
```
asfernandes, line 99: `ucal_clone` is not documented to be thread-safe, so it should not be assumed to be.
I do also not know its performance.
I'd instead implement a pool of unused objects where concurrent usage may do open/close additional copies and single usage would be reverted to the pool after its usage.

hvlad:
> I wonder why all this atomic juggles instead of obtaining UCalendar in constructor? TimeZoneDesc itself is created on demand.

FIrst, all `TimeZoneDesc`'s is created in `TimeZoneStartup` ctor. 
Second, there are 637 items in `timeZoneList` currently, do you offer to create `UCalendar` for all of them ?
Last, define what is "juggles" and why you so afraid of it ? 

BTW, even if TimeZoneDesc were created on demand, some synchronization would still be required.

src/common/TimeZoneUtil.cpp
hvlad: I followed this docs

https://unicode-org.github.io/icu/userguide/icu/design.html#thread-safe-const-apis

https://unicode-org.github.io/icu/userguide/icu/design.html#clone-vs-open

aafemt:
> FIrst, all `TimeZoneDesc`'s is created in `TimeZoneStartup` ctor.

Oops, you are right, I misread getDesc(). Why they create time zones that are never going to be used?..

src/common/TimeZoneUtil.cpp
hvlad: About performance, with patch I have following result:

| TEST | DURATION |
|--------|--------|
TZ -> TS	| 44
TS -> TZ	| 70
TS -> TS	| 13
TZ -> TZ	| 13

asfernandes: "However, you cannot use a reference to an open service object in two threads at the same time if either of them calls any non-const API."

`ucal_clone` is const, but `ucal_setmillis` is not.

In no place it says `ucal_clone` can be called concurrently with others non-const functions, so it's not safe.

hvlad:
> Why they create time zones that are never going to be used?..

Because it is cheap and simple and allows to avoid synchronization later.

src/common/TimeZoneUtil.cpp
hvlad: `ucal_clone` is const by itself thus it could be called with no sync.
`ucal_setmillis` is non-const but it uses own private instance of UCalendar, cloned from existing one.

hvlad: Also, the full cite is:

> However, you cannot use a reference to an open service object in two threads at the same time if either of them calls any non-const API. An individual open service object is not thread-safe for concurrent “writes”. Rather, for non-const use, you must use the clone function to create a copy of the service you want and then pass this copy to the second thread. This procedure allows you to use the same service in different threads, but avoids any thread synchronization or deadlock problems.

asfernandes:
I used your implementation as base for alternative implementation and had better results:

```
Original:
TEST                 DURATION 
================ ============ 
TZ -> TS                  161 
TS -> TZ                  183 
TS -> TS                   12 
TZ -> TZ                   11 

This PR:
TEST                 DURATION 
================ ============ 
TZ -> TS                   36 
TS -> TZ                   56 
TS -> TS                   12 
TZ -> TZ                   11 

https://github.com/FirebirdSQL/firebird/pull/7859
TEST                 DURATION 
================ ============ 
TZ -> TS                   31 
TS -> TZ                   36 
TS -> TS                   11 
TZ -> TZ                   12 
```

hvlad:
> I used your implementation as base for alternative implementation and had better results:

Afraid it will be not as good when run under contention: `ucal_open()` is much more costly than `ucal_clone()`.
So far I didn't run mt-tests for it, though.

asfernandes:
> > I used your implementation as base for alternative implementation and had better results:
> 
> Afraid it will be not as good when run under contention: `ucal_open()` is much more costly than `ucal_clone()`. So far I didn't run mt-tests for it, though.

Calendar is maintained in use for small times. In the end, it could hurt only benchmarks probably.

PR #7864, opened 2023-11-21
The ability to retrieve the total number of pages in the database, the

Danuul3:
The ability to retrieve the total number of pages in the database, the 
number of pages in use, and the number of free pages via the GET_CONTEXT.

### DIFF
--- a/src/jrd/SysFunction.cpp
+++ b/src/jrd/SysFunction.cpp
@@ -430,7 +430,13 @@ const char
 // Boolean values
 static const char
 	FALSE_VALUE[] = "FALSE",
-	TRUE_VALUE[] = "TRUE";
+    TRUE_VALUE[] = "TRUE";
+
+// Get pages
+const char
+    PAGES_ALLOCATED[]="PAGES_ALLOCATED",
+    PAGES_USED[]="PAGES_USED",
+    PAGES_FREE[]="PAGES_FREE";
 
 
 double fbcot(double value) noexcept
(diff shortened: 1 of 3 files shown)


src/jrd/SysFunction.cpp
```diff
@@ -430,7 +430,13 @@ const char
...
 static const char
 	FALSE_VALUE[] = "FALSE",
-	TRUE_VALUE[] = "TRUE";
+    TRUE_VALUE[] = "TRUE";
+
+// Get pages
+const char
+    PAGES_ALLOCATED[]="PAGES_ALLOCATED",
+    PAGES_USED[]="PAGES_USED",
+    PAGES_FREE[]="PAGES_FREE";
```
dyemanov, line 440: Spaces around `=`, please.

src/jrd/SysFunction.cpp
```diff
@@ -4583,6 +4589,25 @@ dsc* evlGetContext(thread_db* tdbb, const SysFunction*, const NestValueArray& ar
...
+
+            resultStr.printf("Number of DB pages used = %" SQUADFORMAT, PageSpace::usedPages(dbb));
+        }
+
+        else if (nameStr == PAGES_FREE)
+        {
+
+            resultStr.printf("Number of DB pages free = %" SQUADFORMAT, PageSpace::maxAlloc(dbb) - PageSpace::usedPages(dbb));
+        }
+
```
dyemanov, line 4610: "Number of ..." text is not needed here, only the number should be returned.

dyemanov: And given that `PageSpace::*` functions return `ULONG`, why `SQUADFORMAT`?

dyemanov: While formally the new code is correct, I'd suggest to use ULONGFORMAT instead of "lu" like it's done for SESSION_IDLE_TIMEOUT or STATEMENT_TIMEOUT.

src/jrd/SysFunction.cpp
```diff
@@ -4835,7 +4853,7 @@ dsc* evlGetContext(thread_db* tdbb, const SysFunction*, const NestValueArray& ar
 	{
 		if (!transaction->tra_context_vars.get(nameStr, resultStr))
 			return NULL;
-	}
+    }
```
dyemanov, line 4856: Please revert spaces back to the tab character here.

dyemanov:
Please add description of the new variables to `/doc/sql.extensions/README.context_variables2`. I will merge this PR then.

dyemanov:
BTW, `src/dsql/parse-conflicts.txt` is not related to this patch, so better remove it from PR.

Issue #7869, opened 2023-11-22
GBAK can write uninitialized data into RDB$RETURN_ARGUMENT and RDB$ARGUMENT_POSITION fields


dmitry-lipetsk:
GBAK does not initialize RDB$RETURN_ARGUMENT, RDB$ARGUMENT_POSITION fields and can write unexpected values into restored database when backup does not have data for ones.

See these two functions:
1. get_function
https://github.com/FirebirdSQL/firebird/blob/cf98c6e97f6fa25eabe81b23b1a7ac6900e51be5/src/burp/restore.epp#L4947

2. get_function_arg
https://github.com/FirebirdSQL/firebird/blob/cf98c6e97f6fa25eabe81b23b1a7ac6900e51be5/src/burp/restore.epp#L5333

I offer to add the initialization of these fields with NULL value.

aafemt:
Backup file without these values is malformed. An error must be raised on attempt to restore such backup.

dmitry-lipetsk:
Without initialization to NULL in GBAK, this "NOT NULL" field will be assigned with NOT NULL garbage.

Am I right?

dmitry-lipetsk:
Hello. Please, let me know if I must create equal PR for FB5 and master. Thanks.

pavel-zotov:
Maybe i did not understand properly: what RDB$ table must be checked ?

Initial DDL:
```
set list on;
set bail on;
shell del g:\temp\tmp4test.fdb 2>nul;
create database 'localhost:g:\temp\tmp4test.fdb' user sysdba password 'masterkey';

set term ^;
create or alter function fn_test(a_x int, a_y timestamp, a_z boolean) returns varchar(50) character set utf8 as
begin
    return '';
end
^
set term ;^
commit;

select f.rdb$return_argument from rdb$functions f where f.rdb$function_name = upper('fn_test');

select g.rdb$argument_name, g.rdb$argument_position from rdb$function_arguments g where g.rdb$function_name = upper('fn_test');
```

2) do backup and restore:
```gbak -b localhost:g:\temp\tmp4test.fdb stdout | gbak -rep stdin localhost:g:\temp\tmp4test.tmp.fdb```

3) check what we have in restored DB:
```
set list on;
select f.rdb$return_argument from rdb$functions f where f.rdb$function_name = upper('fn_test');

select g.rdb$argument_name, g.rdb$argument_position from rdb$function_arguments g where g.rdb$function_name = upper('fn_test');
```

[...]

Issue #7846, opened 2023-11-14
FB4 can't backup/restore int128-array


dmitry-lipetsk:
Hello.

**Problem in backup**

put_array function (backup.epp)

https://github.com/FirebirdSQL/firebird/blob/d1e6921c0c6a0eba82623a450b3cb38d5092a358/src/burp/backup.epp#L962-L969

This switch must include `case blr_int128:`

Without this case, gbak generates wrong SDL and server generates the errors like these:

> gbak:error accessing BLOB column COL_INT128 -- continuing
> gbak: ERROR:invalid slice description language at offset 5
> gbak:error accessing BLOB column COL_NUM_30_2 -- continuing
> gbak: ERROR:invalid slice description language at offset 5

**Same problems in restore**

get_array function (restore.epp)

https://github.com/FirebirdSQL/firebird/blob/d1e6921c0c6a0eba82623a450b3cb38d5092a358/src/burp/restore.epp#L1781-L1788

https://github.com/FirebirdSQL/firebird/blob/d1e6921c0c6a0eba82623a450b3cb38d5092a358/src/burp/restore.epp#L2015-L2022

These switches must have 'case blr_int128'

Error messages:

> gbak:error accessing BLOB column COL_INT128 -- continuing
> gbak: ERROR:column not array or invalid dimensions (expected 0, encountered 1)
> gbak:error accessing BLOB column COL_NUM_30_2 -- continuing
> gbak: ERROR:column not array or invalid dimensions (expected 0, encountered 1)

----
Test database - https://dropmefiles.com/kQeLq (link is valid within 14 days)

Table with arrays: TEST_ARRAY_DIM1

dmitry-lipetsk:
Just for info - test database (TEST_ARRAY_DIM1) contains the following data

```sql
select TEST_ID, COL_INT128,COL_NUM_30_2
from TEST_ARRAY_DIM1
where COL_INT128 is not null or COL_NUM_30_2 is not null
```
<img width="786" alt="image" src="https://github.com/FirebirdSQL/firebird/assets/4594554/7e4451a0-4e60-447d-9970-3a855b6d2f87">

PR #7845, opened 2023-11-14
[FB4] An improved support of UDF SCALAR_ARRAY-arguments

dmitry-lipetsk:
Hello.

It is the implementation of the second solution to issue #7843.

These changes force a server to do two things:

1. Generation of names for INPUT-arguments of UDF and saving these names in RDB$FUNCTION_ARGUMENT::RDB$ARGUMENT_NAME.

**UDF**
```sql
DECLARE EXTERNAL FUNCTION UDF_DUMMY4_RPN_2 smallint, integer, float, double precision
RETURNS PARAMETER 2
ENTRY_POINT 'IB_UDF_abs' MODULE_NAME 'ib_udf';
```

**Metadata**
```sql
select x.RDB$ARGUMENT_POSITION, x.RDB$ARGUMENT_NAME from RDB$FUNCTION_ARGUMENTS x
where x.RDB$FUNCTION_NAME='UDF_DUMMY4_RPN_2'
```
image

2. Passing to client the names of UDF and SCALAR_ARRAY-argument those are linked with DSQL parameter.

``` sql
DECLARE EXTERNAL FUNCTION MY_UDF__GET_DIM
INTEGER BY SCALAR_ARRAY
RETURNS INTEGER BY VALUE
ENTRY_POINT 'fn_get_dim__i4__sa' MODULE_NAME 'lcpi.test.udf.01.dll'

select MY_UDF__GET_DIM(?) from rdb$database;

-- ISQL info:
INPUT message field count: 1
01: sqltype: 540 ARRAY Nullable scale: 0 subtype: 0 len: 8
  :  name: ARG1  alias:
  : table: MY_UDF__GET_DIM  owner:

OUTPUT message field count: 1
01: sqltype: 496 LONG Nullable scale: 0 subtype: 0 len: 4
  :  name: MY_UDF__GET_DIM  alias: MY_UDF__GET_DIM
  : table:   owner:

```
It allows client to read the information about array element (type id, subtype, scale, charset id) and create a correct array.

---
If this patch is OK, I will create more complex tests for it on my side before final merging.

Thanks.

PR #7870, opened 2023-11-22
[FB4, GBAK] Correct handling of NULLs in RDB$RETURN_ARGUMENT and RDB$ARGUMENT_POSITION

dmitry-lipetsk:
This is a fix for issue #7869 - a potential problem of writing into database the uninitialized value in RDB$RETURN_ARGUMENT and RDB$ARGUMENT_POSITION columns.

Of course, GBAK has a lot of equal problems for other columns but I want to fix a problem only with these ones.

UPD1. Symmetric problem in backup is fixed. Thanks to Adriano.


aafemt:
I'm still sure that fld_f_position must be not nullable to prevent UB in engine.

dmitry-lipetsk:
> I'm still sure that fld_f_position must be not nullable to prevent UB in engine.

Me too but it is a change of ODS and I am too young to ask and suggest such changes.

aafemt:
Is it really an ODS change? No field added or deleted, no even unique index/constraint put here (though one won't hurt). Old code should work with new definition flawlessly and vice versa.

asfernandes:
In real life, this looks like more as a code improvement, and a partial one.
Why PR against v4 instead only master?

dmitry-lipetsk:
> In real life, this looks like more as a code improvement, and a partial one. Why PR against v4 instead only master?

I can added these changes to master and fb5, too.

I want to add these changes to FB4, because I work with this version and they are needed to fix another problem with UDF SCALAR-array arguments (#7846, #7845)

dmitry-lipetsk:
Hello. Please, let me know if I must create equal PR for FB5 and master.  It is not a problem for me. Thanks.

Issue #7596, opened 2023-05-22
nbackup -D ON causes problem when operates with some type of drives


pavel-zotov:
Problems can occur in case of sector size <> 1024 with nbackup -D ON.
Ticket created by request of Alexey Kovyazin.
Details were sent to FB team.

dyemanov:
AFAIU, these old PRs already contain the fix: https://github.com/FirebirdSQL/firebird/pull/241, https://github.com/FirebirdSQL/firebird/pull/240

dyemanov:
Closed in favour of the more detailed (and already fixed) #7873.

PR #7874, opened 2023-11-25
Fixed bug #7873 : IO error reading file, The parameter is incorrect.

### LINKED ISSUE #7873: Wrong memory buffer alignment and IO buffer size when working in direct IO mode
Customer reports that nbackup issues error when backup database:

IO error reading file: database.fdb
The parameter is incorrect. 

Investigation shows that:
1. The OS is Windows,
2. error happens with direct IO ON only (switch -D ON), 
3. volume where database file reside have following properties (I show important ones only):

Bytes Per Sector  :                4096
Bytes Per Physical Sector :        4096

hvlad:


src/utilities/nbackup/nbackup.cpp
```diff
@@ -1430,8 +1431,8 @@ void NBackup::backup_database(int level, Guid& guid, const PathName& fname)
 		Array<UCHAR> unaligned_scns_buffer;
 		Ods::scns_page* scns = NULL, *scns_buf = NULL;
 		{ // scope
-			UCHAR* buf = unaligned_scns_buffer.getBuffer(header->hdr_page_size + SECTOR_ALIGNMENT);
-			scns_buf = reinterpret_cast<Ods::scns_page*>(FB_ALIGN(buf, SECTOR_ALIGNMENT));
+			UCHAR* buf = unaligned_scns_buffer.getBuffer(header->hdr_page_size + sectorSize);
```
aafemt, line 1434: Wouldn't it be better to use C11 aligned_alloc or C++17 aligned new instead? Page buffers are big enough to be allocated not from pools IMHO.

hvlad: Why it could be better ? 
There are tens usages of `FB_ALIGN` in the code, do you offer to replace its all in this PR ? 
Also, do you offer to replace RAII class `Array` by `alloc` \ `new` and (a lot of)  `free` \ `delete`'s ?

aafemt: No. Only in places where aligned memory is needed. And yes, replace it with RAII std::unique_ptr.

hvlad: I see no reason to do it in this PR.

dyemanov:
Bug-report duplicates #7596, this PR duplicates PR #241.

hvlad:
> Bug-report duplicates #7596, this PR duplicates PR #241.

Very interesting, thanks for reminding it.

With current case (#7873) I've verified by user fix. This is first time on my memory where issue is reproducible and fix confirmed. 

The #241 covers more cases, but misses some that present in this one.
I'll add some parts from #241 into this PR, if there will be no objections.

Issue #2388, opened 2021-03-08
Implement SQL standard FORMAT clause for CAST between string types and datetime types [CORE6507]


firebird-automations:
Submitted by: @mrotteveel

Implement SQL standard FORMAT clause for CAST between string types and datetime types, to allow custom formatting of datetime values and conversion from string values with a specific format to datetime values\.

"""
<cast specification\> ::=
  CAST <left paren\>
    <cast operand\> AS <cast target\>
    \[ FORMAT <cast template\> \]
    <right paren\>

<cast operand\> ::=
    <value expression\>
  \| <implicitly typed value specification\>

<cast target\> ::=
    <domain name\>
  \| <data type\>

<cast template\> ::=
    <character string literal\>
"""

Where <cast template\> follows the rules of Subclause 9\.42, "Converting a datetime to a formatted character string" or Subclause 9\.43, "Converting a formatted character string to a datetime"\. Specific syntax rules defined in Subclause 9\.44, "Datetime templates":

"""
<datetime template\> ::=
    \{ <datetime template part\> \}\.\.\.

<datetime template part\> ::=
    <datetime template field\>
  \| <datetime template delimiter\>

<datetime template field\> ::=
    <datetime template year\>
  \| <datetime template rounded year\>
  \| <datetime template month\>
  \| <datetime template day of month\>
  \| <datetime template day of year\>
  \| <datetime template 12\-hour\>
  \| <datetime template 24\-hour\>
  \| <datetime template minute\>
  \| <datetime template second of minute\>
  \| <datetime template second of day\>
  \| <datetime template fraction\>
  \| <datetime template am/pm\>
  \| <datetime template time zone hour\>
  \| <datetime template time zone minute\>

<datetime template delimiter\> ::=
    <minus sign\>
  \| <period\>
  \| <solidus\>
  \| <comma\>
  \| <apostrophe\>
  \| <semicolon\>
  \| <colon\>
  \| <space\>

<datetime template year\> ::=
    YYYY \| YYY \| YY \| Y

<datetime template rounded year\> ::=
    RRRR \| RR

<datetime template month\> ::=
    MM

<datetime template day of month\> ::=
    DD

<datetime template day of year\> ::=
    DDD

<datetime template 12\-hour\> ::=
    HH \| HH12

<datetime template 24\-hour\> ::=
    HH24

<datetime template minute\> ::=
    MI

<datetime template second of minute\> ::=
    SS

<datetime template second of day\> ::=
[...]

PR #7881, opened 2023-11-27
Change some aspects of the string-to-date conversion with format to make it more similar to the SQL standard #2388

TreeHunter9:
When converting a formatted character string to a datetime, the default values for the date parts are taken from `CURRENT_TIMESTAMP`.
Change **(A/P)M** to **(A/P).M.** for a 12-hour period.
Fix **RM** pattern.
Add more unit test for conversion function.


TreeHunter9:
I'm using some sort of "hack" in CvtTest.cpp by redefining `TimeStamp::getCurrentTimeStamp()` for tests, and can someone more experienced with C++ say is this even correct?

asfernandes:
> I'm using some sort of "hack" in CvtTest.cpp by redefining `TimeStamp::getCurrentTimeStamp()` for tests, and can someone more experienced with C++ say is this even correct?

IMO this is problematic.

It probably only work due to the way the project is split in common static library, that is implementation detail.

It may also be problematic if this method becomes inline.

Mock in C++ usually is done with static (templates) or dynamic (virtual methods) polymorphism:
- static may make the code difficult, usually everything must be done in header
- dynamic may be slow

I think none of them is appropriate for `TimeStamp::getCurrentTimeStamp()`, but you may instead make the formatter code asks for it trough an indirection (for example, an std::function) that you may replace.

TreeHunter9:
> I think none of them is appropriate for `TimeStamp::getCurrentTimeStamp()`, but you may instead make the formatter code asks for it trough an indirection (for example, an std::function) that you may replace.

Yeah, it would work, but I looked for an alternative of `TimeStamp::getCurrentTimeStamp()` and found `getLocalDate()` from `Callbacks`. I think it's perfect fit for mocking, and I already kinda mocked it in **CvtTestUtils.h**.

dyemanov:
Are there any changes in this branch that's still pending? Or were all suggestions addressed?

TreeHunter9:
I reread that issue and @mrotteveel said that `HH` and `HH12` implemented differently from the standard (A.M. and P.M. should be in pattern, not in formatting string, if I understand it correctly).

asfernandes (approved):

mrotteveel:
> I re-read that issue and @mrotteveel said that `HH` and `HH12` implemented differently from the standard (A.M. and P.M. should be in pattern, not in formatting string, if I understand it correctly).

Yes, the current implementation doesn't implement `<datetime template am/pm>`, and `<datetime template 12-hour>` should not render `A.M.` or `P.M.` after the time, that is solely the responsibility of `<datetime template am/pm>`

dyemanov:
So let's have AM/PM fixed in this branch and only then merge it.

TreeHunter9:
I also added <datetime template rounded year> pattern, but I'm not sure if it matches the standard. I read it multiple times, and still don't get what the standard wants from me. I use Oracle description of <datetime template rounded year> and added it to docs.
@mrotteveel it seems you have some experience of reading and understanding ISO standard, can you please check if Oracle description of <datetime template rounded year> is similar to standard one?

dyemanov:
@asfernandes, @mrotteveel any more comments / suggestions for this PR?

mrotteveel:
I'll try to look at it tomorrow, but it will more likely be in the form of building from the branch and trying some things out than looking at the code.

src/common/cvt.cpp
```diff
@@ -2037,9 +2313,50 @@ static void string_to_format_datetime_pattern_matcher(std::string_view pattern,
...
+		{
+			cb->err(Arg::Gds(isc_pattern_cant_be_used_without_other_pattern_and_vice_versa)
+				<< Arg::Str("HH/HH12") << Arg::Str("A.M./P.M."));
+		}
+	}
+
+	// If CT contains <datetime template second of day>, then CT shall not contain any of the following:
+	// <datetime template 12-hour>, <datetime template 24-hour>, <datetime template minute>, <datetime
+	// template second of minute>, or <datetime template am/pm>.
+	if (static_cast<bool>(formatFlags & FormatPatterns::SSSSS))
```
asfernandes, line 2338: Are these `static_cast<bool>` necessary?

src/common/cvt.cpp
```diff
@@ -136,6 +136,76 @@ const SINT64 INT64_LIMIT = ((((SINT64) 1) << 62) / 5);
 // AP,2012: Looks like there is no need making len zero, but I keep old define for a reference.
 // {if (len) {memcpy(to, from, len); from += len; to += len; len = 0;} }
 
+#define DEFINE_ENUM_BITWISE_OPERATORS(enumType) \
```
asfernandes, line 139: I would move this defines to some generic header as it could be useful for others `enum class` code and `cvt.cpp` already has many things.

src/common/cvt.cpp
TreeHunter9: Yes. I recently read an interesting [discussion](https://chat.stackoverflow.com/rooms/9345/discussion-between-nicol-bolas-and-potatoswatter) about `enum class`, and maybe using it as bitfield wasn't a good idea.
It's interesting to hear what other people think about `enum class` as bitfield, maybe I should change it to simple `enum`?

aafemt: For bitfields I would use constexpressions in namespace. Using enum class for it is strange.

TreeHunter9: > For bitfields I would use constexpressions in namespace.

Yea, it's used a lot in our code base, but I don't know, it doesn't seem very informative when you looking at `int` variable in my opinion.
P.S. Sorry for misleading, **bitfields** is something different, I mean **bitflags**, but i think you get the point.

TreeHunter9: Oh, that's an interesting approach, first I thought about unnamed namespace, but now I get it.

aafemt: Also I would call the namespace "Format". In this case `Format::Pattern` and `Format::SSSSS` would be closer to so-called "literate programming". Code readability and self-documenting in one.

src/common/cvt.cpp
TreeHunter9: Done

mrotteveel:
> @asfernandes, @mrotteveel any more comments / suggestions for this PR?

Unfortunately, I didn't have time or energy yet to look at it in detail. Hopefully later this week, but otherwise maybe it should just be merged and we can fix/correct things later.

PR #7888, opened 2023-11-29
Extend SYSTEM context of RDB$GET_CONTEXT with allocated, free and used pages

Danuul3:
number of pages in use, and the number of free pages via the GET_CONTEXT.

### DIFF
--- a/src/jrd/SysFunction.cpp
+++ b/src/jrd/SysFunction.cpp
@@ -432,6 +432,12 @@ static const char
 	FALSE_VALUE[] = "FALSE",
 	TRUE_VALUE[] = "TRUE";
 
+// Get pages
+const char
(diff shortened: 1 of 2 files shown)


src/jrd/SysFunction.cpp
```diff
@@ -432,6 +432,12 @@ static const char
 	FALSE_VALUE[] = "FALSE",
 	TRUE_VALUE[] = "TRUE";
 
+// Get pages
+const char
+    PAGES_ALLOCATED[] = "PAGES_ALLOCATED",
+    PAGES_USED[] = "PAGES_USED",
+    PAGES_FREE[] = "PAGES_FREE";
+
```
hvlad, line 441: Why these constants put separately ? 
Its belongs to the "SYSTEM namespace: global and database wise items" section.

PR #7890, opened 2023-11-29
Fixed bug #7873 : Wrong memory buffer alignment and IO buffer size when working in direct IO mode

### LINKED ISSUE #7873: Wrong memory buffer alignment and IO buffer size when working in direct IO mode

hvlad:
This PR is based on #7874 and #241

### DIFF
--- a/src/common/classes/array.h
+++ b/src/common/classes/array.h
@@ -425,6 +425,16 @@ public:
 		return data;
 	}
 
+	// prepare array to be used as a buffer of capacity bytes aligned on given alignment
+	T* getAlignedBuffer(const size_type capacityL, const size_type alignL)
+	{
+		static_assert(sizeof(T) == 1, "sizeof(T) != 1");
+
+		ensureCapacity(capacityL + alignL, false);
+		count = capacityL + alignL;
+		return FB_ALIGN(data, alignL);
+	}
+
 	// clear array and release dinamically allocated memory
 	void free()
 	{
--- a/src/jrd/CryptoManager.h
+++ b/src/jrd/CryptoManager.h
@@ -312,16 +312,16 @@ private:
 	public:
 		operator Ods::pag*()
 		{
-			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, PAGE_ALIGNMENT));
+			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, DIRECT_IO_BLOCK_SIZE));
 		}
 
 		Ods::pag* operator->()
 		{
-			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, PAGE_ALIGNMENT));
+			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, DIRECT_IO_BLOCK_SIZE));
 		}
 
 	private:
-		char buf[MAX_PAGE_SIZE + PAGE_ALIGNMENT - 1];
+		char buf[MAX_PAGE_SIZE + DIRECT_IO_BLOCK_SIZE];
 	};
 
 	class DbInfo;
--- a/src/jrd/CryptoManager.h
+++ b/src/jrd/CryptoManager.h
@@ -312,16 +312,16 @@ private:
 	public:
 		operator Ods::pag*()
 		{
-			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, DIRECT_IO_BLOCK_SIZE));
+			return reinterpret_cast<Ods::pag*>(buf);
 		}
 
 		Ods::pag* operator->()
 		{
-			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, DIRECT_IO_BLOCK_SIZE));
+			return reinterpret_cast<Ods::pag*>(buf);
 		}
 
 	private:
-		char buf[MAX_PAGE_SIZE + DIRECT_IO_BLOCK_SIZE];
+		alignas(DIRECT_IO_BLOCK_SIZE) char buf[MAX_PAGE_SIZE];
 	};
 
 	class DbInfo;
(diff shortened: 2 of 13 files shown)


src/common/classes/array.h
```diff
@@ -425,6 +425,16 @@ class Array : protected Storage
 		return data;
 	}
 
+	// prepare array to be used as a buffer of capacity bytes aligned on given alignment
+	T* getAlignedBuffer(const size_type capacityL, const size_type alignL)
+	{
+		static_assert(sizeof(T) == 1, "sizeof(T) != 1");
+
+		ensureCapacity(capacityL + alignL, false);
```
aafemt, line 433: I think that `- ALLOC_ALIGNMENT` can be added here to reduce memory overhead a little.

aafemt: AFAIU your code add some unused space at the beginning and the end of allocated buffer to get right alignment. Because allocated raw buffer already aligned, the difference between raw buffer pointer and returning buffer pointer is guaranteed to be smaller than `alignL` at least by ALLOC_ALIGNMENT (actually ALLOC_ALIGNMENT + 1) bytes, so unused space at the end can be safely reduced.

hvlad: > AFAIU your code add some unused space at the beginning and the end of allocated buffer to get right alignment. Because allocated raw buffer already aligned, the difference between raw buffer pointer and returning buffer pointer is guaranteed to be smaller than `alignL` at least by ALLOC_ALIGNMENT (actually ALLOC_ALIGNMENT + 1) bytes, so unused space at the end can be safely reduced.

Formally you right, but how mush you going to save ? 16 bytes of few KBs ?

src/jrd/jrd.cpp
```diff
@@ -7592,11 +7592,17 @@ static JAttachment* create_attachment(const PathName& alias_name,
...
 static void check_single_maintenance(thread_db* tdbb)
 {
-	UCHAR spare_memory[RAW_HEADER_SIZE + PAGE_ALIGNMENT];
-	UCHAR* header_page_buffer = FB_ALIGN(spare_memory, PAGE_ALIGNMENT);
+	Database* const dbb = tdbb->getDatabase();
+
+	const ULONG ioBlockSize = dbb->getIOBlockSize();
+	const ULONG headerSize = MAX(RAW_HEADER_SIZE, ioBlockSize);
+
+	HalfStaticArray<UCHAR, RAW_HEADER_SIZE + PAGE_ALIGNMENT> temp;
```
asfernandes, line 7600: Here you used HalfStaticArray with fixed alignments, not using the mechanism of its Align parameter and using the base `getAlignedBuffer`. That's the place I'm saying it's better to change HalfStaticArray and use it normally.

src/common/classes/array.h
hvlad: Do you really think that every page buffer in page cache is allocated separately ?

src/utilities/gstat/dba.epp
```diff
@@ -612,9 +612,9 @@ int gstat(Firebird::UtilSvc* uSvc)
 
 	dba_fil* current = db_open(fileName.c_str(), fileName.length());
 
-	SCHAR temp[RAW_HEADER_SIZE];
-	tddba->page_size = sizeof(temp);
-	tddba->global_buffer = (pag*) temp;
+	SCHAR temp[RAW_HEADER_SIZE + DIRECT_IO_BLOCK_SIZE];
```
asfernandes, line 615: You can use `alignas(DIRECT_IO_BLOCK_SIZE)` here.

src/jrd/CryptoManager.h
```diff
@@ -312,16 +312,16 @@ class CryptoManager final : public Firebird::PermanentStorage, public BarSync::I
...
 
 		Ods::pag* operator->()
 		{
-			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, PAGE_ALIGNMENT));
+			return reinterpret_cast<Ods::pag*>(FB_ALIGN(buf, DIRECT_IO_BLOCK_SIZE));
 		}
 
 	private:
-		char buf[MAX_PAGE_SIZE + PAGE_ALIGNMENT - 1];
+		char buf[MAX_PAGE_SIZE + DIRECT_IO_BLOCK_SIZE];
```
asfernandes, line 324: You can use `alignas(DIRECT_IO_BLOCK_SIZE)` here.

src/common/classes/array.h
aafemt: > Do you really think that every page buffer in page cache is allocated separately ?

Probably I'm wrong but yes, I do.

asfernandes: We're talking about generic code in common, so I support the idea of optimize it for general use case.

src/jrd/jrd.cpp
hvlad: The actual alignment could be different (bigger) than used in template, thus I don't see what changes in HalfStaticArray could make the code better or more efficient.

src/common/classes/array.h
aafemt: Ok. In any case the difference can be found on https://firebirdtest.com/oltp-emul/ if someone add this branch there.

src/utilities/gstat/dba.epp
hvlad: Done

src/jrd/CryptoManager.h
hvlad: Done

src/common/classes/array.h
hvlad: > We're talking about generic code in common, so I support the idea of optimize it for general use case.

Sorry, I see any real benefits here but code complication.
Feel free to add such optimization outside of this PR.

dyemanov (approved):

Issue #7909, opened 2023-12-06
Changed plan (possible regression) in FB 5.x / 6.x comparing to FB 3.x / 4.x


pavel-zotov:
Problem was reported by one of our customer privately.
Consider script from attached .zip
[ddl-and-run_-_for-empty-tables.sql.zip](https://github.com/FirebirdSQL/firebird/files/13581104/ddl-and-run_-_for-empty-tables.sql.zip)

Query that caused problems in customer environment has following plans:
1) for FB 3.x and 4.x:
```
Select Expression
    -> Aggregate
        -> Filter
            -> Nested Loop Join (outer)
                -> Nested Loop Join (outer)
                    -> Filter
                        -> Table "PAYMENTSDETAILS" as "PD" Access By ID
                            -> Bitmap
                                -> Index "AAA_PD_PERIODFROMDATE" Range Scan (lower bound: 1/1)
                    -> Filter
                        -> Table "PAYMENTSH" as "PH" Access By ID
                            -> Bitmap
                                -> Index "PK_PAYMENTSH_MOMENT_GENEMPID" Unique Scan
                -> Filter
                    -> Table "INVOICEH" as "IH" Access By ID
                        -> Bitmap Or
                            -> Bitmap
                                -> Index "PK_INVOICEH_INVID" Unique Scan
                            -> Bitmap
                                -> Index "INVOICEHBYPARENTINVOICEIDDESC" Range Scan (full match)
```

2) for FB 5.x and 6.x:
```
Select Expression
    -> Aggregate
        -> Nested Loop Join (inner)
            -> Filter
                -> Table "INVOICEH" as "IH" Full Scan
            -> Filter
                -> Table "PAYMENTSH" as "PH" Access By ID
                    -> Bitmap Or
                        -> Bitmap
                            -> Index "IDX_PAYMENTSH_MASTERDOC_DESC" Range Scan (full match)
                        -> Bitmap
                            -> Index "IDX_PAYMENTSH_MASTERDOC_DESC" Range Scan (full match)
            -> Filter
                -> Table "PAYMENTSDETAILS" as "PD" Access By ID
                    -> Bitmap
                        -> Index "PK_PD_MOMENT_GENEMPID" Unique Scan
```

Customer states that second plan leads to performance problem.
Additional data from customer:
1) number of records per tables:
```
table_name           rec_count
==============================
InvoiceH                953779
PaymentsDetails         783127
PaymentsH             30408945
```
[...]

Issue #7681, opened 2023-07-20
Firebird 5 - wrong plan and sometimes wrong result


EPluribusUnum:
Hi *!

`SELECT lit.gl_fk_tetel14_cpy_run_with_ex(15021, 15021, 'KA', null, null, 'N') FROM rdb$database`

This call builds this SELECT : 

SELECT t.fkisz_id, t.nyitzar, t.ervdat, t.bizonylat, t.szoveg,
    t.szla, t.szla2, t.minosit1, t.minosit2, t.minosit3, t.minosit4, t.minosit5, t.minosit6,
    t.eszla, t.eszla2, t.eminosit1, t.eminosit2, t.eminosit3, t.eminosit4, t.eminosit5, t.eminosit6,
    t.tartozik, t.kovetel, t.mertek, t.megyseg, t.emertek, t.emegyseg, t.dertek, t.dnem, t.edertek, t.ednem,
    t.alrkod, t.gytip, t.gysor, t.ugyf_id, t.kod1, t.kod2, t.kod3, t.kod4, t.kod5,
    (SELECT ubt.id FROM pu_btet ubt WHERE ubt.pbiz_id = 15021 AND ubt.gysor = bt.gysor)
FROM pu_btet bt
    INNER JOIN pu_biz b ON b.id = bt.pbiz_id
    LEFT OUTER JOIN fk_buftet t ON t.alrsor = bt.id
WHERE bt.pbiz_id = 15021 AND
      t.tabla = IIF(b.igaz = 'F', 'FK_TETEL', 'FK_BUFFER') AND
      t.alrsor IS NOT NULL

FB30 plan
PLAN (T EV INDEX (PU_EV_KONYV_FK1))
PLAN (T FK_TETEL INDEX (PU_EV_KONYV_FK2))
PLAN (T FK_BUFFER INDEX (KT_EV_KTBT_FK_FK2, KT_EV_KTBT_FK_FK4, KT_EV_KTBT_FK_FK3, KT_EV_KTBT_FK_FK5))
PLAN (T KP INDEX (KAPENZTAR_FK2))
PLAN (UBT INDEX (PU_BTET_UK))
PLAN JOIN (JOIN (B INDEX (PU_BIZ_PK), BT INDEX (PU_BTET_BIZ_FK)), T FK_BUFFER INDEX (FK_BUFFER_ALRSOR), T FK_TETEL INDEX (FK_TETEL_ALRSOR))

FB40 plan
[...]

PR #7910, opened 2023-12-06
Add backward compatibility option that disables joins transformation

dyemanov:
Given a few tickets showing performance regressions after this improvement and feedback from users at the conference asking about some compatibility option to simplify migration in cases when LEFT JOINs were widely used intentionally just as optimizer hints -- here it is, a backward compatibility option is provided.


dyemanov:
May act like a workaround for tickets #7681 and #7909.

PR #7914, opened 2023-12-10
Update info about quotes in fbtrace.conf

cincuranet:


hvlad:
@AlexPeshkoff: I didn't merged it by myself because I need to check if that quotes required at all.
And its looks as not required when `name = value` syntax is used, i.e. when `=` is not missed.

AlexPeshkoff:
@hvlad 
If trace code does not have some additional requirements - quotes are definitely not required here from config POV. But patch fixed definite bug in the text. And sorry - I did not take a look that PR was already assigned.

PR #7918, opened 2023-12-13
Allow to configure Firebird in posix using relative directories with options --with-fb*

asfernandes:
Relative directories will be interpreted based in the runtime root prefix.

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -1232,11 +1232,19 @@ Firebird::PathName getPrefix(unsigned int prefType, const char* name)
 	{
 		if (prefType != Firebird::IConfigManager::DIR_CONF &&
 			prefType != Firebird::IConfigManager::DIR_MSG &&
+			prefType != Firebird::IConfigManager::DIR_TZDATA &&
 			configDir[prefType][0])
 		{
 			// Value is set explicitly and is not environment overridable
 			PathUtils::concatPath(s, configDir[prefType], name);
-			return s;
+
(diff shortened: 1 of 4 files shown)


AlexPeshkoff (approved):
@asfernandes Do you plan to have it in RC2? If yes contact @dyemanov right now - tag to be set very soon.

asfernandes:
I would like to have it in v5 then next RC would be good.
But if that is problematic, it's always possible to have patches in vcpkg applied before build and for Windows a new solution (via build scripts) should be found or the code would need to be patched too.

I'll first commit this to master, then let's see what @dyemanov thinks.

src/common/utils.cpp
```diff
@@ -1232,11 +1232,19 @@ Firebird::PathName getPrefix(unsigned int prefType, const char* name)
...
 			// Value is set explicitly and is not environment overridable
 			PathUtils::concatPath(s, configDir[prefType], name);
-			return s;
+
+			if (PathUtils::isRelative(s))
+			{
+				gds__prefix(tmp, s.c_str());
+				return tmp;
+			}
+			else
```
dyemanov, line 1246: Without `else` , please.

dyemanov:
I don't see any problems with backport into v5, provided it happens today.

src/common/utils.cpp
asfernandes: Should it be in (the outdated) `doc/README.coding.style`?
It's not there and personally I think it makes things less readable.

PR #7929, opened 2023-12-19
Fixed bug #7927 : Some default values is set incorrectly for SC\CS architectures

### LINKED ISSUE #7927: Some default values is set incorrectly for SC\CS architectures
Default values for `TempCacheLimit`, `DefaultDbCachePages` and `GCPolicy` settings is set as if `ServerMode = Super`, 
even if `ServerMode` is set to 'Classic' or 'SuperClassic'.

The issue is with default values only, if mentioned settings is set explicitly - correct values is used.

hvlad:


src/common/config/config.cpp
```diff
@@ -305,26 +306,45 @@ void Config::setupDefaultConfig()
...
+
+	ConfigValue* pDefault = &defaults[KEY_TEMP_CACHE_LIMIT];
+	ConfigValue* pValue = &values[KEY_TEMP_CACHE_LIMIT];
 	if (pDefault->intVal < 0)
 		pDefault->intVal = (serverMode != MODE_SUPER) ? 8388608 : 67108864;	// bytes
 
-	defaults[KEY_REMOTE_FILE_OPEN_ABILITY].boolVal = bootBuild;
+	if (pValue->intVal < 0)
+		pValue->intVal = pDefault->intVal;
+
```
asfernandes, line 328: Two blank lines inside a function body looks excessive (per function separation in this file is one line).

hvlad: It allows to visually separate handling of each key.
I can change blank line by additional comment, but will it be better ?

Issue #7823, opened 2023-11-02
ISQL command SHOW DATABASE crashes in absence of firebird.msg


aafemt:
Copy fbclient.dll and isql.exe into a separate folder. Connect to any remote database and issue command SHOW DATABASE. Application is aborted. If firebird.msg is also copied - it works.

PR #7936, opened 2023-12-29
Fix for #7823

aafemt:
Instead of putting this BOF off by simple expanding of output buffer I decided to fix it once for good.


aafemt:
Do you see something wrong on this PR?

hvlad (approved):
The only drawback I see is that `isqlGlob.printf()` flushes output - i.e. patch adds some performance penalty. But in this case it is negligible, I believe.

dyemanov (approved):

aafemt:
Fortunately fflush() has little penalty in console mode which is not buffered anyway so the only redirected output into file or pipe is affected.

Issue #7924, opened 2023-12-18
ALTER TABLE ALTER COLUMN <textual_field> can not be changed properly in some cases


pavel-zotov:
In some cases 'ALTER TABLE ... ALTER COLUMN <textual_field> type ... <new_charset>' can lead either to unchanged character set of <textual_field> or this character set is changed but COLLATE attribute gets value that can not be explained. Of course, following normally can be run only against empty tables:

**case-1**.
```
set bail on;
shell del g:\temp\tmp4test.fdb 2>nul;
create database 'localhost:g:\temp\tmp4test.fdb';
show version;
commit;
set echo on;
 
alter character set utf8 set default collation unicode_ci;
alter character set win1250 set default collation win_cz_ci_ai;
alter character set win1252 set default collation win_ptbr;
alter character set win1257 set default collation win1257_ee;
commit;
 
create table test(
    f01 varchar(10) character set win1252
   ,f02 varchar(10) character set win1257
   ,f03 varchar(10) character set utf8
);
commit;

---------------------------------------------------------------
 
commit;
connect 'localhost:g:\temp\tmp4test.fdb';
 
alter table test
    alter column f01 type varchar(10) character set win1250
   -----------------------------------------------------------
   ,alter column f02 type varchar(10) character set win1252
   -----------------------------------------------------------
   ,alter column f03 type varchar(10) character set win1257
;
commit;
connect 'localhost:g:\temp\tmp4test.fdb';
 
 
show table test; -------------------- [ 1 ]

/*
    Output: STRANGE. All remained unchanged:

    f01 ................. win1252 collate win_ptbr    // why not changed to win1250 collate win_cz_ci_ai ?
    f02 ................. win1257 collate win1257_ee  // why not changed to win1252 collate win_ptbr ?
    f03 ................. utf8    collate unicode_ci  // why not changed to win1257 collate win1257_ee ?

*/
```

**case-2**.
```
set bail on;
shell del g:\temp\tmp4test.fdb 2>nul;
create database 'localhost:g:\temp\tmp4test.fdb';
show version;
commit;
set echo on;
 
alter character set utf8 set default collation unicode_ci;
alter character set win1250 set default collation win_cz_ci_ai;
alter character set win1252 set default collation win_ptbr;
alter character set win1257 set default collation win1257_ee;
commit;
 
create table test(
    f_init_1252 varchar(10) character set win1252
   ,f_init_1257 varchar(10) character set win1257
[...]

PR #7940, opened 2024-01-01
Fix both parts of #7924

aafemt:


src/dsql/ddl.cpp
```diff
@@ -276,8 +276,9 @@ void DDL_resolve_intl_type(DsqlCompilerScratch* dsqlScratch, dsql_fld* field,
 		return;
 	}
 
-	if (modifying)
+	if (modifying && !field->charSet.hasData() && !field->collate.hasData())
```
asfernandes, line 279: `! .hasData()` -> `.isEmpty()`

aafemt: May be just `.empty()` then for STL compatibility?

src/jrd/MetaName.h
```diff
@@ -258,6 +258,11 @@ class MetaName
 		return !word;
 	}
 
+	bool empty() const
+	{
+		return !word;
+	}
+
```
dyemanov, line 265: I see no sense to have a duplication. If we're OK with STL-like `empty()`, then let's just rename `isEmpty()` across the codebase.

aafemt: It would be too big and unrelated change for this PR but I can prepare a separate one if agreed.

asfernandes:
What will happen if the field is used in an index and you change its collation?

aafemt:
Most likely some kind of separate issue which should be fixed separately.

asfernandes:
I don't think it is and actually I see more sense in not change the collation or throw error than have an index that will make queries return incorrect data.

aafemt:
But as I said, it is a separate issue. This PR is for #7924 only.

dyemanov (approved):

PR #7952, opened 2024-01-11
Fix for #7950

### LINKED ISSUE #7950: Unable to restore database when .fbk was created on host with other ICU
Ticket has been created after discussion with Alex.
After update Linux host where QA-job is performed to Debian 6.5.13-1 (2023-11-29), attempt to restore database which was created on previous Debian leads to "COLLATION UNICODE for CHARACTER SET UTF8 is not installed" message:
```
 /home/fbqa/qa-rundaily/unpacked-snapshot.tmp/bin/fbsvcmgr
 localhost/3340:service_mgr user SYSDBA password masterkey
 action_restore bkp_file
 /home/fbqa/qa-rundaily/results/fbt_outcomes_fb60_linux.fbk dbname
 /home/fbqa/qa-rundaily/results/fbt_outcomes_fb60_linux.fdb
```
Output:
```
 COLLATION UNICODE for CHARACTER SET UTF8 is not installed
 -An ICU library with collation version 153.88 is required but was not
 found. You may try to install ICU version 63.1, used to register the
 collation in this database or look for 'gfix -icu' in Firebird
 documentation.
 -Exiting before completion due to errors
```
( LI-T6.0.0.199 )

PS.
Part of this message ("look for 'gfix -icu' ...") definitely can not be considered applicable to this case because we operate here with .fbk rather than .fdb

AlexPeshkoff:
Unable to restore database when .fbk was created on host with other ICU

### DIFF
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -2737,11 +2737,11 @@ static void clearAttributes(BlobBuffer& attr)
 	map.remove(ATTR_COLL_VERSION);
 	if (map.remove(ATTR_ICU_VERSION))
 	{
-		string currentIcu = UnicodeUtil::getDefaultIcuVersion();
+		const string currentIcu = UnicodeUtil::getDefaultIcuVersion();
 		map.put(ATTR_ICU_VERSION, currentIcu);
 	}
 
-	string tmp = IntlUtil::generateSpecificAttributes(ascii, map);
+	const string tmp = IntlUtil::generateSpecificAttributes(ascii, map);
 	attr.assign(reinterpret_cast<const UCHAR*>(tmp.c_str()), tmp.length());
 }
 
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -2734,10 +2734,10 @@ static void clearAttributes(BlobBuffer& attr)
 	IntlUtil::SpecificAttributesMap map;
 	IntlUtil::parseSpecificAttributes(ascii, attr.getCount(), attr.begin(), &map);
 
-	map.remove("ICU-VERSION");
-	map.remove("COLL-VERSION");
+	map.remove(ATTR_ICU_VERSION);
+	map.remove(ATTR_COLL_VERSION);
 	string currentIcu = UnicodeUtil::getDefaultIcuVersion();
-	map.put("ICU-VERSION", currentIcu);
+	map.put(ATTR_ICU_VERSION, currentIcu);
 
 	string tmp = IntlUtil::generateSpecificAttributes(ascii, map);
 	attr.assign(reinterpret_cast<const UCHAR*>(tmp.c_str()), tmp.length());
--- a/src/burp/restore.epp
+++ b/src/burp/restore.epp
@@ -160,7 +160,8 @@ bool	get_sql_roles(BurpGlobals* tdgbl);
 bool	get_mapping(BurpGlobals* tdgbl);
 bool	get_db_creator(BurpGlobals* tdgbl);
 bool	get_security_class(BurpGlobals* tdgbl);
-void	get_source_blob(BurpGlobals* tdgbl, ISC_QUAD&, bool, void (*postProcess) (BlobBuffer& segment) = nullptr);
+void	get_source_blob(BurpGlobals* tdgbl, ISC_QUAD&, bool,
+			void (*postProcess) (BurpGlobals* tdgbl, BlobBuffer& segment) = nullptr);
 USHORT	get_text(BurpGlobals* tdgbl, TEXT*, ULONG);
 USHORT	get_text2(BurpGlobals* tdgbl, TEXT* text, ULONG length);
 bool	get_trigger(BurpGlobals* tdgbl);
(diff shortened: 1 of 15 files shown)


src/burp/restore.epp
```diff
@@ -2721,6 +2725,24 @@ bool get_chk_constraint(BurpGlobals* tdgbl)
...
+	IntlUtil::initAsciiCharset(&cs);
+	AutoPtr<CharSet> ascii(Firebird::CharSet::createInstance(*getDefaultMemoryPool(), 0, &cs));
+
+	IntlUtil::SpecificAttributesMap map;
+	IntlUtil::parseSpecificAttributes(ascii, attr.getCount(), attr.begin(), &map);
+
+	map.remove("ICU-VERSION");
+	map.remove("COLL-VERSION");
+	string currentIcu = UnicodeUtil::getDefaultIcuVersion();
+	map.put("ICU-VERSION", currentIcu);
```
asfernandes, line 2740: Isn't this going to add ICU tags even to non ICU collations?

src/burp/restore.epp
```diff
@@ -8906,6 +8928,12 @@ void get_source_blob(BurpGlobals* tdgbl, ISC_QUAD& blob_id, bool glb_trans)
 		--length; // -- or ++ ??? p is decremented, will have to test.
 		const USHORT seg_len = p - buffer;
 
+		if (postProcess)
+		{
+			postProcess(static_buffer);
```
asfernandes, line 8933: Looks like this is wrong place to have this, as it's inside a loop related to segments.

src/burp/restore.epp
AlexPeshkoff: Hops - will fix.

src/burp/restore.epp
AlexPeshkoff: On practice we never have specific attributes with >1 segment. Well, let's be generic & analyze all blob as a whole. Will do.

asfernandes:
I still have a doubt about this PR. Isn't this already being done in DFW (setupSpecificCollationAttributes)? What's wrong there, or, what is different here?

AlexPeshkoff:
What about setupSpecificCollationAttributes() in DFW - yes, it does the job, but it does not help in #7950 case. I.e. code in DFW is for databases, this one - for backups.

asfernandes:
I still don't understand. Restore creates database and collations, so why not DFW does it job?

Another weird thing is that the ticket says about UTF8:UNICODE, but this is system collation and AFAIK is not backed up.

AlexPeshkoff:
On 1/12/24 18:37, Adriano dos Santos Fernandes wrote:
>
> I still don't understand. Restore creates database and collations, so 
> why not DFW does it job?
>

Restore creates database but fails to create one of collations. This is 
user-defined collation, it's based on UNICODE but has numeric sort.
gbak writes backed-up data directly to system table and we get:

RDB$COLLATION_NAME NS_COLL
RDB$CHARACTER_SET_ID            4
RDB$COLLATION_ATTRIBUTES        1
RDB$SYSTEM_FLAG                 0
RDB$BASE_COLLATION_NAME UNICODE
RDB$SPECIFIC_ATTRIBUTES         1d:3cd
COLL-VERSION=153.88;ICU-VERSION=63.1;numeric-sort=1

setupSpecificCollationAttributes() keeps existing collation attributes 
when they exist, and sooner of all it's OK - we may have >1 ICU versions 
on a box and that behavior is a way to support this. In our case an 
attempt to start such collation fails.

I suppose it's possible to change that behavior during restore with some 
kind of DPB tag. Do you think it's better solution?

> Another weird thing is that the ticket says about UTF8:UNICODE, but 
> this is system collation and AFAIK is not backed up.
>

I did not look too much at error message cause did not plan to fix it. 
(Hmm - unfortunately our diags are well known to be bad...)
When analyzing what happens it's clear that problem is with user-defined 
collation NS_COLL.

asfernandes:
I have tried this branch with backup https://github.com/FirebirdSQL/firebird/files/13930774/FBT_OUTCOMES_FB60_WINDOWS.meta-only.fbk.zip and it fails to restore with the same error.

Then I tried master changing call of `setupSpecificCollationAttributes` in dfw.epp `create_collation` to use `dropIcuInfo = true` and it restored fine.

I don't think this is completelly correct (maybe in this case the attributes need to be reset only if load failed) but looks like a much simple solution using same technique as `DFW_reset_icu`.

Where I'm wrong?

AlexPeshkoff:
Did you use -icu switch when restoring? 

What about calling setupSpecificCollationAttributes() with dropIcuInfo == true - yes, that obviously makes gbak restore. Moreover, it can be safely used with remote restore (client has no ICU dependencies). But there some problems with that approach:
1. Always doing that is plain wrong - in some cases restore may be performed (for example) in order to move database to another host.
2. Automatically reset if load failed is wrong for same reason.
3. As I've already mentioned we may pass an option forcing attributes reset. Is it OK for you?

asfernandes:
> Did you use -icu switch when restoring?
> 

No, that is why it falling for me then.


> 3. As I've already mentioned we may pass an option forcing attributes reset. Is it OK for you?

I think it's a good option if it only resets if the original is problematic.

But I'm not sure a new option is necessary. Why not check `isc_dpb_gbak_attach`?

AlexPeshkoff:
As far as I can see createCollation() in DFW is called from 3 cases:
- create database (for system collations);
- create collation;
- restore.

In first 2 cases there are know ICU-specific attributes in it. So if we follow your suggestion - what a need in checking for gbak attach? It will always be true.
On the other hand I'm still unsure whether it's good idea to silently change ICU version on restore. In gfix we always follow policy of explicit upgrade, both for ICU version and minor ODS version. I highly prefer same approach here.

AlexPeshkoff:
OK, I can accept this argument - during restore we may always perform ICU ICU version autoupgrade.

What about why ICU version is stored in backup file - historically :-)

Issue #7962, opened 2024-01-16
System procedure/function inconsistency between ISQL SHOW FUNCTIONS and SHOW PROCEDURES


mrotteveel:
In Firebird 5.0.0.1306 ISQL, the `SHOW FUNCTIONS` command lists system functions in packages, while `SHOW PROCEDURES` correctly does not list system procedures in packages.

`SHOW FUNCTIONS` should be changed to exclude system functions (from `RDB$BLOB_UTIL`, `RDB$PROFILER` and `RDB$TIME_ZONE_UTIL`), as those should be queried with `SHOW SYSTEM FUNCTIONS`.

Reproduction:

```sql
create database 'e:\db\fb5\showfunc.fdb' user sysdba;
set term #;
-- including some functions and package functions for demonstration
create function dummy() returns integer as begin return 1; end#
CREATE OR ALTER PACKAGE APP_VAR
AS
BEGIN
   FUNCTION GET_DATEBEGIN() RETURNS DATE DETERMINISTIC;
   FUNCTION GET_DATEEND() RETURNS DATE DETERMINISTIC;
END#
set term ;#
```
Now, when executing `SHOW FUNCTIONS`, the output includes the system functions:

```none
SQL> show functions;
DUMMY
APP_VAR.GET_DATEBEGIN
APP_VAR.GET_DATEEND
RDB$BLOB_UTIL.IS_WRITABLE
RDB$BLOB_UTIL.NEW_BLOB
RDB$BLOB_UTIL.OPEN_BLOB
RDB$BLOB_UTIL.READ_DATA
RDB$BLOB_UTIL.SEEK
RDB$PROFILER.START_SESSION
RDB$TIME_ZONE_UTIL.DATABASE_VERSION
```

However, the expected output should exclude the system functions, that is, it should produce:

```none
DUMMY
APP_VAR.GET_DATEBEGIN
APP_VAR.GET_DATEEND
```

TreeHunter9:
I've fixed this bug before in RedDatabase, didn't know Firebird had the same issue. I can make PR with this fix.

AlexPeshkoff:
Please do - for both 5 & 6

TreeHunter9:
Made PR #7971, #7972 for both 5 and 6.

TreeHunter9:
I suppose this issue should be closed?

mrotteveel:
Please don't forget to add the fix-version tags (I did that now).

PR #7972, opened 2024-01-19
Fixed bug #7962: Remove system functions from isql SHOW FUNC output

TreeHunter9:


AlexPeshkoff:
As soon as you've started with bringing logic (and harmony) here please compare:

SQL> show func "RDB$BLOB_UTIL.SEEK";
There is no user-defined function RDB$BLOB_UTIL.SEEK in this database
SQL> show sys func "RDB$BLOB_UTIL.SEEK";
RDB$BLOB_UTIL.IS_WRITABLE
RDB$BLOB_UTIL.NEW_BLOB
RDB$BLOB_UTIL.OPEN_BLOB
RDB$BLOB_UTIL.READ_DATA
RDB$BLOB_UTIL.SEEK
RDB$PROFILER.START_SESSION
RDB$TIME_ZONE_UTIL.DATABASE_VERSION
SQL>

with:

SQL> show table "SEC$USERS";
SEC$USER_NAME                   (RDB$USER) CHAR(63) CHARACTER SET UTF8 Nullable 
SEC$FIRST_NAME                  (SEC$NAME_PART) VARCHAR(32) CHARACTER SET UTF8 Nullable 
SEC$MIDDLE_NAME                 (SEC$NAME_PART) VARCHAR(32) CHARACTER SET UTF8 Nullable 
SEC$LAST_NAME                   (SEC$NAME_PART) VARCHAR(32) CHARACTER SET UTF8 Nullable 
SEC$ACTIVE                      (RDB$BOOLEAN) BOOLEAN Nullable 
SEC$ADMIN                       (RDB$BOOLEAN) BOOLEAN Nullable 
SEC$DESCRIPTION                 (RDB$DESCRIPTION) BLOB segment 80, subtype TEXT CHARACTER SET UTF8 Nullable 
SEC$PLUGIN                      (RDB$PLUGIN) CHAR(63) CHARACTER SET UTF8 Nullable 
SQL>

I.e. there is no need to specify system table when providing particular system table name. I think same should work for functions.

TreeHunter9:
> I.e. there is no need to specify 'system' for table when providing particular system table name. I think same should work for functions.

Yea, that makes sense. I'll fix it.

TreeHunter9:
Btw, `show func "RDB$BLOB_UTIL.SEEK"` will not work, because it will think that `"RDB$BLOB_UTIL.SEEK"` is a function name. I thought about solution for this, maybe better to use syntax like this - `"RDB$BLOB_UTIL"."SEEK"`, but still it's kinda hard to parse, because we can have ether `"` and `.` in package/function name.
Anyway, this is not related to this PR, just my thoughts.

AlexPeshkoff:
> Btw, `show func "RDB$BLOB_UTIL.SEEK"` will not work, because it will think that `"RDB$BLOB_UTIL.SEEK"` is a function name. I thought about solution for this, maybe better to use syntax like this - `"RDB$BLOB_UTIL"."SEEK"`, but still it's kinda hard to parse, because we can have ether `"` and `.` in package/function name. Anyway, this is not related to this PR, just my thoughts.

Really unrelated. Btw, I do not see big problems with parsing it - double quote (") is already taken into an account (general rule is used - quoted text "QQ""ZZ" represents name QQ"ZZ), appears enough to add check for dot (.) after quoted text.

PR #7975, opened 2024-01-23
Get rid of the overflow header pages

dyemanov:
We support multiple header pages per every database file, such overflow header pages are allocated if stored clumplets do not fit the available header page space. However, this is practically impossible with the minimal page size of 4KB, even if we double the number of supported clumplet types. We already have some code that ignores overflow header pages and maintaining this feature makes some future development complicated. This PR cleans up support for overflow header pages.

### DIFF
--- a/src/jrd/ods.h
+++ b/src/jrd/ods.h
@@ -460,7 +460,7 @@ struct header_page
 	USHORT hdr_page_size;			// Page size of database
 	USHORT hdr_ods_version;			// Version of on-disk structure
 	ULONG hdr_PAGES;				// Page number of PAGES relation
-	ULONG hdr_next_page;			// Page number of next hdr page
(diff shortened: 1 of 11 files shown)


src/jrd/ods.h
```diff
@@ -460,7 +460,7 @@ struct header_page
 	USHORT hdr_page_size;			// Page size of database
 	USHORT hdr_ods_version;			// Version of on-disk structure
 	ULONG hdr_PAGES;				// Page number of PAGES relation
-	ULONG hdr_next_page;			// Page number of next hdr page
+	ULONG hdr_next_page;			// Page number of next hdr page (unused)
```
aafemt, line 463: I would rename this field to "unused" just to be sure that nothing in codebase refer it.

Issue #7568, opened 2023-05-06
Equivalence of boolean condition in partial index


mrotteveel:
I created a simple partial index example:

```
create table OFFER (
  OFFER_ID bigint generated always as identity primary key,
  PRODUCT_ID bigint not null,
  ARCHIVED boolean default false not null,
  PRICE decimal(9,2) not null
);

create unique index IDX_OFFER_UNIQUE_PRODUCT 
  on OFFER (PRODUCT_ID)
  where not ARCHIVED;
 
insert into OFFER(PRODUCT_ID, ARCHIVED, PRICE) values (1, false, 18.95);
insert into OFFER(PRODUCT_ID, ARCHIVED, PRICE) values (1, true, 17.95);
insert into OFFER(PRODUCT_ID, ARCHIVED, PRICE) values (1, true, 16.95);

-- following fails:
insert into OFFER(PRODUCT_ID, ARCHIVED, PRICE) values (1, false, 19.95);
```

While playing around with this, I noticed the following:

```
set plan on;
select * from offer where not archived;

PLAN (OFFER INDEX (IDX_OFFER_UNIQUE_PRODUCT))

             OFFER_ID            PRODUCT_ID ARCHIVED        PRICE
===================== ===================== ======== ============
                    1                     1 <false>         18.95

select * from offer where archived = false;

PLAN (OFFER NATURAL)

             OFFER_ID            PRODUCT_ID ARCHIVED        PRICE
===================== ===================== ======== ============
                    1                     1 <false>         18.95
```

In other words, the equivalent expression `archived = false` does not use the index. 

Then I made another index:

```
create index IDX_OFFER_ARCHIVED_PRODUCT 
  on OFFER (PRODUCT_ID) 
  where ARCHIVED;

select * from offer where archived;

PLAN (OFFER INDEX (IDX_OFFER_ARCHIVED_PRODUCT))

             OFFER_ID            PRODUCT_ID ARCHIVED        PRICE
===================== ===================== ======== ============
                    2                     1 <true>          17.95
                    4                     1 <true>          16.95

SQL> select * from offer where archived = true;

PLAN (OFFER INDEX (IDX_OFFER_ARCHIVED_PRODUCT))

             OFFER_ID            PRODUCT_ID ARCHIVED        PRICE
===================== ===================== ======== ============
                    2                     1 <true>          17.95
                    4                     1 <true>          16.95
```

[...]

PR #7987, opened 2024-01-30
Solution for #7568: Equivalence of boolean condition in partial index

dyemanov:
Patch consists of two parts:

1. Transform `NOT A = {FALSE | TRUE}` into `A = {TRUE | FALSE}` (with inverted literal) rather than keeping the condition as is or transforming it into `A <> {FALSE | TRUE}`
2. Detect equivalence of different types of comparisons involving (explicitly or implicitly) boolean literals

Using the same test case as in #7568:

```
create table OFFER (
  OFFER_ID bigint generated always as identity primary key,
  PRODUCT_ID bigint not null,
  ARCHIVED boolean default false not null,
  PRICE decimal(9,2) not null
);

create unique index IDX_OFFER_UNIQUE_PRODUCT 
  on OFFER (PRODUCT_ID)
  where not ARCHIVED;
```

all the following SQL queries are now able to use an index:

```
select * from offer where not archived;

select * from offer where archived = false;
select * from offer where false = archived;

select * from offer where archived <> true;
select * from offer where true <> archived;

select * from offer where not archived = true;
select * from offer where not true = archived;
```

### DIFF
--- a/src/dsql/ExprNodes.h
+++ b/src/dsql/ExprNodes.h
@@ -955,6 +955,12 @@ public:
 	virtual ValueExprNode* pass2(thread_db* tdbb, CompilerScratch* csb);
 	virtual dsc* execute(thread_db* tdbb, Request* request) const;
 
+	bool getBoolean() const
+	{
+		fb_assert(litDesc.dsc_dtype == dtype_boolean);
+		return *reinterpret_cast<bool*>(litDesc.dsc_address);
+	}
+
 	SLONG getSlong() const
 	{
 		fb_assert(litDesc.dsc_dtype == dtype_long);
--- a/src/dsql/ExprNodes.h
+++ b/src/dsql/ExprNodes.h
@@ -958,7 +958,7 @@ public:
 	bool getBoolean() const
 	{
 		fb_assert(litDesc.dsc_dtype == dtype_boolean);
-		return *reinterpret_cast<bool*>(litDesc.dsc_address);
+		return *litDesc.dsc_address != '\0';
 	}
 
 	SLONG getSlong() const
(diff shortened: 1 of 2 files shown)


src/dsql/ExprNodes.h
```diff
@@ -955,6 +955,12 @@ class LiteralNode final : public TypedNode<ValueExprNode, ExprNode::TYPE_LITERAL
 	virtual ValueExprNode* pass2(thread_db* tdbb, CompilerScratch* csb);
 	virtual dsc* execute(thread_db* tdbb, Request* request) const;
 
+	bool getBoolean() const
+	{
+		fb_assert(litDesc.dsc_dtype == dtype_boolean);
+		return *reinterpret_cast<bool*>(litDesc.dsc_address);
```
asfernandes, line 961: dtype_boolean is encoded in UCHAR, not `bool` (which has unspecified size).

```suggestion
		return *litDesc.dsc_address != '\0';
```

asfernandes (changes requested):

mrotteveel:
I'm not sure about the correctness for `IS [NOT] DISTINCT FROM`, given `archived is distinct from true` covers both `false` and `null`, and not just `false`.

mrotteveel:
(in the example it works because `archived` is marked not null, but this isn't generally applicable)

dyemanov:
Yes, it appears I was over-optimistic in this regard :-(
And AFAIU the same issue applies to `IS [NOT] {FALSE | TRUE}`, as `A = TRUE` and `A IS NOT FALSE` return different results for `A` being `NULL`.

dyemanov:
At the first glance I thought that `A = false` and `A is not distinct from false` could be considered equivalent, but they also return different values when `A IS NULL` -- `NULL vs FALSE` -- and even if the boolean condition handles them the same way I'm afraid this may strike us back in more complex expressions like `(A = false) is unknown`. So it appears for the safety reasons we should limit our optimization attempts with `=` and `<>` only.

dyemanov:
I've adjusted the patch and the description of possible replacements.

asfernandes (approved):

pavel-zotov:
QA issue: see test for #7568

PR #7988, opened 2024-01-30
Use CFLAGS/CPPFLAGS in extern/decNumber/Makefile

### LINKED ISSUE #7986: extern/decNumber/Makefile does not use CFLAGS
Versions:
firebird 5.0.0.1306
gcc 13.2.1

This means the resulting static library will be built without for example distribution hardening.  This results in all the firebird binaries that link to that static library not supporting hardening such as CET which requires all units to be built with it enabled.

loqs:
Resolves #7986

### DIFF
--- a/extern/decNumber/Makefile
+++ b/extern/decNumber/Makefile
@@ -2,7 +2,7 @@ LIBRARY=libdecFloat$(CROSS).a
 
 $(LIBRARY):		$(wildcard *.c) $(wildcard *.h) Makefile
 	$(RM) -f *.o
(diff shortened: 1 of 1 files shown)


loqs:
I dropped CPPFLAGS as they are not used in / passed from https://github.com/FirebirdSQL/firebird/blob/09b624f38d39a6531d2784236d84b3196046e64a/builds/posix/Makefile.in#L192 or https://github.com/FirebirdSQL/firebird/blob/09b624f38d39a6531d2784236d84b3196046e64a/builds/posix/Makefile.in#L392

PR #7990, opened 2024-01-31
Encapsulate GUID operations

dyemanov:
This PR hides some implementation details and makes operations with GUIDs more user-friendly.

### DIFF
--- a/src/common/os/guid.h
+++ b/src/common/os/guid.h
@@ -32,11 +32,11 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#include "../common/classes/fb_string.h"
+
 #ifdef WIN_NT
 #include <rpc.h>
 #else
-#include "fb_types.h"
-
 struct UUID	// Compatible with Win32 UUID struct layout
 {
 	ULONG Data1;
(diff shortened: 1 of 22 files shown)


src/common/os/guid.h
```diff
@@ -46,41 +46,184 @@ struct UUID	// Compatible with Win32 UUID struct layout
...
+	{
+		Guid guid;
+		GenerateGuid(&guid.m_data);
+		guid.m_initialized = true;
+		return guid;
+	}
+
+private:
+	UUID m_data;
+	bool m_initialized = false;
```
asfernandes, line 225: Why have `initialized` state instead of wrap it in `std::optional` when this functionality is required?

src/common/os/guid.h
```diff
@@ -46,41 +46,184 @@ struct UUID	// Compatible with Win32 UUID struct layout
...
-inline bool StringToGuid(Guid* guid, const char* buffer)
 {
-	const auto result = sscanf(buffer, GUID_FORMAT,
-		&guid->Data1, &guid->Data2, &guid->Data3,
-		&guid->Data4[0], &guid->Data4[1], &guid->Data4[2], &guid->Data4[3],
-		&guid->Data4[4], &guid->Data4[5], &guid->Data4[6], &guid->Data4[7]);
+	// Some versions of MSVC cannot recognize hh specifier but MSVC 2015 has it
+	const char* const GUID_FORMAT =
+		"{%08X-%04hX-%04hX-%02hhX%02hhX-%02hhX%02hhX%02hhX%02hhX%02hhX%02hhX}";
+	const int GUID_FORMAT_ARGS = 11;
```
asfernandes, line 70: Shouldn't these be static?

src/common/os/guid.h
dyemanov: Sorry, the original code comes from times when we didn't allow `std::optional` ;-) Will rework, just need to figure out what would be better - use `optional` as data or derive `Guid` from `optional`.

src/common/os/guid.h
dyemanov: Done

src/common/os/guid.h
dyemanov: Reworked.

src/jrd/replication/Config.cpp
```diff
@@ -391,8 +389,10 @@ void Config::enumerate(ReplicaList& replicas)
 				}
 				else if (key == "source_guid")
 				{
-					if (!config->sourceGuid.fromString(value))
+					Guid guid;
```
asfernandes, line 392: I didn't liked this too much ;)
I mean, the public constructor leaving the class with invalid data.

If I'd not going to add a throwable constructor, I would then use a `static std::optional<Guid> Guid::fromString(const char*)`.

dyemanov: This is doable of course, but the same problem exists with `dbb_guid` which by design always exists but has a delayed initialization (when the header page is read). Generating it always (and replacing afterwards) looks like wasting GUIDs for nothing ;-) Using `AutoPtr` for lazy initialization means an unnecessary memory allocation. Making it `std::optional` means that the whole `Guid` class should be made optional, due to default constructor behaviour, so we get back to my original code (just with `std::optional` hidden inside `Guid`).

asfernandes: > Making it `std::optional` means that the whole `Guid` class should be made optional, due to default constructor behaviour

`std::optional` does not require the class to have a default constructor.

asfernandes:
Can the default Guid constructor now be private?

asfernandes (approved):

PR #7991, opened 2024-02-01
Some enhancements for the file open modes (FS cache usage and forced writes)

dyemanov:
1. As now `UseFileSystemCache` remains the only rule whether filesystem cache should be used or not, it makes sense to consider it immediately when the database file is created/opened rather than adjusting it on demand (and reopen the file for that purpose).
2. The similar approach is also used for the FW mode - guess the likely mode before creating/opening the database file and adjust it later only if required.
3. Given point (1) above, it does not make sense to reopen the database file to change the FW mode on Linux anymore, `fcntl()` again becomes the minimalistic solution for that goal.
4. Some related cleanup.

### DIFF
--- a/src/jrd/os/posix/unix.cpp
+++ b/src/jrd/os/posix/unix.cpp
@@ -133,7 +133,7 @@ using namespace Firebird;
 static const mode_t MASK = 0660;
 
 static jrd_file* seek_file(jrd_file*, BufferDesc*, FB_UINT64*, FbStatusVector*);
-static jrd_file* setup_file(Database*, const PathName&, int, bool, bool, bool, bool, bool);
+static jrd_file* setup_file(Database*, const PathName&, int, USHORT);
 static void lockDatabaseFile(int& desc, const bool shareMode, const bool temporary,
 							 const char* fileName, ISC_STATUS operation);
 static bool unix_error(const TEXT*, const jrd_file*, ISC_STATUS, FbStatusVector* = NULL);
@@ -308,8 +308,13 @@ jrd_file* PIO_create(thread_db* tdbb, const PathName& file_name,
 	PathName expanded_name(file_name);
 	ISC_expand_filename(expanded_name, false);
 
-	return setup_file(dbb, expanded_name, desc, false, shareMode,
-					  forceWrite, notUseFSCache, onRawDevice);
+	const USHORT flags =
+		(shareMode ? FIL_sh_write : 0) |
+		(forceWrite ? FIL_force_write : 0) |
+		(notUseFSCache ? FIL_no_fs_cache : 0) |
+		(onRawDevice ? FIL_raw_device : 0);
+
+	return setup_file(dbb, expanded_name, desc, flags);
 }
 
 
@@ -769,8 +774,14 @@ jrd_file* PIO_open(thread_db* tdbb,
 	}
 #endif // SUPPORT_RAW_DEVICES
 
-	return setup_file(dbb, expandedName, desc, readOnly, shareMode,
-					  forceWrite, notUseFSCache, onRawDevice);
+	const USHORT flags =
+		(readOnly ? FIL_readonly : 0) |
+		(shareMode ? FIL_sh_write : 0) |
+		(forceWrite ? FIL_force_write : 0) |
+		(notUseFSCache ? FIL_no_fs_cache : 0) |
+		(onRawDevice ? FIL_raw_device : 0);
+
+	return setup_file(dbb, expandedName, desc, flags);
 }
 
 
@@ -970,14 +981,7 @@ static void maybeCloseFile(int& desc)
 }
 
 
-static jrd_file* setup_file(Database* dbb,
-							const PathName& file_name,
-							int desc,
-							bool readOnly,
-							bool shareMode,
-							bool forceWrite,
-							bool notUseFSCache,
-							bool onRawDevice)
+static jrd_file* setup_file(Database* dbb, const PathName& file_name, int desc, USHORT flags)
 {
 /**************************************
  *
@@ -996,18 +1000,8 @@ static jrd_file* setup_file(Database* dbb,
 		file = FB_NEW_RPT(*dbb->dbb_permanent, file_name.length() + 1) jrd_file();
 		file->fil_desc = desc;
 		file->fil_max_page = MAX_ULONG;
+		file->fil_flags |= flags;
 		strcpy(file->fil_string, file_name.c_str());
-
-		if (readOnly)
-			file->fil_flags |= FIL_readonly;
-		if (shareMode)
-			file->fil_flags |= FIL_sh_write;
-		if (forceWrite)
-			file->fil_flags |= FIL_force_write;
-		if (notUseFSCache)
-			file->fil_flags |= FIL_no_fs_cache;
-		if (onRawDevice)
-			file->fil_flags |= FIL_raw_device;
 	}
 	catch (const Exception&)
 	{
--- a/src/jrd/os/posix/unix.cpp
+++ b/src/jrd/os/posix/unix.cpp
@@ -124,7 +124,7 @@ using namespace Firebird;
 static const mode_t MASK = 0660;
 
 static jrd_file* seek_file(jrd_file*, BufferDesc*, FB_UINT64*, FbStatusVector*);
-static jrd_file* setup_file(Database*, const PathName&, int, bool, bool, bool, bool);
+static jrd_file* setup_file(Database*, const PathName&, int, bool, bool, bool, bool, bool);
 static void lockDatabaseFile(int& desc, const bool shareMode, const bool temporary,
 							 const char* fileName, ISC_STATUS operation);
 static bool unix_error(const TEXT*, const jrd_file*, ISC_STATUS, FbStatusVector* = NULL);
@@ -218,6 +218,7 @@ jrd_file* PIO_create(thread_db* tdbb, const PathName& file_name,
  *
  **************************************/
 	const auto dbb = tdbb->getDatabase();
+	const bool forceWrite = !temporary && (dbb->dbb_flags & DBB_force_write) != 0;
 	const bool notUseFSCache = !dbb->dbb_config->getUseFileSystemCache();
 	bool onRawDevice = false;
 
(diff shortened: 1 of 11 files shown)


src/jrd/os/win32/winnt.cpp
```diff
@@ -863,14 +874,18 @@ static jrd_file* setup_file(Database* dbb,
 		file->fil_max_page = MAX_ULONG;
 		strcpy(file->fil_string, file_name.c_str());
 
-		if (read_only)
+		if (readOnly)
 			file->fil_flags |= FIL_readonly;
 		if (shareMode)
 			file->fil_flags |= FIL_sh_write;
+		if (forceWrites)
```
hvlad, line 881: It should be `forceWrite` without last `s`

dyemanov: Checked it thrice and still something was missed :-( Sorry, to be fixed.

hvlad: Also, the MacOS build failed with:

> I/O error during "open O_CREAT" operation for file "/Users/runner/work/firebird/firebird/gen/security.tmp"
> -Error while trying to create file
> -No such file or directory
> 

Not sure if patch is the reason. thought.

dyemanov: Both should be fixed now.

hvlad:
Before this patch, when database is created, the order of operations was:
- database file is created with file creation mode set to FW=OFF
- all system relations is created and populated
- system indices is built
- if user specified FW=ON in DPB, then FW mode is switched to ON 
- file buffers is flushed to disk
- if user was not specified FW in DPB, then FW mode is switched to ON 

As you see, new database populated with data with FW=OFF - this was done to make process faster.

Now, with patch, database file will be created with FW=ON and this should slower database creation process.

PS there was complains about slow CREATE DATABASE at the early days of FB3, see thread "Reason for slow CREATE DATABASE in FB 3 (Linux)" in fb-devel for example

dyemanov:
Yes, this may be a problem. I see the following numbers (100 empty databases are created sequentially) with my patch:

NVMe SSD - 18s
SATA3 SSD - 23s
HDD - 33s

while master executes all of them in 13s. The difference is significant. I will revert the part of my changes related to `CREATE DATABASE`.

dyemanov:
BTW, I don't see any sense with different FW activation (before or after `CCH_flush`) depending on whether it was explicitly specified or not. Do I miss anything? And I see zero performance difference in my test between these two locations. So I'd prefer to have a single place but which one should I choose? After `CCH_flush` just to make sure we did our best from the performance POV? What do you think?

AlexPeshkoff (commented):
Main problem with PR is whether fcntl(F_SETFL, O_SYNC) in linux was fixed or not. In actual man (https://man7.org/linux/man-pages/man2/fcntl.2.html) I see it remains broken:

On Linux, this command can change  only the O_APPEND, O_ASYNC, O_DIRECT, O_NOATIME, and O_NONBLOCK flags.  It is not possible to change the O_DSYNC and O_SYNC flags.

The worst with this legacy behavior (i.e. unfixed bug) is that fcntl() reports success when one tries to do the change but O_SYNC is not actually changed.

To check actual state of the art one may compare performance in FW mode before and after this change - if performance grows that means fcntl() still broken.

AlexPeshkoff:
To be precise - taking into an account that PR also changes default file open mode to O_SYNC it should be better to check 4 performance measurements: before PR with FW on/off and after PR with FW on/off.

asfernandes:
> To be precise - taking into an account that PR also changes default file open mode to O_SYNC it should be better to check 4 performance measurements: before PR with FW on/off and after PR with FW on/off.

But does it have sense to do these checks if in some old kernel that user may install Firebird the behavior may be different and unsafe?

AlexPeshkoff:
Probably that makes sense. If currently behavior is fixed we will need to examine when did it happen and decide is it time to do changes in next FB version or not. We should not support too old workarounds for bugs.

hvlad:
> BTW, I don't see any sense with different FW activation (before or after `CCH_flush`) depending on whether it was explicitly specified or not. Do I miss anything? And I see zero performance difference in my test between these two locations. So I'd prefer to have a single place but which one should I choose? After `CCH_flush` just to make sure we did our best from the performance POV? What do you think?

I think you see no perf diffs due to `PIO_flush()` is not called when `DBB_creating` flag is set. 
Looks like one more thing to decide - should we forcible flush just created and fully initialized database or not ?
If yes, then note: `PIO_flush()` is called for FW=OFF only, i.e. flush should run before setting FW to ON.

dyemanov:
@AlexPeshkoff The comment in code said that FCNTL_BROKEN meant inability to successfully change BOTH O_DIRECT and O_SYNC using fcntl(). I got in in the way that one of them can be changed without problem. Now I need to test this deeper in Linux. Actually, I see no big problem to revert reopening back, this is not going to happen 2-3 times after the patch. But avoiding that looked like a good thing.

dyemanov:
> I think you see no perf diffs due to `PIO_flush()` is not called when `DBB_creating` flag is set. Looks like one more thing to decide - should we forcible flush just created and fully initialized database or not ? If yes, then note: `PIO_flush()` is called for FW=OFF only, i.e. flush should run before setting FW to ON.

As for me, the successful return from createDatabase() means the database is now consistent, there should not be any remaining parts in memory not flushed (and thus making the DB corrupted if the host process crashes after that). Maybe `gbak` restore could be the only exception, to avoid a regression for #5385.

aafemt:
An empty database is not much useful, applications usually create them for filling later. In this regard they are not different from gbak and I see no reason to deny high performance for them.

dyemanov:
I also have a feeling that if we use `fcntl()` to set (priorly missing) `O_SYNC`, then we need to call `fsync()` before that. Unless the filesystem cache was disabled.

AlexPeshkoff:
Dmitry, before deciding do we need fsync() before fcntl(O_SYNC) we should decide does it make sense or not to use fcntl(O_SYNC) at all. At least that did not work prior and explicitly denied in man currently.

dyemanov:
> Dmitry, before deciding do we need fsync() before fcntl(O_SYNC) we should decide does it make sense or not to use fcntl(O_SYNC) at all. At least that did not work prior and explicitly denied in man currently.

I'm going to restore the original code there. But as the `fcntl()` code will still be left under #ifdef, it makes sense to add the missing part. I don't want to forget it if one day that code will be enabled.

It seems that other Unixes (HPUX, Solaris, FreeBSD) are [not affected](https://bugzilla.kernel.org/show_bug.cgi?id=5994) by this issue, but this surely does not work for Linux and MacOS and we don't provide official builds for other Unix platforms, so I doubt we should enable `fcntl()` for them.

AlexPeshkoff:
On 2/2/24 16:53, Dmitry Yemanov wrote:
>
>     Dmitry, before deciding do we need fsync() before fcntl(O_SYNC) we
>     should decide does it make sense or not to use fcntl(O_SYNC) at
>     all. At least that did not work prior and explicitly denied in man
>     currently.
>
> I'm going to restore the original code there. But as the |fcntl()| 
> code will still be left under #ifdef, it makes sense to add the 
> missing part. I don't want to forget it if one day that code will be 
> enabled.
>

Yep, that makes sense. Reasonable OS should do that itself - but 
explicit call is better.

> It seems that other Unixes (HPUX, Solaris, FreeBSD) are not affected 
> <https://bugzilla.kernel.org/show_bug.cgi?id=5994> by this issue, but 
> this surely does not work for Linux and MacOS and we don't provide 
> official builds for other Unix platforms, so I doubt we should enable 
> |fcntl()| for them.
>

As a minimum that requires careful check. For FreeBSD testing is quite 
possible, but what about the rest...

src/jrd/os/win32/winnt.cpp
```diff
@@ -107,7 +107,7 @@ using namespace Firebird;
 
 static bool	maybeCloseFile(HANDLE&);
 static jrd_file* seek_file(jrd_file*, BufferDesc*, OVERLAPPED*);
-static jrd_file* setup_file(Database*, const Firebird::PathName&, HANDLE, bool, bool);
+static jrd_file* setup_file(Database*, const Firebird::PathName&, HANDLE, bool, bool, bool, bool);
```
AlexPeshkoff, line 110: Having 4 bool parameters in function makes code bad readable. Even 2 were not too good, but 4 is awful. Suggest to replace with set of flags like PIO_FORCE_WRITE, PIO_USE_FILESYSTEM_CACHE, etc and single flags parameter. 

Also please pay attention to avoiding flags like not_USE_FILESYSTEM_CACHE, flag named USE_FILESYSTEM_CACHE instead is much better readable.

AlexPeshkoff (changes requested):

src/jrd/os/win32/winnt.cpp
dyemanov: We already have corresponding `FIL_*` flags. But this change would mean that we need to convert from multiple bools to flags at least twice inside the code, before every `setup_file` call. So it would complicate the code, thus I avoided that. But I will reconsider.

As for `notSomething`, generally I agree with you, but this particular name was invented before me and I just preserved status quo. And it was added to handle a special and relatively rare case, so optional "NO" flag looked slightly better to me than [almost] mandatory "YES" flag. But I can change this too, if required. Any other opinions?

AlexPeshkoff (approved):

src/jrd/os/posix/unix.cpp
```diff
@@ -955,14 +1000,8 @@ static jrd_file* setup_file(Database* dbb,
 		file = FB_NEW_RPT(*dbb->dbb_permanent, file_name.length() + 1) jrd_file();
 		file->fil_desc = desc;
 		file->fil_max_page = MAX_ULONG;
+		file->fil_flags |= flags;
```
hvlad, line 1003: In winnt.cpp at similar place plain assignment used: `file->fil_flags = flags`
Is it intended difference ?
The final result is the same, I know ;)

dyemanov: Just an oversight, fixed.

hvlad (approved):

PR #8004, opened 2024-02-13
multi-character trim function

MrTeeett:
implementation of support for multi-character trim function


aafemt:
It would be nice to mention in documentation what these functions are supposed to do in addition to a bare syntax example. My best guess is that they cut off any pattern's string character from given string but it is better to know for sure.

hvlad:
We already have TRIM function and I see no good reason to introduce 3 additional similar functions.
Please, explain the difference and purpose of your ones.
Also, please, include just one final commit into PR. 
There is no need to explore whole history of feature development in such cases as this one.

src/include/firebird/impl/blr.h
```diff
@@ -346,6 +346,11 @@
 #define blr_trim_spaces		(unsigned char)0
 #define blr_trim_characters	(unsigned char)1
 
+/* multi-character trim for blr_trim */
+#define blr_trim_btrim		(unsigned char)3
+#define blr_trim_ltrim		(unsigned char)4
+#define blr_trim_rtrim		(unsigned char)5
+
```
dyemanov, line 353: AFAIU, logically LTRIM works similarly to TRIM(LEADING), RTRIM - to TRIM(TRAILING), BTRIM - to TRIM(BOTH). If so, why do we need separate BLR verbs? If you need to distinguish between old and multi-character behaviour, then maybe just add something like:
`#define blr_trim_multi_characters (unsigned char)2`
instead ?

MrTeeett: The functions ltrim, rtrim, btrim have their own implementation different from the standard trim function, they remove all characters no matter in what order they are specified.
Example:
LTRIM('baobab is a three', 'oiba')
is a three

aafemt: Isn't it the way standard TRIM() was supposed to work?

MrTeeett:
> We already have TRIM function and I see no good reason to introduce 3 additional similar functions. Please, explain the difference and purpose of your ones. Also, please, include just one final commit into PR. There is no need to explore whole history of feature development in such cases as this one.

These functions were added in the SQL:2023 -T056 standard
image

hvlad:
> > We already have TRIM function and I see no good reason to introduce 3 additional similar functions. Please, explain the difference and purpose of your ones. Also, please, include just one final commit into PR. There is no need to explore whole history of feature development in such cases as this one.
> 
> These functions were added in the SQL:2023 -T056 standard image

I have no copy of SQL 2023 and can't verify what was stated on the picture above. While I have no doubt that SQL 2023 have these new functions, we need to know exact specification.

@mrotteveel: Mark, could you check it, please ?

src/include/firebird/impl/blr.h
hvlad: According to my old copy of old SQL 2008, standard TRIM allows single character values only in \<trim character\>:
```
11) If <trim function> is specified, then:
  a) Let S be the value of the <trim source>.
  b) If <trim character> is specified, then let SC be the value of <trim character>; otherwise, let SC be 
     <space>.
  c) If either S or SC is the null value, then the result of the <trim function> is the null value.
  d) If the length in characters of SC is not 1 (one), then an exception condition is raised: data exception —
     trim error.
```

hvlad: > AFAIU, logically LTRIM works similarly to TRIM(LEADING), RTRIM - to TRIM(TRAILING), BTRIM - to TRIM(BOTH). If so, why do we need separate BLR verbs? If you need to distinguish between old and multi-character behaviour, then maybe just add something like: `#define blr_trim_multi_characters (unsigned char)2` instead ?

I like it.

> The functions ltrim, rtrim, btrim have their own implementation different from the standard trim function, they remove all characters no matter in what order they are specified.

Modification of existing TrimNode is hard to name "own implementation" :)
And it have no relation with correct (better) usage of BLR verb's proposed above.

mrotteveel:
Some excerpts from SQL:2023-2:

From 4.3.3.2 Operators that operate on character strings and return character strings

> \<single-character trim function\> is a function that returns its first string argument with leading and/or
trailing pad characters removed. The second argument indicates whether leading, or trailing, or both
leading and trailing pad characters shall be removed. The third argument specifies the pad character that
is to be removed.
>
> \<multi-character trim function\> is a function that returns its first string argument with leading pad
characters, trailing pad characters, or both leading and trailing pad characters removed. The second
argument specifies the pad characters that are to be removed. In contrast to the \<single-character trim
function\>, this function accepts a sequence of characters as the second argument and will remove all
leading, trailing, or both occurrences of any of these characters, regardless of their ordering.

From 6.33 \<string value function\>

> ```none
> <trim function> ::=
>       <single-character trim function>
>     | <multi-character trim function>
> 
> <single-character trim function> ::=
>     TRIM <left paren> <trim operands> <right paren>
> 
> <multi-character trim function> ::=
>     { BTRIM | LTRIM | RTRIM }
>       <left paren> <trim source> [ <comma> <trim character> ] <right paren>
> 
> <trim operands> ::=
>     [ [ <trim specification> ] [ <trim character> ] FROM ] <trim source>
> 
> <trim source> ::=
>     <character value expression>
> 
> <trim specification> ::=
>       LEADING
>     | TRAILING
>     | BOTH
> 
> <trim character> ::=
>     <character value expression>
> ```

And syntax rule (only quoting for multi-character as we already implement single character trim):

> 15\) If <multi-character trim function> MTF is specified, then:
> &nbsp;&nbsp;a) Case:
> &nbsp;&nbsp;&nbsp;&nbsp;i) If the declared type of \<character value expression\> simply contained in \<trim source\> TS is fixed-length character string or variable-length character string, then the declared type of MTF is variable-length character string with maximum length equal to the length or maximum length of TS.
> &nbsp;&nbsp;&nbsp;&nbsp;ii) Otherwise, the declared type of MTF is a character large object type with maximum length equal to the maximum length of TS.
> &nbsp;&nbsp;b) Case:
> &nbsp;&nbsp;&nbsp;&nbsp;i) If a <trim character> TC is specified, then:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1) TS and TC shall be comparable.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2) If BTRIM is specified, then:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;A) TC shall not generally contain a \<routine invocation\> whose subject routine is an SQL-invoked routine that is possibly non-deterministic or that possibly modifies SQL-data.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;B) TC shall not generally contain a \<table primary\> that contains a \<data change delta table\>.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;C) BTRIM ( TS, TC ) is equivalent to RTRIM ( LTRIM ( TS, TC ), TC )
> &nbsp;&nbsp;&nbsp;&nbsp;ii) Otherwise:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1) Let TC be \<space\>.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2) If BTRIM is specified, then BTRIM ( TS ) is equivalent to RTRIM ( LTRIM ( TS ) ).
> &nbsp;&nbsp;c) The character set and collation of MTF are those of TS.

And general rule (again, only quoting for multi-character);

> 13\) If \<multi-character trim function\> MTF is specified, then:
> &nbsp;&nbsp;a) Let S be the value of TS.
> &nbsp;&nbsp;b) Let SC be the value of TC.
> &nbsp;&nbsp;c) If at least one of S and SC is the null value, then the result of MTF is the null value and no further General Rules of this Subclause are applied.
> &nbsp;&nbsp;d) If S is the zero-length character string or SC is the zero-length character string, then the result of MTF is S and no further General Rules of this Subclause are applied.
> &nbsp;&nbsp;e) Case:
> &nbsp;&nbsp;&nbsp;&nbsp;i) If RTRIM is specified:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1) Let P be the length in characters of S.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2) Let C be the P-th character of S.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;3) While C is contained in SC, do:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;A) Let P be P − 1 (one).
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;B) Let C be the P-th character of S.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;4) Let LP be the <exact numeric literal> that has the value P.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;5) The result of MTF is SUBSTRING ( TS FROM 1 FOR LP )
> &nbsp;&nbsp;&nbsp;&nbsp;ii) If LTRIM is specified:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1) Let P be 1 (one).
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2) Let C be the P-th character of S.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;3) While C is contained in SC, do:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;A) Let P be P + 1 (one).
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;B) Let C be the P-th character of S.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;4) Let LP be the <exact numeric literal> that has the value P.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;5) The result of MTF is SUBSTRING ( TS FROM LP )

mrotteveel:
Note that the screenshot did not come from the SQL standard itself, but only refers to this feature of the standard.

hvlad:
> Some excerpts from SQL:2023-2:
...

Thanks !

> Note that the screenshot did not come from the SQL standard itself, but only refers to this feature of the standard.

This was a reason to ask for more formal description.

asfernandes:
@mrotteveel are they reserved words?

src/include/firebird/impl/blr.h
```diff
@@ -346,6 +346,11 @@
 #define blr_trim_spaces		(unsigned char)0
 #define blr_trim_characters	(unsigned char)1
 
+/* multi-character trim for blr_trim */
+#define blr_trim_btrim		(unsigned char)3
+#define blr_trim_ltrim		(unsigned char)4
+#define blr_trim_rtrim		(unsigned char)5
```
asfernandes, line 352: Please join this as following:

```
#define blr_trim			(unsigned char)183

/* first sub parameter for blr_trim */
#define blr_trim_both		(unsigned char)0
#define blr_trim_leading	(unsigned char)1
#define blr_trim_trailing	(unsigned char)2
#define blr_trim_btrim		(unsigned char)3
#define blr_trim_ltrim		(unsigned char)4
#define blr_trim_rtrim		(unsigned char)5

/* second sub parameter for blr_trim */
#define blr_trim_spaces		(unsigned char)0
#define blr_trim_characters	(unsigned char)1
```

src/include/firebird/impl/blr.h
asfernandes: > Isn't it the way standard TRIM() was supposed to work? At least in MS SQL it works this way.

Our `TRIM` allowed multi-character string incompatible with the new ones now defined in the standard.

mrotteveel:
> @mrotteveel are they reserved words?

Yes, `BTRIM`, `LTRIM` and `RTRIM` are defined as a reserved word in SQL:2023.

PR #8016, opened 2024-02-22
Free memory issued for isql command list but has never been freed on output file write

AlexeyMochalov:
Such kind of a leak is not a big concern (memory will be freed on isql exit anyway), but it interferes with isql.memdebug.log memory leaks log.

### DIFF
--- a/src/isql/InputDevices.h
+++ b/src/isql/InputDevices.h
@@ -133,8 +133,6 @@ inline InputDevices::~InputDevices()
 	for (unsigned n = 0; n < commands.getCount(); ++n)
 		delete commands[n];
 
-	commands.clear();
-
 	clear();
 }
 
--- a/src/isql/InputDevices.h
+++ b/src/isql/InputDevices.h
@@ -130,6 +130,11 @@ inline InputDevices::InputDevices(Firebird::MemoryPool& p)
 
 inline InputDevices::~InputDevices()
 {
+	for (unsigned n = 0; n < commands.getCount(); ++n)
+		delete commands[n];
+
+	commands.clear();
+
 	clear();
 }
 


src/isql/InputDevices.h
```diff
@@ -130,6 +130,11 @@ inline InputDevices::InputDevices(Firebird::MemoryPool& p)
 
 inline InputDevices::~InputDevices()
 {
+	for (unsigned n = 0; n < commands.getCount(); ++n)
+		delete commands[n];
```
asfernandes, line 134:
```suggestion
	for (auto command : commands)
		delete command;
```

aafemt: `Command*` is an expression simple enough, I see no need for `auto` here.

aafemt: `auto` is bad for readability. It is justified if the iterator is a multilevel template but not in this case.

asfernandes: I made a comment that improves the original code imo. I'll not keep arguing about this minor thing with who is know to argue with every minor things.

src/isql/InputDevices.h
```diff
@@ -130,6 +130,11 @@ inline InputDevices::InputDevices(Firebird::MemoryPool& p)
...
+
+	commands.clear();
```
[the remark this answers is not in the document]
AlexeyMochalov: agree, it's completely redundant

src/isql/InputDevices.h
AlexeyMochalov: Thanks for review. The only reason I chose a non range-based for loop without auto is because I wanted to keep the same code for all major versions, including FB 3.0 (that could be built with C++ 2003 on windows)

Am I wrong in such approach?

dyemanov (approved):

PR #8046, opened 2024-03-17
Postfix for #7989 - Improve performance of external (UDR) functions

asfernandes:
Add missing profiling, savepoint and stack traces.

It became slower than what initially done in https://github.com/FirebirdSQL/firebird/pull/7989, but still much faster than v5.

I edited new timings verified in test in https://github.com/FirebirdSQL/firebird/pull/7989.

### DIFF
--- a/src/jrd/ExtEngineManager.cpp
+++ b/src/jrd/ExtEngineManager.cpp
@@ -898,11 +898,6 @@ ExtEngineManager::Function::~Function()
 void ExtEngineManager::Function::execute(thread_db* tdbb, Request* request, jrd_tra* transaction,
 	unsigned inMsgLength, UCHAR* inMsg, unsigned outMsgLength, UCHAR* outMsg) const
 {
-	AutoSetRestore2<Request*, thread_db> autoSetRequest(
-		tdbb, &thread_db::getRequest, &thread_db::setRequest, request);
-
-	EXE_activate(tdbb, request, transaction);
-
 	fb_assert(inMsgLength == udf->getInputFormat()->fmt_length);
 	fb_assert(outMsgLength == udf->getOutputFormat()->fmt_length);
 
--- a/src/jrd/exe.cpp
+++ b/src/jrd/exe.cpp
@@ -955,7 +955,9 @@ void EXE_execute_function(thread_db* tdbb, Request* request, jrd_tra* transactio
 
 			const auto profilerManager = attachment->isProfilerActive() && !request->hasInternalStatement() ?
 				attachment->getProfilerManager(tdbb) : nullptr;
-			SINT64 profilerInitialTicks = profilerManager ? profilerManager->queryTicks() : 0;
+			const SINT64 profilerInitialTicks = profilerManager ? profilerManager->queryTicks() : 0;
+			const SINT64 profilerInitialAccumulatedOverhead = profilerManager ?
+				profilerManager->getAccumulatedOverhead() : 0;
 
 			try
 			{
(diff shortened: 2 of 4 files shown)


src/jrd/exe.cpp
```diff
@@ -258,6 +258,8 @@ static void execute_looper(thread_db*, Request*, jrd_tra*, const StmtNode*, Requ
 static void looper_seh(thread_db*, Request*, const StmtNode*);
 static void release_blobs(thread_db*, Request*);
 static void trigger_failure(thread_db*, Request*);
+static void savepointForget(thread_db* tdbb, Request* request, jrd_tra* transaction, SavNumber savNumber);
+static SavNumber savepointStart(Request* request, jrd_tra* transaction);
```
dyemanov, line 262: I'd move them to an anonymous namespace, IMHO new functions should not use the old static approach.
And I always prefer naming `doSomething` rather than `somethingDo` -- i.e. `startSavepoint()` and `forgetSavepoint()`.

src/jrd/ExtEngineManager.cpp
```diff
@@ -898,11 +898,6 @@ ExtEngineManager::Function::~Function()
 void ExtEngineManager::Function::execute(thread_db* tdbb, Request* request, jrd_tra* transaction,
 	unsigned inMsgLength, UCHAR* inMsg, unsigned outMsgLength, UCHAR* outMsg) const
 {
-	AutoSetRestore2<Request*, thread_db> autoSetRequest(
-		tdbb, &thread_db::getRequest, &thread_db::setRequest, request);
```
dyemanov, line 902: I don't see this code moved elsewhere, is it not required anymore?

src/jrd/exe.cpp
```diff
@@ -917,10 +920,140 @@ void EXE_activate(thread_db* tdbb, Request* request, jrd_tra* transaction)
...
+			fb_assert(!request->req_caller);
+			request->req_caller = exeState.oldRequest;
+
+			tdbb->tdbb_flags &= ~(TDBB_stack_trace_done | TDBB_sys_error);
+
+			// Execute stuff until we drop
+
+			const auto profilerManager = attachment->isProfilerActive() && !request->hasInternalStatement() ?
+				attachment->getProfilerManager(tdbb) : nullptr;
+			SINT64 profilerInitialTicks = profilerManager ? profilerManager->queryTicks() : 0;
```
dyemanov, line 958: `const` ?

src/jrd/exe.cpp
```diff
@@ -917,10 +920,140 @@ void EXE_activate(thread_db* tdbb, Request* request, jrd_tra* transaction)
...
+				TRA_release_request_snapshot(tdbb, request);
+
+			request->req_flags &= ~(req_active | req_reserved);
+			request->invalidateTimeStamp();
+
+			if (profilerInitialTicks && attachment->isProfilerActive())
+			{
+				ProfilerManager::Stats stats(request->req_profiler_ticks);
+				profilerManager->onRequestFinish(request, stats);
+			}
```
dyemanov, line 1000: I don't see `profilerInitialTicks` accounted in the profiler stats. Why was it requested above? I'd expect some calculation of the final elapsed ticks and `req_profiler_ticks` being incremented before it's passed to `onRequestFinish()`.

src/jrd/exe.cpp
```diff
@@ -917,10 +920,140 @@ void EXE_activate(thread_db* tdbb, Request* request, jrd_tra* transaction)
...
+			const auto profilerManager = attachment->isProfilerActive() && !request->hasInternalStatement() ?
+				attachment->getProfilerManager(tdbb) : nullptr;
+			SINT64 profilerInitialTicks = profilerManager ? profilerManager->queryTicks() : 0;
+
+			try
+			{
+				JRD_reschedule(tdbb);
+
+				function->fun_external->execute(tdbb, request, transaction, inMsgLength, inMsg, outMsgLength, outMsg);
+			}
```
dyemanov, line 965: Even if `att_cancel_lock` is active and cancellation request has been posted to our attachment, the external function cannot be cancelled during its execution. But maybe we should call `tdbb->checkCancelState()` immediately after its execution to react quickly? And maybe `JRD_reschedule()` will be not needed after that change, given that we always check out the engine during function invocation.

dyemanov (changes requested):

src/jrd/ExtEngineManager.cpp
asfernandes: It's done inside `StmtNode::ExeState` constructor.

src/jrd/exe.cpp
asfernandes: Fixed, thanks.

src/jrd/exe.cpp
asfernandes: Please see if it's now correct.

dyemanov (approved):

PR #8047, opened 2024-03-18
Wipe out multi-file database support

dyemanov:
I'm not sure it should be applied to v6 as it supports databases in ODS 13.0 / 13.1 which may be multi-file and the engine will not be able to work with them. The patch raises an error in this case and suggests backup/restore to convert the database into a single file. Another possible reason for delay is that tablespaces are not landed yet. I have no problems merging this PR later.


aafemt:
For v6 I would suggest to drop ability to create multifile database but keep ability to open and work with multifile database until the next major ODS.

sim1984:
> For v6 I would suggest to drop ability to create multifile database but keep ability to open and work with multifile database until the next major ODS.

Don't need this. For this purpose there is EngineXX of previous versions.

dyemanov:
XX will be the same in v4/v5 and v6, this makes things confusing. Different engines are mostly intended for different major ODS versions.

sim1984:
Not anymore. In Firebird 6.0, now the new major version ODS 14.0.

dyemanov:
Yes, but tablespaces are still in the pipeline. Does anyone object to perform the cleanup now?

dyemanov:
Updated, ready for merge.

PR #8049, opened 2024-03-18
Fixed #8048: Crash with SIGBUS when there is no free space on the partition with memory-mapped files

### LINKED ISSUE #8048: Crash with SIGBUS when there is no free space on the partition with memory-mapped files
The crash occurs because the memory-mapped files are extended using ftruncate and it doesn't allocate space on disk for the new file blocks. And later when we try to use the extended space without underlying disk blocks on 100%-filled partition we get SIGBUS error.
Here are some explanations: https://mapdb.org/blog/mmap_file_and_jvm_crash/.
It can be reproduced for example by mounting tmpfs filesystem with a limited size (10 Mb) in /tmp/firebird directory.
After a certain load, the engine will try to extend the memory-mapped file (usually lock or monitoring). This will not cause an error, although there is no disk space in /tmp/firebird for it. But later on an attempt to use the extended memory (memset, memcpy, trivial assignment) the server will crash with SIGBUS.
The problem can be solved by allocating disk blocks for extended space:
https://uk.comp.os.linux.narkive.com/Ve44sO4i/shared-memory-problem-no-space-at-dev-shm-causes-sigbus
Here it is suggested to write zeros to the end of the file, but as testing has shown, fallocate also works.
Windows doesn't have this problem, SetFilePointer call returns error 112 (There is not enough space on the disk.).

dmitry-starodubov:


aafemt:
According to manpages fallocate() is available starting from Linux 2.6.23 and glibc 2.10 i.e. far behind of versions that Firebird officially support. Is there point in HAVE_FALLOCATE check?

AlexPeshkoff:
@aafemt we want to support not only linux:
NAME
       fallocate - manipulate file space
DESCRIPTION
       This is a nonportable, Linux-specific system call...

aafemt:
The code is inside of `#ifdef UNIX`. Other platforms have their own defines.

dyemanov:
`#ifdef UNIX` is not only Linux

AlexPeshkoff:
On 3/18/24 17:05, Dimitry Sibiryakov wrote:
>
> The code is inside of |#ifdef UNIX|. Other platforms have their own 
> defines.
>

Not at all. Mac & FreeBSD are UNIX family.

aafemt:
> This is a nonportable, Linux-specific system call.  For the portable, POSIX.1-specified method of ensuring that space is allocated for a file, see [posix_fallocate(3)](https://man7.org/linux/man-pages/man3/posix_fallocate.3.html).

AlexPeshkoff:
On 3/18/24 17:12, Dimitry Sibiryakov wrote:
>
>     This is a nonportable, Linux-specific system call. For the
>     portable, POSIX.1-specified method of ensuring that space is
>     allocated for a file, see posix_fallocate(3)
>     <https://man7.org/linux/man-pages/man3/posix_fallocate.3.html>.
>

That may look strange, but implementation of posix_fallocate 
<https://man7.org/linux/man-pages/man3/posix_fallocate.3.html> (at least 
in glibc) is far less efficient compared with writing big buffer of 
zero's. It performs write() system call for every 1K block compared with 
much less system calls in out case.

Issue #7699, opened 2023-08-04
Incorrect isql -x script extraction for external procedures with no output parameters.


sim1984:
Let's say we have two external procedures, one selective and one non-selective, written as follows:

```cpp
/***
create procedure insert_row (
	n integer not null
)
	external name 'isql_udr_bug!insert_row'
	engine udr;
***/
FB_UDR_BEGIN_PROCEDURE(insert_row)
	FB_UDR_MESSAGE(InMessage,
		(FB_INTEGER, n)
	);

	FB_UDR_EXECUTE_PROCEDURE
	{
		AutoRelease<IAttachment> att(context->getAttachment(status));
		AutoRelease<ITransaction> tra(context->getTransaction(status));
		AutoRelease<IMessageMetadata> inMetadata(this->procedure->metadata->getInputMetadata(status));
		att->execute(
			status,
			tra,
			0,
			"INSERT INTO test_table(n) VALUES(?)",
			3,
			inMetadata,
			in,
			nullptr,
			nullptr
		);
	}

	FB_UDR_FETCH_PROCEDURE
	{
		return false;
	}
FB_UDR_END_PROCEDURE


/***
create procedure gen_rows (
	start_n integer not null,
	end_n integer not null
) returns (
	n integer not null
)
	external name 'isql_udr_bug!gen_rows'
	engine udr;
***/
FB_UDR_BEGIN_PROCEDURE(gen_rows)
	FB_UDR_MESSAGE(InMessage,
		(FB_INTEGER, start)
		(FB_INTEGER, end)
	);

	FB_UDR_MESSAGE(OutMessage,
		(FB_INTEGER, result)
	);

	FB_UDR_EXECUTE_PROCEDURE
	{
		out->resultNull = FB_FALSE;
		out->result = in->start - 1;
	}

	FB_UDR_FETCH_PROCEDURE
	{
		return out->result++ < in->end;
	}
FB_UDR_END_PROCEDURE
```

Now let's create a new database using these procedures:

```sql
create database 'inet://localhost:3054/D:\fbdata\4.0\isql_bug.fdb'
user SYSDBA password 'masterkey'
default character set utf8;

create table test_table(
   id bigint generated by default as identity,
   n integer,
   constraint pk_test_table primary key(id)
);

create procedure insert_row (
	n integer not null
)
external name 'isql_udr_bug!insert_row'
engine udr;


create procedure gen_rows (
	start_n integer not null,
	end_n integer not null
) returns (
[...]

PR #8054, opened 2024-03-19
Extract external routines in one pass

vasiliy-yashkov:
Extract external procedures and functions in one pass because they have no BLR and cannot have dependencies.

### DIFF
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -2941,7 +2941,7 @@ bool CreateAlterProcedureNode::executeAlter(thread_db* tdbb, DsqlCompilerScratch
 				if (secondPass)
 				{
 					P.RDB$PROCEDURE_TYPE.NULL = FALSE;
-					P.RDB$PROCEDURE_TYPE = (USHORT) prc_selectable;
+					P.RDB$PROCEDURE_TYPE = returns.getCount() != 0 ? (USHORT) prc_selectable : (USHORT) prc_executable;
 				}
 				else
 				{
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -2941,7 +2941,7 @@ bool CreateAlterProcedureNode::executeAlter(thread_db* tdbb, DsqlCompilerScratch
 				if (secondPass)
 				{
 					P.RDB$PROCEDURE_TYPE.NULL = FALSE;
-					P.RDB$PROCEDURE_TYPE = returns.getCount() != 0 ? (USHORT) prc_selectable : (USHORT) prc_executable;
+					P.RDB$PROCEDURE_TYPE = (USHORT) prc_selectable;
 				}
 				else
 				{
--- a/src/isql/extract.epp
+++ b/src/isql/extract.epp
@@ -94,6 +94,8 @@ static void list_package_headers();
 static void list_procedure_bodies(SSHORT default_char_set_id);
 static void list_procedure_headers(SSHORT default_char_set_id);
 static void list_views();
+static void list_external_procedures(SSHORT default_char_set_id);
+static void list_external_functions(SSHORT default_char_set_id);
 
 static const char* const Procterm = "^";	// TXNN: script use only
 
@@ -179,6 +181,8 @@ int EXTRACT_ddl(LegacyTables flag, const SCHAR* tabname)
 		list_domains(default_char_set_id);
 		list_all_tables(flag, default_char_set_id);
 		list_functions_legacy();
+		list_external_functions(default_char_set_id);
+		list_external_procedures(default_char_set_id);
 		list_functions_ods12_headers(default_char_set_id);
 		list_procedure_headers(default_char_set_id);
 		list_package_headers();
@@ -1509,6 +1513,8 @@ static void list_procedure_headers(SSHORT default_char_set_id)
 	FOR PRC IN RDB$PROCEDURES
 		WITH (PRC.RDB$SYSTEM_FLAG NE 1 OR PRC.RDB$SYSTEM_FLAG MISSING)
 		AND PRC.RDB$PACKAGE_NAME MISSING
+		AND PRC.RDB$ENGINE_NAME MISSING
+		AND PRC.RDB$ENTRYPOINT MISSING
 		SORTED BY PRC.RDB$PROCEDURE_NAME
 		if (header)
 		{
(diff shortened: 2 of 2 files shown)


src/dsql/DdlNodes.epp
```diff
@@ -2941,7 +2941,7 @@ bool CreateAlterProcedureNode::executeAlter(thread_db* tdbb, DsqlCompilerScratch
 				if (secondPass)
 				{
 					P.RDB$PROCEDURE_TYPE.NULL = FALSE;
-					P.RDB$PROCEDURE_TYPE = (USHORT) prc_selectable;
+					P.RDB$PROCEDURE_TYPE = returns.getCount() != 0 ? (USHORT) prc_selectable : (USHORT) prc_executable;
```
aafemt, line 2944: Are you sure that procedure without output parameters cannot be selectable?
`select count(*) from anyUDR()`

vasiliy-yashkov: >  `select count(*) from anyUDR()`

This causes ` SQL error code = -84; procedure TEST_UDR does not return any values`. Even if the procedure has `prc_selectable` type.

hvlad, line 2944: How this change is related with PR purposes ?

src/isql/extract.epp
```diff
@@ -3594,3 +3568,172 @@ static void list_views()
...
+ *	l i s t _ e x t e r n a l _ p r o c e d u r e s
+ *
+ **************************************
+ *
+ * Functional description
+ * Create all external procedures declarations.
+ *
+ **************************************/
+
+	fb_assert(isqlGlob.major_ods >= ODS_VERSION12);
```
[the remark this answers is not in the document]
dyemanov: Current ISQL (FB 3-4-5) cannot extract procedures/functions from pre-ODS12 databases, so nothing really changed.

src/dsql/DdlNodes.epp
vasiliy-yashkov: > How this change is related with PR purposes ?

In fact, it's unrelated. Should I delete this change?

dyemanov: A remainder of the older patch. AFAIU, it should not be necessary once external routines are extracted in one pass. @vasiliy-yashkov do you agree?

hvlad: Sure. If it is a part of some other bug fix - feel free to create specific PR.

vasiliy-yashkov: > A remainder of the older patch. AFAIU, it should not be necessary once external routines are extracted in one pass. @vasiliy-yashkov do you agree?

I agree.

vasiliy-yashkov: > Sure. If it is a part of some other bug fix - feel free to create specific PR.

Ok :)

src/isql/extract.epp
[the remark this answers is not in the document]
hvlad: Both `list_external_procedures()` and `list_external_functions()` is called unconditionally by `EXTRACT_ddl()`, see

https://github.com/FirebirdSQL/firebird/pull/8054/files#diff-580a2883156750bdebc94df7a24f090f8b288b49558292ae66fde9b21cbbc432R184

so, what happens in case of pre-ODS13 DB ? 
Error will be raised ? 
Nothing, as query below returns no rows ?
Why assert then ? It will not make DEBUG build happy, AFAIU.
I just want it to be clear.

asfernandes:
I think this is going to cause problems.

When external routines are created, engine immediately calls external code, that can interact with existing metadata.

src/isql/extract.epp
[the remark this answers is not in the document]
dyemanov: Error "field does not exist". You may find the same assert in `list_procedure_bodies()` which is also executed unconditionally. AFAIU, the only purpose of the assert is to warn that the extract function is not able to deal with older ODS. We may remove these asserts separately, if you disagree with their usage in general. They are not related to this PR and existed before.

dyemanov:
> I think this is going to cause problems.
> 
> When external routines are created, engine immediately calls external code, that can interact with existing metadata.

Hmmm. Maybe theoretically this is possible, but is it really our problem? When UDR is created we have no idea about the  logic inside the external module. How can we be expected to pre-create some metadata in advance? What if the external code would depend on some pre-existing data too?

src/isql/extract.epp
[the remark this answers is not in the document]
hvlad: Strange logic, as for me - use assert and runtime error instead of condition and correct handling of old ODS.
I agree, it is not related with this PR. I never looked into this code before and thus wondering.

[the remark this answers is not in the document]
dyemanov: IIRC, it was discussed in fb-devel and we agreed to simplify the code this way, but it was ages ago...

asfernandes:
> > I think this is going to cause problems.
> > When external routines are created, engine immediately calls external code, that can interact with existing metadata.
> 
> Hmmm. Maybe theoretically this is possible, but is it really our problem? When UDR is created we have no idea about the logic inside the external module. How can we be expected to pre-create some metadata in advance? What if the external code would depend on some pre-existing data too?

Nothing different than PSQL routines. We shouldn't care about routines content, but we do to make their actual usage work as we don't know their creation/update order.

For example, it may receive another procedure name in its body, inspect it and precompute things for fast transformation of its result to json in call time.

The thing is that external routines may do things at "compilation" phase, and this things may be interaction with metadata.

Note also that UDR is a layer on top of external routines. External routines may do more things and it still is a public API.

src/dsql/DdlNodes.epp
sim1984: This solves a completely different problem. https://github.com/FirebirdSQL/firebird/issues/7699
As for creating external procedures and functions in one-pass, I agree with @asfernandes. This is not necessary and can even be harmful in some cases.

dyemanov: @sim1984 yes, this is exactly the bug we were fixing. I suggested a single pass extraction as an extension to the original patch.

dyemanov:
I surely missed the "compilation stage" external calls, could you please point me where it happens?

dyemanov:
Thanks. I asked because there may be a difference. PSQL routines are compiled into BLR during DDL execution -- this is the first stage when dependent objects should already exist (METD should be able to find them). Then at DFW time dependencies are tracked (BLR is parsed) and this is the second stage. For external routines we have only the second stage (DFW time) dependencies access. At this time all dependencies already exist in the database. They're not committed yet, but visible for the current transaction. So I'm wondering whether it's enough to satisfy the possible external calls, given that ISQL extracts all metadata to be executed in one transaction.

PR #8061, opened 2024-04-01
Unnest IN/ANY/EXISTS subqueries and optimize them using semi-join algorithm

dyemanov:
The initial implementation was developed by customer request, later it was slightly improved and committed into a separate branch for testing. Some issues were found that prevented it from inclusion into v5.0. Still work in progress.


dyemanov:
A partial implementation (covering only equivalence conditions) is committed into v5.0.1, but disabled by default (see `SubQueryConversion` setting in _firebird.conf_).

dyemanov:
More complete implementation (enabled by default and so far without an option to disable it) is ready for v6. It adds a (limited but still) choice between nested loop and hash semi-join algorithms to be used. Optimization aspects are to be further improved in the master branch later.

PR #8076, opened 2024-04-09
Non-enforced constraints

aafemt:
Implementation of #2358 using standard-compliant clause `ALTER CONSTRAINT [NOT] ENFORCED`.


asfernandes:
What will happen with `isql -x` for `not enforced` constraints?

hvlad:
Why not enforced constaint deactivates corresponding index ? 
It is not necessary at all and requires costly operation (build index) when constraint is enforced again.

aafemt:
> What will happen with `isql -x` for `not enforced` constraints?

Nothing. CREATE TABLE statement does not support creation of inactive constraints.

aafemt:
> Why not enforced constaint deactivates corresponding index ?

Because active unique index won't allow insertion of duplicates by itself.

> It is not necessary at all and requires costly operation (build index) when constraint is enforced again.

AFAIK deactivate index + bulk insert + activate index is less costly than bulk insert with active index.

hvlad:
> > Why not enforced constaint deactivates corresponding index ?
> 
> Because active unique index won't allow insertion of duplicates by itself.

It is could be changed

> > It is not necessary at all and requires costly operation (build index) when constraint is enforced again.
> 
> AFAIK deactivate index + bulk insert + activate index is less costly than bulk insert with active index.

It is about constraints in general, not about singe use-case.

aafemt:
> It is could be changed

Ok, if it is ever changed, this feature can be changed too.

> It is about constraints in general, not about singe use-case.

I know no other use case for this feature.

hvlad:
BTW, it will be good to show here excerpt from standard that describes the feature.

aafemt:
> It would be much better to do it at once, not leaving big piece of work for someone else, IMHO

I do what I can. Sorry, I have no idea how to implement duplicates in unique indices and doubt that it is anyhow useful.

From the standard I have access to BNF only:

> 11.25 <alter table constraint definition>
> Function
> 
> Change the definition of a table constraint.
> Format
> 
> ```
> <alter table constraint definition> ::=
>   ALTER CONSTRAINT <constraint name> <constraint enforcement>
> ```

aafemt:
> Do you have BNF that allows to change ENFORCED state ?

It is exactly quoted. `<alter table constraint definition>` is a clause of `ALTER TABLE` statement:

> 11.10 <alter table statement>
> Function
> 
> Change the definition of a table.
> Format
> 
> ```
> <alter table statement> ::=
>   ALTER TABLE <table name> <alter table action>
> 
> <alter table action> ::=
>     <add column definition>
>   | <alter column definition>
>   | <drop column definition>
>   | <add table constraint definition>
>   | <alter table constraint definition>
>   | <drop table constraint definition>
>   | <add table period definition>
>   | <drop table period definition>
>   | <add system versioning clause>
>   | <drop system versioning clause>
> ```

And even if I get it wrong, I'll just remove "ANSI standard-compliant" from the doc. This PR is still a valid solution for #2358 even without standard compliance.

mrotteveel:
From SQ:2023-2:

> 11.25 <alter table constraint definition>
> 
> Function
> Change the definition of a table constraint.
> 
> Format
> ```
> <alter table constraint definition> ::=
>     ALTER CONSTRAINT <constraint name> <constraint enforcement>
> ```
> 
> Syntax Rules
> 1) Let T be the table identified by the \<table name> in the containing \<alter table statement>.
> 2) The \<constraint name> shall identify a table constraint TC of T.
> 3) TC shall not identify a unique constraint.
>
> Access Rules
> None.
> 
> General Rules
> 1) The table constraint descriptor of TC is modified as follows.
> &nbsp;&nbsp;&nbsp;Case:
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;a) If NOT ENFORCED is specified, then the indication of whether the constraint is enforced or not enforced is replaced with an indication that the constraint is not enforced.
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;b) Otherwise, the indication of whether the constraint is enforced or not enforced is replaced with an indication that the constraint is enforced.
>
>Conformance Rules
>1) Without Feature F492, “Optional table constraint enforcement”, conforming SQL language shall not contain an \<alter table constraint definition> that contains a \<constraint enforcement>.

And as a reminder, the SQL standard defines "unique constraint" as (4.25.3.2 Unique constraints):

> An indication of whether it was defined with PRIMARY KEY or UNIQUE.

In other words, by syntax rule 3, [NOT] ENFORCED is not allowed for unique constraints (i.e. both PRIMARY KEY and UNIQUE constraints), only for referential and check constraints.

Section 4.25.3.1 Introduction to table constraints also says:

> NOTE 52 — A unique constraint is always enforced. A table check constraint or a referential constraint can either
be enforced or not enforced.

aafemt:
> 3. TC shall not identify a unique constraint.

Ah, ok, this really is going to be a non-standard feature.

hvlad:
@mrotteveel: thanks !

IMHO, we could live with [NOT] ENFORCED PRIMARY/UNIQUE KEY constraints. I'm not insisting, just show my current opinion.

And I'm still very concerned about needs of [de]activation of underlying indices in-sync with constraint enforcement.
Proposed implementation should be almost completely changed to avoid it, it can't be just "fixed a bit".

aafemt:
> Proposed implementation should be almost completely changed to avoid it, it can't be just "fixed a bit".

If `ALTER INDEX` is going to be still supported for PK/UK/FK indices I don't see more changes than a couple of lines in `ALTER CONSTRAINT` implementation: just update a field in `RDB$RELATION_CONSTRAINT` (and some flags in cached structures depending on implementation) instead of call to `modifyIndex()`. And one-liner for ISQL. What do I miss?

aafemt:
Exactly because I don't know how engine manage constraints and metadata cache, I've chose the simplest solution: to reuse index deactivation code. For my purposes it is enough.

If you insist I can remove `ALTER CONSTRAINT` part of this PR and limit it to allow deactivation of constraint indices. Then whoever is eager to implement `ALTER CONSTRAINT` can do it from scratch.

aafemt:
> What will happen with `isql -x` for `not enforced` constraints?

BTW, `isql -x` for inactive indexes also extract them as active. Should it be considered as a separate issue?

aafemt:
Now this PR is a complete solution for #2358. Shall you review it in current state or I must make it much bigger implementing creation of not enforced constraints?

livius2:
> Nothing. CREATE TABLE statement does not support creation of inactive constraints.

So after Create Table stetement should be additional statement fo deactivation.

aafemt:
> after Create Table stetement should be additional statement fo deactivation.

No, just creation of a not enforced constraint should be added. The only question is if it will be in this PR or a separate one.

src/dsql/DdlNodes.epp
```diff
@@ -3482,7 +3525,6 @@ bool TriggerDefinition::modify(thread_db* tdbb, DsqlCompilerScratch* dsqlScratch
 		{
 			switch (TRG.RDB$SYSTEM_FLAG)
 			{
-				case fb_sysflag_check_constraint:
```
hvlad, line 3948: Why ?

src/dsql/DdlNodes.epp
```diff
@@ -6605,11 +6647,18 @@ void RelationNode::makeConstraint(thread_db* tdbb, DsqlCompilerScratch* dsqlScra
 				constraint.create = FB_NEW_POOL(pool) Constraint(pool);
 				constraint.create->type = Constraint::TYPE_NOT_NULL;
 				if (clause->constraintType == AddConstraintClause::CTYPE_NOT_NULL)
+				{
 					constraint.name = clause->name;
+					constraint.create->enforced = clause->enforced;
+					*notNull = clause->enforced;
```
hvlad, line 7683: This is wrong

src/dsql/DdlNodes.epp
```diff
@@ -6605,11 +6647,18 @@ void RelationNode::makeConstraint(thread_db* tdbb, DsqlCompilerScratch* dsqlScra
...
+				}
+				// NOT NULL for PRIMARY KEY is always enforced
 			}
 
 			if (clause->constraintType == AddConstraintClause::CTYPE_NOT_NULL)
+			{
 				break;
+			}
```
hvlad: Not needed braces, remove please

src/dsql/DdlNodes.epp
```diff
@@ -7936,6 +7989,118 @@ void AlterRelationNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
...
+						RC IN RDB$RELATION_CONSTRAINTS
+						WITH RC.RDB$CONSTRAINT_NAME EQ clause->name.c_str() AND
+							 RC.RDB$RELATION_NAME EQ name.c_str()
+					{
+						found = true;
+						fb_utils::exact_name(RC.RDB$CONSTRAINT_TYPE);
+						if (strcmp(RC.RDB$CONSTRAINT_TYPE, PRIMARY_KEY) == 0 ||
+							strcmp(RC.RDB$CONSTRAINT_TYPE, UNIQUE_CNSTRT) == 0)
+						{
+							// Deactivation of primary/unique key requires check for active foreign keys
```
hvlad: Is this code works for deactivaiton only ?

hvlad (commented):
It was a quick look, so I can miss something. 
Also, I didn't check extract.epp, show.epp and SystemTriggers.epp

src/dsql/DdlNodes.epp
```diff
@@ -7936,6 +7989,118 @@ void AlterRelationNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
...
+							{
+								MODIFY TRG
+									TRG.RDB$TRIGGER_INACTIVE = clause->enforced ? FALSE : TRUE;
+								END_MODIFY
+							}
+							END_FOR
+						}
+						else if (strcmp(RC.RDB$CONSTRAINT_TYPE, NOT_NULL_CNSTRT) == 0)
+						{
+							AutoRequest requestHandle;
```
hvlad: Should this and below requests be cached ?

src/dsql/DdlNodes.epp
```diff
@@ -10117,6 +10282,8 @@ void AlterIndexNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratch,
 		executeDdlTrigger(tdbb, dsqlScratch, transaction, DTW_BEFORE, DDL_TRIGGER_ALTER_INDEX,
 			name, NULL);
 
+		checkIndexReferenced(tdbb, transaction, name.c_str());
```
hvlad: Is it supposed to run when index is activated ?

doc/sql.extensions/README.ddl.txt
```diff
@@ -718,3 +718,50 @@ CREATE [UNIQUE] [ASC[ENDING] | DESC[ENDING]]
...
+
+Also ALTER CONSTRAINT clause is added to ALTER TABLE statement.
+
+Syntax:
+
+ALTER TABLE ALTER CONSTRAINT <constraint name> <constraint enforcement>
+
+Primary and unique keys cannot be deactivated if they are referenced by any active foreign key.
+
+The corresponding ALTER INDEX and ALTER TRIGGER statements are allowed as well.
```
hvlad: Not clear at all.

src/jrd/ini.epp
```diff
@@ -411,7 +411,7 @@ namespace
 			const char* users[] = {userName, "PUBLIC"};
 			int grantOptions[] = {WITH_GRANT_OPTION, 0};
 
-			for (unsigned i = 0; i < FB_NELEM(users); i++)
+			for (int i = 0; i < FB_NELEM(users); i++)
```
hvlad: Why ?

aafemt: `FB_NELEM` is signed. It fixes a warning.

hvlad: It is `FB_SIZE_T` in master and better to fix if to be at least `unsigned` in v5 too, there is no sence to use signed type, imho.

Are you going to change this place in master ?

Anyway, there is a lot of such misuse, why fix just one of them and at the unrelated PR without any need ?

aafemt: https://groups.google.com/g/firebird-devel/c/D4eu6KGUlRU/m/s-GnU8FxAAAJ

This PR is for `master`.
I fix annoying warnings only in files I touch frequently during work to keep the PR small. Fortunately Mark is going to do this job so may be I'll remove this piece from patch.

hvlad: > This PR is for `master`.

So why you argue ?


> I fix annoying warnings only in files I touch frequently during work to keep the PR small

This file makes NO relation with this PR. Don't do it or you will wait for reviews forever as nobody wants to waste a time.

aafemt: This PR waited more than a year. Forever is just a little longer so why should I care?..

hvlad: You never care. Go wait another year.

src/dsql/DdlNodes.epp
aafemt: Alteration of NOT NULL constraint isn't supposed to be done in mass. These requests don't worth caching.

src/dsql/DdlNodes.epp
aafemt: (De)activation of a check constraint is done by (de)activation of the corresponding triggers. It must be allowed.

src/dsql/DdlNodes.epp
aafemt: No, but activation of the key doesn't require the check.

src/dsql/DdlNodes.epp
aafemt: If NOT NULL constraint is not enforced, notNull flag is set to false. If the constraint is enforced, notNull flag is set to true. What is wrong?

src/jrd/ini.epp
aafemt: You get one thing wrong: it is not me who wait, but Firebird users.

aafemt:
Updated to current upstream.

PR #8088, opened 2024-04-19
Update Linux build to use gcc 13.2, binutils 2.40 and glibc to 2.28

asfernandes:


AlexPeshkoff:
I'm pretty sure that cloop does work with static c++ support, but what was a reason of a change?

asfernandes:
Because when it's built dynamically, it uses newer libstdc++ (from the just built gcc) than the one installed in the container and doesn't work with invalid GLIBCXX symbols.

AlexPeshkoff (approved):
I did not check docker files. What about generic posix builds I see no problems.

PR #8091, opened 2024-04-20
Ability to create an inactive index

### LINKED ISSUE #6233: create index idx [as active | inactive] on ... [CORE5981]
Submitted by: Volker Rehn (vr2_s18)

Syncing metadata of two databases: If the master db has inactive indices which are missing in the target, on creation they will be activated in the target db, do a full data scan, and only afterwards they can then be deactivated by alter index\. An optional "as active/inactive" clause would give more control over when an index is actually built\. Oracle and MySQL have visible/invisible for indices, but that only removes the index from being used by the optimizer, it still performs a data scan on creation\. Apparently the SQL standard does not regulate index creation\. The syntax could be

CREATE \[UNIQUE\] \[ASC\[ENDING\] \| DESC\[ENDING\]\]
INDEX indexname \[AS ACTIVE \| INACTIVE\] ON tablename
\{\(col \[, col \.\.\.\]\) \| COMPUTED BY \(<expression\>\)\};

ACTIVE being the default as it is now\. Similar to how triggers are defined\.

### LINKED ISSUE #8090: Extracting of inactive index
`isql -x` always generates the script that creates indices as active even if they are inactive in the database. It makes a database recreated from such script different from source database.

aafemt:
Implementation of #8090.


aafemt:
This PR also solves #6233.

dyemanov (approved):

dyemanov:
Unless there are objections, I will merge this PR tomorrow.

aafemt:
I would understand if you assigned to me the mentioned issues, but what is the point to assign a pull request? There is nothing I can do with it in absence of write rights for repository.

dyemanov:
This is for documentation purposes (release notes). I prefer to have an explicit assignee everywhere, just for clarity. I've also assigned you to #8090 but cannot do the same for #6233 as non-members can be assigned to an issue only if they commented inside the ticket.

src/dsql/DdlNodes.h
```diff
@@ -1790,6 +1790,7 @@ class CreateIndexNode : public DdlNode
 	MetaName name;
 	bool unique;
 	bool descending;
+	bool active;
```
asfernandes, line 1793: Looks like this field is not initialized.

aafemt: It is unconditionally assigned in parse.y.

aafemt: Assigning value twice in row is just a waste of time.

hvlad: Do you still wonder why most of your PRs nobody wants to review and/or accept ?

aafemt: If you have a testcase that shows a problem - share it.

AlexPeshkoff:
On 10/25/24 16:01, Vlad Khorsun wrote:
>
>
>
> In src/dsql/DdlNodes.h 
> <https://github.com/FirebirdSQL/firebird/pull/8091#discussion_r1816644225>:
>
> > @@ -1790,6 +1790,7 @@ class CreateIndexNode : public DdlNode
>   	MetaName name;
>   	bool unique;
>   	bool descending;
> +	bool active;
>
> Funny (not).
> You should learn and follow basic rules before attempt to contribute 
> something.
>

Dmitry, sorry, but +1.

aafemt:
This class currently has nine members, seven are initialized by explicit constructor, one by implicit constructor and one in the code immediately after instantiation. What exactly do you expect me to do with this mess?

PR #8097, opened 2024-04-28
Implement #8066 : Make protocol schemes case-insensitive.

### LINKED ISSUE #8066: Make protocol schemes case-insensitive.
Currently the protocol schemes in the Firebird 3.0+ connection URLs are case-sensitive (Firebrd/fbclient only accepts the lowercase variant). 

Similar to RFC 3986, I think we should handle schemes case-insensitively, where lowercase is the canonical form. See https://www.rfc-editor.org/rfc/rfc3986#section-3.1

In other words, Firebird should accept `inet://...`, `INET://...`, `inet4://...`, `INET4://...`, `inet6://...`, `INET6://...`, `xnet://...` and `XNET://...` (and mixed case forms) equally.

hvlad:


dyemanov (approved):

aafemt:
Wouldn't be clearer to declare `prefix` as `NoCaseString` and then use it's `compare` (in absence of `starts_with` from C++20)?

hvlad:
> Wouldn't be clearer to declare `prefix` as `NoCaseString` and then use it's `compare` (in absence of `starts_with` from C++20)?

`compare()` takes into account string length and can not work as `starts_with()`

aafemt:
> `compare()` takes into account string length and can not work as `starts_with()`

AFAIU this overload could work:
`int compare(const_pointer s, const size_type n) const`
Because `c_str()` is guaranteed to be zero-terminated, result of `prefix.compare(expanded_name.c_str(), prefix.length())` should do exactly the same what your code does if `prefix` is `NoCaseString`.

hvlad:
> > `compare()` takes into account string length and can not work as `starts_with()`
> 
> AFAIU this overload could work: `int compare(const_pointer s, const size_type n) const` Because `c_str()` is guaranteed to be zero-terminated, result of `prefix.compare(expanded_name.c_str(), prefix.length())` should do exactly the same what your code does if `prefix` is `NoCaseString`.

It could access memory out of `expanded_name` buffer, if `expanded_name.length() < prefix.length()`

aafemt:
> It could access memory out of `expanded_name` buffer, if `expanded_name.length() < prefix.length()`

Only if comparison is greedy and doesn't stop at the first different byte (which would be the expanded_name's terminating `<NUL>`).

PR #8098, opened 2024-04-28
Implement #8010 : Remove gfix -cache option

### LINKED ISSUE #8010: Remove gfix -cache option
Gfix currently lists a `-cache` option which is documented as _"shutdown cache manager"_), and this option - looking at the `alice/exe.cpp` code - can be used in combination with `-shutdown`, but it doesn't seem to be actually doing anything (the associated `isc_dpb_shutdown` bitmap value of `isc_dpb_shut_cache` is never used in Firebird).

This seems to be some remnant from the InterBase days, and as such, it should be removed.

hvlad:


dyemanov:
Any good reason why it's commented out instead of being really removed?

hvlad:
> Any good reason why it's commented out instead of being really removed?

Just follows convention used with other deprecated items (sw_begin_log/IN_SW_ALICE_BEGIN_LOG, etc).
Should I better remove it ?

Issue #8101, opened 2024-05-04
Firebird crashes if a plugin factory returns nullptr and no error in status


aafemt:
If a trace plugin's factory returns nullptr and no error in status - Firebird crashes trying to reference this pointer. Returned value should be checked for nullptr in any case.

hvlad:
What is Firebird version and call stack of crash ?

aafemt:
Firebird 4.0.4.2981

No stack trace but I guess the crash occur on TraceManager.cpp:148 line `info.factory->addRef();` after previous line assigned `nullptr` returned from `traceItr.plugin()` to `info.factory` because `GetPlugins::getPlugin()` check status for error but doesn't check returned value.

And this code is the same in all versions.

hvlad:
It is still not clear what do you speak about - the bug in TraceManager ? in PluginMnager ? in some implementation of custom (trace) plugin ?
It is important to know to fix a root issue, not particular case (what case?)

aafemt:
IMHO it is a bug in TraceManager. It must not trust plugin's sanity and check returned values.

hvlad:
Did you check another callers of `GetPlugins::plugin()` ?

aafemt:
I'm sorry, my guess was wrong. The returned value is checked there by `hasData()` call.

But in `ConfiguredPlugin::factory()` value returned by `createPlugin()` isn't checked before calling `setOwner()` unless I missed something again.

AlexPeshkoff:
On 5/5/24 01:28, Dimitry Sibiryakov wrote:
>
> I'm sorry, my guess was wrong. The returned value is checked there by 
> |hasData()| call.
>
> But in |ConfiguredPlugin::factory()| value returned by 
> |createPlugin()| isn't checked before calling |setOwner()| unless I 
> missed something again.
>

We anyway have to trust code in plugins cause it runs in the same 
process as native FB code. And if plugin factory returned no error it 
must return valid interface pointer. In this case plugin should be 
fixed. It can cause segfault in 100500 other ways, what a sense to fux 
one case?

hvlad:
> I'm sorry, my guess was wrong. 

Don't you think that issue title chould be changed ?

aafemt:
> And if plugin factory returned no error it must return valid interface pointer. In this case plugin should be fixed. It can cause segfault in 100500 other ways, what a sense to fux one case?

[...]

PR #8102, opened 2024-05-05
Fix for #8101

aafemt:


src/yvalve/PluginManager.cpp
```diff
@@ -605,7 +605,7 @@ namespace
 		FbLocalStatus ls;
 		IPluginBase* plugin = module->getPlugin(regPlugin).factory->createPlugin(&ls, par);
 
-		if (!(ls->getState() & Firebird::IStatus::STATE_ERRORS))
+		if (plugin && !(ls->getState() & Firebird::IStatus::STATE_ERRORS))
```
dyemanov, line 608: Why `createPlugin()` may return `nullptr` with successful status? This smells like a bug.

aafemt: Yes, it is a bug. But crash isn't an adequate reaction from Firebird to such bug.

aafemt: Besides API specs doesn't mention that nullptr is not allowed to be returned from `PluginFactory::createPlugin()` so it is legit for the factory to refuse to provide plugin instance because of given configuration or just out of blue.

hvlad: If you want to speak about correct (legal) implementation: plugin factory should set error if it doesn't creates plugin by any reason.
Every method which have `IStatus` as parameter must use it to report errors.

Anyway, I agree that caller should check returned pointer for `nullptr`.
Also, will be good to log message if plugin factory returns `nullptr` and set no error, IMO.
It helps plugin developer to see the bug.

hvlad (changes requested):
log message if plugin factory returns nullptr and set no error

aafemt:
> log message if plugin factory returns nullptr and set no error

Such situation cannot disappear just like that so logging will be endless and log file will overflow. I don't like this idea.

hvlad:
Buggy plugins should be detected and removed by DBA.

aafemt:
It is not given that the plugin is buggy. Signalling Plugin Manager to skip it and try the next plugin in list may be a part of its logic. May be it set a warning (which is not checked by `check()` BTW). There is also no noticeable performance penalty in this case. If plugin is buggy indeed existing DBA will detect it anyway by absence of result. Log overflow in absence of DBA is much more dangerous from my POV.

hvlad (approved):
I looked again at context and agree that factory could return nullptr and set no error in some valid cases.

PR #8106, opened 2024-05-07
Fix for #8105:  Master database could miss replication segments on Windows Classic Server

### LINKED ISSUE #8105: Master database could miss replication segments on Windows Classic Server
When Firebird run on Windows and configured as Classic Server it could not generate some replcation segments.
Segments numbers could be, for example 1,3,5,7, etc.
Also, it could create more files in `journal_directory` than it is allowed by `journal_segment_count`

hvlad:
The main problem is that on Windows it is not allowed to rename open files with mapped memory section.
`rename()` for such file always set `errno` to `EACCES`, `GetLastError()` returns `ERROR_SHARING_VIOLATION`.

### DIFF
--- a/src/jrd/replication/ChangeLog.cpp
+++ b/src/jrd/replication/ChangeLog.cpp
@@ -79,8 +79,6 @@ namespace
 	const char* PATHNAME_WILDCARD = "$(pathname)";
 	const char* ARCHPATHNAME_WILDCARD = "$(archivepathname)";
 
-	SegmentHeader g_dummyHeader;
-
 	static THREAD_ENTRY_DECLARE archiver_thread(THREAD_ENTRY_PARAM arg)
 	{
 		ChangeLog* const log = static_cast<ChangeLog*>(arg);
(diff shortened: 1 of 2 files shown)


src/jrd/replication/ChangeLog.cpp
```diff
@@ -950,11 +1003,26 @@ ChangeLog::Segment* ChangeLog::reuseSegment(ChangeLog::Segment* segment)
...
 		return createSegment();
+	}
 
 	// Re-open the segment using a new name and initialize it
 
 	const auto fd = os_utils::openCreateSharedFile(newname.c_str(), O_BINARY);
+	if (fd == -1)
+	{
+		raiseError("Journal file %s create() failed (error %d)", newname.c_str(), errno);
+	}
```
dyemanov, line 1026: `openCreateSharedFile()` throws itself, so this addition is not necessary.

src/jrd/replication/ChangeLog.cpp
```diff
@@ -893,13 +941,17 @@ void ChangeLog::clearSegments()
...
 
 	PathName filename;
 	filename.printf(FILENAME_PATTERN, m_config->filePrefix.c_str(), sequence);
 	filename = m_config->journalDirectory + filename;
 
 	const auto fd = os_utils::openCreateSharedFile(filename.c_str(), O_EXCL | O_BINARY);
+	if (fd == -1)
+	{
+		raiseError("Journal file %s create() failed (error %d)", filename.c_str(), ERRNO);
+	}
```
dyemanov, line 954: openCreateSharedFile() throws itself, so this addition is not necessary.

hvlad: Done, thanks

src/jrd/replication/ChangeLog.cpp
hvlad: Done

dyemanov (approved):

PR #8121, opened 2024-05-15
Fix debug build failure with assertion in CRT

aafemt:
`isspace` behavior [is undefined](https://en.cppreference.com/w/cpp/string/byte/isspace) with `char` and `signed char`.

### DIFF
--- a/src/isql/FrontendLexer.cpp
+++ b/src/isql/FrontendLexer.cpp
@@ -22,23 +22,20 @@
  */
 
 #include "firebird.h"
+#include "../common/utils_proto.h"
 #include "../isql/FrontendLexer.h"
 #include <algorithm>
 #include <cctype>
 
 
 static std::string trim(std::string_view str);
-static inline int fb_isspace(const char c)
-{
-	return isspace((int)(UCHAR)c);
-}
 
 static std::string trim(std::string_view str)
 {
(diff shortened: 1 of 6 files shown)


src/isql/FrontendLexer.cpp
```diff
@@ -28,22 +28,25 @@
 
 
 static std::string trim(std::string_view str);
-
+static inline int fb_isspace(const char c)
+{
+	return isspace((int)(UCHAR)c);
+}
```
AlexPeshkoff, line 35: We already have one in isql.epp. (And a lot of places in the code where parameter is casted to UCHAR.)
Why duplicate code? Even if it's trivial.

May be better add such function to fb_utils namespace?

aafemt: It would require a common header file. It looked like Adriano was trying to avoid unnecessary dependencies and keep minimal include list.

asfernandes: > I see no problem with having files from src/common anywhere in the project.

Agreed.

AlexPeshkoff:
What about isql/iutils.cpp & dsql/Parser.cpp ?

aafemt:
They already have required cast so I didn't touch them because bigger PRs have smaller probability to be accepted.

PR #8124, opened 2024-05-23
Fixed possible buffer overflow on system privileges lookup during user mapping

dmitry-starodubov:


src/jrd/Mapping.cpp
```diff
@@ -1171,7 +1171,9 @@ class SysPrivCache : public PermanentStorage
 					g |= l;
 				}
 
-				ULONG gg = 0; g.store(&gg);
+				FB_UINT64 gg = 0;
+				fb_assert(sizeof(gg) >= g.BYTES_COUNT);
```
aafemt, line 1175: `static_assert` would be better here because none side of the comparison can be changed in runtime.

dmitry-starodubov: done

AlexPeshkoff:
Why FB4 only? It makes no sense to change old branch and leave fresh one as is. Such change will be lost.

Please provide it at least for master, and better for FB5 too.

dmitry-starodubov:
They were also made yesterday: #8125 and #8126

aafemt:
In my comment I didn't pay attention to target branch assuming that PRs are made for master at the first hand. Is static_assert allowed for usage in v4 branch?

dyemanov:
Yes, it's allowed.

AlexPeshkoff:
On 5/24/24 14:35, Dimitry Sibiryakov wrote:

> Is static_assert allowed for usage in v4 branch?
>

Yes.

PR #8129, opened 2024-05-24
Simplest fix for bug #8120 : Cast dies with numeric value is out of range error (5.0.1 snapshot)

### LINKED ISSUE #8120: Cast dies with numeric value is out of range error (5.0.1 snapshot)
Hi,

5.0.1.1398 snapshot

select cast('27' as numeric(4,2))
from rdb$database

causes
numeric value is out of range.
error.

It works fine with 5.0.0.1306 and 4.0.*, 3.0.*

hvlad:
The problem in #8120 is that `RetValue` updates `value` in own destructor only, thus making impossible to work with `value` in the same scope as `RetValue`.
I found and fixed two occurrences of such code but it looks as fragile to me.
It would be better to not create such indirect dependencies, for example moving `RetValue` instance into `cvt_decompose()`.


AlexPeshkoff (approved):

mrotteveel:
Out of curiosity, why does the introduction of those braces fix this problem?

PR #8148, opened 2024-06-05
Fix compilation issues on Debian 12 and segfault.patch

takenek:
Fix compilation issues on Debian 12 and segfault.patch


AlexPeshkoff:
This branch is unsupported and closed for commits.
PS. Does it make sense to use 15 years old codebase in fresh OS?

PR #8162, opened 2024-06-18
Improvement #8161 : Cardinality estimation should use primary record versions only

### LINKED ISSUE #8161: Cardinality estimation should use primary record versions only
The issue was found on FB3 due to failed QA test for CORE-5602. 
While the test runs successfully on FB4-FB6, the issue is common for all versions.

The test case uses database backup from QA test for [CORE-5602](https://github.com/FirebirdSQL/firebird-qa/blob/master/backups/core5602.fbk).

<details><summary>The test query:</summary>
<p>

```
execute block
as
declare relname varchar(32);
declare cnt int;
begin
  for select X.RDB$RELATION_NAME
        from RDB$RELATION_FIELDS X
       where X.RDB$FIELD_SOURCE = 'BOOL_EMUL'
      into :relname
  do begin
     select count(*)
       from RDB$DEPENDENCIES DEP, RDB$PROCEDURES PRC
      where DEP.RDB$DEPENDED_ON_NAME = :relname
        AND DEP.RDB$FIELD_NAME = :relname
        AND DEP.RDB$DEPENDED_ON_TYPE = 0 /* obj_relation */
        AND DEP.RDB$DEPENDENT_TYPE = 5 /* obj_procedure */
        AND DEP.RDB$DEPENDENT_NAME = PRC.RDB$PROCEDURE_NAME
        AND PRC.RDB$PACKAGE_NAME IS NULL
    into :cnt;
  end
end
```
</p>
</details> 

<details><summary>Bad plan and stats:</summary>
<p>

PLAN JOIN (PRC INDEX (RDB$INDEX_21), DEP INDEX (RDB$INDEX_27, RDB$INDEX_28))
PLAN (X INDEX (RDB$INDEX_3))

Execute       : 94,00 ms
Fetches: 164 751

|          Table Name           |  Indexed reads |
| --- | --- |
|RDB$DEPENDENCIES               |     27000 |
|RDB$INDICES                    |         9 |
|RDB$PROCEDURES                 |     27000 |
|RDB$RELATION_FIELDS            |       900 |

</p>
</details> 

<details><summary>Good plan and stats:</summary>
<p>

PLAN JOIN (DEP INDEX (RDB$INDEX_28), PRC INDEX (RDB$INDEX_21))
PLAN (X INDEX (RDB$INDEX_3))

Execute       : 47,00 ms
Fetches: 29 746

| Table Name |  Indexed   reads |
| --- | --- |
|RDB$DEPENDENCIES               |     27000 |
|RDB$INDICES                    |         7 |
|RDB$RELATION_FIELDS            |       900 |
</p>
</details>

hvlad:


aafemt:
Do I understand right that your code counts records on only one data page because `done` become `true` as soon as `recordCount` get any non-zero value and inner loop is aborted as well? Old code seems to count records for whole table...

hvlad:
> Do I understand right that your code counts records on only one data page because `done` become `true` as soon as `recordCount` get any non-zero value and inner loop is aborted as well? 

Almost correct

> Old code seems to count records for whole table...

No, read it again

src/jrd/dpm.epp
```diff
@@ -245,47 +245,52 @@ double DPM_cardinality(thread_db* tdbb, jrd_rel* relation, const Format* format)
 
 	ULONG recordCount = 0, recordLength = 0;
 
-	const RelationPages* const relPages = relation->getPages(tdbb);
+	RelationPages* const relPages = relation->getPages(tdbb);
 	const vcl* const vector = relPages->rel_pages;
 	if (vector)
```
dyemanov, line 250: It looks like `vector` is not used anymore, so maybe just `if (relPages->rel_pages)` ?
Not sure this check is needed at all, BTW.

dyemanov (approved):

PR #8165, opened 2024-06-21
Added shutdown handler for Classic Server

Zhdanov0:
Added shutdown handler for server to avoid active connections after termination.
If you start the server with classic mode, then at the end of the server, there are still active connections.

### DIFF
--- a/src/remote/server/os/posix/inet_server.cpp
+++ b/src/remote/server/os/posix/inet_server.cpp
@@ -316,8 +316,11 @@ int CLIB_ROUTINE main( int argc, char** argv)
 		ISC_set_prefix(0, 0);
 
 		// set shutdown signals handler for listener
-		set_signal(SIGTERM, shutdown_handler);
-		set_signal(SIGINT, shutdown_handler);
+		if (standaloneClassic)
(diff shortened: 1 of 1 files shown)


src/remote/server/os/posix/inet_server.cpp
```diff
@@ -314,6 +315,10 @@ int CLIB_ROUTINE main( int argc, char** argv)
 		// activate paths set with -e family of switches
 		ISC_set_prefix(0, 0);
 
+		// set shutdown signals handler for listener
```
AlexPeshkoff, line 318: Does it make sense to add this handlers in super & superclassic modes?

Zhdanov0: No, it's only relevant to classic

Zhdanov0: I'll add a condition

AlexPeshkoff (approved):

AlexPeshkoff:
I suggest to backport it to 5 & 4.

dyemanov:
Backported.

PR #8169, opened 2024-06-27
Fixed bug #8168 : MAKE_DBKEY bug after backup/restore

### LINKED ISSUE #8168: MAKE_DBKEY bug after backup/restore
Firebird relation ids are not stable. While some BLRs uses relation ids, BLR of stored code always refers to them by name.

However, system function `MAKE_DBKEY` transforms relation name to id and stores this id even in stored BLR.

After backup/restore, `MAKE_DBKEY(name)` is going to refer to different table.

hvlad:


src/dsql/ExprNodes.cpp
```diff
@@ -12297,6 +12297,30 @@ DmlNode* SysFuncCallNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScrat
...
+	{
+		// Special handling for system function MAKE_DBKEY:
+		// convert constant relation name into ID at the parsing time
+
+		auto literal = nodeAs<LiteralNode>(node->args->items[0]);
+
+		if (literal && literal->litDesc.isText())
+		{
+			const MetaName relName = literal->getText();
+			const jrd_rel* const relation = MET_lookup_relation(tdbb, relName);
```
asfernandes, line 12310: Shouldn't also better register a dependency on the relation?

hvlad: The function itself makes no actions with relation. All it needs - know relation ID.
Thus, I don't think we need any kind of dependency here.

asfernandes: I understand, but if one drops a relation directly referenced by name inside a given procedure, next time this procedure will not be loaded. Dependencies exists just to avoid that.

hvlad: I still see no need in dependency - it is too restrictive in this case. 
I'll better raise error at the execution phase, if relation was not found at parse time.

dyemanov: In this function the relation name is a literal, not an identifier. Thus I agree a dependency is not needed, a runtime check looks enough.

asfernandes (approved):

PR #8170, opened 2024-07-03
Fix #7269: Restore deferred primary key indexes

### LINKED ISSUE #7269: [FR]  Database restore must make every effort on activating deferred indexes
Database restore are aborted on first error of first problematic deferred index.
Database restore must make every effort on activating deferred indexes - build them all and show full problem list for each (#7268)

GeF0rs:
Fixed database recovery aborting on the first failure of the first problematic deferred index gbak ([#7269](https://github.com/FirebirdSQL/firebird/issues/7269))

I took the database to check from [#7269](https://github.com/FirebirdSQL/firebird/issues/7269)
Previously, recovery was interrupted when trying to activate PK_A3.
image

Now the recovery continues, which allows you to smoothly restore PK_A2.
image

The foreign key, however, is still not restored because requires an active primary key, but this still allows you to greatly reduce the number of non-activated indexes (on our industrial base, 3000+ indexes were not activated due to one non-activated one)


GeF0rs:
?

hvlad:
What about extracting code, copied from "creating FK" block below, into separate routine to avoid duplication ?

GeF0rs:
I moved the activation of indexes into a separate function, please take a look

hvlad:
Please, be patient and wait a bit, I'll return to it when time permits.

hvlad:
I will add some additional fix after merge.

aafemt:
JFYI: You can create a pull request to a pull request improving it before merging.

hvlad:
> JFYI: You can create a pull request to a pull request improving it before merging.

Yes. But it is too much bureaucracy and not needed in this case.

GeF0rs:
Thank you for accepting my fix. 
I have studied your additions and taken them into account for future use.

PR #8179, opened 2024-07-10
Do not delete system trace session with missing parent process (fix for https://github.com/FirebirdSQL/firebird/issues/8180)

### LINKED ISSUE #8180: Sometimes a trace session is terminated spontaneously
Sometimes a system session can just disappear. The easiest way to reproduce the problem is to run 2 tests via firebird-qa:
python3 -m pytest --server local --disable-db-cache tests/bugs/core_2940_test.py tests/bugs/core_2981_test.py
The second test will fail with the following result:
Trace session ID 3 not found

Noremos:


Noremos:
Sometimes a system session can just disappear. The easiest way to reproduce the problem is to run 2 tests via firebird-qa:
`python3 -m pytest --server local --disable-db-cache tests/bugs/core_2940_test.py tests/bugs/core_2981_test.py`
The second test will fail with the following result:
`Trace session ID 3 not found`

hvlad (changes requested):
It is not clear what problem this PR going to solve.
Why there is no corresponding issue ?

Noremos:
> It is not clear what problem this PR going to solve. Why there is no corresponding issue ?

Sorry, here is the issue: https://github.com/FirebirdSQL/firebird/issues/8180

hvlad:
> > It is not clear what problem this PR going to solve. Why there is no corresponding issue ?
> 
> Sorry, here is the issue: #8180

Thanks

hvlad (approved):

PR #8188, opened 2024-07-21
Fix #8187 - Performance regression in generating of UUID values after introducing GUID-v7

### LINKED ISSUE #8187: Performance regression in generating of UUID values after introducing GUID-v7
Generating of GUID values become drastically slower since 6.0.0.386 (when improvement related to UUID v7 appeared).
Consider script:
```
set bail on;
set list on;
set blob all;

shell del r:\temp\tmp4test.fdb 2>nul;
create database 'localhost:r:\temp\tmp4test.fdb';
commit;
show version;
/*
61b2d7eb13095a263a70c6db6a09a029a5791eb1 ; Sun Jun 23 08:23:25 2024 -0300
fdac4b8152ba2d16c29ed4c3e9ee46e9b3d2af4f ; Sun Jun 23 08:23:25 2024 -0300
*/
set term ^;
execute block returns(elap_ms int) as
   declare n int = 500000;
   declare t0 timestamp;
   declare e varchar(16) character set octets;
begin
   t0 = 'now';
   while (n > 0) do
   begin
       e = gen_uuid();
       n = n - 1;
   end
   elap_ms = datediff(millisecond from t0 to cast('now' as timestamp));
   suspend;
end
^
set term ;^
commit;
```
On 6.0.0.384-24d99bf (21.06.2024 20:15 UTC) it runs for ~385 ms.
On 6.0.0.386-7c57f2d (23.06.2024 20:14 UTC) its time is about 49 seconds(!).
Changing UUID call to "e = gen_uuid(4);" or "e = gen_uuid(7);" does not matter.

asfernandes:
Replacing deprecated win32 crypto functions.

### DIFF
--- a/src/common/os/win32/guid.cpp
+++ b/src/common/os/win32/guid.cpp
@@ -32,7 +32,7 @@
 #endif
 
 #include <windows.h>
-#include <wincrypt.h>
+#include <bcrypt.h>
 #include <objbase.h>
 #include <stdio.h>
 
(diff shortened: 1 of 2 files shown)


src/common/os/win32/guid.cpp
```diff
@@ -32,46 +32,23 @@
...
+#include <bcrypt.h>
 #include <objbase.h>
 #include <stdio.h>
 
 #include "firebird.h"
 #include "../common/os/guid.h"
 #include "fb_exception.h"
 
+#pragma comment(lib, "bcrypt.lib")
+
```
hvlad, line 44: So far we never uses such a way to declare dependencies.
While it is convenient for code author, it makes hard to collect all required dependencies when needed.
Thus, I prefer to use traditional way and move it into corresponding project file or property sheet.

PR #8192, opened 2024-07-23
Additinal fix for #8180: Make sure trace session slots will not break use counter 

Noremos:
At some point, I ran into the assert fb_assert(header->cnt_uses != 0); in the ConfigStorage::compact() method. I found a way to trigger the assert with Super Server and Embedded user session.

### Steps to reproduce

1. Settup the system session:

firebird.conf
`AuditTraceConfigFile = fbtrace.conf`

fbtrace.conf
```
database
{
	enabled = true
}
```
2. Start a Super Server. (At this point, cnt_uses == 1 and 0 slots)
3. Lunch 3 user session with remote service_mgr  (cnt_uses==1 and there are 3 allocated slots)
4. Lunch a user session with embedded service_mgr (cnt_uses==2 and there are 4 allocated slots)
5. Perform a connection, then disconnect;
6. Restart the server. The remote sessions will be rejected but slots will not be marked as deleted (cnt_uses==2, 4 allocated slots)
7. Shut down the emebdded user session  (cnt_uses==1, 3 allocated slots + 1 marked as deleted)
8. Wait for the compact methid to be invoked. It will find 3 allocated slots with differend pid and perform header->cnt_uses-- for each and we will get cnt_uses==4294967295 (-1).
The same applies in cases where the server is not restarted manually but lunched after a crash.

Additionally, marking all allocated slots as deleted in the shutdown method might be beneficial, but it adds extra overhead and will not protect us from a crash. Therefore, I propose a fix for the compact method for now.

There is only one problem with this PR. As far as I know, LINUX and WINDOWS can reuse PIDs. So in case 2 fork processes of a Classic server crash (leaving cnt_uses undecremented), the compact method will only decrement once after the PR. But this is an extremely rare case to consider, I suppose.

### DIFF
--- a/src/jrd/trace/TraceConfigStorage.cpp
+++ b/src/jrd/trace/TraceConfigStorage.cpp
@@ -470,8 +470,8 @@ void ConfigStorage::compact()
 	ULONG check_used, check_size;
 	check_used = check_size = sizeof(TraceCSHeader);
 
-	// Track undeleted slots from dead storages
-	Firebird::NonPooledMap<ULONG, bool> deadStorages;
+	// Track undeleted slots from dead processes
+	Firebird::SortedArray<ULONG, InlineStorage<ULONG, 16>> deadProcesses;
 
 	// collect used slots, sort them by offset
 	for (TraceCSHeader::Slot* slot = header->slots; slot < header->slots + header->slots_cnt; slot++)
@@ -481,8 +481,8 @@ void ConfigStorage::compact()
 			!ISC_check_process_existence(slot->ses_pid))
 		{
 			// A SUPER server may shut down, but its Storage shared memory continues to live due to an embedded user session.
-			// The storage might allocate multiple slots, so count them carefully.
-			deadStorages.put(slot->ses_pid);
+			// The process might allocate multiple slots, so count them carefully.
+			deadProcesses.add(slot->ses_pid);
 
 			markDeleted(slot);
 		}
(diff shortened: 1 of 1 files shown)


src/jrd/trace/TraceConfigStorage.cpp
```diff
@@ -470,15 +470,20 @@ void ConfigStorage::compact()
 	ULONG check_used, check_size;
 	check_used = check_size = sizeof(TraceCSHeader);
 
+	// Track undeleted slots from dead storages
+	Firebird::NonPooledMap<ULONG, bool> deadStorages;
```
hvlad, line 474: The second field (bool) is not used, why not use `SortedArray` or `SparseBitmap` here ?
I would use `SortedArray` with `InlineStorage` as there should not be too many items.
Also, this container used to collect PID's of dead processes, not storages.
Names are important, please use something more appropriate.

src/jrd/trace/TraceConfigStorage.cpp
```diff
@@ -470,15 +470,20 @@ void ConfigStorage::compact()
...
 	for (TraceCSHeader::Slot* slot = header->slots; slot < header->slots + header->slots_cnt; slot++)
 	{
 		if (slot->used && slot->ses_pid != pid &&
 			((slot->ses_flags & trs_system) == 0) && // System sessions are shared for multiple connections so they may live without the original process
 			!ISC_check_process_existence(slot->ses_pid))
 		{
-			fb_assert(header->cnt_uses != 0);
-			header->cnt_uses--; // Process that created trace session disappeared, count it out
+			// A SUPER server may shut down, but its Storage shared memory continues to live due to an embedded user session.
+			// The storage might allocate multiple slots, so count them carefully.
```
hvlad, line 484: Storage doesn't allocates slots, process does :)

hvlad (changes requested):
The logic is OK, but some better suitable container type and its name is required.

src/jrd/trace/TraceConfigStorage.cpp
Noremos: I didn't find a Firebird::set class in the code, so I used a map. If SortedArray is OK, then I use it

src/jrd/trace/TraceConfigStorage.cpp
Noremos: Corrected the description

hvlad (approved):

PR #8195, opened 2024-07-25
Minor static analysis fixes

marat-iskakov:
4 single-line edits for warnings given by using static analyzer

### DIFF
--- a/src/common/classes/RefCounted.h
+++ b/src/common/classes/RefCounted.h
@@ -175,7 +175,7 @@ namespace Firebird
 			return assign(r.ptr);
 		}
 
-		T* operator=(RefPtr&& r)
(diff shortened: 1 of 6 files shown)


src/common/classes/RefCounted.h
sim1984, line 1: In this case, noexcept is not entirely correct. A function can be marked `noexcept` if it cannot throw an exception. In this case, the `moveFrom` function is called internally, not `noexcept`.

marat-iskakov: yes, it seems while `moveFrom` should work with `noexcept`, it's further internal call of `assign` might complicate things

sim1984 (approved):
I don't see any other problems with this fix. But it's still not up to me to decide whether to merge.

PR #8198, opened 2024-07-28
Switch from const variables to inline constexpr in ods.h

TreeHunter9:
I think it's time to take advantage of new features of C++, and the most obvious and innocuous change is to switch from simple const or macros to inline constexpr variables in the header file.
For the first PR, I decide to take ods.h because it has a lot of const variables and I'd like to see what others think about this initiative.


dyemanov (approved):

aafemt:
While you are at it, may I suggest to change value of pag_header from 1 to 0xDBFB (0xFBDB for big-endian platforms)? This way Firebird database file will get a signature that can be recognizable by external utility or even naked eyes.

AlexPeshkoff:
On 8/9/24 12:41, Dimitry Sibiryakov wrote:
>
> While you are at it, may I suggest to change value of pag_header from 
> 1 to 0xDBFB (0xFBDB for big-endian platforms)? This way Firebird 
> database file will get a signature that can be recognizable by 
> external utility or even naked eyes.
>

No, that will break detection on pages to be or not encrypted in crypto 
manager (static bool array is used, page type - index).

PR #8199, opened 2024-07-29
Checking division by zero

BakhtiyorBahriddini:


aafemt:
Skipping whole print is not a good idea. I would suggest to use
```
if (seconds == 0)
  seconds = 1;
```
in the beginning of this routine instead.

PR #8201, opened 2024-08-01
Try to make cmake work again

raedrizqie:
Built successfully on MinGW-w64 systems (mingw64, ucrt64, clang64):
https://github.com/msys2/MINGW-packages/actions/runs/10094396971?pr=21492

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@
 #
 ################################################################################
 
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.15)
 
 # In-source builds are not possible and so disabled.
 if (${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
@@ -84,6 +84,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${output_dir})
 if (MSVC OR XCODE)
     set(output_dir ${output_dir}/$<CONFIG>)
     set(boot_dir ${boot_dir}/$<CONFIG>)
+elseif (MINGW)
+    set(output_dir ${CMAKE_BINARY_DIR})
 endif()
 
 if (MSVC AND DEFAULT_CONFIGURATION)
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -159,21 +161,55 @@ set(FB_SERVICE_NAME "gds_db")
...
+    if (MSVC)
+        set(FB_PREFIX "c:\\\\Program Files\\\\Firebird\\\\")
+    else()
+        set(FB_PREFIX ${CMAKE_INSTALL_PREFIX})
+    endif()
     set(FB_IPC_NAME "FIREBIRD")
 endif()
 
+if (MINGW)
+    set(FB_BINDIR "${FB_PREFIX}/bin")
```
asfernandes, line 173: My problem with this is that it makes a windows build to use fixed instead of relocatable paths.
In the past mingw was a Windows build tool, but then seems it (together with msys) become a packaging environment too. I'm not sure just a build using mingw should follow its package conventions.

I never used cmake with Firebird, but I like it (or dislike less than the configure+make approach) and maybe we can have a cmake build with mingw compatible with the current Windows MSVC build.

Don't see this comment as an objection, as a short term solution it may be ok for me.

Does it also work using MSVC?

raedrizqie: > My problem with this is that it makes a windows build to use fixed instead of relocatable paths.

`FB_PREFIX` is only used on build-time. The real final installation location is defined by `FB_INSTALL_PREFIX` (at least for mingw). MinGW does not support binreloc, so we add `pathtools.cpp` and `pathtools.h` to get relocatable paths.


>In the past mingw was a Windows build tool, but then seems it (together with msys) become a packaging environment too. I'm not sure just a build using mingw should follow its package conventions.

MinGW generally follows FHS convention.. but, if you guys prefer for it to follow MSVC directory structure, im ok with that. MSys2 can always patch it downstream.

> I never used cmake with Firebird, but I like it (or dislike less than the configure+make approach) and maybe we can have a cmake build with mingw compatible with the current Windows MSVC build.
> 
> Don't see this comment as an objection, as a short term solution it may be ok for me.

Great to hear that :)

> Does it also work using MSVC?

I never tried this setup on msvc. sorry..

hvlad:
I see no need and no way to use MinGW as developer platform on Windows. 
At least until we have free VS and its satisfy our needs.

MinGW could be used by anybody else for any other purposes, of course, but not me and not as main tool for developer.

raedrizqie:
> I see no need and no way to use MinGW as developer platform on Windows. At least as long as we have free VS and it satisfy our needs.
> 
> MinGW could be used by anybody else for any other purposes, of course, but not me and not as main tool for developer.

Fair enough for me :)

AlexPeshkoff:
Did you check cmake build on linux after mentioned changes?

raedrizqie:
> Did you check cmake build on linux after mentioned changes?

No, I didnt.. I only have access to windows. I am hoping someone else could help to improve it for other platform.

AlexPeshkoff (changes requested):
I myself never use cmake build and not ready to fix/check it on linux. I think in current state patch can't be accepted - IMO people sometimes do use it on posix systems while mingw is too exotic case.

PR #8204, opened 2024-08-04
Fixed bug #8203 : Function MAKE_DBKEY may produce random errors if used with relation name

### LINKED ISSUE #8203: Function MAKE_DBKEY may produce random errors if used with relation name
Because function `LiteralNode::getText() `returns string which is not zero-terminated. Particularly QA test gh_8168_test.py sometimes unexpectedly produces "Malformed string" error.

hvlad:
Also, corrected handling of non-ASCII tables names.

### DIFF
--- a/src/dsql/ExprNodes.cpp
+++ b/src/dsql/ExprNodes.cpp
@@ -8092,6 +8092,16 @@ dsc* LiteralNode::execute(thread_db* /*tdbb*/, Request* /*request*/) const
 	return const_cast<dsc*>(&litDesc);
 }
 
+bool LiteralNode::getMetaName(thread_db* tdbb, MetaName& name) const
+{
+	if (!DTYPE_IS_TEXT(litDesc.dsc_dtype))
+		return false;
+
+	CVT2_make_metaname(&litDesc, name, tdbb->getAttachment()->att_dec_status);
+
+	return true;
+}
+
 void LiteralNode::fixMinSInt64(MemoryPool& pool)
 {
 	// MIN_SINT64 should be stored as BIGINT, not 128-bit integer
@@ -12304,9 +12314,9 @@ DmlNode* SysFuncCallNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScrat
 
 		auto literal = nodeAs<LiteralNode>(node->args->items[0]);
 
-		if (literal && literal->litDesc.isText())
+		MetaName relName;
+		if (literal && literal->getMetaName(tdbb, relName))
 		{
-			const MetaName relName = literal->getText();
 			const jrd_rel* const relation = MET_lookup_relation(tdbb, relName);
 
 			if (relation)
--- a/src/dsql/ExprNodes.cpp
+++ b/src/dsql/ExprNodes.cpp
@@ -8092,16 +8092,6 @@ dsc* LiteralNode::execute(thread_db* /*tdbb*/, Request* /*request*/) const
 	return const_cast<dsc*>(&litDesc);
 }
 
-bool LiteralNode::getMetaName(thread_db* tdbb, MetaName& name) const
-{
-	if (!DTYPE_IS_TEXT(litDesc.dsc_dtype))
-		return false;
-
-	CVT2_make_metaname(&litDesc, name, tdbb->getAttachment()->att_dec_status);
-
-	return true;
-}
-
 void LiteralNode::fixMinSInt64(MemoryPool& pool)
 {
 	// MIN_SINT64 should be stored as BIGINT, not 128-bit integer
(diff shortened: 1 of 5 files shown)


src/dsql/ExprNodes.cpp
```diff
@@ -8092,6 +8092,16 @@ dsc* LiteralNode::execute(thread_db* /*tdbb*/, Request* /*request*/) const
 	return const_cast<dsc*>(&litDesc);
 }
 
+bool LiteralNode::getMetaName(thread_db* tdbb, MetaName& name) const
+{
+	if (!DTYPE_IS_TEXT(litDesc.dsc_dtype))
+		return false;
```
asfernandes, line 8098: With this being in the base LiteralNode, I expect it does automatic conversion to string.

hvlad: Not sure I got you point correctly. Are you saying that any non-string literal could be used as metadata name ?

asfernandes: Any? Not sure. It's used in this only place and at least for numbers it's not called.

But generally a place that accept strings may accept convertible things to strings.

If this name comes as blob in a operation involving a substring (also returning a blob) this assumption will be different.

If this method were directly related to make_dbkey only, I would not give the comment.

hvlad: Still not clear - what blobs or substring with literals ?
Could you be more explicit ?
Would you like me to add comment that this method works with string literals only ? 
Or remove check for data type ? Why ?

asfernandes: Take into account too that this method is going to die or have no use. `MAKE_DBKEY` will need to also interpret qualified names with schemas, and the way would be to get it as string and then parse to qualified name.

hvlad: Ok, removed.

asfernandes (approved):

src/dsql/ExprNodes.cpp
aafemt: > `MAKE_DBKEY` will need to also interpret qualified names with schemas, and the way would be to get it as string and then parse to qualified name.

If `MetaName` includes schema name and object name and `CVT2_make_metaname()` is made to parse them right and `MET_lokup_relation()` is aware that `MetaName` contains schema, what will be wrong?

hvlad: I have big doubt that MetaName will be multipart, its purpose is very different.
Also, I don't think that CVT2_make_metaname() will parse anything, it just convert string into metadata charset.

aafemt: The purpose on `MetaName` class as I understand it is to keep a name of a metadata object, right? With introduction of schemas the name become multipart so what may be a reason for `MetaName` not become multipart accordingly? In any case Firebird has to have a routine that convert some string literal into multipart name of metadata object so why `LiteralNode::getMetaName` cannot be/use one?

hvlad: > The purpose on `MetaName` class as I understand it is to keep a name of a metadata object, right? 

This is the secondary purpose. Learn it internals.

> With introduction of schemas the name become multipart so what may be a reason for `MetaName` not become multipart accordingly? 

To avoid huge grows of internal hash table. Learn MetaName.

> In any case Firebird has to have a routine that convert some string literal into multipart name of metadata object so why `LiteralNode::getMetaName` cannot be/use one?

Because such routine will be used in many more places than with literals only.

aafemt: > This is the secondary purpose.

  And the primary one is...? The **only** comment in MetaName.h says "DESCRIPTION:	metadata name holder". Nothing else is mentioned in visible letters.

> To avoid huge grows of internal hash table. Learn MetaName.

Could you please elaborate what exactly will hugely grow if someone add `Dictionary::Word* word2` into the `MetaName`?

hvlad: > > This is the secondary purpose.
> 
> And the primary one is...? The **only** comment in MetaName.h says "DESCRIPTION: metadata name holder". Nothing else is mentioned in visible letters.

  Find the ticket related with introducing MetaName. Its main purpose - save memory used by identifiers and other lexemas which is duplicated across queries many times.

> > To avoid huge grows of internal hash table. Learn MetaName.
> 
> Could you please elaborate what exactly will hugely grow if someone add `Dictionary::Word* word2` into the `MetaName`?

Why word2 only, how about word3 and so on ? schema.package.routine, for example.
MSSQL uses 4-part names, who more ?

MetaName should be used for single part only. QualifiedName is more complex beast, MetaName is a single brick.

hvlad: > but it doesn't answer the question "what will get huge grow"?

Putting all parts of qualified name into single MetaName. It was your first "idea" in this conversation.

> So, is there a reason why we cannot just rename MetaName into QualifiedName?

There are a lot of reasons, just learn the code.

This PR is about different thing and completed.
If you need someone to answer your smart questions, ask elsewhere.

src/dsql/ExprNodes.cpp
```diff
@@ -12306,7 +12306,9 @@ DmlNode* SysFuncCallNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScrat
 
 		if (literal && literal->litDesc.isText())
 		{
-			const MetaName relName = literal->getText();
+			MetaName relName;
+			CVT2_make_metaname(&literal->litDesc, name, tdbb->getAttachment()->att_dec_status);
```
asfernandes, line 12310: Must be `relName`, not `name` here.

hvlad: Thanks. I was sure I checked it :(

PR #8207, opened 2024-08-06
Add manual delete for tra_dbcreators_list

TreeHunter9:
I found that if we try to write a lot of records to `RecordBuffer` (>MIN_TEMP_BLOCK_SIZE) we start increasing `m_tempCacheUsage`. When we don't need `RecordBuffer` anymore, it will be destroyed in `~SnapshotData()`, which will call  `~TempSpace()` for `Firebird::AutoPtr<TempSpace> space`, and in `~TempSpace()` we will decrease `m_tempCacheUsage`.
But this doesn't apply to `tra_dbcreators_list` because we delete it by dropping the transaction pool, so destructor doesn't kick in and we don't decrease `m_tempCacheUsage`. All of this will cause `fb_assert(m_tempCacheUsage == 0)` to be thrown, but I don't know what this might lead to in release build (nothing good, I suppose).
This is also true for Firebird 5.0.

### DIFF
--- a/src/jrd/tra.cpp
+++ b/src/jrd/tra.cpp
@@ -3792,6 +3792,7 @@ jrd_tra::~jrd_tra()
 	delete tra_user_management;
 	delete tra_timezone_snapshot;
 	delete tra_mapping_list;
+	delete tra_dbcreators_list;
 	delete tra_gen_ids;
 
 	if (!tra_outer)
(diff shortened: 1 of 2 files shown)


TreeHunter9:
Maybe it would be better to wrap `tra_dbcreators_list` in a smart pointer?

AlexPeshkoff (approved):
Please also backport this fix to old branches

dyemanov:
I will backport it.

PR #8220, opened 2024-08-20
Fix #8219 - Database creation in 3.0.12, 4.0.5 and 5.0.1 slower than in previous releases

### LINKED ISSUE #8219: Database creation in 3.0.12, 4.0.5 and 5.0.1 slower than in previous releases
While running the Jaybird tests against Firebird 5.0.1, I noticed a slow down in test execution compared to Firebird 5.0.0.

The problem seems to be that database creation in Firebird 5.0.1 is roughly 30% slower compared to Firebird 5.0.0. For example, if I create 500 database against 5.0.0 (on Windows) it takes ~41.7 seconds, and on 5.0.1 it takes ~55.2 seconds.

Both 5.0.0 and 5.0.1 use the same configuration. The databases are created with `isc_dpb_force_write` set to 1, but not setting it makes no significant difference).

Firebird 4.0.5 and 3.0.12 have a similar performance decrease creating database compared to 4.0.4/3.0.11:

- 4.0.4: ~40.5 seconds, 4.0.5: ~56.8 seconds
- 3.0.11: ~41.8 seconds, 3.0.12: ~60.2 seconds 

I used the following test (from within Jaybird's test (placed in src/test/org/firebirdsql/jdbc):

```java
package org.firebirdsql.jdbc;

import org.firebirdsql.common.FBTestProperties;
import org.firebirdsql.management.FBManager;
import org.junit.jupiter.api.RepeatedTest;
import org.junit.jupiter.api.RepetitionInfo;
import org.junit.jupiter.api.io.TempDir;

import java.nio.file.Path;

import static org.firebirdsql.common.FBTestProperties.configureFBManager;

/**
 * @author Mark Rotteveel
 */
class SandboxTest {

    @TempDir
    static Path tempDir;

    @RepeatedTest(500)
    void repeatedDatabaseCreation(RepetitionInfo repetition) throws Exception {
        String dbName = "database-%d.fdb".formatted(repetition.getCurrentRepetition());
        Path dbFile = tempDir.resolve(dbName);

        try (var manager = configureFBManager(new FBManager())) {
            manager.createDatabase(dbFile.toString(), FBTestProperties.DB_USER, FBTestProperties.DB_PASSWORD);
        }
    }

}
```

asfernandes:


AlexPeshkoff (approved):
Longterm MET_get_char_coll_subtype_info() should be optimized to search in the cache before doing system tables lookup, but currently this fix is definitely enough.

PR #8238, opened 2024-09-03
Fix using macro with regex in path parameter in fbtrace.conf

TreeHunter9:
Currently we cannot use macro with regex in path parameter because `PathUtils::fixupSeparators(value.begin());` happens too early, and it will fix `\1` to `/1` before we replace it with regex group, so we get unexpected path.
On Windows, we just get an error, because `PathUtils::fixupSeparators(value.begin());` will change all `/` to `\`, and it will cause  `expandPattern()` to throw an exception.

### DIFF
--- a/src/common/config/config_file.cpp
+++ b/src/common/config/config_file.cpp
@@ -441,9 +441,18 @@ bool ConfigFile::macroParse(String& value, const char* fileName) const
 			return false;
 		}
 
-		// Avoid incorrect slashes in pathnames
-		PathUtils::fixupSeparators(value.begin());
-		PathUtils::fixupSeparators(macro.begin());
+		if (flags & FORCE_SLASHES_INSTEAD_OF_BACKSLASHES)
+		{
+			size_t pos = 0;
+			while ((pos = macro.find('\\', pos)) != String::npos)
+				macro[pos] = '/';
+		}
+		else
+		{
+			// Avoid incorrect slashes in pathnames
+			PathUtils::fixupSeparators(value.begin());
+			PathUtils::fixupSeparators(macro.begin());
+		}
 
 		if (subFrom > 0 && value[subFrom - 1] == PathUtils::dir_sep &&
 			macro.length() > 0 && macro[0] == PathUtils::dir_sep)
--- a/src/common/config/config_file.cpp
+++ b/src/common/config/config_file.cpp
@@ -414,6 +414,41 @@ ConfigFile::LineType ConfigFile::parseLine(const char* fileName, const String& i
  *	Substitute macro values in a string
  */
 
+unsigned ConfigFile::getDirSeparatorLength(const String& value, size_t separatorPosition) const
+{
+	if (separatorPosition >= value.length())
+		return 0;
+
+	const char symbol = value[separatorPosition];
+	if (symbol == '/')
+		return 1;
+
+	if (flags & REGEXP_SUPPORT && symbol == '\\')
+	{
+		// Check forward and backward for second backslash
+		if (separatorPosition + 1 < value.length() && value[separatorPosition + 1] == '\\')
+			return 2;
+		if (separatorPosition > 0 && value[separatorPosition - 1] == '\\')
+			return 2;
+	}
+	else if (symbol == '\\')
+		return 1;
+
+	return 0;
+}
+
+void ConfigFile::adjustMacroReplacePositions(const String& value, const String& macro, String::size_type& from, String::size_type& to) const
+{
+	if (macro.empty())
+		return;
+
+	// Remove dir separators from value string if macro already contains them at start or end
+	if (macro[0] == PathUtils::dir_sep && from > 0)
+		from -= getDirSeparatorLength(value, from - 1);
+	if (macro[macro.length() - 1] == PathUtils::dir_sep)
+		to += getDirSeparatorLength(value, to);
+}
+
 bool ConfigFile::macroParse(String& value, const char* fileName) const
 {
 	String::size_type pos = 0;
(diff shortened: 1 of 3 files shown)


src/common/config/config_file.cpp
```diff
@@ -441,9 +441,18 @@ bool ConfigFile::macroParse(String& value, const char* fileName) const
...
+			size_t pos = 0;
+			while ((pos = macro.find('\\', pos)) != String::npos)
+				macro[pos] = '/';
+		}
+		else
+		{
+			// Avoid incorrect slashes in pathnames
+			PathUtils::fixupSeparators(value.begin());
+			PathUtils::fixupSeparators(macro.begin());
+		}
```
TreeHunter9, line 455: I'm trying to keep the old implementation, so I add FORCE_SLASHES_INSTEAD_OF_BACKSLASHES flag, but maybe we can just remove it because `PathUtils::fixupSeparators();` will happen in `PATH_PARAMETER` after all parsing steps anyway?

AlexPeshkoff: Please provide a sample of causing problems trace configuration and suggested patch in more traditional form (i.e. 'diff -u' acceptable by  patch utility).

TreeHunter9:
```
database = %[\\/](employee).fdb
{
	enabled = true
	log_filename = $(root)\1.log
	log_connections = true
	time_threshold = 0
}
```
Log filename should be `fb_root_path/employee.log`, but we get `fb_root_path/1.log`.

TreeHunter9: On Windows it doesn't even get created because it's trying to parse backslashes as part of regex pattern (\1, \2, etc) but encounters \d[ev] in the file path and throws an error:
```
DESKTOP-86JTPHF	Tue Sep  3 10:37:29 2024
	Trace plugin fbtrace returned error on call trace_create.
	error while parsing trace configuration
	line 195, element "log_filename": pattern is invalid
	 E:\dev\MyFirebird\firebird\output_x64_debug\1.log
```

AlexPeshkoff (changes requested):
I'm afraid suggested solution is not complete. Imagine someone types (in windows):
log_filename = $(root)\audit-logs\\1.log
Some problems are hardly avoidable when same char is used for 2 different purposes, but lets try to minimize them.

I suggest to use something like REGEXP_SUPPORT instead FORCE_SLASHES_INSTEAD_OF_BACKSLASHES and carefully parse both value and macro (moreover, value should be parsed when flag is set even when no macro-subst is done) in order to replace backslash with slash when next char is not a digit. At the first glance it appears more generic. Certainly if someone called directory '1' and tries to use it for logs I do not know how to guess what to do with that.

aafemt:
> Some problems are hardly avoidable when same char is used for 2 different purposes, but lets try to minimize them.

If you are ready to review the initial design, I would suggest to use `$(1)` for subpattern substitutions.

AlexPeshkoff:
On 9/5/24 17:52, Dimitry Sibiryakov wrote:
>
>     Some problems are hardly avoidable when same char is used for 2
>     different purposes, but lets try to minimize them.
>
> If you are ready to review the initial design, I would suggest to use 
> |$(1)| for subpattern substitutions.
>

Looks as a minimum interesting - but discussion is devel is needed for 
such change.

TreeHunter9:
> I'm afraid suggested solution is not complete. Imagine someone types (in windows):
log_filename = $(root)\audit-logs\1.log
Some problems are hardly avoidable when same char is used for 2 different purposes, but lets try to minimize them.

Oh, for some reason I thought only `/` was allowed as a separator in the path. Now I looked more into it and found out that we should use `\\` for backslashes in path (as it says in fbtrace.conf), and here is code snippet from TraceCfgReader::expandPattern:
https://github.com/FirebirdSQL/firebird/blob/72cc0880e25c8fbd8020d56d6e2700f1c7e493c1/src/utilities/ntrace/TraceConfiguration.cpp#L291-L297
So I guess the problem only remains on Linux because of `PathUtils::fixupSeparators(value.begin());` will fixup `\1` regex pattern.
I will do another test tomorrow.

TreeHunter9:
> I'm afraid suggested solution is not complete. Imagine someone types (in windows):
log_filename = $(root)\audit-logs\1.log

We should write path like this: `$(root)audit-logs\\\1.log`, $(root) already have `\\` at the end, and `audit-logs\\\1.log` so we can translate it into `audit-logs\employee.log`. So I guess we don't need to change the semantics of the regexp.
So we really only have one problem, where macro insertion fixes all regexp patterns (on Windows macro insertion uses incorrect backslashes for path) and this patch should fix that.

AlexPeshkoff:
Macro in config design supposed to provide smart behavior when merging paths, i.e. one need not know that $(root) already has \\ at the end. In linux having double dir separators one after another (something like /usr//bin instead /usr/bin) is not dangerous but I'm not sure for other OS-es. Anyway, resulting file name may get visible to user somewhere else, i.e. I prefer to have correctly merged paths, not relying on presence of // at the end. Also we always tried to have config format as OS independent as possible, i.e. one should be able to have
$(root)\\audit-logs\\\1.log on linux or
$(root)/audit-logs/\1.log on windows (may be as a result of file copy when changing OS).

TreeHunter9:
Tested on these file paths:
```
Windows and Linux:
$(root)\1.log -> /path/to/root/employee.log
$(root)/\1.log -> /path/to/root/employee.log
$(root)\\\1.log -> /path/to/root/employee.log
$(root)\\1.log -> /path/to/root/1.log

Only Linux:
/$(root)/\1.log -> /path/to/root/employee.log
\\$(root)/\1.log -> /path/to/root/employee.log
```

AlexPeshkoff (approved):

AlexPeshkoff:
I suggest you backport changes to previous versions

dyemanov:
I will backport it.

PR #8242, opened 2024-09-05
Move the call of the checkArgsMismatch function before the call of th…

Zhdanov0:
…e setParamsFunc function

Why checkArgsMismatch is executed after the setParamsFunc function and not before it? I think it is more logical to make the checkArgsMismatch check at the place where I put it in the PR. This location is more advantageous because there will be no need to do extra work in the form of calling the setParamsFunc function and other things if you know in advance that the number of arguments is incorrect.

Also, thanks to this location, we don't need to add the “if (argsCount > 1)” check to the setParamsMakeDbKey function, which saves us from unnecessary branching. If you are happy with this solution, I can remove these checks in the existing setParams functions.

I doubt about removing the checkArgsMismatch function call in SysFuncCallNode::pass2.
I haven't found what it might affect and don't understand why it was there.

### DIFF
--- a/src/dsql/ExprNodes.cpp
+++ b/src/dsql/ExprNodes.cpp
@@ -12321,6 +12321,8 @@ DmlNode* SysFuncCallNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScrat
 
 	node->args = PAR_args(tdbb, csb);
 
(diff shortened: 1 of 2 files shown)


hvlad:
This PR have no lables, please add

PR #8255, opened 2024-09-17
Catch possible stack overflow when preparing and compiling user statements

hvlad:


hvlad:
It fixed #6467, #6763 and probably more similar issues.
Current code ifdef'ed with WIN_NT by two reasons:
- it can't catch stack overflow on other platforms
- it adds performance penalty on Linux as was shown in fb-devel (see thread "[firebird-devel] UDF vs UDR speed (was Re: Deprecation)" 10.01.2024)

src/dsql/dsql.cpp
```diff
@@ -439,6 +441,23 @@ static dsql_dbb* init(thread_db* tdbb, Jrd::Attachment* attachment)
 	return attachment->att_dsql_instance;
 }
 
+// Use SEH frame when preparing user requests to catch possible stack overflows
+static DsqlRequest* safePrepareRequest(thread_db* tdbb, dsql_dbb* database, jrd_tra* transaction,
+	ULONG textLength, const TEXT* text, USHORT clientDialect, bool isInternalRequest)
+{
+	if (isInternalRequest)
+		return prepareRequest(tdbb, database, transaction, textLength, text, clientDialect, isInternalRequest);
```
dyemanov, line 449: `isInternalRequest -> true` maybe ? And `isInternalRequest -> false` below.

hvlad: No objection, just curious - for what ? ;)

dyemanov: Two absolutely identical calls looks suspicious at the first glance ;-) You already checked for `isInternalRequest`, so the result is known. But I don't insist, it's a highly subjective point.

hvlad: Ok, done

PR #8268, opened 2024-09-26
Fix refetch header data from delta when database in backup lock

XaBbl4:
If for `dbb_sweep_interval` this is not so critical, then for `dbb_repl_sequence `this can damage the replica, since the segments will be saved with the old number (saved in the main header at the time of transition to the backup lock state) after each clean connection to the database.

### Simple example for dbb_sweep_interval:
1 console - run isql
2 console - operations with gfix

```
1> create database 'test';
2> gfix -h 500 test
1> select mon$sweep_interval from mon$database; commit;
MON$SWEEP_INTERVAL
==================
               500
1> alter database begin backup;
2> gfix -h 1000 test
1> select mon$sweep_interval from mon$database; commit;    -- without reconnection to DB
MON$SWEEP_INTERVAL
==================
               1000             -- CORRECT
1> quit;
1> connect 'test';
1> select mon$sweep_interval from mon$database; commit;    -- after reconnection
MON$SWEEP_INTERVAL
==================
               500              -- WRONG (in delta file dbb_sweep_interval = 1000)
```

### DIFF
--- a/src/jrd/pag.cpp
+++ b/src/jrd/pag.cpp
@@ -1055,7 +1055,8 @@ void PAG_header(thread_db* tdbb, bool info, const TriState newForceWrite)
 	fb_assert(attachment);
 
 	WIN window(HEADER_PAGE_NUMBER);
-	header_page* header = (header_page*) CCH_FETCH(tdbb, &window, LCK_read, pag_header);
+	pag* page = CCH_FETCH(tdbb, &window, LCK_read, pag_header);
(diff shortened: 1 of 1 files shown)


livius2:
What about other flags?

XaBbl4:
> What about other flags?

As far as I can see, the other flags do not change after begin backup.
There are some doubts about the `HDR_crypt_...` flags, but they are also not read in PAG_init2, which requires refetch.

Issue #8249, opened 2024-09-11
CAST() ignores collation of target data type


aafemt:
@aafemt this change introduced a problem making this query work:
```
select cast('x' as varchar(10) character set utf8 collate wrong) from rdb$database;
```

_Originally posted by @asfernandes in https://github.com/FirebirdSQL/firebird/issues/7748#issuecomment-2342463807_

asfernandes:
This is a regression introduced in #7748

PR #8270, opened 2024-09-28
Fix for #8249

aafemt:


romansimakov:
Let's please write more details in commit messages. A reference to a ticket is not enough IMO. Ticket describes WHAT is fixed but a commit message should describe HOW it was fixed.

src/isql/isql.epp
```diff
@@ -9597,33 +9597,24 @@ static const char* charset_to_string(unsigned charset)
  **************************************/
 	static Firebird::GlobalPtr<Firebird::GenericMap<Firebird::Pair<Firebird::Right<unsigned, string> > > > csMap;
 
-	charset = TTYPE_TO_CHARSET(charset);
```
asfernandes, line 9600: What is the relation of this isql changes with #8249 ?

aafemt: QA team and me should have a way to check field collation to make sure that #8249 really fixed and result of the `CAST()` has the right collation. The simplest way - to print collation in SQLDA along to the charset. That's why this change.

aafemt:
> commit message should describe HOW it was fixed.

It is written in code: `CastNode` now sends the right collation name instead of `NULL` to `DDL_resolve_intl_type()`.

src/isql/isql.epp
asfernandes: First, ISQL showed the charset/collation id, so it was identifiable.
You want to display also the collation name, but this change makes it incorrectly. Collation ids, different from charset ids, are far from be stable between different databases. I suggest to revert this part of the patch.

asfernandes (changes requested):

PR #8273, opened 2024-09-29
Reorganize public headers

asfernandes:
Reorganize public headers so files in `include/firebird` do not depend on files in `include` and a compiler include dir settings (`-I`) present.

This also deprecates files in `include` recommending that people include files directly present in `firebird/*`.


aafemt:
Are there still compilers in use that don't understand `#pragma once` and require old-style guards?

asfernandes:
> Are there still compilers in use that don't understand `#pragma once` and require old-style guards?

This has nothing to do with this PR and will not be changed here. Please open new thread if you want to discuss this.

But anyway, it's not standard.

PR #8285, opened 2024-10-16
Update the libcxx (llvm project) to version 19.1.1

Noremos:
Update the libcxx custom package to remove the dependency on an old Python version


Noremos:
The ARM build failed during the unit test compilation. It seems like it is necessary to update BOOST. It uses std::unary_function, which was deleted in C++17. I checked, and the updated BOOST version works. It would be nice to update BOOST, but it consists of over 2000 files. I can create another PR, but I'm not sure if all this is really needed by anyone besides me.

aafemt:
If you are sure that this piece of BOOST cannot be eliminated from Firebird tree completely - try to set `_HAS_AUTO_PTR_ETC` to 1. According to the BOOST sources it should switch it to usage of own implementation of `unary_function`.

asfernandes:
Doing boost update in https://github.com/FirebirdSQL/firebird/pull/8286

.github/workflows/main.yml
```diff
@@ -312,15 +312,6 @@ jobs:
...
-        run: |
-          brew uninstall --ignore-dependencies python@3.12
-          [ -e /Library/Frameworks/Python.framework ] && \
-            sudo mv /Library/Frameworks/Python.framework /Library/Frameworks/Python.framework.dont-look-at-this
-          brew install python@3.11
-          ln -s "$(brew --prefix python@3.11)/bin/python3" "$(brew --prefix python@3.11)/bin/python"
-
       - name: Cache - vcpkg
         uses: actions/cache@v4
         with:
```
asfernandes, line 317: Should also change cache key below, to something like `key: ${{ runner.os }}-vcpkg-1`

.github/workflows/main.yml
```diff
@@ -312,27 +312,18 @@ jobs:
...
+          key: ${{ runner.os }}-vcpkg-fb6-0
           path: |
             $HOME/.cache/vcpkg/archives
 
       - name: Build
         run: |
           export LIBTOOLIZE=glibtoolize
           export LIBTOOL=glibtool
-          export PATH="$(brew --prefix python@3.11)/bin:$PATH"
+          export PATH="$(brew --prefix python3)/bin:$PATH"
```
asfernandes, line 326: Is this still necessary?

Noremos:
I'm a little concerned about the “Cache not found for input keys” message in the cache action. Does it really work?

asfernandes:
> I'm a little concerned about the “Cache not found for input keys” message in the cache action. Does it really work?

Seems it was not already working.
The path is invalid.

Noremos:
I tried using a different approach for vcpkg dependency caching, described in https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-actions-cache. 
And it seems to be working: https://github.com/Noremos/firebird/commit/4b5eeb2f799d43ea379a34bc467c85c9525381f2

`Restored 3 package(s) from GitHub Actions Cache in 12 s.`

asfernandes:
It's an experimental flag and is not clear for me what it use as cache key and how it will be in regard to multiple arch and branches.
But we may go with it and see.

asfernandes:
I did some tests and it caches packages individually with their hashs. Seems good.
Could you incorporate it here?
Is it the only pendency for merge?

Noremos:
> Could you incorporate it here?
Commied the change

> Is it the only pendency for merge?
Yes. If there are no problems with CI and there are no additional issues, then everything is ready

PR #8289, opened 2024-10-21
Improve vscode settings for debugging

Noremos:
It is sometimes necessary to debug gpre-generated files, but this is challenging in VS Code due to the current settings. All temporary files are excluded, making it impossible to access them from the VS Code file explorer. Therefore, I propose removing the exclusion option and instead marking these files as read-only.

Excluding generated files is also inconvenient for debugging because it prevents the setup of config files.

Additionally, I have added an association for .epp files with C++. Editing them with syntax errors is less problematic than editing them as plain text.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,5 +15,9 @@
 	"files.insertFinalNewline": true,
 	"files.associations": {
 		"*.epp": "cpp"
+	},
+	"search.exclude":{
+		"temp/**": true,
+		"gen/**": true
 	}
 }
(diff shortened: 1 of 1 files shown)


asfernandes:
I think adding a `"search.exclude"` for them would be good.

.vscode/settings.json
```diff
@@ -9,11 +9,15 @@
...
 	"files.readonlyInclude": {
-		"temp/**":true
+		"temp/**": true
 	},
 
 	"files.insertFinalNewline": true,
 	"files.associations": {
 		"*.epp": "cpp"
+	},
+	"search.exclude":{
```
asfernandes, line 19:
```suggestion
	"search.exclude": {
```

PR #8307, opened 2024-11-05
Wire protocol improvement: prefetch blob info and some data when open blob

hvlad:
This is the first part of wire protocol improvements with regards to reading blobs.

In this part I offer to group op_open_blob2, op_info_blob and op_get_segment logical packets into single physical packet.
It allows to prefetch and cache blob info and some blob data (up to 16KB currently) in single network roundtrip. 
If application ask for blob info after open - return cached blob info without remote access.

This patch requires no protocol changes, no server code changes and should be fully compatible with older Firebird servers.

### DIFF
--- a/src/common/utils_proto.h
+++ b/src/common/utils_proto.h
@@ -272,7 +272,7 @@ namespace fb_utils
 	inline unsigned char* putInfoItemInt(const unsigned char item, T value,
 		unsigned char* ptr, const unsigned char* end)
 	{
-		static_assert(std::is_integral_v<T>, "Intergal type expected");
+		static_assert(std::is_integral_v<T>, "Integral type expected");
 
 		constexpr auto len = sizeof(T);
 		static_assert(len == 1 || len == 2 || len == 4 || len == 8, "unknown data type");
--- a/src/common/utils_proto.h
+++ b/src/common/utils_proto.h
@@ -30,10 +30,13 @@
 #define INCLUDE_UTILS_PROTO_H
 
 #include <string.h>
+#include <type_traits>
+
 #include "../common/classes/fb_string.h"
 #include "../common/classes/array.h"
 #include "iberror.h"
 #include "firebird/Interface.h"
+#include "memory_routines.h"
 
 #ifdef SFIO
 #include <stdio.h>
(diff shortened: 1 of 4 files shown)


src/remote/remote.h
```diff
@@ -221,6 +221,27 @@ struct Rtr : public Firebird::GlobalStorage, public TypedHandle<rem_type_rtr>
 };
 
 
+struct RBlobInfo
+{
+	bool	valid;
+	UCHAR	blob_type;
```
asfernandes, line 227: I believe all of us consider the `camelCase` field as more modern case convention of Firebird.
Many things there still use that underscores separators, but many things there are using camel case.
So why not use it in a new type?

src/remote/remote.cpp
```diff
@@ -855,6 +856,87 @@ ISC_STATUS* Rdb::get_status_vector() throw()
...
+	return true;
+}
+
+
+void RBlobInfo::parseInfo(unsigned int bufferLength, unsigned char* buffer)
+{
+	int c = 0;
+	valid = false;
+
+	Firebird::ClumpletReader p(Firebird::ClumpletReader::InfoResponse, buffer, bufferLength);
```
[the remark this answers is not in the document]
hvlad: WIll do in separate commit after this PR

src/remote/remote.h
hvlad: I usually prefer `camelCase` too (as well as `m_` prefix for class fields) but not in this case: these members named after the public constants `isc_info_blob_XXX` and such usage is very convenient, IMO.
Look at `RBlobInfo::getLocalInfo()` and `RBlobInfo::parseInfo()`, for example.

src/remote/client/interface.cpp
```diff
@@ -5610,7 +5614,79 @@ IBlob* Attachment::openBlob(CheckStatusWrapper* status, ITransaction* apiTra, IS
...
+			information->p_info_items.cstr_length = sizeof(items);
+			information->p_info_items.cstr_address = items;
+			information->p_info_buffer_length = sizeof(infoBuffer);
+
+			send_partial_packet(port, packet);
+
+			// prefetch some data
+			packet->p_operation = op_get_segment;
+			P_SGMT* segment = &packet->p_sgmt;
+			segment->p_sgmt_length = BLOB_LENGTH - 2;
```
dyemanov, line 5645: What `-2` means here?

hvlad: Reserved space for segment length.
Now, when you point here, I not so sure it is necessary.

src/common/utils_proto.h
```diff
@@ -263,6 +266,46 @@ namespace fb_utils
 	// Frequently used actions with clumplets
 	bool isBpbSegmented(unsigned parLength, const unsigned char* par);
 
+
+	// Put integer value into info buffer
+	template<typename T>
+	inline unsigned char* putInfoItemInt(const unsigned char item, T value,
+		unsigned char* ptr, const unsigned char* end)
+	{
+		static_assert(std::is_integral_v<T>, "Intergal type expected");
```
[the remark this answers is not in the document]
hvlad: Will fix, thanks

src/remote/remote.cpp
```diff
@@ -855,6 +856,87 @@ ISC_STATUS* Rdb::get_status_vector() throw()
...
+			// unknown info item, let remote server handle it
+			return false;
+		}
+	}
+
+	return true;
+}
+
+
+void RBlobInfo::parseInfo(unsigned int bufferLength, unsigned char* buffer)
```
AlexPeshkoff, line 904: May be better
const unsigned char* buffer
?

hvlad: Sure, thanks

sim1984:
5.0.0? May be 5.0.2?

hvlad:
> 5.0.0? May be 5.0.2?

Sure, thanks

WarmBooter:
It is valid to mention that initial tests in high latency network (ping 149ms) show an average transfer speed improvement of 33-49% with this patch.

src/common/utils_proto.h
```diff
@@ -263,6 +266,46 @@ namespace fb_utils
...
+		*ptr++ = 0;
+
+		if (len == 8)
+			put_vax_int64(ptr, value);
+		else if (len == 4)
+			put_vax_long(ptr, value);
+		else if (len == 2)
+			put_vax_short(ptr, value);
+		else if (len == 1)
+			*ptr = value;
```
TreeHunter9, line 302: I'm a bit too late, but maybe we can use `if constexpr/else if constexpr` to evaluate this condition at compile time (as I can see `len` is constexpr variable, so this should work)?
Note: This is C++17 feature.

hvlad: Done, thanks

PR #8309, opened 2024-11-07
Add `ALTER PACKAGE BODY` and `CREATE OR ALTER PACKAGE BODY` parse rules

Noremos:
Hello. 
Currently, it is possible to CREATE, ALTER, or RECREATE a PACKAGE 
```SQL
set autoterm;
CREATE PACKAGE TEST_PACKAGE
AS
BEGIN
    FUNCTION DUMMY() RETURNS INT;
END;

CREATE OR ALTER PACKAGE TEST_PACKAGE
AS
BEGIN
    FUNCTION DUMMY() RETURNS INT;
END;

ALTER PACKAGE TEST_PACKAGE
AS
BEGIN
    FUNCTION DUMMY() RETURNS INT;
END;

RECREATE PACKAGE TEST_PACKAGE
AS
BEGIN
    FUNCTION DUMMY() RETURNS INT;
END;
```

However, when it comes to a package body, some operations may fail:

```SQL
set term ^;
ALTER PACKAGE BODY TEST_PACKAGE
AS
BEGIN
    FUNCTION DUMMY() RETURNS INT
    AS
    BEGIN
        RETURN 1;
    END
END^
```

This results in the following error message:

```
Dynamic SQL Error
-SQL error code = -104
-Token unknown - line 1, column 20
-TEST_PACKAGE
```

This is a really confusing error message. The problem is that the PACKAGE BODY does not support the ALTER or CREATE OR ALTER statements, yet the message references the package name, which can be misleading.

I think it is worth adding this parse rules to not confuse the users.
From my perspective, ALTER PACKAGE BODY is essentially equivalent to RECREATE statement, so I used the RECREATE node without changing the package code.

### DIFF
--- a/src/dsql/parse.y
+++ b/src/dsql/parse.y
@@ -863,6 +863,7 @@ using namespace Firebird;
 	Jrd::SetDecFloatTrapsNode* setDecFloatTrapsNode;
 	Jrd::SetBindNode* setBindNode;
 	Jrd::SessionResetNode* sessionResetNode;
+	Jrd::RecreatePackageBodyNode* recreatePackageBodyNode;
 }
 
 %include types.y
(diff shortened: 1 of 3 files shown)


asfernandes:
Please update doc/sql.extensions/README.packages.txt

src/dsql/parse.y
```diff
@@ -863,6 +863,7 @@ using namespace Firebird;
 	Jrd::SetDecFloatTrapsNode* setDecFloatTrapsNode;
 	Jrd::SetBindNode* setBindNode;
 	Jrd::SessionResetNode* sessionResetNode;
+	Jrd::RecreatePackageBodyNode* recreatePackageBodyNode;
```
asfernandes, line 866: Can this declaration be eliminated if you declare `replace_package_body_clause` to return a `ddlNode`?

Noremos: Good idea

PR #8314, opened 2024-11-12
Support for tablespaces

Zhdanov0:
Link to proposal: https://groups.google.com/g/firebird-devel/c/geF0A2_kC28


sim1984:
I didn't find a README.tablespaces file in this PR.

Zhdanov0:
> I didn't find a README.tablespaces file in this PR.

Will be added.

dyemanov (approved):

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    3) Split indices from the database
+
+-----------------
+SYNTAX
+-----------------
+
+    1. TABLESPACE
+    CREATE TABLESPACE [IF NOT EXISTS] <TS NAME> FILE '/path/to/file'
+    ALTER TABLESPACE <TS NAME> SET FILE [TO] '/path/to/file'
+    You can specify either an absolute path or a relative path (relative to the database file)
```
AlexPeshkoff, line 18: CREATE OR ALTER TABLESPACE ???

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+
+    New field in RDB$RELATION_FIELDS:
+    RDB$TABLESPACE_NAME - CHAR (63)
+
+    New fields in RDB$RELATIONS:
+    RDB$TABLESPACE_NAME - CHAR (63)
+
+    By default, the RDB$TABLESPACES table stores a record about the PRIMARY tablespace.
+    PRIMARY tablespace has the system flag and FILE_NAME matches the path to the database file.
+
```
AlexPeshkoff, line 87: What if database file copied to another location?

builds/make.new/config/install-sh~
Cool-Hunter, line 1: Is it bak copy of the file install.sh? Is it to be removed?

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    3. INDEX
+    CREATE INDEX ...
+    [[IN] TABLESPACE {<TS NAME> | PRIMARY}]
+
+    By default, table indexes are created in the same tablespace as the table itself.
+
+    ALTER INDEX ...
+    [SET TABLESPACE [TO] {<TS NAME> | PRIMARY}]
+
+    The index data will be moved to the specified tablespace or to the main database.
```
Cool-Hunter, line 50: May be to add about necessary locks for the index movement.

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    CREATE TABLE ...
+    [[IN] TABLESPACE {<TS NAME> | PRIMARY}]
+
+    It is also possible to specify a tablespace when creating a column or table constraint (unique, primary key, references):
+
+    <column/table constraint> ::= ... UNIQUE ... [[IN] TABLESPACE {<TS NAME> | PRIMARY}] | PRIMARY ... [[IN] TABLESPACE {<TS NAME> | PRIMARY}] | REFERENCES ... [[IN] TABLESPACE {<TS NAME> | PRIMARY}] ...
+
+    ALTER TABLE <TABLE NAME> SET TABLESPACE [TO] {<TS NAME> | PRIMARY}
+
+    The table data will be moved to the specified tablespace or to the main database.
```
Cool-Hunter, line 39: May be add about necessary locks for the table's data movement

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    -ts_orig[inal_paths]
+    To restore tablespaces to the original paths they were on when the backup was created.
+    It is still possible to override paths for some tablespaces using the -ts and -ts_map options.
+    This is an explicit option, not a default action.
+    The option does not overwrite existing files.
+
+    If you do not specify the above options, when restoring a database that has tablespaces,
+    an error about the inability to determine the path to restore tablespaces will occur.
+
+    3. Show
```
Cool-Hunter, line 121: is it a command of isql?

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+  pag_scns and pag_pip are located in every tablespace.
+  pag_root is located in the tablespace where a table is located.
+
+  An algorithm for moving data to another tablespace:
+  First, you have to move all the pages to another tablespace:
+  The main steps are:
+   - allocate necessary number of pointer pages by extents.
+   - allocate the rest of pointer pages by pages.
+   - walking through PPs allocate DPs by pages or extents.
+   - fix every PP by correcting DP numbers and ppg_next pointer and build a map
```
Cool-Hunter, line 145: still allocate new data pages on demand for data pages with "orphan" flags which are not in the PP.

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+   - walking through the map and copy every DP to the new one by fixing
+     b_page and f_page numbers.
+  At the end of work insert new records into RDB$PAGES and mark existing pages for deletion.
+
+-----------------
+CONSTRAINTS
+-----------------
+
+    It's possible to create up to 253 tablespaces.
```
Cool-Hunter: Is the limit still exists?

sim1984:
Will tablespace support for the `nbackup` utility be in a separate PR?

hvlad:
> Will tablespace support for the `nbackup` utility be in a separate PR?

Yes, AFAIK

sim1984:
The new monitoring table `MON$TABLESPACE_STATS` and the new `MON$STAT_GROUP` value for tablespaces are not mentioned in either `doc/README.monitoring_tables` or `doc/README.tablespaces`.

sim1984:
I see in `doc/README.monitoring_tables`

```
-----------------
PLANS
-----------------

    1. TEMPORARY predefined tablespaces.
    2. Grouping page counters by tablespace (For output in trace and monitoring).
    ...
```

Hasn't point 2 already been implemented?

src/dsql/TablespaceNodes.epp
```diff
@@ -0,0 +1,391 @@
...
+				X IN RDB$TABLESPACES USING
+			{
+				X.RDB$TABLESPACE_ID = id;
+				strcpy(X.RDB$TABLESPACE_NAME, name.c_str());
+				X.RDB$SYSTEM_FLAG = 0;
+
+				X.RDB$OWNER_NAME.NULL = FALSE;
+				strcpy(X.RDB$OWNER_NAME, userName.c_str());
+
+				if (fileName.length() >= sizeof(X.RDB$FILE_NAME))
```
Cool-Hunter: It might be checked erlier?

src/jrd/jrd.cpp
```diff
@@ -6908,6 +6918,36 @@ static bool drop_file(const Database* dbb, const jrd_file* file)
 	return status->getState() & IStatus::STATE_ERRORS ? true : false;
 }
 
+static bool drop_files(ObjectsArray<PathName>& tsFiles)
```
Cool-Hunter, line 6921: Just "files" is better than "tsFiles".

src/jrd/met.epp
```diff
@@ -3160,6 +3172,22 @@ ScanResult jrd_rel::scan(thread_db* tdbb, ObjectBase::Flag& flags)
...
+				ULONG pageSpaceId = DB_PAGE_SPACE;
+
+				if (!rel_perm->isSystem())
+				{
+#ifdef DEV_BUILD
+					if (!attachment->isGbak() && fb_utils::hasTablespaceName(rel_t(REL.RDB$RELATION_TYPE)))
+						fb_assert(!REL.RDB$TABLESPACE_NAME.NULL);
+#endif
+					if (!REL.RDB$TABLESPACE_NAME.NULL)
+					{
```
Cool-Hunter, line 3184: remove {}?

src/jrd/met.epp
```diff
@@ -3344,6 +3372,20 @@ ScanResult jrd_rel::scan(thread_db* tdbb, ObjectBase::Flag& flags)
 			}
 		}
 
+		ULONG pageSpaceId = DB_PAGE_SPACE;
```
Cool-Hunter, line 3375: The same code as above. New function or lambda?

src/jrd/Monitoring.cpp
```diff
@@ -879,6 +882,32 @@ void SnapshotData::putField(thread_db* tdbb, Record* record, const DumpField& fi
...
+	{
+		// special case: translate tablespace ID into name
+		fb_assert(field.length == sizeof(ULONG));
+		ULONG pageSpaceId;
+		memcpy(&pageSpaceId, field.data, field.length);
+
+		MetaName name;
+
+		if (pageSpaceId == 0)
+			name = TEMP_TABLESPACE_NAME;
```
Cool-Hunter, line 895: Maybe to move the logic under Tablespace::lookup?

src/jrd/Tablespace.h
```diff
@@ -0,0 +1,238 @@
...
+
+			~Cache() = default;
+
+			Cache(const Cache&) = delete;
+			Cache& operator=(const Cache&) = delete;
+
+			Tablespace* get(ULONG id);
+			Tablespace* get(const MetaName& name);
+
+			void store(Tablespace* tabpeSpace);
```
Cool-Hunter, line 119: typo

src/jrd/Tablespace.h
```diff
@@ -0,0 +1,238 @@
...
+			{
+				return m_mutex;
+			}
+
+		private:
+			SharedReadVector<Tablespace*, 8> m_tablespaces;
+			Firebird::Mutex m_mutex;
+		};
+
+		explicit Tablespace(MemoryPool& pool, ULONG id = DB_PAGE_SPACE)
```
Cool-Hunter, line 134: There is no backward compatibility, so maybe better not to use default value for id?

src/dsql/DdlNodes.epp
```diff
@@ -7020,6 +7033,19 @@ void RelationNode::FieldDefinition::store(thread_db* tdbb, jrd_tra* transaction)
 			DYN_UTIL_find_field_source(tdbb, transaction, relationName, viewContext.value(),
 				baseField.c_str(), RFR.RDB$FIELD_SOURCE_SCHEMA_NAME, RFR.RDB$FIELD_SOURCE);
 		}
+
+		if (tableSpace.hasData() && applyTablespacesDdl(tdbb))
+		{
+			if ((tableSpace != PRIMARY_TABLESPACE_NAME) &&
```
Cool-Hunter, line 7039: Here and later Is this check is necessary? AFAIU checkObjectExist takes into account special case that tablespace is PRIMARY.

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+
+    The index data will be moved to the specified tablespace or to the main database.
+
+-----------------
+SECURITY
+-----------------
+
+    Only administrators and users with the “CREATE TABLESPACE” privilege can create tablespaces (CREATE TABLESPACE).
+    Only administrators and users with the “ALTER ANY TABLESPACE” privilege can change tablespaces file paths (ALTER TABLESPACE <name> SET FILE [TO] <file_name>).
+    Only administrators, domain owners, or users with the ALTER ANY TABLESPACE privilege can comment (COMMENT ON) tablespaces.
```
hvlad, line 58: Who is "domain owners" ?

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    ALTER INDEX ...
+    [SET TABLESPACE [TO] {<TS NAME> | PRIMARY}]
+
+    The index data will be moved to the specified tablespace or to the main database.
+
+-----------------
+SECURITY
+-----------------
+
+    Only administrators and users with the “CREATE TABLESPACE” privilege can create tablespaces (CREATE TABLESPACE).
```
hvlad, line 56: Does tablespace have owner ? If yes, why it is not listed below ?

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+SYNTAX
+-----------------
+
+    1. TABLESPACE
+    CREATE TABLESPACE [IF NOT EXISTS] <TS NAME> FILE '/path/to/file'
+    ALTER TABLESPACE <TS NAME> SET FILE [TO] '/path/to/file'
+    You can specify either an absolute path or a relative path (relative to the database file)
+
+    DROP TABLESPACE [IF EXISTS] <TS NAME>
+    The development of the INCLUDING CONTENTS option has been postponed.
```
hvlad, line 21: It is not clear what happens with objects stored in tablespace,
It is not clear what is "INCLUDING CONTENTS option" and why it is mentioned here.

doc/README.tablespaces
hvlad, line 18: What happens with objects stored in tablespace ? What happens with old file ?

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+ODS CHANGES
+-----------------
+
+    A new table RDB$TABLESPACES:
+    RDB$TABLESPACE_ID - INTEGER # internally it will be pagespaceid.
+    RDB$TABLESPACE_NAME - CHAR (63) # name of a tablespace
+    RDB$SECURITY_CLASS - CHAR (63) # security class for tablespace
+    RDB$SYSTEM_FLAG - SMALLINT # reserved for future
+    RDB$DESCRIPTION - BLOB TEXT # description of a tablespace
+    RDB$OWNER_NAME - CHAR (63) # owner of a tablespace
```
hvlad, line 71: So, tablespace have owner. "Security" section above should be updated.

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    RDB$TABLESPACE_NAME - CHAR (63)
+
+    By default, the RDB$TABLESPACES table stores a record about the PRIMARY tablespace.
+    PRIMARY tablespace has the system flag and FILE_NAME matches the path to the database file.
+
+-----------------
+UTILITIES
+-----------------
+
+    1. Logical backup
```
hvlad, line 92: Logical ??? Please, don't invent terms where it is not required and looks weird or wrong

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    gbak -b works as usual for now. It gets data from a database transparently working with tablespaces.
+
+    2. Logical restore
+    gbak -c
```
hvlad, line 96: As you speak about UTILITIES here, it will be more logical to change paragraphs like 
1. GBAK backup
2. GBAK restore

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+    The tablespace will not be created.
+
+    -ts_orig[inal_paths]
+    To restore tablespaces to the original paths they were on when the backup was created.
+    It is still possible to override paths for some tablespaces using the -ts and -ts_map options.
+    This is an explicit option, not a default action.
+    The option does not overwrite existing files.
+
+    If you do not specify the above options, when restoring a database that has tablespaces,
+    an error about the inability to determine the path to restore tablespaces will occur.
```
hvlad, line 119: It will be good to have option to map all [not mapped] tablespaces to the some common tablespace (it should be already defined, of course).
For example, map all tablespaces to the PRIMARY.
It is just a note for memory, implementation could be done after this PR.

doc/README.tablespaces
hvlad, line 121: ISQL, SHOW

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+
+    3. Show
+    SHOW {TABLESPACES | TABLESPACE <TS NAME>}
+    Displays a list of all tablespaces names in alphabetical order or information about the specified tablespace.
+
+    4. Replication
+    There is an apply_tablespaces_ddl parameter for replication.
+    If this parameter is disabled, tablespaces-related DDL statements and CREATE/ALTER TABLE/INDEX clauses will not be applied to the replica.
+    This is used if the replica has its own set of tablespaces or none at all.
```
hvlad, line 128: Is it not clear if whole statemnet will not be applied or just tablespace-related clauses ?
I.e. `CREATE TABLE T (...) IN TABLESPACE TS` - what will be run on replica: `CREATE TABLE T (...)` or none ?

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+-----------------
+DETAILS
+-----------------
+
+  pag_header in every tablespace is reserved and may be replaced by a
+  new page type.
+  pag_scns and pag_pip are located in every tablespace.
+  pag_root is located in the tablespace where a table is located.
+
+  An algorithm for moving data to another tablespace:
```
hvlad, line 139: It describes moving of table data, not index.

doc/README.tablespaces
```diff
@@ -0,0 +1,168 @@
...
+   - walking through the map and copy every DP to the new one by fixing
+     b_page and f_page numbers.
+  At the end of work insert new records into RDB$PAGES and mark existing pages for deletion.
+
+-----------------
+CONSTRAINTS
+-----------------
+
+    It's possible to create up to 253 tablespaces.
+    Operators to move an index or table to a tablespace require an exclusive database lock.
```
hvlad, line 155: AFAIK, this requirements is no more exists, please check this statement

src/dsql/parse.y
```diff
@@ -1680,8 +1685,11 @@ create_clause
 				node->relation = $8;
 				$$ = node;
 			}
-		index_definition(static_cast<CreateIndexNode*>($9))
+		index_definition(static_cast<CreateIndexNode*>($9)) tablespace_name_clause_opt
 			{
+				if ($11)
+					static_cast<CreateIndexNode*>($9)->tableSpace = *$11;
```
hvlad, line 1691: $11 or $10 ?

src/jrd/relations.h
```diff
@@ -103,6 +104,7 @@ RELATION(nam_indices, rel_indices, ODS_8_0, rel_persistent)
 	FIELD(f_idx_exp_blr, nam_exp_blr, fld_value, 1, ODS_8_0)
 	FIELD(f_idx_exp_source, nam_exp_source, fld_source, 1, ODS_8_0)
 	FIELD(f_idx_statistics, nam_statistics, fld_statistics, 1, ODS_8_0)
+	FIELD(f_idx_ts_name, nam_ts_name, fld_ts_name, 1, ODS_14_0)
```
hvlad, line 107: Is it correct to put new field not at the end of list ?

src/jrd/Relation.h
```diff
@@ -306,118 +308,336 @@ typedef Firebird::SortedArray<ViewContext*, Firebird::EmptyStorage<ViewContext*>
...
+		if (m_pageSpaceId != pageSpaceId)
+		{
+			fb_assert(pageSpaceId != INVALID_PAGE_SPACE);
+			fb_assert(m_instanceId == 0);
+
+			reset(pageSpaceId);
+		}
+	}
+
+	void init(ULONG pageNo)
```
hvlad, line 401: Comment, please

src/jrd/Relation.h
```diff
@@ -306,118 +308,336 @@ typedef Firebird::SortedArray<ViewContext*, Firebird::EmptyStorage<ViewContext*>
...
+	{
+		return m_instanceId;
+	}
+
+	void setInstanceId(InstanceId instanceId)
+	{
+		m_instanceId = instanceId;
+	}
+
+	PageNumber toNumber(ULONG pageNo) const noexcept
```
hvlad, line 432: `toPageNumber` ?

src/jrd/SystemTriggers.epp
```diff
@@ -1609,18 +1609,22 @@ void afterInsertRelation(thread_db* tdbb, Record* record)
 
 		// Here minor optimization is possible - do it in single cache call
 		MetadataCache::newVersion<Cached::Relation>(tdbb, id);
-		auto* relation = MetadataCache::getPerm<Cached::Relation>(tdbb, id, CacheFlag::MINISCAN);
+		const auto relation = MetadataCache::getPerm<Cached::Relation>(tdbb, id, CacheFlag::MINISCAN);
+		relation->rel_flags |= REL_get_dependencies;
```
hvlad, line 1613: This `REL_get_dependencies` flag is set few lines below.
Also here `relation` is accessed before `fb_assert` and before `if (relation)`.
This all looks very weird.

Missed var in `MetadataCache::newVersion` above ?

src/jrd/tra.cpp
```diff
@@ -2502,21 +2510,28 @@ void MetadataCache::retain_temp_tables(thread_db* tdbb, jrd_tra* transaction, Tr
...
+				const auto relPerm = ltt->relation->getPermanent();
+				fb_assert(relPerm);
 
-		if (ltt && ltt->relation && (ltt->relation->getPermanent()->rel_flags & REL_temp_tran))
-			ltt->relation->getPermanent()->retainPages(tdbb, transaction->tra_number, new_number);
+				if (relPerm->rel_flags & REL_temp_tran)
+					relPerm->retainPages(tdbb, transaction->tra_number, new_number);
+			}
+		}
 	}
```
hvlad, line 2535: Same as above

src/jrd/pag.h
```diff
@@ -60,10 +61,11 @@ class PageControl : public pool_alloc<type_pgc>
...
 // INVALID_PAGE_SPACE is to ???
-inline constexpr USHORT INVALID_PAGE_SPACE	= 0;
-inline constexpr USHORT DB_PAGE_SPACE		= 1;
-inline constexpr USHORT TRANS_PAGE_SPACE	= 255;
-inline constexpr USHORT TEMP_PAGE_SPACE	= 256;
+inline constexpr ULONG INVALID_PAGE_SPACE	= 0;
+inline constexpr ULONG DB_PAGE_SPACE		= 1;
+inline constexpr ULONG TRANS_PAGE_SPACE	= 255;
+inline constexpr ULONG TEMP_PAGE_SPACE	= 256;
+inline constexpr ULONG MAX_PAGE_SPACE_ID	= MAX_ULONG;
```
hvlad, line 69: It is documented that only 253 tablespaces could be created.
Is there strong reason to change USHORT by ULONG for tablespace id ?

src/jrd/ods.h
```diff
@@ -410,26 +410,31 @@ struct index_root_page
...
+		void setKill();
+		void setKill(ULONG pageSpaceId, ULONG rootPage);
 		void setKill(TraNumber traNumber);
 		void setNormal();
-		void setNormal(ULONG rootPage/*, ULONG pageSpaceId*/);
+		void setNormal(ULONG pageSpaceId, ULONG rootPage);
 		void setCommit(TraNumber traNumber);
 		void setDrop(TraNumber traNumber);
+		void setMigrate(ULONG pageSpaceId, ULONG rootPage, TraNumber traNumber, USHORT slot);
 		void setEmpty();
```
hvlad, line 438: Consider to use single `PageNumber` instead of separate `pageSpaceId, rootPage`

src/jrd/ods.h
```diff
@@ -410,26 +410,31 @@ struct index_root_page
...
-		USHORT irt_dummy;				// alignment to 8-byte boundary
+		USHORT irt_backslot;			// slot with original page number
 
 	private:
 		void setState(UCHAR newState);
 
 	public:
-		ULONG getRoot() const;
+		ULONG getRootPage() const;
+		ULONG getRootPageSpaceId() const;
```
hvlad, line 420: Why not single `PageNumber getRoot()` ?

src/jrd/ods.h
hvlad, line 492: It will be good to add `irt_migrate` into description below

src/jrd/Tablespace.h
hvlad, line 119: typo: tabpeSpace -> tableSpace

src/jrd/Tablespace.h
```diff
@@ -0,0 +1,238 @@
...
+		void init(const MetaName& name, const Firebird::PathName& fileName, Lock* lock, std::optional<bool> alloc)
+		{
+			fb_assert(m_name.isEmpty() && m_fileName.isEmpty() && !m_lock);
+
+			m_name = name;
+			m_fileName = fileName;
+			m_lock = lock;
+
+			m_flags |= NOALLOC;
+		}
```
hvlad, line 199: Where the arg `alloc` is used ?

src/jrd/pag.h
```diff
@@ -148,32 +159,23 @@ class PageSpace : public pool_alloc<type_PageSpace>
...
 	}
 
-	PageSpace* findPageSpace(const USHORT pageSpaceID) const;
+	PageSpace* findPageSpace(const ULONG pageSpaceID) const;
 
 	void initTempPageSpace(thread_db* tdbb);
-	USHORT getTempPageSpaceID(thread_db* tdbb);
+	ULONG getTempPageSpaceID(thread_db* tdbb);
+
+	void allocTableSpace(thread_db* tdbb, ULONG pageSpaceID, bool create, const Firebird::PathName& fileName);
```
hvlad, line 177: This routine works with PageSpace, it is the member of PageManager, etc -  its name is not consistent.

src/jrd/Tablespace.h
```diff
@@ -0,0 +1,238 @@
...
+	class Lock;
+
+	class Tablespace
+	{
+		static constexpr unsigned CREATED = 1;
+		static constexpr unsigned MODIFIED = 2;
+		static constexpr unsigned DELETED = 4;
+		static constexpr unsigned BLOCKING = 8;
+		static constexpr unsigned OBSOLETE = 16;
+		static constexpr unsigned NOALLOC = 32;
```
hvlad, line 44: AFAIU, `NOALLOC` means "not allocated", correct ? 
I suggest to get better name and add comments for constant values. 
Also, for binary flags it is always better to use hex values.

src/jrd/Tablespace.h
```diff
@@ -0,0 +1,238 @@
...
+			Cache(const Cache&) = delete;
+			Cache& operator=(const Cache&) = delete;
+
+			Tablespace* get(ULONG id);
+			Tablespace* get(const MetaName& name);
+
+			void store(Tablespace* tabpeSpace);
+			void remove(ULONG id);
+			void release(thread_db* tdbb);
+			void rollback(thread_db* tdbb, jrd_tra* transaction);
```
hvlad, line 122: Comments will not harm.

src/jrd/vio.cpp
```diff
@@ -7218,7 +7260,7 @@ static void refresh_changed_fields(thread_db* tdbb, Record* old_rec, record_para
 
 
 static SSHORT set_metadata_id(thread_db* tdbb, Record* record, USHORT field_id, drq_type_t dyn_id,
-	const char* name)
+	const char* name, SLONG shift)
```
hvlad, line 7263: IIRC, shift is required for tablespaces to skip DB_PAGESPACE (1).
Perhaps it will be better to create corresponding sequence with appropriate start value ?

src/jrd/SharedReadVector.h
```diff
@@ -102,6 +102,14 @@ class SharedReadVector
 			return count + needs <= capacity;
 		}
 
+		bool add(const T& value)
+		{
+			if (!hasSpace())
+				return false;
+			data[count++] = value;
```
AlexPeshkoff, line 109: Instead
`    data[count++] = value;`
use
`    data[count] = value;`
`    ++count;`

This makes code a bit safer - we first place data to the location, next increase counter to make that location visible.

src/jrd/btr.cpp
```diff
@@ -2543,6 +2624,7 @@ static bool checkIrtRepeat(thread_db* tdbb, const index_root_page::irt_repeat* i
 		CCH_RELEASE(tdbb, window);
 
 		// Wait for completion
+		if (irtTrans )
```
hvlad, line 2627: Why `irtTrans` could be not set for `irt_in_progress` state ?
If it is ever possible, then code works with not complete index ???

src/jrd/btr.cpp
```diff
@@ -2666,6 +2750,38 @@ static ModifyIrtRepeatValue modifyIrtRepeat(thread_db* tdbb, index_root_page::ir
...
+				backIrtDesc->setKill();
+			}
+			else // state == tra_dead
+			{
+				const auto orgRootPageSpaceId = backIrtDesc->getRootPageSpaceId();
+				const auto orgRootPage = backIrtDesc->getRootPage();
+				const auto newRootPageSpaceId = irt_desc->getRootPageSpaceId();
+				const auto newRootPage = irt_desc->getRootPage();
+				irt_desc->setNormal(orgRootPageSpaceId, orgRootPage);
+				backIrtDesc->setKill(newRootPageSpaceId, newRootPage);
```
hvlad, line 2776: Transaction that migrated index is dead. 
The old slot (backIrtDesc) and new slot (irt_desc) have swapped its root pages. 
B-tree in old slot (former migrating index) is deleted. 
Ok.
What about RDB$INDEX_ID ?
As transaction that set new value for it (new slot id) is dead, I suppose this change was undone, so does RDB$INDEX_ID still points to the old slot ?

src/jrd/btr.cpp
```diff
@@ -3073,26 +3196,41 @@ void BTR_selectivity(thread_db* tdbb, Cached::Relation* relation, MetaId id, Sel
...
 
-	ULONG page = root->irt_rpt[id].getRoot();
-	const bool descending = (root->irt_rpt[id].irt_flags & irt_descending);
-	const ULONG segments = root->irt_rpt[id].irt_keys;
+	const auto indexRpt = getIndexSlot(tdbb, root, id);
+
+	ULONG page = indexRpt->getRootPage();
+	ULONG pageSpaceId = indexRpt->getRootPageSpaceId();
+	if (!pageSpaceId)
+		pageSpaceId = DB_PAGE_SPACE;
```
hvlad, line 3218: Why `pageSpaceId` could be zero ? Problem at another place ?

src/jrd/idx.cpp
```diff
@@ -860,8 +786,6 @@ void IDX_create_index(thread_db* tdbb,
 				 Arg::Gds(isc_wish_list));
 	}
 
-	get_root_page(tdbb, getPermanent(relation));
```
hvlad, line 863: Why this was removed ? It was required to ensure IRT is known for relation.

src/jrd/idx.cpp
```diff
@@ -975,13 +899,13 @@ void IDX_create_index(thread_db* tdbb,
 }
 
 
-bool IDX_activate_index(thread_db* tdbb, Cached::Relation* relation, MetaId id)
+bool IDX_activate_index(thread_db* tdbb, jrd_rel* relation, MetaId id)
 {
 	return BTR_activate_index(tdbb, relation, id);
 }
```
hvlad, line 905: It was removed from idx_proto.h, why leave here ?

src/jrd/btr.cpp
```diff
@@ -230,9 +252,10 @@ static USHORT compress_root(thread_db*, index_root_page*);
 static void copy_key(const temporary_mini_key*, temporary_mini_key*);
 static contents delete_node(thread_db*, WIN*, UCHAR*);
 static void delete_tree(thread_db*, MetaId, MetaId, PageNumber, PageNumber);
-static ULONG fast_load(thread_db*, IndexCreation&, SelectivityList&);
 
-static const index_root_page* fetch_root(thread_db*, WIN*, const RelationPermanent*, const RelationPages*);
+template<typename RS>
+static ULONG fast_load(thread_db*, IndexCreation&, SelectivityList&, RS* scb);
```
hvlad, line 257: Make template from such huge routine for such small goal as call of one function...
Why not virtual method ?

src/jrd/dpm.epp
```diff
@@ -2160,58 +2228,80 @@ void DPM_scan_pages( thread_db* tdbb)
...
 
-	auto* relation = MetadataCache::getPerm<Cached::Relation>(tdbb, 0u, CacheFlag::AUTOCREATE | CacheFlag::NOSCAN);
-	RelationPages* relPages = relation->getBasePages();
-
-	vcl** address = &relPages->rel_pages;
-	vcl* vector = *address;
-	FB_SIZE_T sequence = vector->count() - 1;
-	WIN window(relPages->rel_pg_space_id, (*vector)[sequence]);
-	pointer_page* ppage = (pointer_page*) CCH_FETCH(tdbb, &window, LCK_read, pag_pointer);
+	const auto relation = MetadataCache::getVersioned<Cached::Relation>(tdbb, rel_pages,
```
hvlad, line 2246: Is lines 2246-2288 duplicates contents of `fixupRdbPages` ?

src/jrd/Relation.h
```diff
@@ -306,118 +308,336 @@ typedef Firebird::SortedArray<ViewContext*, Firebird::EmptyStorage<ViewContext*>
...
+
+		return PageNumber(m_pageSpaceId, pageNo);
+	}
+
+	void resetPointerPages(ULONG count, const ULONG* data)
+	{
+		setPointerPages(0, count, data);
+	}
+
+	void setPointerPages(ULONG sequence, ULONG count, const ULONG* data)
```
hvlad, line 500: Name `setPointerPages` is misleading, "append" is more suitable than "set" here.

src/jrd/dpm.epp
```diff
@@ -604,25 +653,34 @@ void DPM_create_relation( thread_db* tdbb, Cached::Relation* relation)
...
 	// Store page numbers in RDB$PAGES
-	DPM_pages(tdbb, relation->getId(), pag_pointer, (ULONG) 0,
-			  (*relPages->rel_pages)[0] /*window.win_page*/);
-	DPM_pages(tdbb, relation->getId(), pag_root, (ULONG) 0,
-			  relPages->rel_index_root /*root_window.win_page*/);
+
+	if (const auto firstPointerPage = relPages->getPointerPage(0))
+		DPM_pages(tdbb, relation->getId(), pag_pointer, (ULONG) 0, firstPointerPage.value());
+	else
+		fb_assert(false);
```
hvlad, line 665: How it is ever possible to have no pointer pages right after `DPM_create_relation_pages` ?

src/jrd/dpm.epp
```diff
@@ -604,25 +653,34 @@ void DPM_create_relation( thread_db* tdbb, Cached::Relation* relation)
...
-	if (auto tra = tdbb->getTransaction())
+	if (const auto transaction = tdbb->getTransaction())
 	{
-		if (auto num = tra->tra_number)
-			set_marker(tdbb, relation->getId(), MRK_rollback, num);
+		if (const auto traNumber = transaction->tra_number)
+			set_marker(tdbb, relation->getId(), pageSpaceId, MRK_rollback, traNumber);
 	}
+	else
+		fb_assert(false);
```
hvlad, line 679: Are you going to add `else fb_assert(false)` after each `if` ?
Sorry, this is visual garbage. 
If you strongly require assert, add it before `if` (and save one line):
```
fb_assert(tdbb->getTransaction());
if (const auto transaction = tdbb->getTransaction())
```

src/jrd/dpm.epp
```diff
@@ -1092,21 +1151,19 @@ void DPM_delete_relation( thread_db* tdbb, RelationPermanent* relation)
  *	Get rid of an unloved, unwanted relation.
  *
  **************************************/
-	fb_assert(!(relation->rel_flags & REL_ltt_created));
```
hvlad, line 1095: Is this assert really not required or it was deleted by accident?

src/jrd/dpm.epp
```diff
@@ -2141,13 +2208,14 @@ ULONG DPM_pointer_pages(thread_db* tdbb, jrd_rel* relation)
...
-	fb_assert(relPages->rel_pages && relPages->rel_pages->count() == sequence + 1);
+
+	fb_assert(relPages->getPointerPageCount() == sequence + 1);
 
 	return sequence + 1;
 }
 
 
-void DPM_scan_pages( thread_db* tdbb)
+void DPM_scan_pages(thread_db* tdbb, SCHAR pagType)
```
hvlad, line 2218: Will be good to add comments what this routine actually does.

src/jrd/dpm.epp
```diff
@@ -2247,12 +2337,81 @@ void DPM_scan_pages( thread_db* tdbb)
...
 
-	for (auto id : markList)
-		DPM_scan_marker(tdbb, id);
+	for (const auto relationId : markList)
+		DPM_scan_marker(tdbb, relationId);
 }
 
 
-void DPM_scan_marker(thread_db* tdbb, MetaId relId)
+void DPM_scan_pages(thread_db* tdbb, RelationPermanent* relation, RelationPages* relPages)
```
hvlad, line 2345: Same as above: will be good to add comments what this routine actually does.

src/jrd/dpm.epp
```diff
@@ -2247,12 +2337,81 @@ void DPM_scan_pages( thread_db* tdbb)
...
+	const auto attachment = tdbb->getAttachment();
+
+#ifdef VIO_DEBUG
+	VIO_trace(DEBUG_TRACE_ALL,
+		"DPM_scan_pages ()\n");
+#endif
+
+	// Special case update of RDB$PAGES pointer page vector to avoid
+	// infinite recursion from this internal request when RDB$PAGES
+	// has been extended with another pointer page.
```
hvlad, line 2357: AFAIU, this comment block should be moved into `fixupRdbPages`

src/jrd/dpm.epp
```diff
@@ -3554,33 +3720,45 @@ static pointer_page* get_pointer_page(thread_db* tdbb,
...
+			if (!relPages->hasData())
+				return nullptr;
 
-			if (sequence < vector->count())
+			const auto ppCount = relPages->getPointerPageCount();
+			if (sequence < ppCount)
 				break;			// we are in business again
 
-			window->win_page = (*vector)[vector->count() - 1];
+			const auto ppNumber = relPages->getPointerPage(ppCount - 1);
```
hvlad, line 3737: Add `fb_assert(ppNumber)` ?

src/dsql/DdlNodes.epp
```diff
@@ -10798,10 +10915,31 @@ void AlterRelationNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratc
...
+
+			const auto newPages = newVersion->getBasePages();
+
+			fb_assert(newPages->getPageSpaceId() == pageSpaceId.value());
+			fb_assert(newPages->getPageSpaceId() != orgPages->getPageSpaceId());
+			newPages->reset(pageSpaceId.value());
+
+			// Protect relation from modification to create a consistent table copy.
+			// The lock is held until the transaction end.
+			ProtectRelations protectRelations(tdbb, transaction, rel, false);
```
hvlad, line 10938: Relation lock is not enough to completely "freeze" relation. GC should be disabled also .

src/jrd/dpm.epp
```diff
@@ -4170,3 +4347,315 @@ static void store_big_record(thread_db* tdbb,
...
+	WIN newWindow(newPages->getPageSpaceId());
+	WIN oldWindow(orgPages->getPageSpaceId());
+	WIN dpWindow(newPages->getPageSpaceId());
+
+	// Copy index root page to the new tablespace
+	oldWindow.win_page = orgPages->getIndexRootPage().value();
+	Ods::pag* oldPage = CCH_FETCH(tdbb, &oldWindow, LCK_read, pag_root);
+
+	Ods::pag* newPage = DPM_allocate(tdbb, &newWindow);
+	CCH_MARK_MUST_WRITE(tdbb, &newWindow);
```
hvlad, line 4387: Is it really necessary ?

src/jrd/dpm.epp
```diff
@@ -4170,3 +4347,315 @@ static void store_big_record(thread_db* tdbb,
...
+			newPages->rel_sec_data_space = newPage;
+
+		// Now copy the page data
+		CCH_MARK(tdbb, &newWindow);
+		memcpy(newPP, oldPP, dbb->dbb_page_size);
+
+		if (sequence < ppCount - 1)
+		{
+			// Fix the pointer to the next PP
+			newPP->ppg_next = newPointerPages[sequence + 1];
```
hvlad, line 4455: Consider to add
```
fb_assert(!(newPP->ppg_header.pag_flags & ppg_eof));
newPP->ppg_header.pag_flags &= ~ppg_eof;

```

src/jrd/dpm.epp
```diff
@@ -4170,3 +4347,315 @@ static void store_big_record(thread_db* tdbb,
...
+		}
+		else
+		{
+			// We rely that rel_pages vector is correct and nobody extend relation when we copy its data!!!
+			fb_assert(oldPP->ppg_header.pag_flags & ppg_eof);
```
hvlad, line 4460:
```
fb_assert(!newPP->ppg_next);
newPP->ppg_next = 0;
```

src/jrd/dpm.epp
```diff
@@ -4170,3 +4347,315 @@ static void store_big_record(thread_db* tdbb,
...
+
+				// All non empty slots we replace by new data page numbers in the allocated extent.
+				// Check if there are empty slots. In general case all slots in extent must be used.
+				// If such slot is we remove it and replace by empty data page.
+				for (unsigned i = 0; i < PAGES_IN_EXTENT; i++)
+				{
+					if (!newPP->ppg_page[slot + i])
+					{
+						// Original PP has empty slot here. We replace it by empty data page
+						PPG_DP_BIT_SET(bits, slot + i, ppg_dp_empty);
```
hvlad, line 4516: Clear other bits ?

src/jrd/dpm.epp
```diff
@@ -4170,3 +4347,315 @@ static void store_big_record(thread_db* tdbb,
...
+							const bool r = pagesMap.get(header->rhd_b_page, header->rhd_b_page);
+							fb_assert(r);	// Maybe to drop bugcheck?
+						}
+					}
+					if (header->rhd_flags & rhd_incomplete)
+					{
+						Ods::rhdf* fheader = (rhdf*) header;
+						if (fheader->rhdf_f_page)
+						{
+							const bool r = pagesMap.get(fheader->rhdf_f_page, header->rhd_b_page);
```
hvlad, line 4630: typo: header->rhd_b_page -> header->rhd_f_page

But... fragments could contain chain of a few pages, and all of them are not present on pointer pages.
See `store_big_record()`.
I.e. fragment pages should be allocated, not mapped.

src/jrd/validation.cpp
```diff
@@ -2448,117 +2485,124 @@ void Validation::walk_pip()
...
-		if (fixme)
-		{
-			CCH_MARK(vdr_tdbb, &window);
+			bool fixme = false;
 			if (pipMin < page->pip_min)
 			{
-				page->pip_min = pipMin;
-				vdr_fixed++;
+				corrupt(VAL_PIP_WRONG_MIN, NULL, pageSpaceId, page_number, sequence, page->pip_min, pipMin);
+				fixme = (vdr_flags & VDR_update);
```
hvlad, line 2566: Why this (and below) change is required ?
What wrong with unconditionally fix of minor inconsistencies (warnings) ?

hvlad (changes requested):
I didn't check all the code, missed DSQL and metadata manage parts.
Else it will took at least one week more.
Please, remove not needed files (.vscode, etc) and resolve merge conflicts before next round.

src/jrd/pag.cpp
```diff
@@ -1289,15 +1287,16 @@ void PAG_release_page(thread_db* tdbb, const PageNumber& number, const PageNumbe
...
+//	fb_assert(number.getPageSpaceID() == prior_page.getPageSpaceID() ||
+//			  prior_page == ZERO_PAGE_NUMBER);
 
 	const ULONG pgNum = number.getPageNum();
 	PAG_release_pages(tdbb, number.getPageSpaceID(), 1, &pgNum, prior_page.getPageNum());
 }
 
 
-void PAG_release_pages(thread_db* tdbb, USHORT pageSpaceID, int cntRelease,
+void PAG_release_pages(thread_db* tdbb, ULONG pageSpaceID, int cntRelease,
```
AlexPeshkoff, line 1299: May be better add global typedef to store ID of page space? Something like MetaId?

src/jrd/Relation.h
```diff
@@ -1053,150 +1210,227 @@ inline FB_UINT64 CacheElement<jrd_rel, RelationPermanent>::makeId<NoData>(MetaId
...
-{
-	return rel_perm->isLTT();
-}
+	std::atomic<SSHORT> rel_scan_count;		// concurrent sequential scan count
 
-inline bool jrd_rel::isVirtual() const noexcept
-{
-	return rel_perm->isVirtual();
-}
+	GCLock rel_gc_lock;						// garbage collection lock
```
AlexPeshkoff, line 1246: You moved GC lock into versioned part of relation. At the first look that means that one attach can perform GC with one version of relation and another attachment - validation with another version. Is it OK?

src/jrd/SystemTriggers.epp
AlexPeshkoff: > Missed var in `MetadataCache::newVersion` above ?

MetadataCache::newVersion() reserves entry in list of versions but does not guarantee that versioned part of object is created and Versioned::scan() is called. We can add flags parameter to MetadataCache::newVersion() but currently this is single known to me place where such behavior is needed - therefore I've preferred for the time being to have 2 sequential calls here.

Looking later in the code I see that relation->getVersioned() is called soon, i.e. yes - with this change 2 calls really not needed.

AlexPeshkoff (commented):
I've reviewed what seems to be related to MetadataCache and SharedReadVector. Did not pay much attention to other places. One general thing - is is worth to add to such PR massive changes like this:
```
	auto* rel = ....
	const auto rel = .....

```
IMO makes problems during review.

src/jrd/SharedReadVector.h
aafemt: It doesn't make sense. According to C++ standard right expression of an assignment operator is evaluated before left and if the evaluation of `value` throws, `count` is not incremented. If the assignment itself throws, value of `count` is the smallest problem. Code like `*p++ = c;` is very wide spread along of the codebase and I see no point to change it into `*p = c; ++p;`.

AlexPeshkoff: Remember - it's shared read data structure. One can read from it at any moment.

Thread A: count++
Thread B: garbage = data[count - 1]
Thread A: data[count - 1 /* i.e. count before ++ */ ] = correctValue

'B' got garbage instead correct value cause 'A' first incremented count and only next assigned.

I agree in general case barriers are needed here but for Intel CPU it's OK as is.

aafemt: I see. Quite dangerous. Even on Intel thorn read is impossible only for types up to the pointer size. Worth a detailed comment with this explanation.

dyemanov: But vector modification is always done under a mutex, isn't it? Or do you mean a reader/writer conflict?

AlexPeshkoff: A reader/writer conflict. Read can take place in parallel with write. The only guarantee from readAccessor() is that data, returned by it, will not be freed. Avoiding dangerous modification of data, stored in SharedReadVector, is callers responsibility. 

That's really low-level class, but avoiding need in mutex to access data in it is sometimes worth efforts.

doc/README.tablespaces
dyemanov: Database owners were meant, fixed.

doc/README.tablespaces
dyemanov: `INCLUDING CONTENTS` was originally suggested but rejected (in the suggested form) by the project during initial discussion. I've removed all references to it. Added more details about what's happening under the hood.

doc/README.tablespaces
dyemanov: Done.

doc/README.tablespaces
dyemanov: Fixed

doc/README.tablespaces
dyemanov: Noted. IIRC, we have something like that in a separate branch, to be reviewed and added later.

doc/README.tablespaces
dyemanov: Added these details.

doc/README.tablespaces
dyemanov: Fixed the docs to match the reality.

src/dsql/parse.y
dyemanov: I agree this looks weird, but `$11` is correct

src/jrd/relations.h
dyemanov: It does not matter for a new major ODS, but I moved it to the end for consistency.

src/jrd/Relation.h
dyemanov: Renamed, thanks.

src/jrd/SystemTriggers.epp
dyemanov: I have simplified this piece of code.

src/dsql/parse.y
aafemt: Nothing weird: `{}` block above is $9, `index_definition` is $10 and `tablespace_name_clause_opt` is $11.

src/jrd/pag.h
dyemanov: 253 was a temporary limit -- mostly to avoid changing the existing `TRANS_PAGE_SPACE / TEMP_PAGE_SPACE` constants and due to known sub-optimal tracking of used tablespaces in the code. I've improved the latter now. It was mentioned in fb-devel that 253 could be an unpleasant limitation, so we may switch to any bigger limit any time. Honestly, I don't think more than 64K tablespaces is something manageable, so `USHORT` should be pretty enough. But given we need some room for temporary pagespace IDs `USHORT` does not look enough anymore -- unless we limit tablespaces by 32K and leave the remaining 32K for temporary pagespaces. This could be OK in practice, but I'm not sure we should restrict ourselves this way without any good reason.

src/jrd/ods.h
dyemanov: This was my first attempt, but due to cyclic dependencies between `pag.h` and `ods.h` it would require moving some inline `idx_repeat` methods into `ods.cpp` which I didn't like much at the first glance. Do you think it's acceptable?

src/jrd/ods.h
dyemanov: See above.

src/jrd/ods.h
dyemanov: Documented

src/jrd/Tablespace.h
dyemanov: Fixed

src/jrd/Tablespace.h
dyemanov: Fixed (unrelated copy-paste from another method)

src/jrd/dpm.epp
dyemanov: Fixed both (chained fragments handling was implemented in another branch and skipped this PR, now ported here)

src/jrd/dpm.epp
dyemanov: Done

src/jrd/dpm.epp
dyemanov: Done

src/jrd/dpm.epp
dyemanov: Done

doc/README.tablespaces
dyemanov: Exists, documented now

doc/README.tablespaces
dyemanov: Agree this decision was very questionable, now NULL is stored as a pathname for the primary tablespace

src/jrd/SharedReadVector.h
dyemanov: Fixed with a comment

src/jrd/pag.cpp
dyemanov: I will consider this is the next review iteration

src/jrd/Relation.h
dyemanov: The idea was that two different pagespaces of a single table should not lock each other, i.e. one may be read or validated while another is locked. But I will take one more look here. Maybe some new tests should be implemented.

doc/README.tablespaces
dyemanov: Locks are documented now

doc/README.tablespaces
dyemanov: Locks are documented now

src/jrd/jrd.cpp
dyemanov: Fixed and slightly refactored to simplify the code even more.

src/jrd/met.epp
dyemanov: Done

src/jrd/met.epp
dyemanov: Removed the redundancy

src/jrd/Tablespace.h
dyemanov: Fixed

src/jrd/Tablespace.h
dyemanov: Replaced with an explicit ID passed (only a single place is affected)

src/dsql/DdlNodes.epp
dyemanov: Simplified and made consistent with the handling of other object types.

src/jrd/validation.cpp
dyemanov: `VDR_update` is documented as "fix simple things" and we usually check it in other places before fixing something. And looking at the patch closer, I see that the original logic was the same, just some code was moved. `fixme` and `corrupt()` calls exist in master too.

dyemanov:
I've addressed most of the review comments, a few are still in the pipeline and will be finished this week.

doc/README.tablespaces
dyemanov: Done

doc/README.tablespaces
dyemanov: Fixed

doc/README.tablespaces
dyemanov: Both are documented now

src/jrd/Relation.h
dyemanov: Done

src/jrd/ods.h
hvlad: Move `PageNumber` into separate header or into `ods.h` ?

src/jrd/pag.h
dyemanov: It allocates a pagespace, but expects a valid tablespace ID (which is also a valid pagespace ID) -- and this is documented inside this method. If we treat a tablespace being kinda specialization of a pagespace, then the name becomes consistent -- we cannot allocate an arbitrary pagespace by this method, only the one backing a specific tablespace. Maybe the `pageSpaceID` argument should be renamed to `tableSpaceId` for better consistency, but I'm not sure.

src/jrd/Relation.h
AlexPeshkoff: Well, but how is related with moving lock to versioned part?

dyemanov: Every relation version now has its own pagespace ID and all relation locks have this pagespace ID inside the lock key, so `getRelLockKey()` was moved into the versioned part to generate a proper key and it looked logical to move `rel_gc_lock` there as well.

dyemanov: Now I see that I've confused instance ID with pagespace ID. Will rethink this change.

src/jrd/btr.cpp
dyemanov: RDB$INDEX_ID is not changed during migration, it's kinda DBKEY and it always points to the latest slot (whether it's visible to the current transaction or the back slot is returned instead -- this is checked separately by transaction ID).

src/jrd/btr.cpp
dyemanov: This was outdated code (some old commit), no longer relevant now.

src/jrd/Relation.h
dyemanov: `append` usually means to keep the existing data "as is" while `setPointerPages()` may be called to overwrite the page number list. Maybe `updatePointerPages()`?

src/jrd/idx.cpp
dyemanov: `Relation::getIndexRootPage()` ensures that the IRP is known before returning, i.e. lazy initialization is used. The IRP number is now private and accessible only via `Relation::getIndexRootPage()`, so there's no chance someone will get uninitialized IRP number.

src/jrd/idx.cpp
dyemanov: Just an oversight, sorry - to be wiped out.

src/jrd/dpm.epp
dyemanov: I prefer to be over-paranoid ;-)

src/jrd/ods.h
dyemanov: Done.

src/jrd/pag.h
dyemanov: Renamed for better consistency.

src/jrd/Tablespace.h
dyemanov: Renamed `NOALLOC`, improved readability, added comments.

src/jrd/Tablespace.h
dyemanov: Done

src/jrd/vio.cpp
dyemanov: Reworked

src/jrd/btr.cpp
dyemanov: This was an old code from the pre-PR times (page space was missing on IRP), removed now

src/jrd/dpm.epp
dyemanov: Fixed

src/jrd/Relation.h
dyemanov: Renamed

src/jrd/dpm.epp
dyemanov: Fixed

src/jrd/dpm.epp
dyemanov: Maybe just an accident. I see nothing wrong in CI both with/without this assertion, but restored it back.

src/jrd/dpm.epp
dyemanov: Done

src/jrd/dpm.epp
dyemanov: Done

src/jrd/dpm.epp
dyemanov: Done

src/dsql/DdlNodes.epp
dyemanov: Done

src/jrd/dpm.epp
dyemanov: I don't think so, thus removed

dyemanov:
For `gstat`, we have also added a `-ts` switch that allows to analyze only particular tablespace. It will be added as a separate PR (unless you think it's not needed for Firebird).

doc/README.tablespaces
```diff
@@ -0,0 +1,178 @@
...
+    If you do not specify the above options, when restoring a database that has tablespaces,
+    an error about the inability to determine the path to restore tablespaces will occur.
+
+    3. ISQL
+    SHOW {TABLESPACES | TABLESPACE <TS NAME>}
+    Displays a list of all tablespaces names in alphabetical order or information about the specified tablespace.
+
+    4. Replication
+    There is an apply_tablespaces_ddl parameter for replication, it may be configured for the replica side.
+    If this parameter is disabled, CREATE/ALTER/DROP TSABLESPACE statements will not be ignored on the replica and
```
hvlad, line 132: will **not** be ignored ?

PR #8316, opened 2024-11-12
Cost-based decision between ORDER and SORT plans

dyemanov:
Implement cost calculation for the ORDER vs SORT choice when ordering is required for the query. The visible effects are:

1) `ORDER ... INDEX` plan is almost never used, being inferior to `SORT` in most cases -- maybe except some queries with `OPTIMIZE FOR FIRST ROWS` hint, explicit or implicit (e.g. `SELECT FIRST`). This is already proven in practice to work better.
2) `SORT` plan may be preferred to `ORDER` for the default `OPTIMIZE FOR ALL ROWS` scenario if its estimated as being cheaper.
3) Queries should not imply that `ORDER` plan will be used always if a suitable index exists. We have a number of QA tests that are affected by this change.

The cost calculation will be adjusted once more statistics will be available to the optimizer (e.g. clustering factor).

### DIFF
--- a/src/jrd/optimizer/Optimizer.h
+++ b/src/jrd/optimizer/Optimizer.h
@@ -46,30 +46,30 @@ namespace Jrd {
 
 // AB: 2005-11-05
 // Constants below needs some discussions and ideas
-constexpr double REDUCE_SELECTIVITY_FACTOR_EQUALITY = 0.001;
-constexpr double REDUCE_SELECTIVITY_FACTOR_BETWEEN = 0.0025;
-constexpr double REDUCE_SELECTIVITY_FACTOR_LESS = 0.05;
-constexpr double REDUCE_SELECTIVITY_FACTOR_GREATER = 0.05;
-constexpr double REDUCE_SELECTIVITY_FACTOR_STARTING = 0.01;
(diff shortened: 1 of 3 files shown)


src/jrd/optimizer/Optimizer.h
```diff
@@ -46,23 +46,24 @@ namespace Jrd {
 
 // AB: 2005-11-05
 // Constants below needs some discussions and ideas
-const double REDUCE_SELECTIVITY_FACTOR_EQUALITY = 0.001;
-const double REDUCE_SELECTIVITY_FACTOR_BETWEEN = 0.0025;
-const double REDUCE_SELECTIVITY_FACTOR_LESS = 0.05;
-const double REDUCE_SELECTIVITY_FACTOR_GREATER = 0.05;
-const double REDUCE_SELECTIVITY_FACTOR_STARTING = 0.01;
-const double REDUCE_SELECTIVITY_FACTOR_OTHER = 0.01;
+constexpr double REDUCE_SELECTIVITY_FACTOR_EQUALITY = 0.001;
```
asfernandes, line 49: Should not better be `inline constexpr`?

PR #8318, opened 2024-11-14
Send small blobs inline.

hvlad:
The feature allows to send small blob contents in the same data stream as main resultset.
This lowers number of roundtrips required to get blob data and significantly improves performance on high latency networks.

The blob metadata and data is send using new type of packet `op_inline_blob` and new structure `P_INLINE_BLOB`.
The `op_inline_blob` packet is send before corresponding `op_sql_response` (in case of answer on `op_execute2` or `op_exec_immediate2`), or `op_fetch_response` (answer on `op_fetch`).
There could be as much `op_inline_blob` packets as number of blob fields in output format.
NULL blobs and too big blobs are not sent. 
The blob send as a whole, i.e. current implementation doesn't support sending of part of blob. The reasons - attempt to not over-complicate the code and the fact that `seek` is not implemented for segmented blobs.

Current, initial, implementation send all blobs that total size is not greater than 16KB.

The open questions is what API changes is required to allow user to customize this process:
- allow to enable and disable inline blob sending
- allow to set inline blob size limit
- decide on what level should be applicable settings above: per-attachment, per-statement, etc
- decide default and maximum values for inline blob size limit.

Also, will good to have but not required:
- allow to set BPB in advance
- allow to enable blob in-lining on per-field basis, if output format contains many blob fields.

This PR currently in draft state and published for the early testers and commenters.

### DIFF
--- a/src/remote/client/interface.cpp
+++ b/src/remote/client/interface.cpp
@@ -1866,7 +1866,7 @@ IBlob* Attachment::createBlob(CheckStatusWrapper* status, ITransaction* apiTra, 
 		p_blob->p_blob_bpb.cstr_length = 0;
 		p_blob->p_blob_bpb.cstr_address = NULL;
 
-		Rbl* blob = FB_NEW Rbl();
+		Rbl* blob = FB_NEW Rbl(BLOB_LENGTH);
 		blob->rbl_blob_id = *blob_id = packet->p_resp.p_resp_blob_id;
 		blob->rbl_rdb = rdb;
 		blob->rbl_rtr = transaction;
@@ -6009,7 +6009,7 @@ IBlob* Attachment::openBlob(CheckStatusWrapper* status, ITransaction* apiTra, IS
 		//p_blob->p_blob_bpb.cstr_length = 0;
 		//p_blob->p_blob_bpb.cstr_address = NULL;
 
-		Rbl* blob = FB_NEW Rbl;
+		Rbl* blob = FB_NEW Rbl(BLOB_LENGTH);
 		blob->rbl_rdb = rdb;
 		blob->rbl_rtr = transaction;
 		blob->rbl_blob_id = *id;
@@ -9486,7 +9486,7 @@ static void release_blob( Rbl* blob)
 	if (blob->isCached())
 	{
 		// Assume buffer was not resized while blob was cached
-		rdb->decBlobCache(blob->rbl_data.getCapacity());
+		rdb->decBlobCache(blob->getCachedSize());
 	}
 	else
 		rdb->rdb_port->releaseObject(blob->rbl_id);
--- a/src/remote/client/interface.cpp
+++ b/src/remote/client/interface.cpp
@@ -699,6 +699,26 @@ public:
 	Batch* createBatch(CheckStatusWrapper* status, IMessageMetadata* inMetadata,
 		unsigned parLength, const unsigned char* par) override;
 
+	unsigned getMaxInlineBlobSize(CheckStatusWrapper* status) override
+	{
+		if (statement->rsr_rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+		{
+			status->setErrors(Arg::Gds(isc_wish_list).value());
+			return 0;
+		}
+		return statement->rsr_inline_blob_size;
+	}
+
+	void setMaxInlineBlobSize(CheckStatusWrapper* status, unsigned size) override
+	{
+		if (statement->rsr_rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+		{
+			status->setErrors(Arg::Gds(isc_wish_list).value());
+			return;
+		}
+		statement->rsr_inline_blob_size = size;
+	}
+
 public:
 	Statement(Rsr* handle, Attachment* a, unsigned aDialect)
 		: metadata(getPool(), this, NULL),
@@ -908,6 +928,12 @@ public:
 		IMessageMetadata* inMetadata, unsigned parLength, const unsigned char* par) override;
 
 	Replicator* createReplicator(CheckStatusWrapper* status) override;
+
+	unsigned getMaxBlobCacheSize(CheckStatusWrapper* status) override;
+	void setMaxBlobCacheSize(CheckStatusWrapper* status, unsigned size) override;
+
+	unsigned getMaxInlineBlobSize(CheckStatusWrapper* status) override;
+	void setMaxInlineBlobSize(CheckStatusWrapper* status, unsigned size) override;
 
 public:
 	Attachment(Rdb* handle, const PathName& path)
@@ -2445,6 +2471,50 @@ Batch* Attachment::createBatch(CheckStatusWrapper* status, ITransaction* transac
 }
 
 
+unsigned Attachment::getMaxBlobCacheSize(CheckStatusWrapper* status)
+{
+	if (rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+	{
+		status->setErrors(Arg::Gds(isc_wish_list).value());
+		return 0;
+	}
+	return rdb->rdb_blob_cache_size;
+}
+
+
+void Attachment::setMaxBlobCacheSize(CheckStatusWrapper* status, unsigned size)
+{
+	if (rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+	{
+		status->setErrors(Arg::Gds(isc_wish_list).value());
+		return;
+	}
+	rdb->rdb_blob_cache_size = size;
+}
+
+
+unsigned Attachment::getMaxInlineBlobSize(CheckStatusWrapper* status)
+{
+	if (rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+	{
+		status->setErrors(Arg::Gds(isc_wish_list).value());
+		return 0;
+	}
+	return rdb->rdb_inline_blob_size;
+}
+
+
+void Attachment::setMaxInlineBlobSize(CheckStatusWrapper* status, unsigned size)
+{
+	if (rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+	{
+		status->setErrors(Arg::Gds(isc_wish_list).value());
+		return;
+	}
+	rdb->rdb_inline_blob_size = size;
+}
+
+
 Batch* Statement::createBatch(CheckStatusWrapper* status, IMessageMetadata* inMetadata,
 	unsigned parLength, const unsigned char* par)
 {
@@ -3573,6 +3643,7 @@ ITransaction* Statement::execute(CheckStatusWrapper* status, ITransaction* apiTr
 		sqldata->p_sqldata_out_message_number = 0;	// out_msg_type
 		sqldata->p_sqldata_timeout = statement->rsr_timeout;
 		sqldata->p_sqldata_cursor_flags = 0;
+		sqldata->p_sqldata_inline_blob_size = statement->rsr_inline_blob_size;
 
 		send_packet(port, packet);
 
@@ -3752,6 +3823,7 @@ ResultSet* Statement::openCursor(CheckStatusWrapper* status, ITransaction* apiTr
 		sqldata->p_sqldata_out_message_number = 0;	// out_msg_type
 		sqldata->p_sqldata_timeout = statement->rsr_timeout;
 		sqldata->p_sqldata_cursor_flags = flags;
+		sqldata->p_sqldata_inline_blob_size = statement->rsr_inline_blob_size;
 
 		{
 			Cleanup msgClean([&message] {
@@ -3945,6 +4017,8 @@ ITransaction* Attachment::execute(CheckStatusWrapper* status, ITransaction* apiT
 		ex_now->p_sqlst_out_blr.cstr_length = out_blr_length;
 		ex_now->p_sqlst_out_blr.cstr_address = const_cast<unsigned char*>(out_blr);
 		ex_now->p_sqlst_out_message_number = 0;	// out_msg_type
+		ex_now->p_sqlst_inline_blob_size = (packet->p_operation == op_exec_immediate2) ?
+			rdb->rdb_inline_blob_size : 0;
 
 		send_packet(port, packet);
 
@@ -4156,6 +4230,7 @@ Statement* Attachment::createStatement(CheckStatusWrapper* status, unsigned dial
 
 	statement->rsr_next = rdb->rdb_sql_requests;
 	rdb->rdb_sql_requests = statement;
+	statement->rsr_inline_blob_size = rdb->rdb_inline_blob_size;
 
 	Statement* s = FB_NEW Statement(statement, this, dialect);
 	s->addRef();
@@ -9287,7 +9362,10 @@ static void release_blob( Rbl* blob)
 	Rtr* transaction = blob->rbl_rtr;
 	Rdb* rdb = blob->rbl_rdb;
 
-	if (!blob->isCached())
+	if (blob->isCached())
+		// Assume buffer was not resized while blob was cached
+		rdb->decBlobCache(blob->rbl_buffer_length);
+	else
 		rdb->rdb_port->releaseObject(blob->rbl_id);
 
 	if (transaction->rtr_blobs.locate(blob->rbl_blob_id))
--- a/src/remote/client/interface.cpp
+++ b/src/remote/client/interface.cpp
@@ -699,25 +699,8 @@ public:
 	Batch* createBatch(CheckStatusWrapper* status, IMessageMetadata* inMetadata,
 		unsigned parLength, const unsigned char* par) override;
 
-	unsigned getMaxInlineBlobSize(CheckStatusWrapper* status) override
-	{
-		if (statement->rsr_rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
-		{
-			status->setErrors(Arg::Gds(isc_wish_list).value());
-			return 0;
-		}
-		return statement->rsr_inline_blob_size;
-	}
-
-	void setMaxInlineBlobSize(CheckStatusWrapper* status, unsigned size) override
-	{
-		if (statement->rsr_rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
-		{
-			status->setErrors(Arg::Gds(isc_wish_list).value());
-			return;
-		}
-		statement->rsr_inline_blob_size = size;
-	}
+	unsigned getMaxInlineBlobSize(CheckStatusWrapper* status) override;
+	void setMaxInlineBlobSize(CheckStatusWrapper* status, unsigned size) override;
 
 public:
 	Statement(Rsr* handle, Attachment* a, unsigned aDialect)
@@ -2500,13 +2483,18 @@ void Attachment::setParamsFromDPB(ClumpletReader& dpb)
 		switch (item)
 		{
 		case isc_dpb_max_blob_cache_size:
-			if (rdb->rdb_port->port_protocol >= PROTOCOL_INLINE_BLOB)
-				rdb->rdb_blob_cache_size = dpb.getInt();
-			break;
-
 		case isc_dpb_max_inline_blob_size:
 			if (rdb->rdb_port->port_protocol >= PROTOCOL_INLINE_BLOB)
-				rdb->rdb_inline_blob_size = dpb.getInt();
+			{
+				SLONG val = dpb.getInt();
+				if (val < 0)
+					val = 0;
+
+				if (item == isc_dpb_max_blob_cache_size)
+					rdb->rdb_blob_cache_size = val;
+				else
+					rdb->rdb_inline_blob_size = MIN(val, MAX_INLINE_BLOB_SIZE);
+			}
 			break;
 
 		default:
@@ -2585,7 +2573,61 @@ void Attachment::setMaxInlineBlobSize(CheckStatusWrapper* status, unsigned size)
 		if (rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
 			unsupported();
 
+		if (size > MAX_INLINE_BLOB_SIZE)
+			size = MAX_INLINE_BLOB_SIZE;
+
 		rdb->rdb_inline_blob_size = size;
+	}
+	catch (const Exception& ex)
+	{
+		ex.stuffException(status);
+	}
+}
+
+
+unsigned Statement::getMaxInlineBlobSize(CheckStatusWrapper* status)
+{
+	try
+	{
+		reset(status);
+
+		Rsr* statement = getStatement();
+		CHECK_HANDLE(statement, isc_bad_req_handle);
+		Rdb* rdb = statement->rsr_rdb;
+		CHECK_HANDLE(rdb, isc_bad_db_handle);
+
+		if (rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+			unsupported();
+
+		return statement->rsr_inline_blob_size;
+	}
+	catch (const Exception& ex)
+	{
+		ex.stuffException(status);
+	}
+
+	return 0;
+}
+
+
+void Statement::setMaxInlineBlobSize(CheckStatusWrapper* status, unsigned size)
+{
+	try
+	{
+		reset(status);
+
+		Rsr* statement = getStatement();
+		CHECK_HANDLE(statement, isc_bad_req_handle);
+		Rdb* rdb = statement->rsr_rdb;
+		CHECK_HANDLE(rdb, isc_bad_db_handle);
+
+		if (rdb->rdb_port->port_protocol < PROTOCOL_INLINE_BLOB)
+			unsupported();
+
+		if (size > MAX_INLINE_BLOB_SIZE)
+			size = MAX_INLINE_BLOB_SIZE;
+
+		statement->rsr_inline_blob_size = size;
 	}
 	catch (const Exception& ex)
 	{
@@ -9444,7 +9486,7 @@ static void release_blob( Rbl* blob)
 	if (blob->isCached())
 	{
 		// Assume buffer was not resized while blob was cached
-		rdb->decBlobCache(blob->rbl_buffer_length);
+		rdb->decBlobCache(blob->rbl_data.getCapacity());
 	}
 	else
 		rdb->rdb_port->releaseObject(blob->rbl_id);
--- a/src/remote/client/interface.cpp
+++ b/src/remote/client/interface.cpp
@@ -9363,8 +9363,10 @@ static void release_blob( Rbl* blob)
 	Rdb* rdb = blob->rbl_rdb;
 
 	if (blob->isCached())
+	{
 		// Assume buffer was not resized while blob was cached
 		rdb->decBlobCache(blob->rbl_buffer_length);
+	}
 	else
 		rdb->rdb_port->releaseObject(blob->rbl_id);
 
--- a/src/remote/client/interface.cpp
+++ b/src/remote/client/interface.cpp
@@ -954,6 +954,9 @@ public:
 	Transaction* remoteTransactionInterface(ITransaction* apiTra);
 	Statement* createStatement(CheckStatusWrapper* status, unsigned dialect);
 
+	// Set params that was set in DPB, ignoring unknown and not applicable tags.
+	void setParamsFromDPB(ClumpletReader& dpb);
+
 	Replicator* replicator;
 
 private:
@@ -965,7 +968,7 @@ private:
 
 	// Returns nullptr if all items was handled or if user buffer is full, else
 	// returns pointer into unused buffer space. Handled info items are removed.
-	unsigned char* getWireStatsInfo(UCharBuffer& info, unsigned int buffer_length,
+	unsigned char* getLocalInfo(UCharBuffer& info, unsigned int buffer_length,
 								unsigned char* buffer);
 
 	Rdb* rdb;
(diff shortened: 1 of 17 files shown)


aafemt:
Why a new packet instead of sending them in the response message itself? IIRC response packets contain its own format so inline BLOBs can be described individually as strings and then transformed to cached BLOBs on client.

AlexPeshkoff:
Vlad, I suppose that content of op_inline_blob is cached by remote provider in order to serve requests for data in that blobs w/o network access. If yes - how long is data in that cache kept?

hvlad:
> Vlad, I suppose that content of op_inline_blob is cached by remote provider in order to serve requests for data in that blobs w/o network access. If yes - how long is data in that cache kept?

Yes, sure. Cached blob is bound to the transaction object and will be released (what happens first):
- at transaction end, or
- when user opens the blob with non-empty BPB, or
- user opens the blob with empty BPB and then closed it.

hvlad:
> Imagine RO-RC transaction which leasts VERY long (nothing prevents from keeping it open for client application lifetime). Would not such long life of cache be an overhead?

It is supposed that cached blobs will be read by application.
Anyway, it will be good to have a way to set limit on blobs cache size, is it your point ?

AlexPeshkoff:
Telling true my first thought was that cache is **very** tiny - just blobs from last fetched row, but this appears inefficient when we try to support various grids.

First of all let's think about binding cache not to transaction but to request/statement. It's hardly typical to close statement and read blobs from it after close. Moreover, in the worst case that will anyway work - in old way over the wire.

With limiting cache size arrives one more tunable parameter and I'm afraid there are already too much of them: blob size limit per-attachment or per-statement, may be on per-field basis (at least on/off), default BPB, may be on per-field basis too. (Hmm - are there too many cases when >1 blob per row is returned ?)

Last but not least - is blob's inlining enabled by default? On my mind yes, but very reasonable (ie not too big) defaults should be used.

sim1984:
There should be cache size limits in any case. If you loaded 1000000 records (1 blob per record) at 16K, that's already 16G. But if I understand correctly, this will be provided that the user does not read these cached blobs as the records are fetched. Maybe it's worth limiting the blob cache to some amount, for example 1000 (configurable) and when the number of blobs becomes greater than this value, the oldest of them are removed from the cache.

And of course, this should be disabled at the statement level. And perhaps some `dpb` to set the default parameter.

hvlad:
> Telling true my first thought was that cache is **very** tiny - just blobs from last fetched row, but this appears inefficient when we try to support various grids.

Yes, it was my thoughts too. Also, consider batch fetching, when whole batch of rows should be read from the wire - it will cache all corresponding blobs anyway.

> First of all let's think about binding cache not to transaction but to request/statement. It's hardly typical to close statement and read blobs from it after close. Moreover, in the worst case that will anyway work - in old way over the wire.

It was in my very first version of code. Until I started to handle `op_exec_immediate2` - it have no statement :)

It is possible to mark blobs by `stmt id` (when possible) and remove such blobs from transaction cache on statement close.
But I prefer to avoid it, so far. It gives a chance for the "not typical" apps to access cached blobs after statement close - I guess is it not so non-typical when there is no cursor, i.e. for 'EXECUTE PROCEDURE', etc.

> With limiting cache size arrives one more tunable parameter and I'm afraid there are already too much of them: blob size limit per-attachment or per-statement, may be on per-field basis (at least on/off), default BPB, may be on per-field basis too. (Hmm - are there too many cases when >1 blob per row is returned ?)

If there will be too many parameters, we can put them into separate dedicated interface, say IClientBlobCache, that will be implemented by Remote provider only.

And I'm sure there is applications that have many blobs in its resultsets. Look at monitoring tables, for example: MON$STATEMENTS have two blobs, there are other.
 
> Last but not least - is blob's inlining enabled by default? On my mind yes, but very reasonable (ie not too big) defaults should be used.

Currently it is enabled - else nobody could be able to test the feature ;) 

One of the goals of this PR is to discuss and then implement necessary numbers of parameters and corresponding API to customize the blobs cache.

So far, I see two really required parameters: 'maximum blob size for inline sending' (per-statement or per-attachment- to be decided, it should be known to the server) and 'size of blob cache' (per-attachment, client-only). Others is 'good to have' but not highly required : BPB, per-field inlining.

hvlad:
The builds for testing could be found here: 
https://github.com/FirebirdSQL/firebird/actions/runs/11836803458
Scroll page down to the 'Artifacts' section

src/remote/remote.h
```diff
@@ -192,23 +193,42 @@ struct Rdb : public Firebird::GlobalStorage, public TypedHandle<rem_type_rdb>
 };
 
 
+const SQUAD NULL_BLOB = {0, 0};
+
+inline bool operator==(const SQUAD& s1, const SQUAD& s2)
```
asfernandes, line 198: I think these operators should better be defined in the same header as the type.

src/remote/remote.h
```diff
@@ -192,23 +193,42 @@ struct Rdb : public Firebird::GlobalStorage, public TypedHandle<rem_type_rdb>
...
+{
+	return s1.gds_quad_high == s2.gds_quad_high &&
+		s2.gds_quad_low == s1.gds_quad_low;
+}
+
+inline bool operator>(const SQUAD& s1, const SQUAD& s2)
+{
+	return s1.gds_quad_high > s2.gds_quad_high ||
```
asfernandes, line 206: For readability, please use parenthesis when mixing different logical operators.

src/remote/remote.h
hvlad: `SQUAD` is defined in _src\include\fb_types.h_, I thought it is public header, but it seems not.
So, if it is OK to move these operators into _src\include\fb_types.h_ - I'll do it.

src/remote/remote.h
hvlad: Ok

src/remote/remote.h
```diff
@@ -242,15 +265,24 @@ struct RBlobInfo
 		unsigned int bufferLength, unsigned char* buffer);
 };
 
+// Used in XDR
+class RemBlobBuffer : public Firebird::HalfStaticArray<UCHAR, BLOB_LENGTH>
+{
+public:
+	RemBlobBuffer(Firebird::MemoryPool& pool) :
```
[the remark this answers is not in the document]
hvlad: It is used in _protocol.h_ and should be moved or included there. Currently, _protocol.h_ is clean (includes nothing) and I prefer to keep it clean.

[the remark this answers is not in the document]
asfernandes: What this makes different base class vs typedef/using?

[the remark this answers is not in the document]
hvlad: Forward declaration is required in _protocol.h_ but it is not possible to have forward declaration for class alias:

`error C2371: 'RemBlobBuffer': redefinition; different basic types`

src/remote/remote.h
hvlad: Done

sim1984:
I tried to conduct experiments on a local network. There are no problems with latency there, however, I will provide some results of the experiment.

Run the query in different variants

```
select
  remark
from horse
where remark is not null
```

It contains 66794 small BLOBs.

Run IBExpert with this query and do FetchAll

Results Firebird-5.0.2.1567-0-9fbd574-windows-x64 (server + client):
640ms Memory consumption 38 MB (IBExpert)

Results Firebird-6.0.0.526-0-Initial-windows-x64 (server + client):
1s 187ms Memory consumption 385 MB (IBExpert)

Probably there will be a gain in networks with high latency. I will try to see in the near future. In the meantime, the experiment shows that default blob prefetching is not always useful and at least consumes more memory.

PS

```
select
  sum(octet_length(remark)) as len
from horse
where remark is not null
```

```
LEN
=========
6 558 101
```
Overhead seems quite large to save 6 MB.

Am I right in understanding that 16K of memory is always allocated for each BLOB? I also don't know how exactly BLOBs are handled in IBExpert, perhaps it doesn't close a fully read BLOB until the end of the request/transaction. What about the limitation of storing the last N BLOBs in the cache?

hvlad:
> Am I right in understanding that 16K of memory is always allocated for each BLOB?

Yes, and it was not introduced by this PR.

BTW, 66794 blobs should consume near 1GB, while you see about 350MB - what memory counter you look at ?
I tried with 67000 of blobs of 1024 bytes and see about 1.4GB increase of 'Private Bytes' and about 1.1GB increase of 'Virtual Memory' (it was with DEBUG build).

> I also don't know how exactly BLOBs are handled in IBExpert, perhaps it doesn't close a fully read BLOB until the end of the query/transaction. 

  I doubt IBE reads any blob contents when shows data in grid - until user explicitly ask for it by moving mouse cursor over grid cell or by pressing '...' button in the cell. And debugger confirms it.

> What about the limitation of storing the last N BLOBs in the cache?

It was proposed but we still have not defined what settings and API to manage them we need.

Thanks for testing !

hvlad:
@AlexPeshkoff : I think the time overhead is related with memory allocations.

sim1984:
I just looked at the task manager. It is clear that it does not display memory quite correctly, but here the difference is visible to the naked eye. And I have no claims to performance, I understand that slightly different conditions need to be tested (primarily in networks with high latency). Nevertheless, I consider this test useful to understand that without proper settings we can get excessive memory consumption at least.

hvlad:
As there is no better ideas, I offer following API changes

```
interface Statement : ReferenceCounted
{
...
version:	// 6.0
	// Inline blob transfer
	uint getMaxInlineBlobSize(Status status);
	void setMaxInlineBlobSize(Status status, uint size);
}
```

```
interface Attachment : ReferenceCounted
{
...

version:	// 6.0
	// Blob caching by client
	uint getBlobCacheSize(Status status);
	void setBlobCacheSize(Status status, uint size);

	// Inline blob transfer
	uint getMaxInlineBlobSize(Status status);
	void setMaxInlineBlobSize(Status status, uint size);
}

```

AlexPeshkoff:
On 11/19/24 18:47, Vlad Khorsun wrote:
>
> As there is no better ideas, I offer following API changes
>
> |interface Statement : ReferenceCounted { ... version: // 6.0 // 
> Inline blob transfer uint getMaxInlineBlobSize(Status status); void 
> setMaxInlineBlobSize(Status status, uint size); } |
> |interface Attachment : ReferenceCounted { ... version: // 6.0 // Blob 
> caching by client uint getBlobCacheSize(Status status); void 
> setBlobCacheSize(Status status, uint size); // Inline blob transfer 
> uint getMaxInlineBlobSize(Status status); void 
> setMaxInlineBlobSize(Status status, uint size); } |

ok for me

aafemt:
I see no need for new methods in IAttachment, it can be handled by backward-compatible way using DPB and info items unless someone want to make such adjustments dynamically during attachment lifetime.

sim1984:
> I see no need for new methods in IAttachment, it can be handled by backward-compatible way using DPB and info items unless someone want to make such adjustments dynamically during attachment lifetime.

The presence of methods in `IAttachment` does not cancel the need for `dpb` tags to initially set these parameters when connecting. And yes, since the cache itself is for each transaction, it makes sense to change these parameters during the connection. If I understand correctly, the value from `setBlobCacheSize` is passed to the transaction at startup, and `IAttachment::setMaxInlineBlobSize` is used during `IAttachment::execute` and `IAttachment::OpenCursor`, and passes the default value to `IStatement` when calling `IAttachment::prepare`.

hvlad:
New methods was added:

```
interface Attachment
...
	// Blob caching by client
	uint getMaxBlobCacheSize(Status status);
	void setMaxBlobCacheSize(Status status, uint size);

	// Inline blob transfer
	uint getMaxInlineBlobSize(Status status);
	void setMaxInlineBlobSize(Status status, uint size);
...
interface Statement
...
	// Inline blob transfer
	uint getMaxInlineBlobSize(Status status);
	void setMaxInlineBlobSize(Status status, uint size);
```

New DPB and info items will be added later, after interface changes above stabilized finally.

**Common behaviour**

All methods above is implemented by both `Remote` and `Engine` providers. 

`Engine` provider set `isc_wish_list` error in status and returns zero, when appropriate.
`Remote` provider check protocol version and get/set internal object data (no network roundtrip) or set `isc_wish_list` error in status and returns zero, when appropriate.

**Inline blob size**

`Attachment::setMaxInlineBlobSize()` set default value for inline blob size. This value is used by `Attachment::execute()` and `Attachment::openCursor()`.

Also, this value is assigned to the new `Statement` instance created by `Attachment::prepare()`. It can be changed for given statement using `Statement::setMaxInlineBlobSize()` but it should be done before call of `Statement::execute()` or `Statement::openCursor()`.

Default value for inline blob size value is 16KB. To disable inline blob transfer, set inline blob size to zero. 

Currently, maximum value of inline blob size is not limited. It is open for discussion if some limit should be introduced or not and what value to choose. Obvious value of maximum possible segment size (64KB-2, or 65534 bytes) could be recommended for cursors (many blobs to cache). But in case of single row resultset it is not so obvious. Protocol is not limited by 2-bytes length, if I'm not mistaken.

The value of inline blob size is transferred within `op_execute`, `op_execute2` and `op_exec_immediate2` packets for supported protocol versions only.

**Client blobs caching**

The content of inline blobs is cached on the `Attachment` level. The blob is removed from cache after application uses it (opened and then closed) or if application opened same blob using custom BPB.

The size of client cache of blobs is limited. Default size is 10MB and can be changed using `Attachment::getMaxBlobCacheSize()`. There is no upper or lower limit for this value. The limit change is not applied immediately, i.e. if new limit is less than currently used size - nothing happens. If blob cache have no space for new inlined blob - such blob is discarded silently. 

Note, currently per-blob buffer is pre-allocated and its size is 16KB. It means that smaller blobs requires no additional memory re-allocations but occupy 16KB in memory (and in blobs cache) despite of its real size. I considering changes in this regards.

hvlad:
The updated builds for testing could be found here: 
https://github.com/FirebirdSQL/firebird/actions/runs/11986926791
Scroll page down to the 'Artifacts' section

src/remote/server/server.cpp
```diff
@@ -3511,6 +3512,10 @@ ISC_STATUS rem_port::execute_immediate(P_OP op, P_SQLST * exnow, PACKET* sendL)
 	{
 		this->port_statement->rsr_format = this->port_statement->rsr_select_format;
 
+		if (out_msg && exnow->p_sqlst_inline_blob_size)
```
asfernandes, line 3515: Please use braces in multi-line if.

src/remote/server/server.cpp
```diff
@@ -3946,10 +3951,16 @@ ISC_STATUS rem_port::execute_statement(P_OP op, P_SQLDATA* sqldata, PACKET* send
 			iMsgBuffer.metadata, iMsgBuffer.buffer, oMsgBuffer.metadata, oMsgBuffer.buffer);
 	}
 
+	statement->rsr_inline_blob_size = sqldata->p_sqldata_inline_blob_size;
+
 	if (op == op_execute2)
 	{
 		this->port_statement->rsr_format = this->port_statement->rsr_select_format;
 
+		if (out_msg && statement->rsr_inline_blob_size)
```
asfernandes, line 3960: Please use braces in multi-line if.

src/remote/client/interface.cpp
```diff
@@ -9217,16 +9361,15 @@ static void release_blob( Rbl* blob)
...
 
-	for (Rbl** p = &transaction->rtr_blobs; *p; p = &(*p)->rbl_next)
-	{
-		if (*p == blob)
-		{
-			*p = blob->rbl_next;
-			break;
-		}
-	}
+	if (blob->isCached())
```
asfernandes, line 9365: Please use braces in multi-line if.

src/remote/server/server.cpp
hvlad: Ok

sim1984:
I have compiled a small application and tested Firebird 5.0 and 6.0 with different fbclient (parameters were not changed). The following query was tested:

```
select
  code_horse,
  remark
from horse
where remark is not null
```

where `remark` - `BLOB SUB_TYPE TEXT`.

Each test executed the query twice, in the first case the blob itself was not read (only its identifier), in the second case the BLOB was read entirely and closed and the total length of all BLOBs was calculated. In order for the compiler not to optimize the empty loop in the first case, I calculated the code_horse sum.

Here are the results I got:

```
==== Blob inline test ====

Firebird server version
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574/tcp (Server)/P18:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-V5.0.1.1469 Firebird 5.0/tcp (station9)/P18:C"
on disk structure version 13.1

Test without read blob
Elapsed time: 491ms
sum: 72304481272

Test with read blob
Elapsed time: 66672ms
sum: 72304481272
Blob size: 6558101

==== Blob inline test ====

Firebird server version
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-T6.0.0.533 Firebird 6.0 Initial"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-T6.0.0.533 Firebird 6.0 Initial/tcp (Server)/P18:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-V5.0.1.1469 Firebird 5.0/tcp (station9)/P18:C"
on disk structure version 14.0

Test without read blob
Elapsed time: 500ms
sum: 72304481272

Test with read blob
Elapsed time: 66288ms
sum: 72304481272
Blob size: 6558101

==== Blob inline test ====

Firebird server version
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574/tcp (Server)/P18:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574/tcp (station9)/P18:C"
on disk structure version 13.1

Test without read blob
Elapsed time: 512ms
sum: 72304481272

Test with read blob
Elapsed time: 46041ms
sum: 72304481272
Blob size: 6558101

==== Blob inline test ====

Firebird server version
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-T6.0.0.533 Firebird 6.0 Initial"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-T6.0.0.533 Firebird 6.0 Initial/tcp (Server)/P18:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574/tcp (station9)/P18:C"
on disk structure version 14.0

Test without read blob
Elapsed time: 491ms
sum: 72304481272

Test with read blob
Elapsed time: 45596ms
sum: 72304481272
Blob size: 6558101

==== Blob inline test ====

Firebird server version
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-V5.0.2.1567 Firebird 5.0 9fbd574/tcp (Server)/P18:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-T6.0.0.533 Firebird 6.0 Initial/tcp (station9)/P18:C"
on disk structure version 13.1

Test without read blob
Elapsed time: 513ms
sum: 72304481272

Test with read blob
Elapsed time: 48868ms
sum: 72304481272
Blob size: 6558101

==== Blob inline test ====

Firebird server version
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-T6.0.0.533 Firebird 6.0 Initial"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-T6.0.0.533 Firebird 6.0 Initial/tcp (Server)/P19:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-T6.0.0.533 Firebird 6.0 Initial/tcp (station9)/P19:C"
on disk structure version 14.0

Test without read blob
Elapsed time: 1167ms
sum: 72304481272

Test with read blob
Elapsed time: 1704ms
sum: 72304481272
Blob size: 6558101
```

It is clear that embedded BLOBs (Firebird 6.0, fbclient 6.0) significantly reduce the time of their full reading, but it also becomes clear that if the blobs themselves are not read, but only their id is read, the reading slows down. This is not surprising given that network packets for sending embedded blobs are sent idle. That is why the ability to disable this optimization via parameters is important (BLOBs can be read deferred and it is not known when).

fbclient 5.0.2 (BLOB sending optimization) is faster than 5.0.1 by about 30%.

sim1984:
last test with optimization (`stmt->setMaxInlineBlobSize(status, 0);` for Test without read blob)

```
==== Blob inline test ====

Firebird server version
Firebird/Windows/AMD/Intel/x64 (access method), version "WI-T6.0.0.533 Firebird 6.0 Initial"
Firebird/Windows/AMD/Intel/x64 (remote server), version "WI-T6.0.0.533 Firebird 6.0 Initial/tcp (Server)/P19:C"
Firebird/Windows/AMD/Intel/x64 (remote interface), version "WI-T6.0.0.533 Firebird 6.0 Initial/tcp (station9)/P19:C"
on disk structure version 14.0

Test without read blob
MaxInlineBlobSize = 0
Elapsed time: 504ms
sum: 72304481272

Test with read blob
Elapsed time: 2178ms
sum: 72304481272
Blob size: 6558101
```

hvlad:
@sim1984, many thanks for testing !

At the last test, with setMaxInlineBlobSize(0), there is confusion for 2nd pass (with reading blobs) result: what was the MaxInlineBlobSize value ?

sim1984:
> @sim1984, many thanks for testing !
> 
> At the last test, with setMaxInlineBlobSize(0), there is confusion for 2nd pass (with reading blobs) result: what was the MaxInlineBlobSize value ?

There is a default value. I just output it.

```cpp
std::cout << std::format("MaxInlineBlobSize = {}", stmt->getMaxInlineBlobSize(status)) << std::endl;
```

livius2:
Thank you @hvlad for implementing this and @sim1984 for testing.
Is there a chance that memory consumption will be equal to current content, not full 16k? Our apps are still 32bit (using 64 bit Firebird of course). As this fix one problem, but create second.

hvlad:
> Is there a chance that memory consumption will be equal to current content, not full 16k?

Yes, I going to change this.

> As this fix one problem, but create second.

Is it measured fact or just a guess ? Note, blobs cache size is limited by 10MB by default.

livius2:
> > As this fix one problem, but create second.
> 
> Is it measured fact or just a guess ? Note, blobs cache size is limited by 10MB by default.

No, I haven't tested it yet; I'm only referring to @sim1984  information about high memory usage.

sim1984:
> > > As this fix one problem, but create second.
> > 
> > 
> > Is it measured fact or just a guess ? Note, blobs cache size is limited by 10MB by default.
> 
> No, I haven't tested it yet; I'm only referring to @sim1984 information about high memory usage.

Increased memory consumption was in the very first implementation. In the second, when @hvlad introduced buffer size restrictions for BLOBs, such effects no longer exist. The client consumes slightly more memory per buffer size (approximately).

hvlad:
Implementing new DPB items to set blobs cache size and inline blob size and have a question.
What is expected / correct behavior when DPB item(s) is not supported by underlying wire protocol but not crucial for attachment itself (i.e. could be ignored with no harm) : raise error on attach, put warning into status or silently ignore such items ?

AlexPeshkoff:
On 1/27/25 13:23, Vlad Khorsun wrote:
>
> Implementing new DPB items to set blobs cache size and inline blob 
> size and have a question.
> What is expected / correct behavior when DPB item(s) is not supported 
> by underlying wire protocol but not crucial for attachment itself 
> (i.e. could be ignored with no harm) : raise error on attach, put 
> warning into status or silently ignore such items ?
>

Initial JS approach was to ignore unknown & not supported DPB items.

hvlad:
> Initial JS approach was to ignore unknown & not supported DPB items.

This is a bit different case - items is known and supported (by the code), it just can't be used with current protocol version (that gets known after attach only, btw).
Thus I had a doubts. 
Thanks for the opinions - I agree, ignoring such items is the correct way to go.

AlexPeshkoff:
On 1/27/25 14:18, Vlad Khorsun wrote:
>
>     Initial JS approach was to ignore unknown & not supported DPB items.
>
> This is a bit different case - items is known and supported (by the 
> code), it just can't be used with current protocol version (that gets 
> known after attach only, btw).
> Thus I had a doubts.
>

Just to help you avoid doubts - 99% your case:)
REMOTE_get_timeout_params() is invoked in inet but obviously skipped in 
xnet, i.e. timeout from DPB is ignored.

hvlad:
One more question: wire protocol is already allows to pass buffers larger than `MAX_USHORT`, but blob-related code in Remote subsystem uses USHORT's for various lengths (see `struct Rbl` and its usage). 
I can limit max blob inline size by `MAX_USHORT` or change `struct Rbl` and related code to handle larger buffers. 
Is there will be benefit when sending large buffers (>= 64KB) at once ?
Opinions ?

AlexPeshkoff:
On 1/28/25 14:09, Vlad Khorsun wrote:
>
> One more question: wire protocol is already allows to pass buffers 
> larger than |MAX_USHORT|, but blob-related code in Remote subsystem 
> uses USHORT's for various lengths (see |struct Rbl| and its usage).
> I can limit max blob inline size by |MAX_USHORT| or change |struct 
> Rbl| and related code to handle larger buffers.
> Is there will be benefit when sending large buffers (>= 64KB) at once ?
> Opinions ?
>

Increasing limits in remote is in general good idea but on my mind 
should be done for all protocoal (certainly with new version) rather 
than single place.
I.e. if you do not plan to rework everything  better limit max blob 
inline size by MAX_USHORT.

sim1984:
I would like to make one more improvement. It is not directly related to the ticket, but it has been long overdue.

We have a search function in streaming BLOBs

```
int IBlob::seek(StatusType* status, int mode, int offset)
```

I think it is high time to move away from the 2G limit and introduce a new function that does not have such a limit

```
int64_t IBlob::seek2(StatusType* status, int mode, int64_t offset)
```

See also https://github.com/FirebirdSQL/firebird/issues/550

dyemanov:
Larger buffers are usually better compressible, although I doubt the difference will be significant. But if this change decreases number of *our* (application-level) round-trips it may be a good bonus.

hvlad:
> I would like to make one more improvement. It is not directly related to the ticket, but it has been long overdue.

For a good reason, I think ;)

> We have a search function in streaming BLOBs
> 
> ```
> int IBlob::seek(StatusType* status, int mode, int offset)
> ```
> 
> I think it is high time to move away from the 2G limit and introduce a new function that does not have such a limit

For what ? Who store 2GB+ blob at database ? Is it really practical ?
And I not going to address it in this PR anyway.
 
> ```
> int64_t IBlob::seek2(StatusType* status, int mode, int64_t offset)
> ```
> 
> See also #550
> 
> And also add
> 
> ```
> void IBlob::getInfo2(StatusType* status,
>   unsigned itemsLength,
>   const unsigned char* items,
>   unsigned bufferLength,
>   unsigned char* buffer)
> ```
> 
> Which would return 64-bit length to get BLOBs over 2GB.

This is not necessary as API already allows to pass 64-bit ints at info buffers.
The question is if apps ready to get it.

aafemt:
> The question is if apps ready to get it.

If it is using `isc_portable_integer()` instead of deprecated `isc_vax_integer()` - most likely yes.

AlexPeshkoff:
On 1/28/25 18:24, Vlad Khorsun wrote:
> For what ? Who store 2GB+ blob at database ? Is it really practical ?
> And I not going to address it in this PR anyway.

In addition I have big doubts that engine is ready for such blobs.

dyemanov:
> On 1/28/25 18:24, Vlad Khorsun wrote: For what ? Who store 2GB+ blob at database ? Is it really practical ? And I not going to address it in this PR anyway.
> In addition I have big doubts that engine is ready for such blobs.

It does work (except incorrect length stored/reported but this is fixable).

dyemanov:
> > > On 1/28/25 18:24, Vlad Khorsun wrote: For what ? Who store 2GB+ blob at database ? Is it really practical ? And I not going to address it in this PR anyway.
> > > In addition I have big doubts that engine is ready for such blobs.
> > 
> > 
> > It does work (except incorrect length stored/reported but this is fixable).
> 
> Sure. But this doesn't answer my question ;)

I'm not going to argue whether it's practical or not ;-) Personally, I'd rather avoid that if possible. But anyway, blobs beyond 2GB are documented as allowed thus they should be usable.

sim1984:
> > > > On 1/28/25 18:24, Vlad Khorsun wrote: For what ? Who store 2GB+ blob at database ? Is it really practical ? And I not going to address it in this PR anyway.
> > > > In addition I have big doubts that engine is ready for such blobs.
> > > 
> > > 
> > > It does work (except incorrect length stored/reported but this is fixable).
> > 
> > 
> > Sure. But this doesn't answer my question ;)
> 
> I'm not going to argue whether it's practical or not ;-) Personally, I'd rather avoid that if possible. But anyway, blobs beyond 2GB are documented as allowed thus they should be usable.
> 
> But surely this is for another day and unrelated to this PR.

That's exactly what I wanted to say. I don't really think it's practical to store even 1GB in a BLOB, but it's allowed. As far as I know, our BLOBs can store up to 32GB. However, the API and network protocol do not allow seek in such BLOBs. This is more of a desire to make the API more consistent with the capabilities of the engine. And of course, this is not the PR for this. I just decided to remind you since we're already dealing with BLOBs.

Sorry for digressing from the topic of discussion.

hvlad:
Maximum inline blob size is limited by MAX_USHORT (65535) bytes now.
The limit is applied to the buffer size, not to the "clear" blob data, i.e. it includes additional 2 bytes per each segment.

When max inline blob value is set (using API or DPB) and the passed value is greater than MAX_USHORT, it is decreased to MAX_USHORT silently.

hvlad:
> Note, currently per-blob buffer is pre-allocated and its size is 16KB. It means that smaller blobs requires no additional memory re-allocations but occupy 16KB in memory (and in blobs cache) despite of its real size. I considering changes in this regards.

Done.

hvlad:
The updated builds for testing could be found here:
https://github.com/FirebirdSQL/firebird/actions/runs/13076211167
Scroll page down to the 'Artifacts' section.

I consider this PR as feature-complete now.

hvlad:
For documentation purposes:

The improvement was sponsored by the donations collected in the 21 Firebird Developers Day, many thanks to everyone involved!

pavel-zotov:
QA note: more checks are performed in the [functional/blob/test_small_blob_caching.py](https://github.com/FirebirdSQL/firebird-qa/blob/master/tests/functional/blob/test_small_blob_caching.py)

PR #8319, opened 2024-11-15
Use a 64bit counter for the records written during the backup

dmitry-starodubov:


hvlad:
Two quick questions:
- did you test if really large values (>2^32) printed correctly ?
- what about restore ?

dmitry-starodubov:
I did not test it on a real database, but changed the value of records variable in the debugger to 5*10^9. The output was correct.
64bit counters are already used in restore code.

hvlad (approved):

dyemanov:
Frontported to v6.

PR #8320, opened 2024-11-15
Do not allow concurrent run of sweep instances

TreeHunter9:
Old code has a bug where if we already have a sweep instance running, and some one call sweep once more, `allowSweepRun()` will return false and we are gonna call `clearSweepFlags()`, that will reset sweep flags and release a sweep lock. This will open an opportunity to run a second sweep instance for both Super and Classic server.

### DIFF
--- a/src/jrd/Database.cpp
+++ b/src/jrd/Database.cpp
@@ -316,7 +316,7 @@ namespace Jrd
 			if (old & DBB_sweep_in_progress)
 			{
 				clearSweepStarting();
-				return false;
+				ERR_post(Arg::Gds(isc_sweep_concurrent_instance));
 			}
 
 			if (dbb_flags.compareExchange(old, old | DBB_sweep_in_progress))
@@ -336,7 +336,7 @@ namespace Jrd
 				fb_utils::init_status(tdbb->tdbb_status_vector);
 
 				dbb_flags &= ~DBB_sweep_in_progress;
-				return false;
+				ERR_post(Arg::Gds(isc_sweep_concurrent_instance));
 			}
 		}
 		else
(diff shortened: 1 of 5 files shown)


AlexPeshkoff:
As far as I can see isc_sweep_concurrent_instance is returned to the client which started sweep. Where is is handled in a case when sweep started automatically? IMO it should be silently ignored.

TreeHunter9:
As I can see, `allowSweepRun()` will not allow us to run auto sweep if we have concurrent instances, but I added a check for isc_sweep_concurrent_instance anyway, just in case.

TreeHunter9:
> If you want to add diags for sweep startup failure all cases should better be taken into an account.

Yeah, I agree. I will add more error messages.

hvlad:
> But suggested patch contains new feature - user is notified what happened, without it sweep silently finishes in a moment.

How it works for auto-sweep ?

AlexPeshkoff:
> > But suggested patch contains new feature - user is notified what happened, without it sweep silently finishes in a moment.
> 
> How it works for auto-sweep ?

See
https://github.com/FirebirdSQL/firebird/pull/8320#issuecomment-2482790184

src/jrd/Database.cpp
```diff
@@ -47,6 +47,14 @@
 
 using namespace Firebird;
 
+namespace
+{
+	void unableToRunSweepException(const Arg::Gds& reason)
```
AlexPeshkoff, line 52: Passing ISC_STATUS to exception raise will make code simpler, cleaner and cheaper.

TreeHunter9: Done

PR #8323, opened 2024-11-21
Add AUTO RELEASE TEMP BLOBID transaction option

ilya071294:
It makes the transaction release temporary ID of user BLOB just after its materialization. It's useful for massive insertions of records with user-defined BLOBs because it eliminates the memory overhead caused by creating and keeping temporary IDs until the transaction ends. This option is used during the database restore.


src/dsql/parse.y
```diff
@@ -9749,10 +9753,12 @@ non_reserved_word
 	// added in FB 4.0.2
 	| BLOB_APPEND
 	// added in FB 5.0
+	| BLOBID
 	| LOCKED
 	| OPTIMIZE
 	| QUARTER
 	| TARGET
+	| TEMP
```
hvlad, line 9761: Its added in FB 6.0

src/burp/restore.epp
```diff
@@ -421,7 +421,7 @@ int RESTORE_restore (const TEXT* file_name, const TEXT* database_name)
 		// Block added to verbose index creation by Toni Martir
 		// Always try to activate deferred indices - it helps for some broken backups,
 		// and in normal cases doesn't take much time to look for such indices. AP-2008.
-		EXEC SQL SET TRANSACTION ISOLATION LEVEL READ COMMITTED NO_AUTO_UNDO;
+		EXEC SQL SET TRANSACTION ISOLATION LEVEL READ COMMITTED NO_AUTO_UNDO AUTO_RELEASE_TEMP_BLOBID;
```
hvlad, line 424: Is it really required for every transaction in restore.epp ? 

For example, if gbak run against pre-FB6 server, this will fail (for every transaction!) and gbak will start every transaction second time and without NO_AUTO_UNDO also. Is it good ?

Another way would be to remember at first failure that target server not supported AUTO_RELEASE_TEMP_BLOBID or NO_AUTO_UNDO and don't ever try to start transaction with unsupported options. 
Though, such change it out of scope of this PR...

doc/sql.extensions/README.set_transaction.txt
```diff
@@ -27,6 +27,12 @@ commit fails. This option is mostly used by gfix.
 LOCK TIMEOUT nonneg_short_integer: it's the time (measured in seconds) that a
 transaction waits for a lock in a record before giving up and reporting an error.
 
+AUTO RELEASE TEMP BLOBID: makes the transaction release temporary ID of user BLOB
+just after its materialization. It's useful for massive insertions of records with
+user-defined BLOBs because it eliminates the memory overhead caused by creating and
+keeping temporary IDs until the transaction ends. This option is used during the
+database restore.
+
```
hvlad, line 35: I would add something like: use with care and only if there is no need to open just inserted (or updated) blob using temporary blob ID obtained after blob creation.

src/dsql/parse.y
```diff
@@ -163,6 +163,7 @@ using namespace Firebird;
 %token <metaNamePtr> BEGIN
 %token <metaNamePtr> BETWEEN
 %token <metaNamePtr> BLOB
+%token <metaNamePtr> BLOBID
```
asfernandes, line 166: Add it to correct section (version). This place is for v3/v4.

src/common/ParserTokens.h
```diff
@@ -492,6 +493,7 @@ PARSER_TOKEN(TOK_TAN, "TAN", true)
 PARSER_TOKEN(TOK_TANH, "TANH", true)
 PARSER_TOKEN(TOK_TARGET, "TARGET", true)
 PARSER_TOKEN(TOK_TEMPORARY, "TEMPORARY", true)
+PARSER_TOKEN(TOK_TEMP, "TEMP", true)
```
asfernandes, line 496: TEMP sorts before TEMPORARY, so should (by convention) be first.

doc/sql.extensions/README.set_transaction.txt
```diff
@@ -27,6 +27,12 @@ commit fails. This option is mostly used by gfix.
 LOCK TIMEOUT nonneg_short_integer: it's the time (measured in seconds) that a
 transaction waits for a lock in a record before giving up and reporting an error.
 
+AUTO RELEASE TEMP BLOBID: makes the transaction release temporary ID of user BLOB
```
asfernandes, line 30: May mention the tpb constant here too.

src/dsql/parse.y
ilya071294: Do we need this feature in FB5? Considering that it was initially made to optimize the database restore in the first place. Without MON$TRANSACTION changes, of course.

hvlad: I would like to have it in FB5.

ilya071294: Then it's better to leave them in FB5 section, I think. I will fix other places where I clearly chose a wrong section.

src/burp/restore.epp
ilya071294: > Is it really required for every transaction in restore.epp ?

It's not really. But it's not so obvious to determine particular places where the new transaction option should be used. Ideally, this needs to be tested carefully (also considering gbak options which affect commit points) to make sure there are no missed places. So now I think it would be better to set transaction options at one place. Even if AUTO_RELEASE_TEMP_BLOBID is not going to make a difference for some transaction, it won't harm anyway. And I decided to solve this problem using the approach you suggested.

doc/sql.extensions/README.set_transaction.txt
ilya071294: I don't know if it's an appropriate place to mention TPB constants. But if it is, then it should be added for other options as well.

src/burp/restore.epp
```diff
@@ -11625,6 +11610,52 @@ void fix_system_generators(BurpGlobals* tdgbl)
...
+ *	supported by the target server.
+ *
+ **************************************/
+
+	while (true)
+	{
+		if (tdgbl->gbl_use_auto_release_temp_blobid && tdgbl->gbl_use_no_auto_undo)
+		{
+			EXEC SQL SET TRANSACTION NO_AUTO_UNDO AUTO_RELEASE_TEMP_BLOBID;
+			if (gds_status->hasData())
```
hvlad, line 11632: I understand you follow existing (old) code, but check for `isc_bad_tpb_form` makes this code more robust:

`
if (gds_status->hasData() && fb_utils::containsErrorCode(gds_status->getErrors(), isc_bad_tpb_form))
`

src/burp/restore.epp
hvlad: Thanks!

src/burp/BurpTasks.cpp
```diff
@@ -893,11 +895,12 @@ void RestoreRelationTask::initItem(BurpGlobals* tdgbl, Item& item)
 			if (status->getState() & IStatus::STATE_ERRORS)
 				BURP_abort(&status);
 
-			// SET TRANSACTION NO_AUTO_UNDO, see at the end of get_data()
+			// SET TRANSACTION NO_AUTO_UNDO AUTO_RELEASE_TEMP_BLOBID, see at the end of get_data()
 
 			ClumpletWriter tpb(ClumpletReader::Tpb, 128, isc_tpb_version3);
 			tpb.insertTag(isc_tpb_concurrency);
 			tpb.insertTag(isc_tpb_no_auto_undo);
+			tpb.insertTag(isc_tpb_auto_release_temp_blobid);
```
hvlad, line 903: Looks like check for new `gbl_use_XXX` will not harm here too.

hvlad (approved):

doc/sql.extensions/README.set_transaction.txt
ilya071294: We may decide this beyond the PR if no one minds.

pavel-zotov:
::: QA notes :::
1. Test verifies only syntax extension of SET TRANSACTION, i.e. ability to use 'AUTO RELEASE TEMP BLOBID' clause.
    Handling with temporary BLOBID can not be tested in ISQL and will be checked when firebird-driver will support this.
2. Presense of mon$transactions.mon$auto_release_temp_blobid column not checked: FB 5.x currently missed it.

PR #8327, opened 2024-11-25
Fix use :@ characters and add sub-section to configure username and password for sync_replica

XaBbl4:
There is a problem with using :@ characters in the connection string to a sync_replica

For example, if the username contains the @ symbol: `jonh@it-corp:qwe@rty@localhost:employee` - the parser split it as:
```
username = jonh                                 // wrong, should be jonh@it-corp
password = <null>                               // wrong, should be qwe@rty
database = it-corp:qwe@rty@localhost:employee   // wrong, should be localhost:employee
```

or only the password contains the @ symbol: `jonh:qwe@rty@db_alias`, then:
```
username = jonh                     // is correct
password = qwe                      // wrong, should be qwe@rty
database = rty@localhost:employee   // wrong, should be localhost:employee
```

In addition to fix it, this PR add a subsection to the sync_replica configuration. You can specify a username and password without any character restrictions, or specify where they should be read: from the first non-empty line of the file or from an environment variable.

Now in the replication.conf can:
```python
database = <master_path>
{
  # 1) old format with restrictions:
  #    @ can't be used in database
  #    : can't be used in username
  #    all characters can be used in password
  sync_replica = jonh@it-corp:qwe@rty@localhost:employee
    #  username = jonh@it-corp        // is correct
    #  password = qwe@rty             // is correct
    #  database = localhost:employee  // is correct

  # 2) new sub-stream config with plain username and password without restrictions (username can be : character)
  sync_replica = localhost/3051:employee
  {
    username = jonh:smith@it-corp
    password = qwe@rty
  }

  # 3) new sub-stream when read values from environment and file
  sync_replica = localhost/3052:employee
  {
    username_env = replication_user
    password_file = replication_pass
  }
}
```

### DIFF
--- a/src/jrd/replication/Config.cpp
+++ b/src/jrd/replication/Config.cpp
@@ -246,12 +246,10 @@ Config* Config::get(const PathName& lookupName)
 								}
 								else if (key_source.equals(KEY_FILE))
 								{
-									const PathName sub_filename =
-										fb_utils::getPrefix(IConfigManager::DIR_CONF, sub_value.c_str());
-
+									const PathName sub_filename = sub_value.c_str();
 									AutoPtr<FILE> file(os_utils::fopen(sub_filename.c_str(), "rt"));
 									if (!file)
-										configError("missing file", value, sub_filename.c_str());
+										configError("missing or inaccessible file", value, sub_filename.c_str());
 
 									// skip first empty lines
 									sub_value = "";
--- a/src/jrd/replication/Config.cpp
+++ b/src/jrd/replication/Config.cpp
@@ -246,7 +246,11 @@ Config* Config::get(const PathName& lookupName)
 								}
 								else if (key_source.equals(KEY_FILE))
 								{
-									const PathName sub_filename = sub_value.c_str();
+									PathName sub_filename = sub_value.c_str();
+									PathUtils::fixupSeparators(sub_filename);
+									if (PathUtils::isRelative(sub_filename))
+										sub_filename = fb_utils::getPrefix(IConfigManager::DIR_CONF, sub_filename.c_str());
+
 									AutoPtr<FILE> file(os_utils::fopen(sub_filename.c_str(), "rt"));
 									if (!file)
 										configError("missing or inaccessible file", value, sub_filename.c_str());
--- a/src/jrd/replication/Config.cpp
+++ b/src/jrd/replication/Config.cpp
@@ -103,6 +103,69 @@ namespace
 
 		status->setErrors(sv.value());
 	}
+
+	void parseSyncReplica(const ConfigFile::Parameters& params, SyncReplica& output)
+	{
+		for (const auto& el : params)
+		{
+			string key(el.name.c_str());
+			string value(el.value);
+
+			if (value.isEmpty())
+				continue;
+
+			auto pos = key.rfind('_');
+			if (pos != string::npos)
+			{
+				const string key_source = key.substr(pos + 1);
+
+				if (key_source.equals(KEY_ENV))
+				{
+					fb_utils::readenv(value.c_str(), value);
+					if (value.isEmpty())
+						configError("missing environment variable", output.database, value);
+
+					key = key.substr(0, pos);
+				}
+				else if (key_source.equals(KEY_FILE))
+				{
+					PathName filename = value.c_str();
+					PathUtils::fixupSeparators(filename);
+					if (PathUtils::isRelative(filename))
+						filename = fb_utils::getPrefix(IConfigManager::DIR_CONF, filename.c_str());
+
+					AutoPtr<FILE> file(os_utils::fopen(filename.c_str(), "rt"));
+					if (!file)
+						configError("missing or inaccessible file", output.database, filename.c_str());
+
+					// skip first empty lines
+					value = "";
+					do
+					{
+						if (feof(file))
+							break;
+
+						if (!value.LoadFromFile(file))
+							break;
+
+						value.alltrim(" \t\r");
+					} while (value.isEmpty());
+
+					if (value.isEmpty())
+						configError("empty file", output.database, filename.c_str());
+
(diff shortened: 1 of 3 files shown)


aafemt:
User names are SQL identifiers and cannot include `@` without being delimited. You path just move problem from user part to database part (which also can include legal `@`). Instead you must properly parse delimiters.

dyemanov:
I'd suggest that if the new (sub-section) syntax is used for `sync_replica`, the database name is not parsed at all (and thus it may *not* contain username and/or password). In this case a database path containing @ will work perfectly. Is it already the case in your patch?

src/jrd/replication/Config.cpp
```diff
@@ -216,7 +220,69 @@ Config* Config::get(const PathName& lookupName)
...
+								if (key_source.equals(KEY_ENV))
+								{
+									fb_utils::readenv(sub_value.c_str(), sub_value);
+									if (sub_value.isEmpty())
+										configError("missing environment variable", value, sub_value);
+								}
+								else if (key_source.equals(KEY_FILE))
+								{
+									const PathName sub_filename =
+										fb_utils::getPrefix(IConfigManager::DIR_CONF, sub_value.c_str());
```
dyemanov, line 250: What if the specified value is an absolute pathname? AFAIU, it will be prefixed anyway and become wrong.

src/jrd/replication/Config.cpp
```diff
@@ -216,7 +220,69 @@ Config* Config::get(const PathName& lookupName)
...
+
+									AutoPtr<FILE> file(os_utils::fopen(sub_filename.c_str(), "rt"));
+									if (!file)
+										configError("missing file", value, sub_filename.c_str());
```
dyemanov, line 254: Maybe "missing or inaccessible file" ?

src/jrd/replication/Config.cpp
XaBbl4: Fixed, since the configs are stored in the root directory, I don't to add a file check and a separate prefix addition.

dyemanov: I'd suggest using `PathUtils::isRelative()` and call `getPrefix()` for relative paths and leave absolute paths "as is".
Also I believe `PathUtils::fixupSeparators()` would be worth adding before opening the file.

src/jrd/replication/Config.cpp
```diff
@@ -216,7 +220,71 @@ Config* Config::get(const PathName& lookupName)
...
+							if (sub_key == "username")
+								syncReplica.username = sub_value.c_str();
+							else if (sub_key == "password")
+								syncReplica.password = sub_value.c_str();
+							else
+								configError("unknown parameter", value, sub_key);
+						}
+					}
+					else
+						splitConnectionString(value, syncReplica.database, syncReplica.username, syncReplica.password);
```
TreeHunter9, line 285: Maybe move it to a separate function, because it blows up Config::get and has a lot of nested blocks?

dyemanov: I agree.

XaBbl4: Thanks, I did.

dyemanov (approved):

src/jrd/replication/Config.h
```diff
@@ -33,6 +33,21 @@
 
 namespace Replication
 {
+	struct SyncReplica
+	{
+		SyncReplica(MemoryPool& p)
```
dyemanov, line 38: `explicit` c'tor, please

dyemanov:
Any opinion whether it should be backported into v5?

hvlad:
Where it should be documented ? I see no changes in replication.conf (at least).

Also, it is not completely clear how environment or file could be used - is it allowed to use `username_file` and\or `password_env` ?

I support backporting it into v5.

src/jrd/replication/Config.cpp
```diff
@@ -99,6 +103,69 @@ namespace
...
+					key = key.substr(0, pos);
+				}
+			}
+
+			if (key == "username")
+				output.username = value.c_str();
+			else if (key == "password")
+				output.password = value.c_str();
+			else
+				configError("unknown parameter", output.database, key);
```
hvlad, line 166: Is it correct to read the file in advance, before check of key name ?

dyemanov:
I will add a description to replication.conf, thanks for the reminder.

dyemanov:
> Also, it is not completely clear how environment or file could be used - is it allowed to use `username_file` and\or `password_env` ?

Username / password (independently of each other) may be specified as either a string (key without suffix) or via envvar (key suffixed by `_env`) or via filename (key suffixed by `_file`). So both options in your example above are valid.

dyemanov:
Please review: https://github.com/FirebirdSQL/firebird/commit/3575d917386476c8d6fe7f345ea7e1dc62a32b7f

XaBbl4:
> Please review: [3575d91](https://github.com/FirebirdSQL/firebird/commit/3575d917386476c8d6fe7f345ea7e1dc62a32b7f)

That's all right. But now I've found out that there is uncertainty when several parameters with the same key are specified in a subsection, for example:
```
{
  username = jonh
  username_file = repl_user.txt
  password_env = FB_REPL_PWD
}
```

In this case, username will first be equal to `jonh`, and then read from the file and replaced in the order of the processing queue, either add this point to the description, or output an error when keys are duplicated.

src/jrd/replication/Config.cpp
hvlad: > I think so. Since the configuration is set only by the administrator, it doesn't matter when to read from the file, before or after the key check, but in this case there is no need for double checking the key.

There will be confusion with error diags when wrong key name points to the non-existing file, for example.
Consider case: `pasword_file = deleted_file_name`

> But can be refactor this into a separate function if need?

Sure.

PR #8340, opened 2024-12-04
Increase index flags capacity in ODS14

dyemanov:
This PR changes the `irt_repeat` layout to allow more bits for index flags and native `TraNumber` storage (and also fixes a few related things). The `irt_repeat` size increases from 12 bytes to 16 bytes, but I find it acceptable. We can avoid the union and remove the `irt_in_progress` flag, but the `irt_repeat` size will grow to 24 bytes which doubles the initial storage, so I'm not sure whether it's worth it (as it limits the maximum number of indices per table).

I tried to partially synchronize with the `metacache` branch to simplify the merge later, as Alex already needed to have some of these changes in his branch. I'm ready to adjust the patch if needed.

Other PRs for ODS14 will follow, so I'd suggest to review them separately but merge together in a batch, to avoid recreating all existing ODS14 databases often due to ongoing ODS changes.


src/jrd/ods.h
```diff
@@ -368,39 +368,39 @@ struct index_root_page
...
-		void setRoot(ULONG root_page);
-
-		TraNumber getTransaction() const;
-		void setTransaction(TraNumber traNumber);
-
+		union
+		{
+			FB_UINT64 irt_transaction;	// transaction in progress
+			ULONG irt_root;				// page number of index root
+		};
```
hvlad, line 375: When it was private there was guarantee of consistent usage of `irt_root`, `irt_transaction `and flag `irt_in_progress`. 
Why make it public and open way for errors ?

src/jrd/validation.cpp
```diff
@@ -1976,25 +1976,26 @@ Validation::RTN Validation::walk_index(jrd_rel* relation, index_root_page& root_
...
 
-	temporary_key nullKey, *null_key = 0;
+	temporary_key nullKey, *null_key = nullptr;
 	if (unique)
 	{
 		index_desc idx;
 		{
 			// No need to evaluate index expression and/or condition
-			AutoSetRestoreFlag<UCHAR> flags(&root_page.irt_rpt[id].irt_flags,
+			AutoSetRestoreFlag<USHORT> flags(&root_page->irt_rpt[id].irt_flags,
```
hvlad, line 1995: Is it necessary to specify template arg, could it be auto deducted from ctor arguments by compiler ?

asfernandes: It's C++17 feature (https://en.cppreference.com/w/cpp/language/class_template_argument_deduction)
AFAIK, we were not using it, but we could.

src/jrd/ods.h
dyemanov: Honestly, private data members (and getters/setters) just look unnatural (to me) in ODS structures, IMHO they complicate things. That said, I agree about usage consistency. Now it's also consistent if you check for `isEmpty() / inProgress()` before using `irt_root / irt_transaction` and I thought that's enough. I can revert to private members if your disagree.

hvlad: > Honestly, private data members (and getters/setters) just look unnatural (to me) in ODS structures, IMHO they complicate things. That said, I agree about usage consistency. Now it's also consistent if you check for `isEmpty() / inProgress()` before using `irt_root / irt_transaction` and I thought that's enough. I can revert to private members if your disagree.

For me, it is good case when consistency is over complexity.

src/jrd/validation.cpp
hvlad: I've found at least one such usage of `AutoSetRestore` in remote.cpp - see `PortsCleanup::closePorts()`.

dyemanov: Auto deduction works for `irt_expression`, but fails for `irt_expression | irt_condition`, it should be explicitly casted to `USHORT` to work and it somewhat defeats the idea.

Nevertheless, I'd suggest (outside this PR) to remove `AutoSetRestoreFlag` at all and instead add the needed lookup flags to `BTR_description`. Because the index root page may be locked as `LCK_read` but we modify data there (even if temporarily) -- this looks hackery.

src/jrd/ods.h
dyemanov: OK, to be changed back.

PR #8341, opened 2024-12-04
Cleanup batches inside the engine if they were not released explicitly before disconnection

dyemanov:
This avoids a resource leak (it's mostly about `TempSpace`). Easy to reproduce - just open a batch, push a significant number of records through the batch and disconnect. You'll get an assertion `fb_assert(m_tempCacheUsage == 0);` inside `~GlobalObjectHolder()` and may also notice temporary files still existing.

Init/cleanup logic of `JBatch` is changed to match `JResultSet` and `JReplicator`.


AlexPeshkoff (approved):

dyemanov:
It it worth backporting into v4/v5?

PR #8360, opened 2024-12-24
Fix windows warnings (C4554, C4018) and mark error message constants as constexpr (C26814)

Noremos:


aafemt:
You should add check that `__cplusplus` macro has value high enough. They appeared only in C++11.

aafemt:
Yes, but iberrors.h is a **public** header. Firebird build requirements are not applied to public headers.

Noremos:
> You should add check that `__cplusplus` macro has value high enough. They appeared only in C++11.

Yes, I was sure these errors are used only in internal code and did not pay enough attention to the include location of the file.

aafemt:
Actually I see no reason to mark integer constants `inline` (or even `constexpr` at all). Anybody does?

TreeHunter9:
> Actually I see no reason to mark integer constants `inline` (or even `constexpr` at all). Anybody does?

I think if you have C++17 you should mark every constexpr variable in header file as inline constexpr, because it will use external linkage. For visual difference of external and internal linkage of variables in header file you can watch [this video](https://youtu.be/QVHwOOrSh3w?si=0ideJF9cE118kvkE&t=299).
And about constexpr, it doesn't have any disadvantages and give you permission to use this variable in constexpr function.

aafemt:
External linkage is good for complex types such as strings because it leaves only one instance of data in executable image. But integers? They used to be embedded into assembler code directly even is declared as simple `const`.

TreeHunter9:
IIRC if someone will try to take an address or pass by reference this const variable it will be allocated in memory, so in this case, inline will do it job.
And I think the right question about inline should be "why not to use it", not "why should we use it", because inline has no drawbacks as far as I know.

src/common/CvtFormat.cpp
```diff
@@ -75,7 +75,7 @@ namespace
 			const TrieNode* currentNode = m_root;
 			FB_SIZE_T valueLength = fb_strlen(value);
 
-			for (outParsedTimezoneLength = 0; outParsedTimezoneLength < valueLength; outParsedTimezoneLength++)
+			for (outParsedTimezoneLength = 0; static_cast<FB_SIZE_T>(outParsedTimezoneLength) < valueLength; outParsedTimezoneLength++)
```
TreeHunter9: I think it's not the right way to remove the warning, we are just replacing implicit cast to explicit. I think it would be better to change type of `outParsedTimezoneLength` to FB_SIZE_T, so we can be 100% sure about this comparison. Yeah, we would have to change more code, but I think this is the right approach to handle warnings.

src/isql/FrontendLexer.cpp
```diff
@@ -142,7 +142,7 @@ std::variant<FrontendLexer::SingleStatement, FrontendLexer::IncompleteTokenError
 
 		while (pos < end)
 		{
-			if (end - pos >= term.length() && std::equal(term.begin(), term.end(), pos))
+			if (static_cast<decltype(term.length())>(end - pos) >= term.length() && std::equal(term.begin(), term.end(), pos))
```
asfernandes, line 145: Use a fixed type, please.

Noremos: Changed it to FB_SIZE_T

src/jrd/sqz.cpp
```diff
@@ -558,7 +558,7 @@ ULONG Difference::apply(ULONG diffLength, ULONG outLength, UCHAR* const output)
 			BUGCHECK(177);	// msg 177 applied differences will not fit in record
 	}
 
-	const auto length = p - output;
+	const FB_UINT64 length = p - output;
```
dyemanov, line 561: `ULONG` is enough here.

src/include/firebird/iberror.h
```diff
@@ -17,32 +17,38 @@
...
+STATUS_EXPR isc_arg_unix			= 7;	// UNIX error code
+STATUS_EXPR isc_arg_domain			= 8;	// Apollo/Domain error code
+STATUS_EXPR isc_arg_dos				= 9;	// MSDOS/OS2 error code
+STATUS_EXPR isc_arg_mpexl			= 10;	// HP MPE/XL error code
+STATUS_EXPR isc_arg_mpexl_ipc		= 11;	// HP MPE/XL IPC error code
+STATUS_EXPR isc_arg_next_mach		= 15;	// NeXT/Mach error code
+STATUS_EXPR isc_arg_netware			= 16;	// NetWare error code
+STATUS_EXPR isc_arg_win32			= 17;	// Win32 error code
+STATUS_EXPR isc_arg_warning			= 18;	// warning argument
+STATUS_EXPR isc_arg_sql_state		= 19;	// SQLSTATE
```
dyemanov, line 46: Maybe name it `STATUS_CONST` instead?

PR #8361, opened 2024-12-24
Fix linux warnings Wformat

DmEye:


AlexPeshkoff:
If you anyway touch this code - may be better remove cast to SLONG?

src/include/fb_types.h
```diff
@@ -41,11 +41,13 @@
 	typedef unsigned int ULONG;
 	const SLONG SLONG_MIN = INT_MIN;
 	const SLONG SLONG_MAX = INT_MAX;
+	#define SLONG_STR_FORMAT "%d"
 #elif SIZEOF_LONG == 4
 	typedef long SLONG;
 	typedef unsigned long ULONG;
 	const SLONG SLONG_MIN = LONG_MIN;
 	const SLONG SLONG_MAX = LONG_MAX;
+	#define SLONG_STR_FORMAT "%ld"
```
dyemanov, line 50: We already have `SLONGFORMAT` defined in `common.h`, why yet another define?

DmEye:
> If you anyway touch this code - may be better remove cast to SLONG?

Perhaps it may be better. But the cause of warning was the usage of wrong format. So, I assumed the fix is usage of proper format.

src/include/fb_types.h
DmEye: Thanks for the note. I didn't know.

dyemanov (approved):

PR #8362, opened 2024-12-25
Make return-type a compilation error

ilya071294:
This protects from undefined behavior cases that can be added to the code by mistake.

### DIFF
--- a/src/common/CvtFormat.cpp
+++ b/src/common/CvtFormat.cpp
@@ -1124,6 +1124,7 @@ namespace
 			return twelveHours == 12 ? twelveHours : 12 + twelveHours;
 
 		cb->err(Arg::Gds(isc_incorrect_hours_period) << string(period.data(), period.length()));
+		return 0; // suppress compiler warning/error
 	}
 
 	constexpr int roundYearPatternImplementation(int parsedRRValue, int currentYear)
--- a/builds/posix/make.defaults
+++ b/builds/posix/make.defaults
@@ -111,7 +111,7 @@ GLOB_OPTIONS:=
 #____________________________________________________________________________
 
 # Global c++ flags: firebird needs no RTTI, choose build standard and c++ specific warnings level
-PLUSPLUS_FLAGS:= -fno-rtti -std=c++17 -Werror=delete-incomplete
+PLUSPLUS_FLAGS:= -fno-rtti -std=c++17 -Werror=delete-incomplete -Werror=return-type
 
 # If this is defined then we use special rules useful for developers only
 IsDeveloper = @DEVEL_FLG@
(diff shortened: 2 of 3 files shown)


src/common/CvtFormat.cpp
```diff
@@ -1124,6 +1124,7 @@ namespace
 			return twelveHours == 12 ? twelveHours : 12 + twelveHours;
 
 		cb->err(Arg::Gds(isc_incorrect_hours_period) << string(period.data(), period.length()));
```
aafemt, line 1126: I wonder if this line can do anything but throw. If no it would be better to mark it as [noreturn](https://en.cppreference.com/w/cpp/language/attributes/noreturn) and undo addition of `return`.

ilya071294: Normally it should throw. But we still can't be 100% sure. If we add `noreturn` and `cb->err` won't throw, the server will crash. I prefer to avoid such cases as well.

AlexPeshkoff: I tend to agree with @aafemt - err function always throws, that's unconditional behavior.

But there is another side of a question. My old hack with passing different error raising calls was (among others) added to support descriptors in qli, at the same time avoid calls (like MOB_/CVT_) from qli directly to jrd and keep correct diags in engine. Currently, when qli is gone away, ideally that hack should better be rolled back. After that we will get here [[ noreturn ]] function in usual way.

I.e. for the time being I suggest to add return 0 as suggested in PR - it will be cleaned out together with both cb and err in it.

ilya071294: > I.e. please mark err function as [[ noreturn ]], it will be cleaned out together with both cb and err in it. IMO further _deep_ modifications of this code make no sense.

BTW, I have no idea how to mark `err` as `noreturn` while it's a function pointer.

AlexPeshkoff: Damned, looks like c++ standard is really bad regarding attributes. OK, let's live with 'return 0' till cleanup of class Callbacks.

AlexPeshkoff (approved):

src/common/CvtFormat.cpp
aafemt: Not quite, you just have to keep at C++ instead of mixing it with C. Look at this testcase:
```
#include <stdio.h>
#include <stdlib.h>

struct Fnoret
{
  [[noreturn]] static int foo(int a)
  {
    printf("Fnoret(%d)\n", a);
    throw 3;
  }
};

struct Fret
{
  static int foo(int a)
  {
    printf("Fret(%d)\n", a);
    throw 3;
  }
};

template <class T>
int bar(const T& f)
{
  if (rand() < 2) return 0;

  f.foo(2);
}

int main()
{
  try
  {
    printf("bar = %d\n", bar(Fnoret()));
    printf("bar = %d\n", bar(Fret()));
  }
  catch(...)
  {
    printf("caught\n");
  }
}
```
Only second invocation leads to the compilation warning.

AlexPeshkoff:
On 12/26/24 15:31, Dimitry Sibiryakov wrote:||

> Only second invocation leads to the compilation warning.
>

As I've said I do not think that spending too much efforts with 
Callbacks is good idea - better cleanup it at all. Do not think we ever 
return to qli.

PR #8366, opened 2024-12-27
Fix MSVC C4146 warnings

ilya071294:
The warning description: "unary minus operator applied to unsigned type, result still unsigned".

### DIFF
--- a/src/dsql/WinNodes.cpp
+++ b/src/dsql/WinNodes.cpp
@@ -432,7 +432,7 @@ void FirstValueWinNode::aggInit(thread_db* tdbb, Request* request) const
 
 dsc* FirstValueWinNode::winPass(thread_db* tdbb, Request* request, SlidingWindow* window) const
 {
-	if (!window->moveWithinFrame(-(window->getRecordPosition() - window->getFrameStart())))
+	if (!window->moveWithinFrame(-static_cast<SINT64>(window->getRecordPosition() - window->getFrameStart())))
 		return NULL;
 
 	dsc* desc = EVL_expr(tdbb, request, arg);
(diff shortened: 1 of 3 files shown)


dyemanov (approved):

src/dsql/WinNodes.cpp
```diff
@@ -432,7 +432,7 @@ void FirstValueWinNode::aggInit(thread_db* tdbb, Request* request) const
 
 dsc* FirstValueWinNode::winPass(thread_db* tdbb, Request* request, SlidingWindow* window) const
 {
-	if (!window->moveWithinFrame(-(window->getRecordPosition() - window->getFrameStart())))
+	if (!window->moveWithinFrame(-static_cast<SINT64>(window->getRecordPosition() - window->getFrameStart())))
```
dyemanov, line 435: I\d suggest a temporary variable for better readability, e.g.:
```C++
const SINT64 offset = window->getRecordPosition() - window->getFrameStart(); // with static_cast, if necessary
if (!window->moveWithinFrame(-offset))
```

src/dsql/WinNodes.cpp
```diff
@@ -583,7 +583,7 @@ dsc* NthValueWinNode::winPass(thread_db* tdbb, Request* request, SlidingWindow*
 	const SLONG fromPos = desc ? MOV_get_long(tdbb, desc, 0) : FROM_FIRST;
 
 	if (fromPos == FROM_FIRST)
-		records += -(window->getRecordPosition() - window->getFrameStart()) - 1;
+		records += -static_cast<SINT64>(window->getRecordPosition() - window->getFrameStart()) - 1;
```
dyemanov, line 586: Maybe just `-=` with a positive offset here?

src/dsql/WinNodes.cpp
hvlad: Expression  `window->getRecordPosition() - window->getFrameStart()` used many times, makes sence to move it into separate function, like:

```
	class SlidingWindow
	{
	public:
...
	FB_UINT64 getInFrameOffset() const
	{
		return savedPosition - frameStart;
	}

```

ilya071294:
As I have no ability to update the branch, I would suggest to merge it. And I will commit all minor fixes separately.

PR #8380, opened 2025-01-09
Fix race condition in shutdownThread start that can cause server to be unable to stop

TreeHunter9:
Previously we assign value to `shutdownSemaphore` after `shutdownThread` is started, where it was already needed. So we can have situation where `shutdownThread` instantly leaving due to `shutdownSemaphore == nullptr`, and we are left with a server that can only be stopped with kill -9.
I think this problem was hidden behind `fbguard` timeout and it's really hard to reproduce due to nature of race conditions, but in a synthetic environment using `sleep` to emulate high load and process switching in OS it is easily reproducible.

### DIFF
--- a/src/yvalve/why.cpp
+++ b/src/yvalve/why.cpp
@@ -831,11 +831,11 @@ private:
 		explicit CtrlCHandler(MemoryPool& p)
 			: ShutdownInit(p)
 		{
+			shutdownSemaphore = &semaphore;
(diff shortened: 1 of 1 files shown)


AlexPeshkoff (approved):

dyemanov:
Should it be backported?

TreeHunter9:
> Should it be backported?

I think so, this bug appears in at least v4 and v5 (haven't looked at v3 and more older versions).

PR #8393, opened 2025-01-16
remove redundant plugin reload on super server

### LINKED ISSUE #8389: Unnecessary reload of the encryption plugin in the Super architecture
Hello, colleagues,  

During the debugging of my encryption plugin, I identified an unnecessary reload of the encryption plugin in `CryptoManager::lockAndReadHeader()` (CryptoManager.cpp) for the Super architecture. This reload was unintentionally introduced by me in PR #7423. It can result in the plugin's key being overwritten during its operation, as occurred in my case.  

This issue is primarily related to the specific implementation of my plugin and should not affect other plugins if they securely manage their keys and prevent spontaneous changes during encryption or decryption processes. However, for the Super architecture, this reload is redundant—it simply reloads the same plugin with the same key, serving no useful purpose. For this reason, I have decided to address it.  

The reload is only required for the Classic architecture when the encryption state is modified from another process after an AST trigger.  

This issue affects all major versions: 3, 4, 5, and 6. A PR with the fix will be provided shortly.

MochalovAlexey:
Fix #8389. It removes unnecessary plugin reloads in the SuperServer and introduces a special flag to update plugins in the Classic Server after the stateLock ast event.

### DIFF
--- a/src/jrd/CryptoManager.cpp
+++ b/src/jrd/CryptoManager.cpp
@@ -367,7 +367,7 @@ namespace Jrd {
 			return;
 
 		fb_assert(tdbb);
-		lockAndReadHeader(tdbb, CRYPT_HDR_NOWAIT);
+		lockAndReadHeader(tdbb, CRYPT_HDR_NOWAIT | CRYPT_RELOAD_PLUGIN);
 	}
 
 	void CryptoManager::lockAndReadHeader(thread_db* tdbb, unsigned flags)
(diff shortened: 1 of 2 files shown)


MochalovAlexey:
I decided to create a PR in fb5 first because that's the version where I discovered the behavior described in issue #8389. If there are no questions about the code, I can create PRs for other branches as well.

dyemanov:
No need in extra PRs, I will front-port the commit myself.

AlexPeshkoff:
> No need in extra PRs, I will front-port the commit myself.

OK, please do. BTW, suppose old versions may also be affected.

PR #8399, opened 2025-01-17
Improvement #8356 : Make Trace use HEX representation for parameter values ​​of types [VAR]CHAR CHARACTER SET OCTETS and [VAR]BINARY.

### LINKED ISSUE #8356: Make Trace use HEX representation for parameter values ​​of types [VAR]CHAR CHARACTER SET OCTETS and [VAR]BINARY
From firebird support list: https://groups.google.com/u/2/g/firebird-support/c/kPgAW-BtpQo

When tracing, parameters are logged, very nice.
Currently we use CHAR(16) CHARACTER SET OCTETS to store ID-values.

Tracing log them like this.
param0 = varchar(16), " ßFÃ/¿J¹Z3¡õñ]["

I can manually convert this to hex by Win1252 decoding, (DF46C32FBF4AB95A33A1F5F15D5B) but it would be much easier to have a Hex presentation when parameter is of Octets charset. 
Please implement encoding with hex-values when parameter is of type CHAR(16) CHARACTER SET OCTETS.

hvlad:


src/jrd/trace/TraceObjects.cpp
```diff
@@ -56,6 +56,67 @@
...
+		return false;
+	}
+
+	if (param->getCharSet() == CS_BINARY)
+	{
+		// Convert OCTETS and [VAR]BINARY to HEX string
+
+		char* hex = result.getBuffer(length * 2);
+
+		for (UCHAR* p = address; p < address + length; p++)
```
asfernandes, line 92:
```suggestion
		for (const UCHAR* p = address; p < address + length; p++)
```

asfernandes (changes requested):

src/jrd/trace/TraceObjects.cpp
hvlad: Done, thanks

PR #8401, opened 2025-01-20
ODS14: header page refactoring

dyemanov:
- Make transaction/attachment counters natively 64-bit
- Refactor backup/shutdown/replica flags into dedicated modes
- Move GUID to the core part of the header
- Cleanup the remaining stuff from multi-file databases
- Some other misc changes


sim1984:
Apparently https://github.com/FirebirdSQL/firebird/pull/8340 also has an alignment issue for 32-bit builds.

dyemanov:
> Apparently #8340 also has an alignment issue for 32-bit builds.

Yes, it will be fixed too.

PR #8403, opened 2025-01-20
Fix potential deadlock when starting the encryption thread

### LINKED ISSUE #8402: Potential deadlock when starting the encryption thread
Inside the encryption thread, before creating an internal connection, the non-recursive mutex `Jrd::Database::dbb_thread_mutex` is locked. This can lead to a deadlock if an exception is thrown during the connection process and the `release_attachment` function is called, as it also attempts to lock this mutex. 

This issue affects all major versions: 3, 4, 5, and 6. A PR with the fix will be provided shortly.

MochalovAlexey:
fixes #8402
add a check to verify thread matching between the encryption thread and the thread where we release the attachment. If they match, use a dummy mutex instead of the actual `dbb_thread_mutex` to avoid a deadlock

### DIFF
--- a/src/jrd/CryptoManager.h
+++ b/src/jrd/CryptoManager.h
@@ -303,7 +303,7 @@ public:
 	const char* getPluginName() const;
 	Thread::Handle getCryptThreadHandle() const
 	{
-		return cryptThreadId;
+		return cryptThreadHandle;
 	}
 
 private:
@@ -382,7 +382,7 @@ private:
 	AttachmentsRefHolder keyProviders, keyConsumers;
 	Firebird::string hash;
 	Firebird::RefPtr<DbInfo> dbInfo;
-	Thread::Handle cryptThreadId;
+	Thread::Handle cryptThreadHandle;
 	Firebird::IDbCryptPlugin* cryptPlugin;
 	Factory* checkFactory;
 	Database& dbb;
--- a/src/jrd/CryptoManager.h
+++ b/src/jrd/CryptoManager.h
@@ -301,6 +301,10 @@ public:
 	UCHAR getCurrentState(thread_db* tdbb) const;
 	const char* getKeyName() const;
 	const char* getPluginName() const;
+	Thread::Handle getCryptThreadHandle() const
+	{
+		return cryptThreadId;
+	}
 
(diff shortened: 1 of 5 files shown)


src/jrd/CryptoManager.h
```diff
@@ -301,6 +301,10 @@ class CryptoManager final : public Firebird::PermanentStorage, public BarSync::I
 	UCHAR getCurrentState(thread_db* tdbb) const;
 	const char* getKeyName() const;
 	const char* getPluginName() const;
+	Thread::Handle getCryptThreadHandle() const
+	{
+		return cryptThreadId;
+	}
```
hvlad, line 307: If `cryptThreadId` actually is a thread handle - make its name consistent.
Else we have a problem here.

src/jrd/jrd.cpp
```diff
@@ -7755,14 +7755,26 @@ void release_attachment(thread_db* tdbb, Jrd::Attachment* attachment, XThreadEns
...
 	// avoid races with special threads
 	// take into an account lock earlier taken in DROP DATABASE
 	XThreadEnsureUnlock threadGuard(dbb->dbb_thread_mutex, FB_FUNCTION);
 	XThreadEnsureUnlock* activeThreadGuard = dropGuard;
 	if (!activeThreadGuard)
 	{
-		threadGuard.enter();
-		activeThreadGuard = &threadGuard;
+		if (dbb->dbb_crypto_manager
+			&& Thread::isCurrent(Thread::getIdFromHandle(dbb->dbb_crypto_manager->getCryptThreadHandle())))
```
hvlad, line 7769: Isn't it clearer to make `isCurrent()` to accept handle, not id ?

src/jrd/CryptoManager.h
MochalovAlexey: Ok, tried to leave it as it was, but I agree that it's necessary here. Will fix that

src/jrd/jrd.cpp
MochalovAlexey: Maybe, but the non-static isCurrent() used to work with the thread id, and I don't want to change it.

PR #8413, opened 2025-01-28
Fix incorrect maximum size when reading dbb parameter values in SHOW DATABASE

TreeHunter9:
Incorrect maximum size occurs for `isc_info_oldest_transaction`, `isc_info_oldest_active`, `isc_info_oldest_snapshot`, `isc_info_next_transaction` and `isc_info_limbo`. This leads to an error if the value is greater than `int`.
Also change maximum size for all parameter values, so we don't get the same error in future.

This bug also occurs in v3.0, v4.0 and v5.0. In v4.0 and v5.0 it can be easily cherry picked, but in v3.0 TraNumber is a signed type, so printf has to be changed.

### DIFF
--- a/src/isql/show.epp
+++ b/src/isql/show.epp
@@ -438,12 +438,12 @@ void SHOW_dbb_parameters(Firebird::IAttachment* db_handle,
 			break;
 
 		case isc_info_page_size:
-			value_out = p.getInt();
+			value_out = p.getBigInt();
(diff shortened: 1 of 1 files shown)


dyemanov (approved):

dyemanov:
Backported to v5 in the simplified form, without transaction numbers format change.

PR #8418, opened 2025-01-29
New 'UNLIST' function

### LINKED ISSUE #8005: UNLIST table-valued function
Implement a built-in function working the opposite way to the existing aggregate function `LIST` - i.e. parse the delimited string and return a record set containing the parsed values.

It may be a possible solution for #3821, as it would be possible to do `WHERE ID IN (SELECT * FROM UNLIST(...))`

The name `UNLIST` is suggested due to (1) being an antipode to the existing `LIST` function (even if we add a standard synonym `LISTAGG` one day) and (2) SQL standard having the `UNNEST` table-value function which does the similar job just for arrays/multi-sets.

Syntax:

```
<list derived table> ::= UNLIST(<value> [, <separator>] [<data type conversion>])
<separator> ::= <value>
<data type conversion> ::= RETURNING <data type>
```

If `<separator>` is omitted, comma is implied (as for `LIST`). If `<data type conversion>` is omitted, resulting field has the data type of the first input parameter.

The RETURNING syntax look weird, but this is exactly how SQL specification declares the output data type in JSON functions, so we used the same approach here.

Remaining questions:

1. I suppose it's OK to imply `UNLIST` as relation name if the `AS` clause is omitted. But I'm not sure about the column name - should we also imply `UNLIST`, or maybe use keyword `VALUE` or maybe some other ideas?
```
select unlist from unlist(...)
select value from unlist(...)
select unlist.unlist from unlist(...)
select unlist.value from unlist(...)
```

2. Should we trim the parsed strings under the hood, or make it optional, or leave it up to developers?
`select * from unlist('1, 2, 3')`

3. If `<data type conversion>` is omitted and the input argument is BLOB, should we return also BLOB or maybe VARCHAR(MAX)?

ChudaykinAlex:
1) Purpose

The function parses the input string using the specified delimiter (comma "," is implied by default) and returns the identified substrings as discrete records containing a single field. Additionally, the desired type of the returned field can be specified. If the specified data type conversion is impossible, an error is raised at runtime.

See also: https://github.com/FirebirdSQL/firebird/issues/8005

2) Syntax and rules

```
<table value function> ::= UNLIST(<input> [, <separator>] [, <data type conversion>]) [AS] <correlation name> [(<derived column name>)]
<input> ::= <value>
<separator> ::= <value>
<data type conversion> ::= RETURNING <data type>

```
)Arguments:

- ```<input>```: any value expression that returns a string/blob of characters (or may be converted to a string), including string literals, table columns, constants, variables, expressions, etc. This parameter is mandatory.
- ```<separator>```: optional value expression that returns a string which is used as a delimiter (i.e. it separates one value from another inside the input string). It may also be a BLOB TEXT value, limited to 32KB. If an empty string is specified, the output will be just one record containing the input string. If omitted, the comma character is used as a delimiter.
- ```<data type>```: target data type to convert the output values into. Alternatively, a domain can be specified as the returned type. If omitted, VARCHAR(32) is implied. Feel free to suggest any better alternative default.
- ```<correlation name>```: alias of the record set returned by the UNLIST function. It is a mandatory parameter (per SQL standard).
- ```<derived column name>```: optional alias of the column returned by the UNLIST function. If omitted, UNLIST is used as an alias.

3) Changes in the code

1. New rules are added to the parser (parse.y).
2. New BLR code is added to support table-value functions with a sub-code for the UNLIST function.
3. New class TableValueFunctionSourceNode is added, inherited from RecordSourceNode. This class is a generic implementation of table-value functions inside the engine (to be reused later by other table-value functions defined in the SQL specification). Its descendant classes are responsible for implementing particular functions (UNLIST is just the first one in this list, more to follow).
4. Extend csb_repeat with jrd_tab_func* pointing to a new structure defining the compile-time info about the function (name, fields, format).
5. New class TableValueFunctionScan is added, inherited from RecordStream. It generates records for the function stream. Class UnlistFunctionScan is inherited from TableValueFunctionScan to implement the record generator specific to the UNLIST function.

4) Examples
```
SELECT * FROM UNLIST('1,2,3,4,5') AS U;

UNLIST                           
================================ 
1                                
2                                
3                                
4                                
5    


SELECT * FROM UNLIST('100:200:300:400:500', ':' RETURNING INT) AS U;

      UNLIST 
============ 
         100 
         200 
         300 
         400 
         500 


SELECT U.* FROM UNLIST('text1,text2,text3') AS U;

UNLIST                           
================================ 
text1                            
text2                            
text3     


SELECT C0 FROM UNLIST('text1,text2,text3') AS U(C0);

C0                               
================================ 
text1                            
text2                            
text3 


SELECT U.C0 FROM UNLIST('text1,text2,text3') AS U(C0);

C0                               
================================ 
text1                            
text2                            
text3 


SET TERM ^ ;
RECREATE PROCEDURE TEST_PROC RETURNS (PROC_RETURN_INT INT)
AS
DECLARE VARIABLE text VARCHAR(11);
BEGIN
	text = '123:123:123';
	FOR SELECT * FROM UNLIST( :text, ':' RETURNING INT) AS A INTO :PROC_RETURN_INT DO
	SUSPEND;
END^
SET TERM ; ^

SELECT * FROM TEST_PROC;
PROC_RETURN_INT 
=============== 
            123 
            123 
            123 


CREATE DOMAIN D1 AS INT;
SELECT TEST_DOMAIN FROM UNLIST('1,2,3,4' RETURNING D1) AS A(TEST_DOMAIN);

 TEST_DOMAIN 
============ 
           1 
           2 
           3 
           4 
           

CREATE VIEW TEST_VIEW AS SELECT * FROM UNLIST('1,2,3,4') AS A(B);
SELECT B FROM TEST_VIEW;

B                                
================================ 
1                                
2                                
3                                
4      


Statement failed, SQLSTATE = 42S22
Dynamic SQL Error
-SQL error code = -206
-Column unknown
-UNLIST
-At line 2, column 9
At line 24 in file ...

```


sim1984:
What about interaction with the IN predicate? In the initial discussion, the idea was put forward that it would be nice to be able to perform queries like this

SELECT * FROM table WHERE id IN UNLIST('1,2,3')

That is, without additional wrapping of the UNLIST results in a derived table.

ChudaykinAlex:
I didn't do anything specifically for IN.
I may not understand the specific task at hand. Here is an example of how I understood your question.
```
recreate table test1 (s01 INT);
insert into test1(s01)values('1');
insert into test1(s01)values('2');
insert into test1(s01)values('3');
insert into test1(s01)values('4');
insert into test1(s01)values('5');
insert into test1(s01)values('6');
insert into test1(s01)values('7');
insert into test1(s01)values('8');

SELECT * FROM test1 WHERE s01 IN (SELECT * FROM UNLIST('1,8' RETURNING INT) AS U);

         S01 
============ 
           1 
           8 


```

src/jrd/RecordSourceNodes.cpp
```diff
@@ -3651,6 +3657,331 @@ RseNode* SelectExprNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
 	return PASS1_derived_table(dsqlScratch, this, NULL);
 }
 
+TableValueFunctionSourceNode*
```
asfernandes, line 3660: This is not a Firebird code convention. Please start the method name in the same line as the return type and break later as necessary.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -3651,6 +3657,331 @@ RseNode* SelectExprNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
...
+
+TableValueFunctionSourceNode* TableValueFunctionSourceNode::parse2(thread_db* tdbb,
+																   const SSHORT blrOp)
+{
+	MemoryPool& pool = *tdbb->getDefaultPool();
+	if (blrOp == blr_table_value_fun_unlist)
+		return FB_NEW_POOL(pool) UnlistFunctionSourceNode(pool);
+
+	fb_assert(false);
+	return nullptr;
```
asfernandes, line 3729: This should generate an error.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -3651,6 +3657,331 @@ RseNode* SelectExprNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
...
+	dsqlScratch->appendMetaString(dsqlContext->ctx_alias.c_str());
+
+	dsqlScratch->appendUShort(dsqlContext->ctx_proc_inputs->items.getCount());
+	for (auto& arg : dsqlContext->ctx_proc_inputs->items)
+		GEN_expr(dsqlScratch, arg);
+
+	Array<const dsql_fld*> arrayFld;
+	for (const auto* field = tableValueFunctionContext->outputField; field;
+		 field = field->fld_next)
+		arrayFld.add(field);
```
asfernandes, line 3791: Add braces to multiline `for` body.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -3651,6 +3657,331 @@ RseNode* SelectExprNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
...
+		field = newField;
+
+		auto inputItem = inputList->items.begin()->getObject();
+		dsc desc;
+		DsqlDescMaker::fromNode(dsqlScratch, &desc, inputItem);
+
+		USHORT ttype = desc.getCharSet();
+
+		if (ttype == CS_NONE)
+			ttype = CS_ASCII;
```
asfernandes, line 3964: Why?

src/jrd/RecordSourceNodes.cpp
```diff
@@ -3651,6 +3657,331 @@ RseNode* SelectExprNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
...
+
+		desc.makeText(32, ttype);
+		MAKE_field(newField, &desc);
+		newField->fld_id = 0;
+	}
+
+	if (dsqlNameColumns.hasData())
+	{
+		if (dsqlNameColumns.getCount() > 1)
```
asfernandes, line 3973: Use braces.

src/jrd/RecordSourceNodes.h
```diff
@@ -963,6 +963,68 @@ class SelectExprNode final : public TypedNode<RecordSourceNode, RecordSourceNode
...
+public:
+	explicit TableValueFunctionSourceNode(MemoryPool& pool)
+		: TypedNode<RecordSourceNode, RecordSourceNode::TYPE_TABLE_VALUE_FUNCTION>(pool),
+		  dsqlName(pool), alias(pool), dsqlField(nullptr), dsqlNameColumns(pool),
+		  m_csbTableValueFun(nullptr)
+	{
+	}
+	static TableValueFunctionSourceNode* parse(thread_db* tdbb, CompilerScratch* csb,
+											   const SSHORT blrOp);
+	static TableValueFunctionSourceNode* parse2(thread_db* tdbb, const SSHORT blrOp);
```
asfernandes, line 978: Use better name.

src/jrd/exe.h
```diff
@@ -431,6 +431,28 @@ class ItemInfo : public Printable
 typedef Firebird::LeftPooledMap<MetaNamePair, FieldInfo> MapFieldInfo;
 typedef Firebird::RightPooledMap<Item, ItemInfo> MapItemInfo;
 
+// Table value function block
+
+class jrd_tab_func
```
asfernandes, line 436: Use better name.

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+	rpb->rpb_number.setValue(BOF_NUMBER);
+
+	fb_assert(m_inputList->items.getCount() >= UNLIST_INDEX_LAST);
+
+	auto valueItem = m_inputList->items[UNLIST_INDEX_STRING];
+	const auto valueDesc = EVL_expr(tdbb, request, valueItem);
+	if (valueDesc == nullptr)
+	{
+		rpb->rpb_number.setValid(true);
```
asfernandes, line 162: Why `true`?

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -0,0 +1,277 @@
...
+		rpb->rpb_number.setValid(false);
+		return;
+	}
+
+	auto toDesc = m_format->fmt_desc.begin();
+	fb_assert(toDesc);
+
+	const auto textType = toDesc->getTextType();
+
+	auto setStringToRecord = [&] (string str, USHORT length = 0)
```
asfernandes, line 179: Whould not be a string reference?

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -0,0 +1,277 @@
...
+		string resultStr;
+
+		while (!(blob->blb_flags & BLB_eof))
+		{
+			UCHAR buffer[BUFFER_LARGE];
+			const auto length = blob->BLB_get_data(tdbb, buffer, sizeof(buffer), false);
+			if (length == 0)
+				continue;
+
+			string valueStr(buffer, length);
```
asfernandes, line 212: Please do not copy the buffer to another string.

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+			auto end = AbstractString::npos;
+			do
+			{
+				auto size = end = valueStr.find(separatorStr);
```
asfernandes, line 217: Is this code compatible with multi-character separator and the list string split in different blob segments?

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -0,0 +1,277 @@
...
+	string separatorStr(MOV_make_string2(tdbb, separatorDesc, textType, true));
+
+	if (separatorStr.isEmpty())
+	{
+		string valueStr(MOV_make_string2(tdbb, valueDesc, textType, false));
+		setStringToRecord(valueStr);
+		return;
+	}
+
+	if (valueDesc->isBlob())
```
asfernandes, line 198: Others Firebird functions compare strings using collation rules.
Here this is disregard. Should it be?

dyemanov:
> In the discussion of #8005 it was suggested to use the shortened syntax for `IN UNLIST` for convenience, but it is not necessary.
> 
> One more question.
> 
> ```
> SELECT *
> FROM UNLIST(?, ',' RETURNING INT)
> ```
> 
> What type of input parameter will be returned to the client after preparing the query?
> 
> `BLOB SUB_TYPE TEXT` or `VARCHAR(N)`. If the second, what will be N?
> 
> For the output parameter, unless otherwise specified, it will be `VARCHAR(32)` judging by the description above.



> In the discussion of #8005 it was suggested to use the shortened syntax for `IN UNLIST` for convenience, but it is not necessary.
> 
> One more question.
> 
> ```
> SELECT *
> FROM UNLIST(?, ',' RETURNING INT)
> ```
> 
> What type of input parameter will be returned to the client after preparing the query?
> 
> `BLOB SUB_TYPE TEXT` or `VARCHAR(N)`. If the second, what will be N?
> 
> For the output parameter, unless otherwise specified, it will be `VARCHAR(32)` judging by the description above.

Good question to be resolved together. Something like `VARCHAR(1024)` looks OK at the first glance, but it would lead to a runtime error if overflowed by the user (coercion will fail). `CLOB` looks safer but has a noticeable overhead if the input string is actually a `VARCHAR`.

aafemt:
> it would lead to a runtime error if overflowed by the user (coercion will fail).

In #8145 I already pass user-provided buffers to looper. One step further and functions can accept user input directly as descriptor avoiding coercion and thus overflow.

AlexPeshkoff:
On 1/30/25 13:01, Dimitry Sibiryakov wrote:
>
>     it would lead to a runtime error if overflowed by the user
>     (coercion will fail).
>
> In #8145 <https://github.com/FirebirdSQL/firebird/pull/8145> I already 
> pass user-provided buffers to looper. One step further and functions 
> can accept user input directly as descriptor avoiding coercion and 
> thus overflow.
>

This looks like solution.

src/jrd/recsrc/TableValueFunctionScan.cpp
dyemanov: No, it's not. To be improved, I guess. While `LIST` would never split the separator, the input may be provided from the different source as well and we cannot predict the layout.

src/jrd/recsrc/TableValueFunctionScan.cpp
dyemanov: Good question to be discussed. Theoretically, one may ask to split the text into chunks by e.g. both 'A' and 'a' used as a case-insensitive separator 'a'. But honestly, I cannot imagine that being used in practice. And collation-aware compares would add some extra overhead. So maybe it's worth comparing byte-wise (converting both value and separator to the common data type, if necessary). Or maybe we should consider two code branches depending on whether byte-wise compare is possible or not.

doc/sql.extensions/README.unlist
```diff
@@ -0,0 +1,69 @@
...
+		text = '123:123:123';
+		FOR SELECT * FROM UNLIST( :text, ':' RETURNING INT) AS A INTO :PROC_RETURN_INT DO
+		SUSPEND;
+	END^
+	SET TERM ; ^
+	SELECT * FROM TEST_PROC;
+
+G)
+	CREATE DOMAIN D1 AS INT;
+	SELECT TEST_DOMAIN FROM UNLIST('1,2,3,4' RETURNING D1) AS A(TEST_DOMAIN);
```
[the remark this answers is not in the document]
ChudaykinAlex: Check. Granted. Added example in readme

src/dsql/ExprNodes.cpp
```diff
@@ -6446,9 +6458,28 @@ dsql_fld* FieldNode::resolveContext(DsqlCompilerScratch* dsqlScratch, const Meta
 		return nullptr;
 	}
 
+	const TEXT* dsqlName = NULL;
```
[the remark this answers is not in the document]
ChudaykinAlex: My bad.

src/dsql/parse.y
```diff
@@ -6595,6 +6597,84 @@ outer_noise
 	| OUTER
 	;
 
+%type <recSourceNode> table_value_function
+table_value_function
+	: table_value_function_clause table_value_function_correlation_name table_value_function_columns_name
+		{
+			auto *node = nodeAs<TableValueFunctionSourceNode>($1);
```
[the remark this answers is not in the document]
ChudaykinAlex: Сorrected.

src/dsql/parse.y
```diff
@@ -6595,6 +6597,84 @@ outer_noise
...
+
+
+%type <metaNamePtr> table_value_function_correlation_name
+table_value_function_correlation_name
+	: as_noise symbol_table_alias_name	{ $$ = $2; }
+	;
+
+
+%type <metaNameArray> table_value_function_columns_name
+table_value_function_columns_name
```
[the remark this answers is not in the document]
ChudaykinAlex: Apparently I didn't notice this rule.

src/jrd/RecordSourceNodes.cpp
ChudaykinAlex: I used the clang-format utility to do the formatting. That's what he decided. I'll be more careful

src/jrd/RecordSourceNodes.cpp
ChudaykinAlex: The code should not reach that line, and the error would not be logical to the user. I would be grateful if you could suggest a solution.

src/jrd/RecordSourceNodes.cpp
ChudaykinAlex: Corrected.

src/jrd/RecordSourceNodes.cpp
ChudaykinAlex: Corrected.

src/jrd/RecordSourceNodes.cpp
ChudaykinAlex: I think it should be announced somehow.

src/jrd/RecordSourceNodes.h
ChudaykinAlex: I've seen similar naming. I haven't come up with a better name yet. I'll keep thinking

src/jrd/exe.h
ChudaykinAlex: I tried to fix it

src/jrd/recsrc/TableValueFunctionScan.cpp
ChudaykinAlex: Thank you.

src/jrd/RecordSourceNodes.cpp
asfernandes: That will crash release build with direct BLR manipulation.
It should generate a proper error.

src/jrd/RecordSourceNodes.cpp
asfernandes: Why should it transform NONE to ASCII here?

src/jrd/RecordSourceNodes.cpp
ChudaykinAlex: I hear you.

src/jrd/recsrc/TableValueFunctionScan.cpp
ChudaykinAlex: Are you telling me to read the data directly into a string?

sim1984: I think it's better to use something lightweight like string_view here, so as not to allocate memory and not do unnecessary copying.

ChudaykinAlex: I'm going to try an experiment with string_view

src/jrd/RecordSourceNodes.cpp
ChudaykinAlex: I think in this case the result charset should not be CS_NONE (0), but the charset of the input string/blob. If the input is not a string/blob, then CS_ASCII.

src/jrd/recsrc/TableValueFunctionScan.cpp
dyemanov: @asfernandes, do we generally consider OK to read/process the blob as a whole? If so, this would be the easiest solution (and IIRC it worked this way in our original implementation) and we do the same in some built-in functions. Or should it be considered a bad practice to be revisited/fixed and thus the new code should be able to process blobs in chunks?

dyemanov: It requires to move the processing from `open()` to `getRecord()`. Currently implemented differently, but it could be worth the effort.

sim1984: I wrote my UDR Split exactly like this. Not perfect, but works well https://github.com/sim1984/split_udr/blob/master/src/SplitProcedure.cpp

asfernandes: > @asfernandes, do we generally consider OK to read/process the blob as a whole? If so, this would be the easiest solution (and IIRC it worked this way in our original implementation) and we do the same in some built-in functions. Or should it be considered a bad practice to be revisited/fixed and thus the new code should be able to process blobs in chunks?

For this function, the cost/benefit to incrementally read the input is very clear for me.

dyemanov: Shouldn't it depend on `OPTIMIZE FOR {FIRST | ALL} ROWS`? I.e. process/return records in chunks for `ALL ROWS` (not necessarily read/cache the whole blob, maybe just some sliding buffer) but one by one for `FIRST ROWS`?

asfernandes: > Shouldn't it depend on `OPTIMIZE FOR {FIRST | ALL} ROWS`? I.e. process/return records in chunks for `ALL ROWS` (not necessarily read/cache the whole blob, maybe just some sliding buffer) but process/return records one by one for `FIRST ROWS`?

I do not see a clear relation of `OPTIMIZE FOR` and `UNNEST`. `OPTIMIZE FOR` is top level query feature, while `UNNEST` may be in inner queries.

dyemanov: I mean the mechanics, not the clause itself. For example, a subquery with a `FIRST` clause implies `FOR FIRST ROWS` when compiling/optimizing its own RSE.

sim1984: The current implementation has two drawbacks:
1. Increased memory consumption due to result buffering
2. Slowing down queries with the FIRST ROWS strategy

In general, it is not very clear why TableValueFunction is a buffered data source. It seems to be related to primary, and all our primary sources are cover. If you need to buffer the result, then the optimizer of the higher level does it.

dyemanov: These issues are being addressed.

src/jrd/recsrc/TableValueFunctionScan.cpp
dyemanov: I'd even say it should be a const string reference.

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -0,0 +1,279 @@
...
+		return false;
+	}
+
+	rpb->rpb_number.increment();
+
+	if (!impure->m_recordBuffer->fetch(rpb->rpb_number.getValue(), rpb->rpb_record))
+	{
+		rpb->rpb_number.setValid(false);
+		return false;
+	}
```
dyemanov, line 86: `rpb->rpb_number.setValid(true);`

src/jrd/recsrc/RecordSource.h
```diff
@@ -1462,6 +1462,59 @@ namespace Jrd
 		NestConst<BoolExprNode> const m_boolean;
 	};
 
+	class TableValueFunctionScan : public RecordStream
+	{
+	protected:
+		struct Impure : public RecordSource::Impure
+		{
+			RecordBuffer* m_recordBuffer;
+			blb* m_blob;
```
asfernandes, line 1471: Aren't you mixing specifics of UnlistFunctionScan's impure in the base TableValueFunctionScan?

pavel-zotov:
PS.
Same for
```select * from unlist('abcdefghi,abcdefghj', ',') as u;```
(but not for 'abcdefgh' - for this case function works OK)

**Comment by sim1984** [comment]
This is not surprising considering that the default return type is This is not surprising considering that the default return type is `varchar(32) character set none`. It is better to always specify the type that is needed.

**Comment by dyemanov** [comment]
Resulting type should be VARCHAR(32) of the input character set, but it appears it's described as 32 bytes, not characters. To be fixed, AFAIU.

As for the length, maybe 32 is a short limit, but this function is to be commonly used with numbers and this limit is enough for that usage case. If you want to deal with strings, better specify the length yourself (nobody knows how many characters gonna fit).

**Comment by pavel-zotov** [comment]
QA note: deferred waiting for fix of resulting type (expect  [32 characters](https://github.com/FirebirdSQL/firebird/pull/8418#issuecomment-2755462383) rather than bytes)

File: src/dsql/parse.y [hunk 19]
```diff
@@ -6589,6 +6591,77 @@ outer_noise
...
+	: /*nothing*/							{ $$ = NULL; }
+	| RETURNING data_type_descriptor		{ $$ = $2; }
+	;
+
+%type <metaNamePtr> table_value_function_correlation_name
+table_value_function_correlation_name
+	: as_noise symbol_table_alias_name	{ $$ = $2; }
+	;
+
+%type <metaNameArray> table_value_function_columns
```
asfernandes, line 6645: There are warnings about rules `table_value_function_column*` never used.

asfernandes:
@ChudaykinAlex there is also a missing entry in the final of src/jrd/blp.h for BLR printing.

pavel-zotov:
Consider following script (save it in some.sql):
```
connect '/:employee';
set heading off;
set sqlda_display on;
set planonly;
select u.* from unlist(?, ':' returning varchar(8190) character set utf8) as u;
```

If we run ISQL several times with misc. values of '-ch [charset]' then we can see:
```
C:\FB\60SS\isql -q -i unlist-sqlda.sql -ch win1250
INPUT message field count: 1
01: sqltype: 448 VARYING scale: 0 subtype: 0 len: 1024 charset: 51 WIN1250

OUTPUT message field count: 1
01: sqltype: 448 VARYING scale: 0 subtype: 0 len: 8190 charset: 51 WIN1250
---------------------------------------------------------------------------
C:\FB\60SS\isql -q -i unlist-sqlda.sql -ch win1251
INPUT message field count: 1
01: sqltype: 448 VARYING scale: 0 subtype: 0 len: 1024 charset: 52 WIN1251

OUTPUT message field count: 1
01: sqltype: 448 VARYING scale: 0 subtype: 0 len: 8190 charset: 52 WIN1251
---------------------------------------------------------------------------
C:\FB\60SS\isql -q -i unlist-sqlda.sql -ch utf8
INPUT message field count: 1
01: sqltype: 448 VARYING scale: 0 subtype: 0 len: 1024 charset: 4 UTF8

OUTPUT message field count: 1
01: sqltype: 448 VARYING scale: 0 subtype: 0 len: 32760 charset: 4 UTF8
```

Q-1.
UNLIST() by default implies that input value has 1024 characters - is it expected ? If yes, then what "[Fix 32 bytes to 32 characters](https://github.com/FirebirdSQL/firebird/pull/8418#issuecomment-2761038685)" means ?

Q2.
Why OUTPUT argument is always is the same as connection charset (rather than what was specified in 'returning' clause) ?

aafemt:
A1: Yes, 32 characters is default for **output** value.
A2: Because this is a common datatype coercion rule: all string values are delivered to client in requested charset (which is by default the connection charset).

pavel-zotov:
@aafemt 
yes, me wrong (did not read carefully doc).
But what about INPUT: "01: sqltype: 448 ... len: 1024" ?

pavel-zotov:
from doc/sql.extensions/README.unlist.txt
```
2) <separator>: optional value expression that returns a string which is used as a delimiter (i.e. it separates one value from another inside the input string). 
It may also be a BLOB TEXT value, limited to 32KB. 
If an empty string is specified, the output will be just one record containing the input string.
If omitted, the comma character is used as a delimiter.
```

Am i right in guess that one may to use *any* unicode character as ```<separator>``` ?
(maybe except from 'surrogate' range and control characters from [0 ... 31])

pavel-zotov:
Consider attached .zip (there is .fbk that was created after some tests with employee on 6.0.0.716).
Unpack it, restore, connect with charset = utf8  and run following query:
```
select u.* from unlist((select blob_fld from t_longblob), unicode_char(0x227b) returning blob character set utf8) as u(x)
where x containing unicode_char(0x227b)
;
```
You will see one record whereas expected is empty rowset because delimiter must not appear in any rows produced by UNLIST() function (AFAIU).
// unicode_char(0x227b) - valid character, it is math sign with name ["Succeeds"](https://www.compart.com/en/unicode/U+227B)

Screen:
изображение


[employee-fbk-unlist-unexpected-result_-_WI-6.0.0.716-a5b25c1.fbk.zip](https://github.com/user-attachments/files/19618017/employee-fbk-unlist-unexpected-result_-_WI-6.0.0.716-a5b25c1.fbk.zip)

PS
Yes, I understand that such data / separator cannot exist in real applications.
But maybe this [wrong] result will help to find some other errors (e.g. in FB unicode characters handling ?)

ChudaykinAlex:
The default input parameter is VARCHAR(1024) 
CS_dynamic. Why exactly 1024?  There is no definitive answer. This can be corrected.
There may be an answer here.
https://github.com/FirebirdSQL/firebird/pull/8418#issuecomment-2623955138

ChudaykinAlex:
I'll definitely look into it.

pavel-zotov:
One more example - unpack attached zip, restore from .fbk and run this query:
```
set count on;
set list on;
set blob all;
set echo on;
select u.x, octet_length(u.x), position(unicode_char(0x2114) in u.x), unicode_char(0x2114)
from unlist((select blob_fld from t_longblob), unicode_char(0x2114) returning blob character set utf8) as u(x)
where x containing unicode_char(0x2114)
;
```

Unicode char 0x2114 is "LB Bar", https://www.compart.com/en/unicode/U+2114:
изображение

Output will be:
изображение

It seems that minimal length of unicode string on which this can be reproduced is ~11'800 characters.
Total number of characters which are needed to be used as 'alphabet' to generate random delimiter is 8894 (i could not reproduce on less values during several hours).
[employee-fbk-unlist-unexpected-result_-_WI-6.0.0.717-fbb6b0c.fbk.zip](https://github.com/user-attachments/files/19665275/employee-fbk-unlist-unexpected-result_-_WI-6.0.0.717-fbb6b0c.fbk.zip)

ChudaykinAlex:
Good day. Thank you for your comments. I have looked into the previous example. I found an error in my code. I made a patch, now locally checking the build and tests. And will prepare for publishing.
I have executed your script from the example.
image

pavel-zotov:
0xFF.
IMO, it is inconveniently that one need to repeat some parts of code (e.g. "unicode_char(0x2114)").
Why this function could not be used as *other* functions, like this:
```
with
d as (
    select
         blob_fld
        ,unicode_char(0x2114) as separator
    from t_longblob
)
, e as (
    select unlist(d.blob_fld, d.separator returning blob character set utf8) as x
    from d
)
select e.x
from e
where e.x containing d.separator
;
```
-- ?

pavel-zotov:
It seems that ASCII_CHAR(0) can not be used as separator (at least currently):
```
echo set list on; set count on; select * from unlist('1', ascii_char(1)) as u(x); | isql /:employee
X                               1
Records affected: 1
(OK, expected)

echo set list on; set count on; select * from unlist('1', ascii_char(0)) as u(x); | isql /:employee
-- NO OUTPUT. FB hangs with 100% load of one CPU core.
-- Ctrl-Creak in ISQL does not help: FB process continues its activity
```
Checked on Windows.
This is URL to FB dump, stack trace, snapshot 6.0.0.725-a2b05f4-x64.7z and  other info:
https://drive.google.com/drive/folders/1BgTSsRqvyD-cbmnvyE43xlG0vpxuOwk5?usp=sharing

aafemt:
ascii_char(0) must not be used anywhere outside of OCTETS. Perhaps this function should explicitly prohibit zero value.

But the infinite loop is definitely a bug.

pavel-zotov:
One more example:
```
connect '/:employee';
set echo on;
set list on;
set count on;

select '1' || ascii_char(26) || '2' as chr_26_lst from rdb$database;

select /* check ascii_char(26) */ * from unlist( '1' || ascii_char(26) || '2', ascii_char(26)) as u(x);

select /* check ascii_char(26) */ * from unlist('1→2', '→') as u(x);
```

изображение

Output will be:
```
set list on;
set count on;
select '1' || ascii_char(26) || '2' as chr_26_lst from rdb$database;
CHR_26_LST                      1→2
Records affected: 1

select /* check ascii_char(26) */ * from unlist( '1' || ascii_char(26) || '2', ascii_char(26)) as u(x);
X                               1
X                               2
Records affected: 2

select /* check ascii_char(26) */ * from unlist('1
Expected end of statement, encountered EOF
```

изображение

aafemt:
"This function" I meant `ascii_char()`.

pavel-zotov:
> "This function" I meant ascii_char().

Why to decline the whole function usage ? IMO, it is enough to restrict only some values that it returns, at least chr(0).

aafemt:
> IMO, it is enough to restrict only some values that it returns, at least chr(0).

Yes, this is exactly what I wrote: `ascii_char` would better to refuse zero,

pavel-zotov:
> This is a separate ISQL problem, not UNLIST one.

https://github.com/FirebirdSQL/firebird/issues/8512

ChudaykinAlex:
> It seems that ASCII_CHAR(0) can not be used as separator (at least currently):
> 
> ```
> echo set list on; set count on; select * from unlist('1', ascii_char(1)) as u(x); | isql /:employee
> X                               1
> Records affected: 1
> (OK, expected)
> 
> echo set list on; set count on; select * from unlist('1', ascii_char(0)) as u(x); | isql /:employee
> -- NO OUTPUT. FB hangs with 100% load of one CPU core.
> -- Ctrl-Creak in ISQL does not help: FB process continues its activity
> ```
> 
> Checked on Windows. This is URL to FB dump, stack trace, snapshot 6.0.0.725-a2b05f4-x64.7z and other info: https://drive.google.com/drive/folders/1BgTSsRqvyD-cbmnvyE43xlG0vpxuOwk5?usp=sharing

I found the problem with the infinite loop, I will fix it.
As for the ban ascii_char(0), I won't give you an answer yet.

aafemt:
> As for the ban ascii_char(0), I won't give you an answer yet.

Perhaps you can answer an other question: is `UNLIST` supposed to work with `(VAR)BINARY` datatype (charset `OCTETS`) and binary BLOBs?

pavel-zotov:
It seems that semicolon character ( '```;```' = ascii_char(59)) has some effect on time of PREPARE statement when source string has lot of such separators.
Consider attached .zip file, it contains four queries with different characters used as separator: '```:```', '```;```', '```<```' and '```=```'.
In each case extremely long source string is used for parsing:
```q'#t;y;;;<skipped>;;;#'``` -- where ```<skipped>``` is sequence of ~32K similar characters.

Queries look like this:
```
select /* check ascii_char(58) */ * from unlist(q'#t:y::: ... :::#')', ':') ;
select /* check ascii_char(59) */ * from unlist(q'#t;y;;; ... ;;;#')', ';') ;
select /* check ascii_char(60) */ * from unlist(q'#t<y<<< ... <<<#')', '<') ;
select /* check ascii_char(61) */ * from unlist(q'#t=y=== ... ===#')', '=') ;
```

Then run trace with config:
```
    log_connections = true
    log_transactions = true
    log_statement_prepare = true
    log_statement_start = true
    log_statement_finish = true
    log_statement_free = true
    print_perf = true
```
(its log also is in attached .zip)

For 1st, 4rd and 4th statements trace will show similar parts which executes almost instantly:
```
2025-04-10T17:26:13.2550 (232:0000000001FE24C0) START_TRANSACTION
...
2025-04-10T17:26:13.2550 (232:0000000001FE24C0) EXECUTE_STATEMENT_FINISH
...
SET TRANSACTION

2025-04-10T17:26:13.2550 (232:0000000001FE24C0) FREE_STATEMENT
...
2025-04-10T17:26:13.2550 (232:0000000001FE24C0) START_TRANSACTION
...
2025-04-10T17:26:13.2650 (232:0000000001FE24C0) PREPARE_STATEMENT
...
Statement 59:
-------------------------------------------------------------------------------
select /* check ascii_char(58) */ * from unlist(q'#t:y:::::::::::::::::::::::::::::::::
      3 ms

2025-04-10T17:26:13.2660 (232:0000000001FE24C0) EXECUTE_STATEMENT_START
...
select /* check ascii_char(58) */ * from unlist(q'#t:y:::::::::::::::::::::::::::::::::

2025-04-10T17:26:13.3730 (232:0000000001FE24C0) EXECUTE_STATEMENT_FINISH
...
```

But for SECOND statement (when ```;``` is used as delimiter) trace will show that there is weird "gap" in timestamps after START_TRANSACTION and PREPARE_STATEMENT.
Size of this "gap" is about 8 seconds.

Appropriate lines are marked as **[ 1 ]** and **[ 2 ]**, their stamps: 2025-04-10T17:26:23.0550 and 2025-04-10T17:26:31.1140:

```
2025-04-10T17:26:23.0550 (232:0000000001FE24C0) START_TRANSACTION
   employee (ATT_245, SYSDBA:NONE, NONE, TCPv6:fe80::8974:2b63:cd85:3f68%17/56873)
   C:\FB\60SS\isql.exe:10040
       (TRA_276, CONCURRENCY | WAIT | READ_WRITE)

2025-04-10T17:26:23.0550 (232:0000000001FE24C0) EXECUTE_STATEMENT_FINISH
   employee (ATT_245, SYSDBA:NONE, NONE, TCPv6:fe80::8974:2b63:cd85:3f68%17/56873)
   C:\FB\60SS\isql.exe:10040
       (TRA_276, CONCURRENCY | WAIT | READ_WRITE)

-------------------------------------------------------------------------------
SET TRANSACTION
0 records fetched
      0 ms, 1 write(s), 1 fetch(es), 1 mark(s)

2025-04-10T17:26:23.0550 (232:0000000001FE24C0) FREE_STATEMENT
   employee (ATT_245, SYSDBA:NONE, NONE, TCPv6:fe80::8974:2b63:cd85:3f68%17/56873)
   C:\FB\60SS\isql.exe:10040

-------------------------------------------------------------------------------
SET TRANSACTION

2025-04-10T17:26:23.0550 (232:0000000001FE24C0) START_TRANSACTION ------------------------ [  1  ]
   employee (ATT_245, SYSDBA:NONE, NONE, TCPv6:fe80::8974:2b63:cd85:3f68%17/56873)
   C:\FB\60SS\isql.exe:10040
       (TRA_277, READ_COMMITTED | NO_REC_VERSION | WAIT | READ_WRITE)

2025-04-10T17:26:31.1140 (232:0000000001FE24C0) PREPARE_STATEMENT ------------------------ [  2  ]
   employee (ATT_245, SYSDBA:NONE, NONE, TCPv6:fe80::8974:2b63:cd85:3f68%17/56873)
   C:\FB\60SS\isql.exe:10040
       (TRA_277, READ_COMMITTED | NO_REC_VERSION | WAIT | READ_WRITE)

Statement 59:
-------------------------------------------------------------------------------
select /* check ascii_char(59) */ * from unlist(q'#t;y;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
      2 ms

2025-04-10T17:26:31.1140 (232:0000000001FE24C0) EXECUTE_STATEMENT_START
   employee (ATT_245, SYSDBA:NONE, NONE, TCPv6:fe80::8974:2b63:cd85:3f68%17/56873)
   C:\FB\60SS\isql.exe:10040
       (TRA_276, CONCURRENCY | WAIT | READ_WRITE)

Statement 59:
-------------------------------------------------------------------------------
select /* check ascii_char(59) */ * from unlist(q'#t;y;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

2025-04-10T17:26:31.2350 (232:0000000001FE24C0) EXECUTE_STATEMENT_FINISH
   employee (ATT_245, SYSDBA:NONE, NONE, TCPv6:fe80::8974:2b63:cd85:3f68%17/56873)
   C:\FB\60SS\isql.exe:10040
       (TRA_276, CONCURRENCY | WAIT | READ_WRITE)
```

[unlist-long-duplicated-separators-at-end.zip](https://github.com/user-attachments/files/19686940/unlist-long-duplicated-separators-at-end.zip)

изображение

**Comment by mrotteveel** [comment]
That thing with `;` is probably the new ISQL feature to automatically recognize when a statement is complete.

**Comment by dyemanov** [comment]
> > As for the ban ascii_char(0), I won't give you an answer yet.
> 
> Perhaps you can answer an other question: is `UNLIST` supposed to work with `(VAR)BINARY` datatype (charset `OCTETS`) and binary BLOBs?

Yes, it's expected to work.

**Comment by dyemanov** [comment]
It's not about UNLIST at all, you're comparing different cases. See:

```sql
SQL> select cast('2.225073858507202e-308' as double precision) from rdb$database;

                   CAST 
======================= 
-numeric value is out of range
SQL> 

```

pavel-zotov:
0xFF. Is this a bug or no ? (i mean: cast('2.225073858507202e-308' as double precision) ==> numeric value is out of range)

dyemanov:
IMO: if the numeric literal fits the double precision range, CAST from string should work. But maybe the underlying conversion is trickier and may cause undesired side effects.

pavel-zotov:
QA note: deferred because of
https://github.com/FirebirdSQL/firebird/pull/8418#issuecomment-2789284597
https://github.com/FirebirdSQL/firebird/pull/8418#issuecomment-2792358627
(waiting for fix / resolution)

dyemanov:
Please retest with the new snapshot.

pavel-zotov:
Checked on 6.0.0.744-e883a89: all fine now. New test for UNLIST will be added soon.

pavel-zotov:
::: QA note :::
See functional/intfunc/unlist/ folder with several tests related to this function.

Issue #8426, opened 2025-02-03
Error while parsing trace configuration with 2 sections


Noremos:
I have the following configuration in fbtrace.conf:  
```
database = /var/primary-db.fdb  
{  
  enabled = true  
  log_errors = true  
}  

database = /var/secondary.fdb  
{  
  enabled = true  
  log_connections = true  
}  

database  
{  
  enabled = true  
}  
```

It works fine when connecting to primary-db.fdb. However, the error appears when connecting to secondary.fdb or any other database.  

```
art  Mon Feb  3 10:19:12 2025
  Trace plugin fbtrace returned error on call trace_create.
  error while parsing trace configuration
  line 1: error while compiling regular expression "/tmp/primary-db.fdb"
```

The Firebird configuration:  
`AuditTraceConfigFile = fbtrace.conf  `

I think this is unclear behavior because the trace works all the time but suddenly stops working when connecting to another database.

Noremos:
I suggest a possible solution: https://github.com/FirebirdSQL/firebird/pull/8427

pavel-zotov:
QA note: see test for #8427

PR #8427, opened 2025-02-03
Allow specifying exactly what type of database string is expected in the database section for trace configuration

Noremos:
The new syntax:  
  ```
database = name /var/primary-db.fdb  
{  
  enabled = true  
  log_errors = true  
}  

database = name /var/secondary.fdb  
{  
  enabled = true  
  log_connections = true  
}  

database = regex (%[\\/](e[[:DIGIT:]]{{2}}).fdb)  
{  
  enabled = true  
}  

# Default behavior for backward compatibility  
database = %[\\/]my_database.fdb  
{  
  enabled = true  
}  
```
  

I chose `database = name path.fdb` over `database name = path.fdb` because it is easier to implement and parse, but it doesn't look as great, so it can be reimplemented.

### DIFF
--- a/src/utilities/ntrace/TraceConfiguration.cpp
+++ b/src/utilities/ntrace/TraceConfiguration.cpp
@@ -90,7 +90,7 @@ void TraceCfgReader::readConfig()
 		const ConfigFile::Parameter* section = &params[n];
 
 		const ConfigFile::SectionType sectionType = section->parseSectionKey();
-		const bool isDatabase = sectionType != ConfigFile::SectionType::SERVICE;
+		const bool isDatabase = (sectionType != ConfigFile::SectionType::SERVICE);
 
 		const ConfigFile::String pattern = section->value;
 		bool match = false;
--- a/src/utilities/ntrace/TraceConfiguration.cpp
+++ b/src/utilities/ntrace/TraceConfiguration.cpp
@@ -70,15 +70,6 @@ void TraceCfgReader::readTraceConfiguration(const char* text,
 
 #define ERROR_PREFIX "error while parsing trace configuration\n\t"
 
-enum class AliasType
-{
-	DETECT,
-	NAME,
-	REGEX
-};
-inline constexpr std::string_view REGEX_PREFIX = "regex ";
-inline constexpr std::string_view NAME_PREFIX = "name ";
-
 void TraceCfgReader::readConfig()
 {
 	ConfigFile cfgFile(ConfigFile::USE_TEXT, m_text, ConfigFile::HAS_SUB_CONF | ConfigFile::NATIVE_ORDER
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/utilities/ntrace/TraceConfiguration.cpp [hunk 1]
```diff
@@ -70,6 +70,15 @@ void TraceCfgReader::readTraceConfiguration(const char* text,
...
 
+enum class AliasType
+{
+	DETECT,
+	NAME,
+	REGEX
+};
+inline constexpr std::string_view REGEX_PREFIX = "regex ";
+inline constexpr std::string_view NAME_PREFIX = "name ";
+
```
hvlad, line 81: I can't find `#include <string_view>` anywhere in the whole source tree...

src/utilities/ntrace/TraceConfiguration.cpp
```diff
@@ -127,18 +136,41 @@ void TraceCfgReader::readConfig()
...
 		{
-			PathName noQuotePattern = pattern.ToPathName();
+			PathName noQuotePattern;
+
+			AliasType specifier = AliasType::DETECT;
+			const std::string_view patternView(pattern.data(), pattern.length());
+
+			// Check for "name" or "regex" specific in the path.
+			// Compare path in both ways if specifier is not specified
+			if (patternView.find_first_of(REGEX_PREFIX) == 0)
```
hvlad, line 146: Is it case-sensitive ? Note, section name is not case-sensitive

hvlad (commented):
Please, avoid massive not related changes such as end space trimming.
It makes harder to read the patch looking for valuable changes.

hvlad:
> I chose `database = name path.fdb` over `database name = path.fdb` because it is easier to implement and parse, but it doesn't look as great, so it can be reimplemented.

Consider `databaseName = path` or `databaseByName = path`.

Note, in proposed syntax `database name path.fdb` also accepted, as `=` is not required by config code.

Also, your sample is wrong as default section must be the first one.

src/utilities/ntrace/TraceConfiguration.cpp
Noremos: It looks like some other include was implicitly connecting string_view. Anyway, I have removed the usage of string_view

src/utilities/ntrace/TraceConfiguration.cpp
Noremos: Change it to compare keys with NoCaseString

Noremos:
> > I chose `database = name path.fdb` over `database name = path.fdb` because it is easier to implement and parse, but it doesn't look as great, so it can be reimplemented.
> 
> Consider `databaseName = path` or `databaseByName = path`.

Now databaseName and databaseRegex are used

```
database
{
	enabled = false
}

databaseName = /var/primary-db.fdb
{
	enabled = true
	log_errors = true
}

databaseRegex = (%[\\/](e[[:DIGIT:]]{{2}}).fdb)
{
	enabled = true
}

database = employee
{
	enabled = true
}
```

hvlad (approved):

Noremos:
The correct term is "services." I misspelled it, but I'll prepare a PR to fix it.

Noremos:
https://github.com/FirebirdSQL/firebird/pull/8500

pavel-zotov:
```
#	database = (%[\\/](e[[:DIGIT:]]{{2}}).fdb)
vs
#	databaseRegex = (%[\\/](e[[:DIGIT:]]{{2}}).fdb)
```
What's the difference between them ? (i.e. for what one more section was introduced - just for ease of perception when read trace config ?)

Noremos:
> ```
> #	database = (%[\\/](e[[:DIGIT:]]{{2}}).fdb)
> vs
> #	databaseRegex = (%[\\/](e[[:DIGIT:]]{{2}}).fdb)
> ```
> 
> What's the difference between them ? (i.e. for what one more section was introduced - just for ease of perception when read trace config ?)

This change is a solution to a problem when there is a database that has a correct path but the path is an incorrect regex: https://github.com/FirebirdSQL/firebird/issues/8426

PR #8428, opened 2025-02-03
ODS14: Store (and return via *_LENGTH functions) blob length longer than 4GB

dyemanov:
This PR contains only ODS-related change (to be merged together with #8401 and #8340). The `blh` layout has been slightly modified to preserve the blob header size.

The code requires more changes to support bigger BLOBs properly, appropriate PR will be offered later.

### DIFF
--- a/src/dsql/ExprNodes.cpp
+++ b/src/dsql/ExprNodes.cpp
@@ -11079,7 +11079,7 @@ dsc* StrLenNode::execute(thread_db* tdbb, Request* request) const
 		switch (blrSubOp)
 		{
 			case blr_strlen_bit:
-				length = (FB_UINT64) blob->blb_length * 8;
+				length = blob->blb_length * 8;
 				break;
 
 			case blr_strlen_octet:
--- a/src/jrd/blb.h
+++ b/src/jrd/blb.h
@@ -76,7 +76,7 @@ public:
 	jrd_rel* blb_relation;			// Relation, if known
 	JBlob* blb_interface;
 
-	ULONG blb_length;				// Total length of data sans segments
+	FB_UINT64 blb_length;			// Blob's total length (in bytes)
 	USHORT blb_flags;				// Interesting stuff (see below)
 
 	SSHORT blb_sub_type;			// Blob's declared sub-type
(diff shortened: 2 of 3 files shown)


dyemanov:
Answering Vlad's [comment](https://github.com/FirebirdSQL/firebird/commit/bcafa978b99e3b56acfae98dd667f32e95e3b514#commitcomment-152073165) inside the commit:

> I highly doubt this is required.
> IMHO, we should document max blob length as 2 or 4GB and add corresponding checks in code.
> This must be properly discussed first.

IMHO, we must store and retrieve blobs of arbitrary size and given that we have them officially (as documented) allowed up to 128GB (IIRC) I see no good excuse to just shrink the documented limit, provided that it costs us nearly nothing.

That said, I'd seriously question the need to **process** (inside the engine) blobs bigger than 2/4GB. Currently we have a number of functions that read the whole blob contents into memory for processing and it may easily lead to OOM condition even for 4GB blobs. So I'd agree to add some max length for blobs being processed and throw an error for longer blobs. Just not sure whether it should happen everywhere or only in specific places that cannot process blobs chunk-by-chunk.

dyemanov:
All, may we have more opinions on this issue, please? Should we support blobs over 2-4GB everywhere, only in storage or disallow at all?

sim1984:
I don't see any reason to prohibit storing blobs larger than 2-4GB. Of course, it's not desirable from a common sense point of view, but since it was allowed before, are you sure that there are no databases where users already store such BLOBs?

As for the functions for working with BLOBs, where it is possible, it is desirable to provide the ability to work with large BLOBs. Where this is difficult and requires loading such BLOBs into memory entirely, you can throw an exception when a certain reasonable limit is exceeded.

AlexPeshkoff:
On 2/10/25 10:26, Dmitry Yemanov wrote:
>
> All, may we have more opinions on this issue, please? Should we 
> support blobs over 2-4GB everywhere, only in storage or disallow at all?
>

IMHO that's not critical feature for us and might such support require 
relatively serious efforts I'd say 'no'. But taking into an account that 
they do work at storage level, 32Tb HDD is today's reality and they will 
sooner of all grow on and on - why not?

PR #8433, opened 2025-02-09
Improve code of class BePlusTree 

hvlad:
Maked it usabe with debugger, removed a lot of casts and a few other misc improvemens.


aafemt:
Access to a member of unions that wasn't recently assigned is UB in C++. For reliability it would be better to remove `void* pointer` from `NodePtr` and make explicit separate constructors for node pointer and item pointer.

src/common/classes/tree.h
```diff
@@ -1007,10 +1020,10 @@ void BePlusTree<Value, Key, Allocator, KeyOfValue, Cmp>::_removePage(const int n
 	}
 
 	if (nodeLevel)
-		((NodeList *)node)->~NodeList();
+		node.nodes->~NodeList();
```
aafemt, line 1023: `delete(pool) node.nodes` and `delete(pool) node.items` here and above, please.

src/common/classes/tree.h
```diff
@@ -106,24 +93,24 @@ enum LocType { locEqual, locLess, locGreat, locGreatEqual, locLessEqual };
 // an indexed dynamic array without increase of algorithm calculation costs (this is one
 // more classical B+ tree feature). This is also not done to improve tree performance a little
 //
-template <typename Value, typename Key = Value, typename Allocator = MallocAllocator,
+template <typename Value, typename Key = Value, typename Allocator = Firebird::MemoryPool,
```
AlexPeshkoff, line 96: Why not remove "Allocator" parameter at all?

src/common/classes/tree.h
```diff
@@ -106,24 +93,24 @@ enum LocType { locEqual, locLess, locGreat, locGreatEqual, locLessEqual };
...
 class BePlusTree
 {
-	static const FB_SIZE_T LeafCount = LEAF_PAGE_SIZE / sizeof(Value);
-	static const FB_SIZE_T NodeCount = NODE_PAGE_SIZE / sizeof(void*);
+	static const FB_SIZE_T LEAF_COUNT = LEAF_PAGE_SIZE / sizeof(Value);
+	static const FB_SIZE_T NODE_COUNT = NODE_PAGE_SIZE / sizeof(void*);
 public:
 	explicit BePlusTree(Allocator *_pool)
-		: pool(_pool), level(0), root(NULL), defaultAccessor(this)
+		: pool(_pool), level(0), defaultAccessor(this)
```
AlexPeshkoff, line 105: Also suggest to use reference (not pointer) to MemoryPool in all c-tors - like in the rest of our classes.

AlexPeshkoff (changes requested):

src/common/classes/tree.h
hvlad: Good idea

src/common/classes/tree.h
hvlad: Also OK

src/common/classes/tree.h
hvlad: Done

src/common/classes/tree.h
hvlad: Done

AlexPeshkoff (approved):

dyemanov (approved):

hvlad:
Should it be backported and, if yes, into which branches ?

dyemanov:
Usually, we don't backport refactoring commits. If you think it's really needed, I may agree with backporting into v5 only.

hvlad:
The main goal of this PR is not the refactoring. It is about debugging experience, which was close to impossible.

dyemanov:
OK, I don't mind (except maybe v3 which is nearly end-of-life). I just suppose that the need of debugging there may be over-estimated, given the last bugfix in this class was about 20 years ago. And even if we have something fixed in master/v5, it can be easily backported "as is".

hvlad:
Let me clarify - it is not about debugging `BePlusTree` only, it is mostly about debugging everything else where it is used.
I.e. before this patch it was hard to impossible to inspect tree or map contents.

AlexPeshkoff:
@hvlad +1

PR #8434, opened 2025-02-11
Fix implicitly used collations after ALTER DOMAIN

### LINKED ISSUE #1627: "alter domain" supporting collation change [CORE1202]
Submitted by: Jorge Andres Brugger (jbrugger)

Relate to [CORE6324](https://github.com/FirebirdSQL/firebird/issues?q=CORE6324+in%3Atitle)

Votes: 12

Related to [CORE1058](https://github.com/FirebirdSQL/firebird/issues?q=CORE1058+in%3Atitle): "ALTER DOMAIN and ALTER TABLE don't allow to change character set and/or collation"

CREATE DOMAIN D\_TEST AS
VARCHAR\(10\) CHARACTER SET ISO8859\_1
COLLATE ES\_ES

You can do:
alter DOMAIN d\_test type
VARCHAR\(10\) CHARACTER SET iso8859\_1

But you can't do:
alter DOMAIN d\_test type
VARCHAR\(10\) CHARACTER SET iso8859\_1
COLLATE ES\_ES

Adding collate support for alter domain would be useful

dyemanov:
This is a postfix for PR #7748. Before this PR, if the field's collation was derived from some domain, RDB$COLLATION_ID was not filled inside RDB$RELATION_FIELDS and collation was inherited from the domain definition. Now it's broken.

Test case:

```sql
CREATE DOMAIN d1 VARCHAR(30) CHARACTER SET UTF8 COLLATE UNICODE;
CREATE TABLE test(data1 d1);
COMMIT;
INSERT INTO test VALUES('A');
INSERT INTO test VALUES('a');
COMMIT;

ALTER DOMAIN d1 TYPE VARCHAR(30) CHARACTER SET UTF8 COLLATE UNICODE_CI;
COMMIT;

SELECT data1 FROM test WHERE data1 = 'a';
```

```
DATA1                          
============================== 
a
```

Expected:

```
DATA1                          
============================== 
A                              
a
```

### DIFF
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -6541,7 +6541,8 @@ void RelationNode::defineField(thread_db* tdbb, DsqlCompilerScratch* dsqlScratch
 		}
 
 		fieldDefinition.defaultValue = defaultValue;
-		fieldDefinition.collationId = field->collationId;
+		if (field->typeOfName.isEmpty())
+			fieldDefinition.collationId = field->collationId;
 		fieldDefinition.store(tdbb, transaction);
 
 		// Define the field constraints.
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -6480,11 +6480,10 @@ void RelationNode::defineField(thread_db* tdbb, DsqlCompilerScratch* dsqlScratch
 			storeGlobalField(tdbb, transaction, fieldDefinition.fieldSource, field,
 				computedSource, computedValue);
 		}
-		else
+		else if (field->collate.hasData())
 		{
 			// Resolve possible additional collation for domains. For plain types it is already resolved above.
-			if (field->collate.hasData())
-				DDL_resolve_intl_type(dsqlScratch, field, field->collate);
+			DDL_resolve_intl_type(dsqlScratch, field, field->collate);
 		}
 
 		if ((relation->rel_flags & REL_external) &&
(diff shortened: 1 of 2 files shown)


aafemt:
IIRC this change can break the case when field is defined with domain and explicit collation. Most likely both conditions must be checked here.

dyemanov:
> IIRC this change can break the case when field is defined with domain and explicit collation. Most likely both conditions must be checked here.

True. But this seems to be against the SQL standard:

> 11.4 "column definition"
> 12) Case:
> a) If "column definition" immediately contains "domain name", then it shall not also immediately contain "collate clause".

So perhaps this should be disallowed rather than fixed.

aafemt:
I see no rationale behind this piece of standard bu the decision is on you.

BTW, changing of collation requires index rebuild. If you change collation of domain, will every index on every domain-based filed rebuilt automatically?

dyemanov:
IIRC, yes, indices are rebuilt.

dyemanov:
> I see no rationale behind this piece of standard but the decision is on you.

Personally, I don't see much value in this feature, but it seems it was supported before your PR, so this becomes a backward compatibility issue.

aafemt:
In this case IMHO it must be a rule at the parser level that COLLATE clause cannot be used with domains.

dyemanov:
As such a change wasn't suggested in the original discussion (although mentioned as deviation from the standard), I will not change it for the time being.

pavel-zotov:
Have questions related to some single-byte charsets.
Sent letter to dimitr, 11-feb-2025 21:35. Waiting for reply.

Issue #8291, opened 2024-10-24
NULLs should be skipped during index navigation when there's no lower bound and matched conditions are known to ignore NULLs


dyemanov:
Since v2.0 we have an optimization that skips NULL keys in the beginning of the ASC index if the retrieval is known to ignore NULL values, i.e. it does not involve IS [NOT] NULL and IS [NOT] DISTINCT FROM conditions matched with an index. However, it was disabled for index navigation (i.e. ORDER BY mapped to an index). This is correct for a full index scan (no conditions matched), but it should be enabled if the navigational retrieval has also upper bound key(s) defined and these matches are known to treat NULLs as FALSE. Currently it returns every NULL key found inside the index to be evaluated at the record level and then ignored, just wasting CPU cycles.

Example:

```sql
create table t (id int);

insert into t
select case when mod(val, 2) = 0 then null else val end from (
  select row_number() over() as val
  from rdb$types, rdb$types, rdb$types
  rows 1000000
);

commit;
create index it on t(id);

set stat on;
set explain on;

SELECT count(*) FROM T WHERE ID < 3;

Select Expression
    -> Aggregate
        -> Filter
            -> Table "T" Access By ID
                -> Bitmap
                    -> Index "IT" Range Scan (upper bound: 1/1)

                COUNT 
===================== 
                    1 

-- Fetches = 4

SELECT count(*) from (SELECT ID FROM T WHERE ID < 3 ORDER BY ID);

Select Expression
    -> Aggregate
        -> Filter
            -> Table "T" Access By ID
                -> Index "IT" Range Scan (upper bound: 1/1)

                COUNT 
===================== 
                    1 

-- Fetches = 1007094 (1775 when NULLs are skipped)
```

dyemanov:
The current NULLs skipping code is still sub-optimal, number of page fetches shows that (still more than expected). But this is to be addressed separately and only in the master branch.

PR #8446, opened 2025-02-22
Ignore NULLs (if desired) while scanning keys during index navigation

dyemanov:
See also #8291. This improvement completes the solution by extending the "ignore null" checks to the scan phase. The logic mostly matches `btr.cpp` (`BTR_evaluate()` and `scan()`).

With the same test case, results are:

before this PR:

```sql
SELECT count(*) from (SELECT ID FROM T WHERE ID < 3 ORDER BY ID);

Select Expression
    -> Aggregate
        -> Filter
            -> Table "T" Access By ID
                -> Index "IT" Range Scan (upper bound: 1/1)

                COUNT 
===================== 
                    1 

-- Fetches = 1775
```

after this PR:

```sql
-- Fetches = 7
```

### DIFF
--- a/src/jrd/recsrc/IndexTableScan.cpp
+++ b/src/jrd/recsrc/IndexTableScan.cpp
@@ -198,6 +198,7 @@ bool IndexTableScan::internalGetRecord(thread_db* tdbb) const
 	}
 
 	index_desc* const idx = (index_desc*) ((SCHAR*) impure + m_offset);
+	const bool descending = (idx->idx_flags & idx_descending);
 
 	// find the last fetched position from the index
 	const USHORT pageSpaceID = m_relation->getPages(tdbb)->rel_pg_space_id;
@@ -205,6 +206,8 @@ bool IndexTableScan::internalGetRecord(thread_db* tdbb) const
 
 	const IndexRetrieval* const retrieval = m_index->retrieval;
 	const USHORT flags = retrieval->irb_generic & (irb_descending | irb_partial | irb_starting);
+	const bool ignoreNulls =
+		(retrieval->irb_generic & irb_ignore_null_value_key) && (idx->idx_count == 1);
 
 	do
 	{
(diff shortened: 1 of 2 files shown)


src/jrd/recsrc/IndexTableScan.cpp
```diff
@@ -252,6 +255,14 @@ bool IndexTableScan::internalGetRecord(thread_db* tdbb) const
 				continue;
 			}
 
+			// If we're walking in a descending index and we need to ignore NULLs
+			// then stop at the first NULL we see (only for single segment!)
+			if (descending && ignoreNulls && node.prefix == 0 &&
+				node.length >= 1 && node.data[0] == 255)
```
hvlad, line 261: Accordingly to the `compress()`: 
// Further a NULL state is always returned as 1 byte 0xFF (descending index).
So, why check for `>= 1`, not `== 1` ? 
And, for consistency, please use 0xFF, not 255.

hvlad (approved):

src/jrd/recsrc/IndexTableScan.cpp
dyemanov: This is a copy-paste from btr.cpp :) where such a check is used twice. I don't mind changing it, but I'd suggest this to be done in btr.cpp too.

dyemanov: I'm also wondering whether we really should check for `(node.prefix == 0 && node.length == 1 && node.data[0] == 0xFF)` -- i.e. the first NULL encountered -- or better for `(node.prefix + node.length == 1 && node.data[0] == 0xFF)` -- i.e. any NULL encountered, as more reliable. Is it theoretically possible that we somehow jump to the bucket in the middle of the NULL duplicates chain and start scanning from there, thus skipping the first NULL node?

hvlad: > I'm also wondering whether we really should check for 
`(node.prefix == 0 && node.length == 1 && node.data[0] == 0xFF)` -- i.e. the first NULL encountered -- or better 
for 
`(node.prefix + node.length == 1 && node.data[0] == 0xFF)` -- i.e. any NULL encountered, as more reliable. 

Second way is not correct: when `node.length == 0` we should not access `node.data[0]` at all.
Instead, we could look at `key` contents, if necessary.

> Is it theoretically possible that we somehow jump to the bucket in the middle of the NULL duplicates chain and start scanning from there, thus skipping the first NULL node?

The position is based on the last key processed, so, we should be safe.

hvlad: > This is a copy-paste from btr.cpp :) where such a check is used twice. I don't mind changing it, but I'd suggest this to be done in btr.cpp too.

Ok, then, let it be this way

pavel-zotov:
QA note: ticket issue is covered by test for [8291](https://github.com/FirebirdSQL/firebird-qa/blob/master/tests/bugs/gh_8291_test.py) (it was enough to reduce MAX_ALLOWED_IDX_READS threshold, see notes there).

PR #8450, opened 2025-02-26
Slightly better checking for a valid database file

dyemanov:
1) Validate a few more header fields before the ODS version check
2) Check for the too small file size and report a proper error instead of I/O error


pavel-zotov:
::: QA note :::
Have an issue about FB crashes when invalid DB is a file with size from 1024 to 8191 bytes.
Sent example to dimitr, waiting for reply.

PR #8453, opened 2025-02-26
Fixed #8444: GBAK: GDS error batch_big_seg2 when restoring a table with many BLOBs

### LINKED ISSUE #8444: GBAK: GDS error batch_big_seg2 when restoring a table with many BLOBs
I'm trying to restore a FBK created with Firebird 2.5 using GBAK into a Firebird 5.0.1 Server.
During the restore operation GBAK just segfaults when restoring BLOBs for a specific table.
That table has several BLOB fields (total BLOB size for each row about 400kB).

I compiled gbak myself to investigate the error.
It turns out the server sends an error which causes an exception and getCleanBuffer inside RestoreRelationTask::checkSpace (restore.cpp) returns a NULL pointer.
But that is just a consequence of the actual error from the server, which is:

```
-901
335545196 
batch_big_seg2
"Size of segment (@1) greater than available data (@2)"
@1 = 16384
@2 = 4150
```
Investigating further, I checked the function "get_blob" where it checks if there's enough room for the BLOB to be inlined or not.
When I just set blobInline to false to keep it from inlining BLOBs, the restore succeeds without a problem. 
So I assume there's something off with the calculation which checks if there's enough room in the batch buffer for the BLOB.
It seems it checks the blob size limit (m_batchInlineBlobLimit) for the current BLOB only and does not take into account other BLOBs from the same record which causes the buffer to grow too big and gets truncated, resulting in the GDS error?
I.e. the sum of all BLOB sizes where the size is smaller than m_batchInlineBlobLimit exceeds m_batchInlineBlobLimit.

Inside WriteRelationMeta::createBatch, the values retrieved from the server are:
```
IBatch::INF_BUFFER_BYTES_SIZE: 268435456
IBatch::INF_BLOB_ALIGNMENT: 4
IBatch::INF_BLOB_HEADER: 16
```

Which gives the following other values:
```
m_batchInlineBlobLimit: 268415
m_batchStep: 1000
```

When using the "-parallel" switch with e.g. a value of 8 the restore works as well. I guess this is just a coincidence because the batches fill up differently to not cause this error.

Sorry for the long text, I tried to be as detailed as I can.

AlexPeshkoff:


src/remote/client/interface.cpp
```diff
@@ -9313,6 +9313,16 @@ static void receive_packet_noqueue(rem_port* port, PACKET* packet)
...
+
+			case op_batch_blob_stream:
+				stmt_id = p->packet.p_batch_blob.p_batch_statement;
+				bCheckResponse = true;
+				break;
+
+			case op_batch_regblob:
+				stmt_id = p->packet.p_batch_regblob.p_batch_statement;
+				bCheckResponse = true;
+				break;
```
hvlad, line 9325: Looks like all `op_batch_XXX` labels have the same handler, why not unite them ?

hvlad: Now it is clear, thanks

hvlad (approved):

PR #8469, opened 2025-03-12
fix(cast-format): Throw exception for non-existing date in string to datetime conversion

TreeHunter9:
- Also add a small unit test to verify exception throwing for non-existent date


TreeHunter9:
Problem was found by @aafemt.

PR #8472, opened 2025-03-17
Ability to set up the source GUID for the replication stream

dyemanov:
This allows GUID-specific code being supported to the replica side. Currently only a check against configuration is performed.


dyemanov:
Not sure the API method name is really good, maybe `setSource()` or `reset()` would suit better.

aafemt:
For the beginning it would be good to document what is the purpose of this method and the meaning of parameter(s), how it is intended to be used and in which situations. Just to determine if this method really needed or there is a better alternative(s).

PR #8473, opened 2025-03-17
Fix #8452 - Disallow creation of object names with only spaces.

### LINKED ISSUE #8452: Disallow creation of object names with only spaces
This is currently valid:
```
create table " " (n integer)
```

Considering that empty names are not allowed and trailing spaces are discharged, names with only spaces should be considered as empty.

asfernandes:


pavel-zotov:
From #8452 :
```
 > It seems to me that we should prohibit not only names consisting of only
 >  spaces, but also the use of special characters in names, such as line
 >  feeds. For example, it is now completely legal to create a table with
 >  such a name, but it does not make sense.

 >  |CREATE TABLE "T " (ID INT); |

Yes, that is part of std requirement:

A <non-double quote character> is any character of the source language
character set other than a <double quote> that is not included in the
Unicode General Categories “Cc”, “Cf”, “Cn”, “Cs”, “Zl”, or “Zp”.
```

It seems that currently we CAN create table with totally weird names that include lot of characters from above mentioned Unicode General Categories.
Consider script and its log from attachment file (they both are in UTF8):
изображение

All of such tables except first *can* be created.
[test-for-names-from-disallowed-unicode-categories.zip](https://github.com/user-attachments/files/19351079/test-for-names-from-disallowed-unicode-categories.zip)

asfernandes:
> It seems that currently we CAN create table with totally weird names that include lot of characters from above mentioned Unicode General Categories.

While I recognized the mentioned problem, I fixed exactly what is mentioned in the ticket subject ("Disallow creation of object names with only spaces"), as it will cause big problems with the schemas implementation.

We can create another ticket, but it will have lower priority at least from me.

Issue #6912, opened 2021-07-30
Protocol.h is not in distribution packages


aafemt:
Replication protocol header is supposed to be public but is not included into distributed include/firebird directory.

PR #8489, opened 2025-03-27
Replace internal 'dsc' with public 'paramdsc' in the API

dyemanov:
So far they're binary compatible and this breaks nothing. But it prevents problems when `dsc` is to be extended.


aafemt:
The problem is that because of backward compatibility of public API dsc won't be able to be extended at all.

src/jrd/trace/TraceObjects.cpp
```diff
@@ -83,7 +83,7 @@ bool descToUTF8(const dsc* param, string& result)
 		return false;
 	}
 
-	if (param->getCharSet() == CS_BINARY)
+	if (param->dsc_sub_type == CS_BINARY)
```
mrotteveel, line 86: This change, and others like it, seems to indicate this change is **not** a compatible change.

aafemt: It is, but public `dsc` is a POD-type so used getters are replaced with direct access to members.

dyemanov: It's compatible as long as only string data types are processed. And given that `descToUTF8()` returns `false` for all non-string data types before entering the code you mention, I see it as safe.

mrotteveel: Yes, but it does mean that if people used this, they will need to change their code more than just replacing references to `dsc` with `paramdsc`.

aafemt: `dsc.h` header never was a part of distribution so officially people had no way to use it.

mrotteveel: So, I'm not talking specifically about this code, but about the fact that - judging by this code - calling `param->getCharSet()`, `param->setNull()`, `param->getTextType()` (and possibly other methods) is no longer possible as those are defined for `dsc` and (I guess) not for `paramdsc`.

aafemt: It is hard to believe, but it was never possible. This part of public API really wasn't useable without some hacks.

PS: It was a bad surprise during development of my own trace plugin.
PPS: `Firebird::IReplicator` API has the same problem because of #6912.

mrotteveel: OK, then I guess that is no issue and doesn't need to be documented in the release notes.

PR #8494, opened 2025-04-01
Fix issues with longish concatenated context aliases

dyemanov:
Currently nested contexts (e.g. derived tables) have context aliases concatenated through the whole stack, so the final string is not limited to the maximum metadata name length.

1) Fix the comparison order to avoid an alias being truncated while compared
2) Make `planEntry.alias` a string rather than `MetaName`
3) Truncate the concatenated alias when generating BLR

All these issues now assert: (1) and (2) for aliases longer than 252 bytes and (3) for aliases longer than 255 bytes.


dyemanov:
Test case is available upon request.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -54,6 +54,20 @@ static ValueExprNode* resolveUsingField(DsqlCompilerScratch* dsqlScratch, const
...
+	void appendContextAlias(DsqlCompilerScratch* dsqlScratch, const string& alias)
+	{
+		const auto len = alias.length();
+		if (len <= MAX_UCHAR)
+			dsqlScratch->appendMetaString(alias.c_str());
+		else
+		{
+			string truncatedAlias(alias);
+			truncatedAlias.resize(MAX_UCHAR - 3);
+			truncatedAlias += "...";
```
asfernandes, line 66: How this change of a name overflow to ellipsis will resolve problems?

asfernandes (commented):
The adaptation of plans to the schemas PR was really trick.
Maybe you can do the master changes based on it or in the same branch?

src/jrd/RecordSourceNodes.cpp
dyemanov: It avoids an assertion (debug build) or weird alias truncation (release build) inside `BlrWriter::appendString` when called from `genBlr()`. Of course, we could just truncate the string at the 255 bytes boundary, but IMHO ellipsis look better as clearly shows the fact of truncation. Now BLR is generated (and later parsed) properly and ellipsis is shown inside the very long plan's alias, AFAIS without any additional side effects.

dyemanov:
> The adaptation of plans to the schemas PR was really trick. Maybe you can do the master changes based on it or in the same branch?

No problem. Once approved, I will commit to v5 and delay the master PR until schemas are merged.

asfernandes (approved):

src/jrd/RecordSourceNodes.cpp
```diff
@@ -54,6 +54,20 @@ static ValueExprNode* resolveUsingField(DsqlCompilerScratch* dsqlScratch, const
 
 namespace
 {
+	void appendContextAlias(DsqlCompilerScratch* dsqlScratch, const string& alias)
+	{
+		const auto len = alias.length();
+		if (len <= MAX_UCHAR)
```
Noremos, line 60: Could it be more advantageous to use MAX_SQL_IDENTIFIER_LEN rather than MAX_UCHAR, considering the possibility of MetaName length expanding in the future?

Noremos: Got it, thanks for the clarification.

pavel-zotov:
::: QA note :::
Explained plan become truncated when number of derived tables >= 36.
Sent report (test+logs) to dimitr, 20.07.2025 12:41. Waiting for reply.

PR #8497, opened 2025-04-02
Fix windows warnings (C4554, C4018) and mark error message constants as constexpr (C26814)

Noremos:
This is a replacement for my previous PR: FirebirdSQL/firebird/pull/8360.
I apologize for having to reopen it. The code is the same except for corrections addressing the last two issues from the original PR.

### DIFF
--- a/src/include/firebird/iberror.h
+++ b/src/include/firebird/iberror.h
@@ -18,37 +18,37 @@
 #ifdef __cplusplus /* c++ definitions */
 
 #if __cplusplus >= 201703L // C++17 or later
-#define STATUS_CONST inline constexpr ISC_STATUS
+#define FB_STATUS_CONST inline constexpr ISC_STATUS
 #else
-#define STATUS_CONST const ISC_STATUS
+#define FB_STATUS_CONST const ISC_STATUS
 #endif
 
-STATUS_CONST isc_facility = 20;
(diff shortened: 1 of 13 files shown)


src/include/firebird/iberror.h
```diff
@@ -17,32 +17,38 @@
...
-const ISC_STATUS isc_arg_dos			= 9;	// MSDOS/OS2 error code
-const ISC_STATUS isc_arg_mpexl			= 10;	// HP MPE/XL error code
-const ISC_STATUS isc_arg_mpexl_ipc		= 11;	// HP MPE/XL IPC error code
-const ISC_STATUS isc_arg_next_mach		= 15;	// NeXT/Mach error code
-const ISC_STATUS isc_arg_netware		= 16;	// NetWare error code
-const ISC_STATUS isc_arg_win32			= 17;	// Win32 error code
-const ISC_STATUS isc_arg_warning		= 18;	// warning argument
-const ISC_STATUS isc_arg_sql_state		= 19;	// SQLSTATE
+#if __cplusplus >= 201703L // C++17 or later
+#define STATUS_CONST inline constexpr ISC_STATUS
```
asfernandes, line 21: This is public header, so I think some FB-specific prefix in the macro is good.
Also, it should be good to undefine it after usage.

PR #8500, opened 2025-04-02
Correct section name for services in fbtrace config

Noremos:


src/common/config/config_file.cpp
```diff
@@ -979,14 +979,14 @@ ConfigFile::SectionType ConfigFile::Parameter::parseSectionKey() const
 	{
 		return SectionType::DATABASE_REGEX;
 	}
-	else if (name == "service")
```
AlexPeshkoff, line 982: This change will break existing trace configs. I strongly disagree.

AlexPeshkoff (changes requested):

src/common/config/config_file.cpp
hvlad: When reviewing #8427 I missed that trace-specific routine was put into common class. 
It is `ConfigFile::Parameter::parseSectionKey()` and related `enum class SectionType`.
I think both must be moved out of `ConfigFile::Parameter` into `TraceCfgReader`.

AlexPeshkoff: > I think both must be moved out of `ConfigFile::Parameter` into `TraceCfgReader`.

+1

Noremos: > This change will break existing trace configs. I strongly disagree.

Existing configurations are using the key 'services'. It was my mistake in the original PR where I incorrectly set the key as 'service'."

Noremos: > When reviewing #8427 I missed that trace-specific routine was put into common class. It is `ConfigFile::Parameter::parseSectionKey()` and related `enum class SectionType`. I think both must be moved out of `ConfigFile::Parameter` into `TraceCfgReader`.

I have moved the method

hvlad (approved):

hvlad:
I don't think that PR with unresolved change request should be merged.
Also, there was no need in hurry in this case.

dyemanov:
Sorry, but I've seen your approvement and both change requests have been addressed.

PR #8501, opened 2025-04-03
fix(cast-format): Throw exception when value cannot be found for specific pattern in string to datetime conversion

TreeHunter9:
- Example: CAST('A.M. 8' as time format 'MI A.M. HH12') where value for MI pattern is missing;
- Handle integer overflow when converting string to int by using std::from_chars;
- Add tests for missing values and integer overflow;


dyemanov:
Looks like CvtTest.cpp fails on Alpine and Windows 2019, while works OK on other platforms. Could you please take a look? And please also resolve the conflicts.

TreeHunter9:
As I can see, all tests where we need to parse number are failing, but tests where we parse strings are fine. I guess the problem is `std::from_chars`, maybe I'm handling its result incorrectly somehow, I'll look into it.

pavel-zotov:
i have misunderstanding about these examples from [src/common/tests/CvtTest.cpp](https://github.com/FirebirdSQL/firebird/pull/8501/files#diff-4c41a1e71a7a12168ddce05a6e570ee08d64e36c981fca8e7274ebac4c41edb1)
```
        testExceptionCvtStringToFormatDateTime("00:60", "TZR", cb);
	testExceptionCvtStringToFormatDateTime("15:00", "TZH:TZM", cb);
	testExceptionCvtStringToFormatDateTime("15:00", "TZR", cb);
	testExceptionCvtStringToFormatDateTime("-15:00", "TZH:TZM", cb);
	testExceptionCvtStringToFormatDateTime("-15:00", "TZR", cb);
```
What kind of datatype and which string of FORMAT must be used for them ?
I have tried these:
```
select cast('00:60' as time with time zone format 'TZR') from rdb$database
select cast('15:00' as time format 'TZH:TZM') from rdb$database
```
-- but got the same error messages in 6.0.0.726 (intermediate snapshot from "Actions") and 6.0.0.725:
Value for TZR pattern is out of range [0, 59]
and
Cannot use "TZH" format with current date type

TreeHunter9:
> What kind of datatype and which string of FORMAT must be used for them ?

These tests only check that there will be an exception if the input time zone offset is out of range, except for the first one, it only checks bounds for minutes value. These formats can be used for any `... with time zone' types.

PR #8504, opened 2025-04-07
Simplify clumplet reader/writer

mrotteveel:
Simplify code surrounding tagged kind determination (and fixing some issues in ClumpletWriter not covering all untagged variants).

Also fixes some code quality warnings regarding switches.


mrotteveel:
I guess I need to go back to the drawing board of my change.

Issue #8505, opened 2025-04-07
fb_interpret returns "Unknown ISC error"


P1cheneG:
Hello, I am writing a C++ program using the Firebird API and encountered an 'Unknown ISC error' message instead of the correct error message. The problem can be reproduced with the following code.

```
#include <iostream>
#include <string>
#include <vector>
#include "firebird/Interface.h"
#include "firebird/IdlFbInterfaces.h"

using namespace Firebird;

const char* USER = "SYSDBA";
const char* PASSWORD = "1234";
constexpr int MAX_DIAG_MSG_LEN = 300;

static IMaster* master = fb_get_master_interface();
static IProvider* prov = master->getDispatcher();
static IStatus* st = master->getStatus();

class PrintStatus : public BaseStatusWrapper<PrintStatus>
{
public:
    PrintStatus(IStatus* aStatus) : BaseStatusWrapper(aStatus) {}

    static void checkException(PrintStatus* status)
    {
        if (status && status->isDirty() && (status->getState() & IStatus::STATE_ERRORS))
        {
            status->hasError();
        }
    }

    bool hasError() const
    {
        const bool hasErr = isDirty() && (status->getState() & IStatus::STATE_ERRORS);
        if (hasErr)
        {
            auto statusPtr = status->getErrors();
            int code = statusPtr[1];
            std::cout << "SQL CODE = " << code << std::endl;

            char buffer[MAX_DIAG_MSG_LEN];
            while (fb_interpret(buffer, MAX_DIAG_MSG_LEN, &statusPtr))
            {
                std::cout << buffer << std::endl;
            }
        }
        return hasErr;
    }
};

int main()
{
    IAttachment* att = nullptr;
    PrintStatus status(st);

    IUtil* utl = master->getUtilInterface();
    IXpbBuilder* dpb = utl->getXpbBuilder(&status, IXpbBuilder::DPB, NULL, 0);
    dpb->insertString(&status, isc_dpb_user_name, USER);
    dpb->insertString(&status, isc_dpb_password, PASSWORD);

    std::string connectStr = "localhost:employee";
    att = prov->attachDatabase(
        &status,
        connectStr.c_str(),
        dpb->getBufferLength(&status),
        dpb->getBuffer(&status));

    if (status.hasError()) {
        dpb->dispose();
        return 1;
    }

    IXpbBuilder* tpb = utl->getXpbBuilder(&status, IXpbBuilder::TPB, NULL, 0);
[...]

PR #8506, opened 2025-04-07
Add missing implementation for SYMBOL and NO_SYMBOL messages

P1cheneG:
This is a fix for https://github.com/FirebirdSQL/firebird/issues/8505. In the fix, I added the missing message to the exported one. Previously, the program displayed the error: `"unknown ISC error 336067776"`, Now the program outputs: `"Function TEST_FUNC already exists"`. The original problem was reproduced in Windows

### DIFF
--- a/builds/win32/make_boot.bat
+++ b/builds/win32/make_boot.bat
@@ -177,9 +177,9 @@ goto :EOF
 @echo Building ttmath (%FB_OBJ_DIR%)...
 @mkdir %FB_ROOT_PATH%\extern\ttmath\%FB_CONFIG% 2>nul
 if /I "%FB_CONFIG%"=="debug" (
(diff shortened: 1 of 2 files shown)


AlexPeshkoff:
Instead this changes you should simply have firebird.msg available.

P1cheneG:
Hello. Thank you for the answer, but the other messages are working fine. In my opinion, either all messages should require the firebird.msg file, or none of them should. There is an example of error output for creating a function and a syntax error:
```
SQL CODE = 335544351
unsuccessful metadata update
CREATE FUNCTION TEST_FUNC failed
unknown ISC error 336068876
SQL CODE = 335544569
Dynamic SQL Error
SQL error code = -104
Token unknown - line 1, column 1
creat
```
Using this code:
```
att->execute(&status, tra, 0,
    "create function TEST_FUNC RETURNS float as BEGIN RETURN 1; END;",
    3, NULL, NULL, NULL, NULL);
att->execute(&status, tra, 0,
    "creat function TEST_FUNC RETURNS float as BEGIN RETURN 1; END;",
    3, NULL, NULL, NULL, NULL);
```

AlexPeshkoff:
No idea why some messages were put directly to client library. That's historically so since interbase times.

PR #8519, opened 2025-04-14
Adding constexpr/const/noexcept where relevant

mrotteveel:
Also replaces a memset on a struct with empty list-initializers on its fields, and replaces some define struct followed by initializing fields with using a list-initializer.

### DIFF
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -1519,7 +1519,7 @@ static rem_port* aux_connect(rem_port* port, PACKET* packet)
 		if (port->port_channel == INVALID_SOCKET)
 			return NULL;
 
-		SOCKET n = os_utils::accept(port->port_channel, NULL, NULL);
+		const SOCKET n = os_utils::accept(port->port_channel, NULL, NULL);
 		inetErrNo = INET_ERRNO;
 
 		if (n == INVALID_SOCKET)
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -79,7 +79,7 @@
 
 
 #ifdef	WIN_NT
-constexpr int FD_SETSIZE = 2048;
+#define FD_SETSIZE 2048
 #endif
 
 #ifndef WIN_NT
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -1543,7 +1543,7 @@ static rem_port* aux_connect(rem_port* port, PACKET* packet)
 	port->port_async = new_port;
 	new_port->port_dummy_packet_interval = port->port_dummy_packet_interval;
 	new_port->port_dummy_timeout = new_port->port_dummy_packet_interval;
-	const P_RESP* response = &packet->p_resp;
+	P_RESP* response = &packet->p_resp;
 
 	// NJK - Determine address and port to use.
 	//
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -955,7 +955,7 @@ rem_port* INET_connect(const TEXT* name,
 	// Prepare hints
 	const bool ipv6 = os_utils::isIPv6supported();
 
-	struct addrinfo gai_hints {};
+	addrinfo gai_hints {};
 	if (packet)
 		gai_hints.ai_family = af;
 	else
@@ -1107,7 +1107,7 @@ static rem_port* listener_socket(rem_port* port, USHORT flag, const addrinfo* pa
 
 	if (flag & SRVR_multi_client)
 	{
-		const struct linger lingerInfo {};
+		const linger lingerInfo {};
 
 		// Get any values for SO_LINGER so that they can be reset during
 		// disconnect.  SO_LINGER should be set by default on the socket
@@ -1490,7 +1490,8 @@ static rem_port* aux_connect(rem_port* port, PACKET* packet)
 
 	if (port->port_server_flags)
 	{
-		struct timeval timeout { port->port_connect_timeout, 0 };
+		timeval timeout {};
+		timeout.tv_sec = port->port_connect_timeout;
 
 		Select slct;
 		slct.set(port->port_channel);
@@ -2300,7 +2301,7 @@ static bool select_wait( rem_port* main_port, Select* selct)
 						// in current fdset. Search and return it to caller to close
 						// broken connection correctly
 
-						struct linger lngr {};
+						linger lngr {};
 						socklen_t optlen = sizeof(lngr);
 #ifdef WIN_NT
 						constexpr bool badSocket = false;
@@ -2359,7 +2360,8 @@ static bool select_wait( rem_port* main_port, Select* selct)
 
 			// Some platforms change the timeout in the select call.
 			// Reset timeout for each iteration to avoid problems.
-			struct timeval timeout { SELECT_TIMEOUT, 0 };
+			timeval timeout {};
+			timeout.tv_sec = SELECT_TIMEOUT;
 			selct->select(&timeout);
 			const int inetErrNo = INET_ERRNO;
 
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -79,7 +79,7 @@
 
 
 #ifdef	WIN_NT
-#define FD_SETSIZE 2048
+constexpr int FD_SETSIZE = 2048;
 #endif
 
 #ifndef WIN_NT
@@ -97,7 +97,7 @@
 
 #endif // !WIN_NT
 
-const int INET_RETRY_CALL = 5;
+constexpr int INET_RETRY_CALL = 5;
 
 #include "../remote/remote.h"
 #include "../remote/SockAddr.h"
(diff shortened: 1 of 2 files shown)


src/remote/inet.cpp
```diff
@@ -1097,7 +1096,7 @@ static rem_port* listener_socket(rem_port* port, USHORT flag, const addrinfo* pa
 	//			e.g. while it's listening. This is surely not what we want.
 	//			We set this options for any kind of listener, including standalone Classic.
 
-	int optval = TRUE;
+	constexpr int optval = TRUE;
```
aafemt, line 1099: Are you sure that getting address of `constexpr` is a defined behavior? Cast below perhaps raise warning about qualifier being dropped.

src/remote/remote.h
```diff
@@ -348,16 +348,11 @@ struct Rtr : public Firebird::GlobalStorage, public TypedHandle<rem_type_rtr>
...
-
-	RBlobInfo()
-	{
-		memset(this, 0, sizeof(*this));
-	}
+	bool	valid{};
+	UCHAR	blob_type{};
+	ULONG	num_segments{};
+	ULONG	max_segment{};
+	FB_UINT64	total_length{};
```
aafemt, line 355: These are simple types I would prefer `num_segment = 0` syntax because of better readability and explicit blob type constant for the same sake.

src/remote/inet.cpp
mrotteveel: And that is, BTW, because the parameter to setopt is defined as a pointer to const.

src/remote/remote.h
mrotteveel: C++ Core Guidelines: [ES.23: Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-list). I could replace it with `{false}`, `{0}` but as that is already the default value when using `{}`, that seems unnecessary to me.

aafemt: Not every guideline worth following. In this case narrowing surely cannot happen. And it is better not to make readers guess which blob type is default.

mrotteveel: Point taken for blob_type, but for the rest, I think this is explicit enough if you're familiar with the initializer syntax.

src/remote/inet.cpp
TreeHunter9: > Are you sure that getting address of constexpr is a defined behavior?

Yes, it's basically the same as a const variable, but its value should be calculated at compile time.

src/remote/inet.cpp
```diff
@@ -1524,12 +1519,12 @@ static rem_port* aux_connect(rem_port* port, PACKET* packet)
 		if (port->port_channel == INVALID_SOCKET)
 			return NULL;
 
-		const SOCKET n = os_utils::accept(port->port_channel, NULL, NULL);
+		SOCKET n = os_utils::accept(port->port_channel, NULL, NULL);
```
TreeHunter9, line 1522: Why was `const` removed there?

src/remote/remote.h
asfernandes: I'm also against these `{}`. And the team never discussed about following C++ Core Guidelines.

mrotteveel: OK, what do you want then? Explicit list-initialization (`{ false }`/`{ 0 }`), or copy-initialization (`= false`/`= 0`). I did it primarily to get rid of that memset ([C.90: Rely on constructors and assignment operators, not memset and memcpy](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-memset)).

And re C++ Core Guidelines, given list-initializers (or initialization in general) are not mentioned in https://geosoft.no/development/cppstyle.html (which is referenced from https://www.firebirdsql.org/en/coding-style), I referenced a more complete style guide, maintained by people who designed C++ to get an impression of what desirable style is. I'm not saying it is the bible, or that we must follow it, I'm using it to explain _why_ I did something

AlexPeshkoff:
On 4/14/25 14:04, Mark Rotteveel wrote:
>
>
>
> In src/remote/remote.h 
> <https://github.com/FirebirdSQL/firebird/pull/8519#discussion_r2041917454>:
>
> > +	bool	valid{};
> +	UCHAR	blob_type{};
> +	ULONG	num_segments{};
> +	ULONG	max_segment{};
> +	FB_UINT64	total_length{};
>
> OK, what do you want then? Explicit list-initialization (|{ false 
> }|/|{ 0 }|), or copy-initialization (|= false|/|= 0|). I did it 
> primarily to get rid of that memset (C.90: Rely on constructors and 
> assignment operators, not memset and memcpy 
> <https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-memset>).
>
> And re C++ Core Guidelines, given list-initializers (or initialization 
> in general) are not mentioned in 
> https://geosoft.no/development/cppstyle.html (which is referenced from 
> https://www.firebirdsql.org/en/coding-style), I referenced a more 
> complete style guide, maintained by people who designed C++ to get an 
> impression of what desirable style is. I'm not saying it is the bible, 
> or that we must follow it, I'm using it to explain /why/ I did something
>

In most of new modules we typically use the following style:

struct RBlobInfo
{
     bool    valid = false;
     UCHAR   blob_type = 0;
     ULONG   num_segments = 0;
     ULONG   max_segment = 0;
     FB_UINT64   total_length = 0;

BTW, memset here is absolutely safe - I see no reason to waste time for 
such changes.

src/remote/remote.h
asfernandes: `name = <value>` is fine for me.

hvlad: > `name = <value>` is fine for me.

+1

hvlad:
> BTW, memset here is absolutely safe - I see no reason to waste time for such changes.

+1

src/remote/inet.cpp
```diff
@@ -2365,9 +2359,7 @@ static bool select_wait( rem_port* main_port, Select* selct)
 
 			// Some platforms change the timeout in the select call.
 			// Reset timeout for each iteration to avoid problems.
-			timeout.tv_sec = SELECT_TIMEOUT;
-			timeout.tv_usec = 0;
-
+			struct timeval timeout { SELECT_TIMEOUT, 0 };
```
asfernandes, line 2362: IMO it's not good to be order-dependend on field here.

src/remote/inet.cpp
```diff
@@ -955,8 +955,7 @@ rem_port* INET_connect(const TEXT* name,
 	// Prepare hints
 	const bool ipv6 = os_utils::isIPv6supported();
 
-	struct addrinfo gai_hints;
-	memset(&gai_hints, 0, sizeof(gai_hints));
+	struct addrinfo gai_hints {};
```
asfernandes, line 958: These `struct` are required only in C.

mrotteveel: What do you mean with this? 

Using `struct addrinfo gai_hints {};` will zero-initialize, just like the original call to memset did, but just with C++ idioms.

asfernandes:
```suggestion
	addrinfo gai_hints {};
```

mrotteveel: Ah, ok.

mrotteveel: Do you want me to remove all occurrences of such constructs in inet.cpp?

src/remote/inet.cpp
asfernandes: 1) You can't know that thinking in multi-platform code
2) It maked code more difficult to read

src/remote/inet.cpp
asfernandes: Looks like a good thing to do at least in the lines/blocks you are changing.

mrotteveel: Done (for the lines I already touched).

src/remote/inet.cpp
mrotteveel: And done

AlexPeshkoff:
On 4/15/25 13:17, Adriano dos Santos Fernandes wrote:
>
>
>
> In src/remote/inet.cpp 
> <https://github.com/FirebirdSQL/firebird/pull/8519#discussion_r2044187292>:
>
> > @@ -2365,9 +2359,7 @@ static bool select_wait( rem_port* main_port, Select* selct)
>   
>   			// Some platforms change the timeout in the select call.
>   			// Reset timeout for each iteration to avoid problems.
> -			timeout.tv_sec = SELECT_TIMEOUT;
> -			timeout.tv_usec = 0;
> -
> +			struct timeval timeout { SELECT_TIMEOUT, 0 };
>
>  1. You can't know that thinking in multi-platform code
>  2. It maked code more difficult to read
>

+1

aafemt:
Someday C++20 is allowed and [designated initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers) will be available.

mrotteveel:
Any objections if I merge?

asfernandes (approved):

PR #8528, opened 2025-04-17
Replace p_batch_free_cancel with p_rlse

mrotteveel:


src/remote/protocol.cpp
```diff
@@ -1084,8 +1084,8 @@ bool_t xdr_protocol(RemoteXdr* xdrs, PACKET* p)
 	case op_batch_rls:
 	case op_batch_cancel:
 		{
-			P_BATCH_FREE_CANCEL* b = &p->p_batch_free_cancel;
-			MAP(xdr_short, reinterpret_cast<SSHORT&>(b->p_batch_statement));
+			P_RLSE* b = &p->p_rlse;
```
mrotteveel, line 1087: @AlexPeshkoff This switch branch could be merged with the generic `p_rlse` branch if it weren't for the debugging print done

```
if (xdrs->x_op != XDR_FREE)
	DEB_RBATCH(fprintf(stderr, "BatRem: xdr release/cancel %d\n", p->p_operation));
```

Would it be OK with the general branch (line 547)?

mrotteveel: That should have been: Would it be OK to merge with the general branch (line 547)?

AlexPeshkoff: I suggest to add this lines to master under
#ifdef NEVERDEF
into generic protocol for release package.

If sometimes I need to use DEB_RBATCH I will find a way to make it be usable

mrotteveel: OK, committed the change. Was that what you meant?

AlexPeshkoff (approved):

PR #8534, opened 2025-04-24
Update ICU to version 77.1 and VS BuildTools to 17.13.6

asfernandes:


asfernandes:
My local built binaries are working, but I can't make the Windows x86/x64 binaries artifacts produced here in this PR to work.
Virus warnings and even with AV turned off, things does not work.
Attaching the debugger to the downloaded isql.exe, I see a special ICU mutex being used before initialization.

asfernandes:
It started to work after update VS BuildTools used in the build to the same version I used to build ICU, so I'm maintaining the change in this PR.

hvlad (approved):
My working environment uses a bit outdated VS 17.12.5
I built binaries from code of this branch (with ICU provided in this branch) - works OK.
Also, I've run binaries from builds artifact (VS 2022) - also works OK.
Thus, I found no problem at Windows side.

mrotteveel:
Running a build (on Windows) adds two files that probably should be ignored (I guess?):

- extern/icu/LICENSE
- extern/icu/icudt77l.dat

mrotteveel:
Or otherwise, those files need to be committed to, or is something missing in a cleanup somewhere?

mrotteveel:
> The data file is of course needed. I didn't changed the structure of this zip, so I believe a discussion should not happen inside this change.

This change introduced this behaviour, so it seems to me the logic place to discuss it. Where do you want to discuss it instead?

hvlad:
> Running a build (on Windows) adds two files that probably should be ignored (I guess?):
> 
>     * extern/icu/LICENSE
> 
>     * extern/icu/icudt77l.dat

I've adjusted extern/icu/.gitignore

mrotteveel:
@hvlad Thanks!

PR #8555, opened 2025-05-06
fix(config): Prevent printf injection in replication.conf

TreeHunter9:
The problem can be reproduced with this incorrect replication.conf:
```
database = <path/to/database>
{
	source_guid = 12%d%s%s%s3456
}
```
The server will crash on startup.

v5 has this bug too.

### DIFF
--- a/src/jrd/replication/Config.cpp
+++ b/src/jrd/replication/Config.cpp
@@ -81,9 +81,7 @@ namespace
 
 	void configError(const string& type, const string& key, const string& value)
 	{
-		string msg;
-		msg.printf("%s specifies %s: %s", key.c_str(), type.c_str(), value.c_str());
-		raiseError("%s", msg.c_str());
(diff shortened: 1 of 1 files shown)


src/jrd/replication/Config.cpp
```diff
@@ -83,7 +83,7 @@ namespace
 	{
 		string msg;
 		msg.printf("%s specifies %s: %s", key.c_str(), type.c_str(), value.c_str());
-		raiseError(msg.c_str());
+		raiseError("%s", msg.c_str());
```
aafemt, line 86: May be just eliminate `msg` completely moving whole formatting into `raiseError`?..

TreeHunter9: Done

AlexPeshkoff:
Please backport to v5.

PR #8560, opened 2025-05-10
Cache vcpkg binary cache for macOS

mrotteveel:
Due to https://github.com/microsoft/vcpkg-tool/pull/1662, the original caching is no longer working.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -389,9 +389,8 @@ jobs:
         id: restore-vcpkg-cache
         with:
           path: ~/.cache/vcpkg/archives
-          key: vcpkg-cache-${{ runner.os }}-${{ matrix.arch }}-${{ hashFiles('vcpkg-custom/**') }}
-          restore-keys: |
-            vcpkg-cache-${{ runner.os }}-${{ matrix.arch }}-
+          # We're always matching on a partial key here!
+          key: vcpkg-cache-${{ runner.os }}-${{ matrix.arch }}-
 
       - name: Build
         run: |
@@ -421,10 +420,10 @@ jobs:
 
       - name: Save vcpkg cache
         uses: actions/cache/save@v4
-        # if: steps.restore-vcpkg-cache.outputs.cache-hit != 'true'
+        if: steps.restore-vcpkg-cache.outputs.cache-matched-key != 'vcpkg-cache-${{ runner.os }}-${{ matrix.arch }}-${{ hashFiles('~/.cache/vcpkg/archives/**') }}'
         with:
           path: ~/.cache/vcpkg/archives
-          key: ${{ steps.restore-vcpkg-cache.outputs.cache-primary-key }}
+          key: vcpkg-cache-${{ runner.os }}-${{ matrix.arch }}-${{ hashFiles('~/.cache/vcpkg/archives/**') }}
 
       - name: Upload installer
         uses: actions/upload-artifact@v4
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -421,7 +421,7 @@ jobs:
 
       - name: Save vcpkg cache
         uses: actions/cache/save@v4
-        if: steps.restore-vcpkg-cache.outputs.cache-hit != 'true'
+        # if: steps.restore-vcpkg-cache.outputs.cache-hit != 'true'
         with:
           path: ~/.cache/vcpkg/archives
           key: ${{ steps.restore-vcpkg-cache.outputs.cache-primary-key }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -362,8 +362,6 @@ jobs:
   build-macos:
     name: build-macos-${{ matrix.arch }}
     runs-on: ${{ (matrix.arch == 'arm64' && 'macos-14') || 'macos-13' }}
-    env:
-      VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


mrotteveel:
@asfernandes Do you think generating the key based on the contents of vcpkg-custom is sufficient, or do you think I should take anything else into consideration?

asfernandes:
I don't think you need to hash vcpkg-custom. vcpkg will hash things by hemself.

.github/workflows/main.yml
```diff
@@ -419,6 +419,13 @@ jobs:
           mkdir gen/artifacts
           mv gen/Release/*.pkg gen/artifacts
 
+      - name: Save vcpkg cache
+        uses: actions/cache/save@v4
+        if: steps.restore-vcpkg-cache.outputs.cache-hit != 'true'
```
asfernandes, line 424: This condition does not seems correct for me.
Just me vcpkg do its job. Just use `actions/cache@v4` only and let it restore and save the cache.

.github/workflows/main.yml
```diff
@@ -384,14 +382,16 @@ jobs:
...
-        uses: actions/github-script@v7
+      - name: Restore vcpkg cache
+        uses: actions/cache/restore@v4
+        id: restore-vcpkg-cache
         with:
-          script: |
-            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
-            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+          path: ~/.cache/vcpkg/archives
+          key: vcpkg-cache-${{ runner.os }}-${{ matrix.arch }}-${{ hashFiles('vcpkg-custom/**') }}
```
asfernandes, line 392: I would use `key: vcpkg-cache-${{ runner.os }}-${{ matrix.arch }}`

.github/workflows/main.yml
mrotteveel: GitHub Action caches are write once. If you remove this condition, you don't need to write it again *under a different key*. The alternative, which I had in an earlier iteration, was to just use the run id as part of the cache key, but that has the downside that each run will write its own 100 MiB cache, even if nothing changed.

.github/workflows/main.yml
mrotteveel: See my earlier comment about GitHub Action caches being write once, so the key needs to change if you want to write a new one.

mrotteveel: With the current setup, if vcpkg-custom doesn't change, it will use the existing cache and only write a new one if the cache didn't exist (or not for the same hash), and if it does change, it will use the existing cache and then write the new cache, which it will use for subsequent actions as well.

.github/workflows/main.yml
mrotteveel: Actually, that might only apply to the cache action, and not the cache/save action. I'll verify my assumption.

.github/workflows/main.yml
asfernandes: This is IMO wrong and will not update cache when vcpkg submodule is updated.

mrotteveel: Then that would be an indication that vcpkg needs to be included in the hash as well.

.github/workflows/main.yml
mrotteveel: In other words, if you want to have any chance of replacing the cache when something changes, you need to use a key that changes when something important changes (or have some additional code/action to detect cache changes, delete the existing cache and the upload the modified cache under the same key).

.github/workflows/main.yml
mrotteveel: Alternatively, the hash needs to be generated over `~/.cache/vcpkg/archives`. I'll try that as well.

mrotteveel: OK, that doesn't work, because `hashFiles` can only hash content within the workspace directory. That would mean moving the cache into a location in the workspace and using `VCPKG_DEFAULT_BINARY_CACHE` to specify that location. I'll first try with hashing over vcpkg and vcpkg-custom, before I try and do that.

mrotteveel: As an aside, do we really need that vcpkg submodule? Can't we use [brew to install vcpkg](https://formulae.brew.sh/formula/vcpkg)?

mrotteveel: Or am I missing what that submodule actually is?

asfernandes: submodule is the recommended approach

asfernandes:
From one side, vcpkg handles all things about differences with its own hashing, so it would be a matter of get and update the cache.
From another side, that would accumulate things when we start to change versions (vcpkg, libraries, custom ports).
You don't need to hash the git submodule. We have a reference of it in vcpkg-configuration.json. So we can hash it and maybe vcpkg.json too.

mrotteveel:
> From one side, vcpkg handles all things about differences with its own hashing, so it would be a matter of get and update the cache. From another side, that would accumulate things when we start to change versions (vcpkg, libraries, custom ports). You don't need to hash the git submodule. We have a reference of it in vcpkg-configuration.json. So we can hash it and maybe vcpkg.json too.

In the current setup, even if the hash changes, it will fetch a previous version of the cache, so accumulation can occur here as well (but I consider that not so big a problem, or we can sometimes do some manual cache maintenance by deleting (all) caches).

In any case, I'll add vcpkg-configuration.json and vcpkg.json to the hash (and remove vcpkg/**). 

Sidequest: I noticed that vcpkg.json has `"version-string": "5.0.0"`. Should that be updated to 6.0.0?

mrotteveel:
@asfernandes I think this is good to go. Do you agree?

asfernandes (approved):

mrotteveel:
I'll also apply these changes to v5.0-release.

PR #8561, opened 2025-05-10
Fixed bug #8524 : ISQL will truncate lines longer than 255 when pasting

### LINKED ISSUE #8524: ISQL will truncate lines longer than 255 when pasting
Firebird 5 ISQL will truncate lines longer than 255 when pasting (tested on Windows 10)

1. Open command prompt
2. Execute ISQL connecting to a database, i.e. isql server:dbpath
3. Open a script file in any text editor, select all and copy
4. Paste it back to ISQL prompt

If any script line has more than 255 chars, it will be truncated and script execution will fail. Attached example script.

Same script will run fine if you paste into Firebird 2.5 ISQL.

[sp_test_long_line.txt](https://github.com/user-attachments/files/19760098/sp_test_long_line.txt)

hvlad:


src/isql/isql.epp
```diff
@@ -378,34 +383,42 @@ static int win32ReadConsole(FILE* file, char* buffer, size_t bufferSize)
...
-			if (!ReadConsoleW(handle, wideBuf + 1, 1, &charsRead, 0))
-				return -1;
-
 			if (charsRead)
-				++wideBufLen;
+				charsRead--;
+			else
+			{
+				DWORD read2;
+				if (!ReadConsoleW(handle, pWide + 1, 1, &read2, NULL))
```
aafemt, line 408: You already read whole line into `wideBuf`, there is nothing to read here. `WideCharToMultiByte` below will take care about surrogates automatically. Whole this `if` is meaningless AFAIK (as well as outer loop, perhaps - you just read the line and convert it, no piecewise read is possible).

src/isql/isql.epp
```diff
@@ -329,6 +329,11 @@ static int win32ReadConsole(FILE* file, char* buffer, size_t bufferSize)
 		return -1;
 	}
 
+	const size_t MAX_LINE_LENGTH = 32765;
+	static WCHAR wideBuf[MAX_LINE_LENGTH +1];
```
mrotteveel, line 333: Curiosity question: should this be allocated on the heap instead of on the stack? 

For example Visual Studio analysis will issue a C6262 warning for this, and its caller (readNextInputLine) also allocates a buffer of 64 KiB on the stack, so these two methods combined will allocate 128 KiB+ on the stack.

hvlad: > Curiosity question: should this be allocated on the heap instead of on the stack?
> 
> For example Visual Studio analysis will issue a C6262 warning for this, and its caller (readNextInputLine) also allocates a buffer of 64 KiB on the stack, so these two methods combined will allocate 128 KiB+ on the stack.

This is `static` buffer, it is not on stack. And it is done intentionally.

src/isql/isql.epp
hvlad: The code is generic and doesn't assume that whole line is read by single call of `ReadConsoleW()`. 
I see no need to break this.

aafemt: First, surrogates aren't supported in modern Unicode standard at all.

Second, you cannot enter them from console in a way that only a half of surrogate pair is been read exactly because of the issue: the console disallow to enter more characters than fits the requested buffer. For old code such situation was theoretically possible because it read by tiny pieces and the system console buffer was bigger than requested buffer. In the new code this call will wait for the next line requesting one character which, in turn, will again limit such continuation line by 254 characters, no?.. If you want to preserve such logic, move the incomplete character into the beginning of the buffer and read `MAX_LINE_LENGTH - 1` here at least.

hvlad: I've tested the code with small buffer and actually able to enter surrogate pair in the way you consider impossible.
There is no documenattion and no guarantee that next Windows version keeps internal buffer of 255 characters, so I prefer to keep existing working already tested code.
I see no reason to convince you here, it is senseless.

PS ask yourself - why wideBuf is declared with MAX_LINE_LENGTH +1 length

mrotteveel: > First, surrogates aren't supported in modern Unicode standard at all.

That is not correct, Unicode 16 (the latest version) still covers surrogates, and explicitly says they should only be used with UTF-16, and UTF-16 needs surrogates to be able to encode all characters defined by Unicode. And guess what `ReadConsoleW` produces: UTF-16.

mrotteveel: And I'd guess that pasting a U+10000 or higher character in a console that supports it will produce surrogate pairs.

src/isql/isql.epp
mrotteveel: OK, I had missed that, but the similar buffer in `readNextInputLine` is not static. Should it be?

aafemt: `pWide` and `charsRead` are more obvious candidates for static duration. Or at least a comment why they are fine as local variables.

hvlad: > OK, I had missed that, but the similar buffer in `readNextInputLine` is not static. Should it be?

It is not necessary.

Issue #8565, opened 2025-05-14
Silent truncation of user-provided data of type SQL_TEXT


aafemt:
This is an initial architectural problem which surfaced after apply #8145.

For multibyte encodings engine validates data in input message limiting it to a supposed character length `dsc_length / BPC`. If an user has provided in the message metadata the real data length, this data gets truncated without even a warning.

Testcase is attached.

[ctest.zip](https://github.com/user-attachments/files/20210459/ctest.zip)

asfernandes:
The problem is #8145 then, and you were warned in https://github.com/FirebirdSQL/firebird/issues/8082.

I see no sense in ask Pavel Zotov to test a change like #8566 that is not a fix, but a bug!

Issue #8082, opened 2024-04-14
Transliteration error if connection charset is narrower that storage and requested charsets


aafemt:
Also a temporary BLOB is created unnecessarily.

Test application is attached.
[ctest.zip](https://github.com/FirebirdSQL/firebird/files/14971237/ctest.zip)

asfernandes:
This is by design. When prepare is called, there is no additional information about charsets, then output message is generated using the connection charset.

For external routines, they can override client charset. I remember many (all) were against it, fortunately I insisted and maintained it allowing external routines to not be broken by client options.

aafemt:
Problem is not with output message generated during prepare, it is informational only. Problem is with output message received during `openCursor()` which has to be used for data delivery.

asfernandes:
You are wrong.

```
isql ctest -ch utf8
SQL> set exec_path_display blr;
SQL> select * from a;

Execution path (BLR):

    0 blr_version5,
    1 blr_begin,
    2    blr_message, 1, 11,0,
    6       blr_short, 0,
    8       blr_short, 0,
   10       blr_blob2, 1,0, 4,0,
   15       blr_short, 0,
   17       blr_blob2, 1,0, 4,0,
   22       blr_short, 0,
   24       blr_varying2, 4,0, 40,0,
   29       blr_short, 0,
   31       blr_varying2, 4,0, 40,0,
   36       blr_short, 0,
   38       blr_long, 0,
   40    blr_for,
   41       blr_stall,
   42       blr_rse, 1,
   44          blr_rid, 128,0, 0,
   48          blr_end,
   49    blr_send, 1,
   51       blr_begin,
   52          blr_assignment,
   53             blr_literal, blr_short, 0, 1,0,
   58             blr_parameter, 1, 0,0,
   62          blr_assignment,
   63             blr_fid, 0, 4,0,
   67             blr_parameter2, 1, 2,0, 1,0,
   73          blr_assignment,
   74             blr_fid, 0, 3,0,
   78             blr_parameter2, 1, 4,0, 3,0,
   84          blr_assignment,
   85             blr_fid, 0, 2,0,
   89             blr_parameter2, 1, 6,0, 5,0,
   95          blr_assignment,
   96             blr_fid, 0, 1,0,
  100             blr_parameter2, 1, 8,0, 7,0,
  106          blr_assignment,
  107             blr_fid, 0, 0,0,
  111             blr_parameter2, 1, 10,0, 9,0,
[...]

Issue #7611, opened 2023-05-31
Can't backup/restore database from v3 to v4 with SEC$USER_NAME field longer than 10 characters


dmitry-starodubov:
If I have such a simple database 

CREATE TABLE T (ID INTEGER NOT NULL,
        USER_NAME SEC$USER_NAME,
        PASSWD VARCHAR(256) CHARACTER SET OCTETS,
        ALG VARCHAR(32) CHARACTER SET UNICODE_FSS);
		
insert into t values(1, 'aaabbbcccddd', '12345678', '123');

I can backup/restore in on v3.
But when I try to restore backup from v3 on v4 I get an error:
gbak: ERROR:arithmetic exception, numeric overflow, or string truncation
gbak: ERROR:    string right truncation
gbak: ERROR:    expected length 10, actual 12
gbak: ERROR:gds_$send failed
gbak:Exiting before completion due to errors

Because maximum length of USER_NAME field is calculated wrong.

sim1984:
v3 and v4, the SEC$USER_NAME domain has a different definition

```sql
CREATE DOMAIN SEC$USER_NAME AS
VARCHAR(31) CHARACTER SET UNICODE_FSS;
```

vs

```sql
CREATE DOMAIN SEC$USER_NAME AS
VARCHAR(63) CHARACTER SET UTF8;
```

aafemt:
Both of them should be enough for 12 characters. Expected 10 in the error message is strange.

pavel-zotov:
I could not reproduce described problem.
What *exact* versions of FB 3.x and FB 4.x / 5.x ?

asfernandes:
Try restore in v4 using embedded.

pavel-zotov:
Adriano, thanks. Now all OK.

PR #8566, opened 2025-05-14
Fix for #8565

aafemt:
So far this is a simplest quick fix to let Pavel Zotov test it.


aafemt:
It looks like this fix also solves #7611, so workaround in gbak can be removed.

aafemt:
I think it would be better to merge this PR now because it will make firebirdtest.com operational and do the rest of postfixes in separate PRs one per issue.

asfernandes:
You will break even more things than you already broke in #8082 ;)

```
create table tests (
    id integer,
    str varchar(10) character set utf8,
    char_len integer,
    octet_len integer
);

create procedure proc (
    id integer,
    str char(10) character set utf8
)
as
begin
    insert into tests (id, str, char_len, octet_len) values (:id, :str, char_length(:str), octet_length(:str));
end;
```

aafemt:
Reworked.

PR #8571, opened 2025-05-20
Postfixes for PR#8145

aafemt:


aafemt:
Commit "Return isc_dsql_error prefix" has been undone because it caused more problems than solved.

mrotteveel:
@dyemanov @hvlad Can one on you review and merge this?

aafemt:
@asfernandes Do you have a testcase that fail on this build?

dyemanov:
Just one question: are your changes really needed for VARCHAR parameters?

dyemanov (approved):

aafemt:
IMHO, yes. Some libraries do with them the same as with CHAR: overrun maximum length in characters as derived from maximum length in bytes while still fit to the storage. Besides, it suits to the standard's requirement to silently cut trailing blanks if they don't fit on store. Lastly it may help with API backward compatibility if longer VARCHARs are ever allowed.

PR #8586, opened 2025-06-04
Fix for #8139: Add lookup index name by constraint for correct resolve conflict on replica

### LINKED ISSUE #8139: Conflict resolution code uses constraint name instead of index name
https://github.com/FirebirdSQL/firebird/blame/95442bdfff76d22aafb57b58894047be2a89c6ea/src/jrd/replication/Applier.cpp#L637

If these names are different debug build consistently throw assertion error during execution of test_blob_not_found_in_rw_replica_if_target_row_exists.py.

XaBbl4:


src/jrd/replication/Applier.cpp
```diff
@@ -634,10 +634,18 @@ void Applier::insertRecord(thread_db* tdbb, TraNumber traNum,
...
 	fb_assert(error[3] != 0);
 
-	const char* idxName = reinterpret_cast<const char*>(error[3]);
+	const MetaName nameFromErr(reinterpret_cast<const char*>(error[3]));
+	MetaName idxName;
+	if (error[1] == isc_no_dup)
+		idxName = nameFromErr;
+	else if (!m_constraintIndexMap.get(nameFromErr, idxName))
+	{
+		MET_lookup_index_for_cnstrt(tdbb, idxName, nameFromErr);
```
hvlad, line 643: What happens if `idxName` is empty (constraint not found) ?

src/jrd/replication/Applier.cpp
```diff
@@ -1099,16 +1107,13 @@ bool Applier::lookupRecord(thread_db* tdbb,
...
 	bool haveIdx = false;
-	if (idxName)
+	if (idxName && idxName->hasData())
 	{
 		SLONG foundRelId;
 		IndexStatus idxStatus;
-		SLONG idx_id = MET_lookup_index_name(tdbb, idxName, &foundRelId, &idxStatus);
-
-		fb_assert(idxStatus == MET_object_active);
-		fb_assert(foundRelId == relation->rel_id);
```
hvlad, line 1109: Why remove these asserts ? It helps in DEBUG build to detect "impossible" case.

src/jrd/met.epp
```diff
@@ -2037,6 +2037,37 @@ void MET_load_trigger(thread_db* tdbb,
...
+{
+/**************************************
+*
+*      M E T _ l o o k u p _ i n d e x _ f o r _ c n s t r t
+*
+**************************************
+*
+* Functional description
+*      Lookup index name from constraint name, if one exists.
+*      Calling routine must pass a buffer of at least 32 bytes.
```
hvlad: Oh, really ? :)

XaBbl4: Copy/Paste, and did not correct this part, can also change this line from the place from which I copied :)
Change it in all same places.

src/jrd/replication/Applier.cpp
XaBbl4: I first remove them, and then made lookup the index by the name of the constraint, in principle, return it, since in this situation they should not occur in the debug build.

src/jrd/replication/Applier.cpp
XaBbl4: This situation cannot happen, because we will get here only with the error `isc_unique_key_violation`, and in this case the error will always contain the name of the constraint

If it was meant that the function would return `idxName == ""`, then the behavior would be the same as before, `lookupKey` would be called inside `lookupRecord`.

hvlad (approved):

hvlad:
Please, backport into v5

XaBbl4:
> Please, backport into v5

Separate PR? Or what is the best way to do it?

hvlad:
Backported into v4 and v5

PR #8591, opened 2025-06-10
Fix #8590 - Line "BLR to Source mapping:" may look broken when procedures.rdb$debug_info is queried using remote protocol.

asfernandes:
When transforming a blob of an internal type (say, isc_blob_debug_info) to text, the segments are generated without new line information.

ISQL treats it segment as a line, putting some leading spaces and the new line.

I have no idea why.

But this causes problems when remote batches segments and some are split, causing the extra new line as reported in #8590.

This change is not always backward incompatible, specially when different versions of engine x ISQL is used or non-ISQL clients converting internal blobs to text.

### DIFF
--- a/src/isql/isql.epp
+++ b/src/isql/isql.epp
@@ -7755,23 +7755,7 @@ processing_state ISQL_print_item_blob(FILE* fp, const IsqlVar* var, Firebird::IT
 			break;
 
 		buffer[length] = 0;
-
-		// Special displays for blr or acl subtypes
-
-		if (blob_subtype > isc_blob_text && blob_subtype < isc_blob_max_predefined_subtype)
-		{
-			for (char* b = buffer + length - 1; b >= buffer;)
-			{
-				if (*b == '\n' || *b == '\t' || *b == BLANK)
-					*b-- = 0;
-				else
-					break;
-			}
-
-			IUTILS_printf2(fp, "%s	%s%s", TAB_AS_SPACES, buffer, NEWLINE);
-		}
-		else
-			IUTILS_printf(fp, buffer);
+		IUTILS_printf(fp, buffer);
 
 	} while (true);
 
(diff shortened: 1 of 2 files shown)


hvlad:
> ISQL treats it segment as a line, putting some leading spaces and the new line.
> 
> I have no idea why.

I guess - because new line on server could be different from new line on client, for example - Linux server and Windows client.

Your change broke this. I would just make ISQL correctly handle RESULT_SEGMENT.

asfernandes:
This problem has been resolved in stdout around the 1900 years.

hvlad:
> This problem has been resolved in stdout around the 1900 years.

How your patch is relatied with it ?

asfernandes:
> I would just make ISQL correctly handle RESULT_SEGMENT.

If you looked at details of the problem, you would saw that ISQL does not receives `RESULT_SEGMENT` at all.

aafemt:
> If you looked at details of the problem, you would saw that ISQL does not receives `RESULT_SEGMENT` at all.

Then this is an another bug to be fixed.

hvlad:
> > I would just make ISQL correctly handle RESULT_SEGMENT.
> 
> If you looked at details of the problem, you would saw that ISQL does not receives `RESULT_SEGMENT` at all.

Then, what do you mean here ?

>But this causes problems when remote batches segments and some are split, causing the extra new line as reported in https://github.com/FirebirdSQL/firebird/issues/8590.

hvlad:
Lets fix the real bug - in `string_filter()`

hvlad:
See #8594

asfernandes:
The Firebird team thanks you for effort duplication.
And I'd say this is a problem here too, the way it does is inefficient and more complex than necessary.

hvlad:
> The Firebird team thanks you for effort duplication. 

Are you present the whole team ? Or you going to tell me what to do ?

> And I'd say this is a problem here too, the way it does is inefficient and more complex than necessary.

The issue should be fixed correctly and not broke old code, when possible, like you it or not.
Your fix is wrong, by my opinion, and I offer a correct one.

PR #8600, opened 2025-06-14
Implement #8598 : Don't fire referential integrity triggers if primary or unique keys haven't changed

### LINKED ISSUE #8598: Don't fire referential integrity triggers if primary or unique keys haven't changed
We encountered the following issue on a large database:

[https://github.com/FirebirdSQL/firebird/discussions/8593](https://github.com/FirebirdSQL/firebird/discussions/8593)

We needed to update a table containing 172 million records. The update did not modify the primary key of any records — it only updated three integer fields. However, since the table was referenced by 1,358 foreign keys, the operation triggered:

1.72 × 10⁸ × 1.358 × 10³ ≈ 10¹¹

system trigger calls. These trigger calls alone caused the update to take 19 hours.

The proposed solution is to skip referential integrity triggers when the primary key or any other unique key is not modified.

hvlad:


TechAnton:
Hi everyone,

Could you please let me know why the pull request with the bug fix has not been accepted yet? Are there any issues or is there a queue for review? 
When can we expect it to be merged into the main branch?

Thank you!

dyemanov:
Review is pending, to be completed this week.

TechAnton:
Thank you! I’ll wait for the review.

src/jrd/req.h
```diff
@@ -136,6 +136,7 @@ const USHORT RPB_undo_data		= 0x02;	// data got from undo log
 const USHORT RPB_undo_read		= 0x04;	// read was performed using the undo log
 const USHORT RPB_undo_deleted	= 0x08;	// read was performed using the undo log, primary version is deleted
 const USHORT RPB_just_deleted	= 0x10;	// record was just deleted by us
+const USHORT RPB_uk_updated		= 0x20;	// set by IDX_modify if it insert key into any primary or unique index
```
dyemanov, line 139: Shouldn't it be added to `RPB_CLEAR_FLAGS`? Where does it get cleared for the next record processed?

hvlad: It is cleared before record processing, see IDX_modify().
So far I see no need to add it into RPB_CLEAR_FLAGS.

dyemanov: I agree that the current code looks safe, while I'd still be surprised to see in the debugger this flag raised after the new record is fetched (but before `IDX_modify` is called). Not insisting though.

dyemanov (approved):

src/jrd/req.h
hvlad: I don't see how **new** record could be fetched :)
Anyway, I see no harm in including RPB_uk_updated into RPB_CLEAR_FLAGS, so let is be so.

PR #8606, opened 2025-06-17
Refactor join types

dyemanov:
1) Replace BLR join verbs with enum inside RseNode
2) Introduce ANTI JOIN
3) Separate special join types (SEMI | ANTI) from INNER JOINs


src/jrd/RecordSourceNodes.h
```diff
@@ -715,6 +715,21 @@ class WindowSourceNode final : public TypedNode<RecordSourceNode, RecordSourceNo
 
 class RseNode final : public TypedNode<RecordSourceNode, RecordSourceNode::TYPE_RSE>
 {
+	enum : UCHAR							// storage is BLR-compatible
+	{
+		INNER_JOIN = 0,
```
asfernandes: Isn't better to assign the blr constants?
```
INNER_JOIN = blr_inner,
...
```

pavel-zotov:
Improvement affects only FB internals. Letter from dimitr 22.07.2025 16:56.

PR #8609, opened 2025-06-18
Move impure-related string allocation to helper methods

Noremos:
This is also a preparation for the JSON Type implementation

### DIFF
--- a/src/common/dsc.h
+++ b/src/common/dsc.h
@@ -228,6 +228,11 @@ typedef struct dsc
 		return dsc_dtype == dtype_unknown;
 	}
 
+	bool hasDynamicLength() const
+	{
+		return isText(); // TODO: add isJson();
+	}
+
 	SSHORT getBlobSubType() const
 	{
 		if (isBlob())
--- a/src/common/dsc.h
+++ b/src/common/dsc.h
@@ -228,11 +228,6 @@ typedef struct dsc
 		return dsc_dtype == dtype_unknown;
 	}
 
-	bool hasDynamicLength() const
-	{
-		return isText(); // TODO: add isJson();
-	}
-
 	SSHORT getBlobSubType() const
 	{
 		if (isBlob())
--- a/src/jrd/evl.cpp
+++ b/src/jrd/evl.cpp
@@ -573,22 +573,10 @@ void EVL_make_value(thread_db* tdbb, const dsc* desc, impure_value* value, Memor
 
 	// Allocate a string block of sufficient size.
 
-	VaryingString* string = value->vlu_string;
+	if (!pool)
+		pool = tdbb->getDefaultPool();
 
-	if (string && string->str_length < length)
-	{
-		delete string;
-		string = NULL;
-	}
-
-	if (!string)
-	{
-		if (!pool)
-			pool = tdbb->getDefaultPool();
-
-		string = value->vlu_string = FB_NEW_RPT(*pool, length) VaryingString();
-		string->str_length = length;
-	}
+	VaryingString* string = value->getAllocateString(from.dsc_length, *pool);
 
 	value->vlu_desc.dsc_length = length;
 	UCHAR* target = string->str_data;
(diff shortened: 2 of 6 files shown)


src/common/dsc.h
```diff
@@ -228,6 +228,11 @@ typedef struct dsc
 		return dsc_dtype == dtype_unknown;
 	}
 
+	bool hasDynamicLength() const
+	{
+		return isText(); // TODO: add isJson();
+	}
+
```
dyemanov: This does not look proper to me, as `CHAR` is not dynamic-length generally (except inside the impure). IMHO, it should be up to the caller (impure class) to check what string types are stored as variable-length.

src/jrd/evl.cpp
```diff
@@ -573,22 +573,10 @@ void EVL_make_value(thread_db* tdbb, const dsc* desc, impure_value* value, Memor
...
-
-	if (!string)
-	{
-		if (!pool)
-			pool = tdbb->getDefaultPool();
-
-		string = value->vlu_string = FB_NEW_RPT(*pool, length) VaryingString();
-		string->str_length = length;
-	}
+	VaryingString* string = value->getAllocateString(from.dsc_length, *pool);
```
dyemanov: Shouldn't it be `value->getAllocateString(length, *pool)` ?

src/jrd/val.h
```diff
@@ -168,6 +168,13 @@ struct impure_value
 	void make_double(const double val);
 	void make_decimal128(const Firebird::Decimal128 val);
 	void make_decimal_fixed(const Firebird::Int128 val, const signed char scale);
+
+	template<class T>
+	VaryingString* getAllocateString(T requeuedLength, MemoryPool& pool) = delete; // Prevent dangerous length shrink
+	VaryingString* getAllocateString(USHORT requeuedLength, MemoryPool& pool);
```
dyemanov: I'd call it just `getString()` for simplicity but I'm not going to insist.

src/jrd/val.h
```diff
@@ -168,6 +168,13 @@ struct impure_value
...
+
+	void makeImpureDscAddress(MemoryPool& pool);
+	void allocateTextImpureDscAddress(MemoryPool& pool);
```
dyemanov: We're already inside the impure class, so please simplify the names, e.g. `makeDscAddress()` / `setupDynamicDscAddress()`.

src/jrd/val.h
```diff
@@ -221,6 +228,40 @@ inline void impure_value::make_decimal_fixed(const Firebird::Int128 val, const s
 	this->vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&this->vlu_misc.vlu_int128);
 }
 
+inline VaryingString* impure_value::getAllocateString(USHORT requeuedLength, MemoryPool& pool)
```
dyemanov: I'd suggest `newLength` or just `length` instead of `requeuedLength`.

dyemanov (changes requested):

src/jrd/val.h
asfernandes: getAllocateString is not a name that sounds good.

asfernandes: Usually pool comes first in parameter list.

src/jrd/val.h
asfernandes: These names are weird and maybe due to it, the methods too.
With current names, I prefer the original code.

src/common/dsc.h
Noremos: By this method I mean that dsc_length of content is not fixed. Now I remembered that there is an array `type_lengths` that checks exactly this, so I replace this method with the array

src/jrd/val.h
Noremos: Names are my weak point. I tried others in the last commit, but it would be nice to get your suggestions.

src/jrd/val.h
Noremos: Fixed

src/jrd/val.h
Noremos: Changed the name to `getString` with the pool as the first parameter

dyemanov (approved):

asfernandes (approved):

PR #8610, opened 2025-06-18
Merge putDtype and putType; remove duplicated code

Noremos:
This is also a preparation for the JSON Type implementation


src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -239,299 +239,180 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
+	putType(*field, useSubType, field->explicitCollation);
+}
+
+void DsqlCompilerScratch::putType(const TypeClause* type, bool useSubType)
+{
+	fb_assert(type);
+	putType(*type, useSubType, type->collate.object.hasData());
+}
+
+void DsqlCompilerScratch::putType(const TypeClause& type, bool useSubType, bool useExplicitCollate)
```
asfernandes, line 253: It's weird that the two methods with the same name one expects the type as a pointer and the other as reference.

asfernandes (changes requested):
If you not choose better names for methods you created to not duplicate code, I prefer to have the code duplicated.
I understand nothing what the refactored methods do by their names.

src/dsql/DsqlCompilerScratch.cpp
Noremos: This is a design issue. A pointer means that the value can be null, and the method should check for that. But these methods explicitly treat values ​​as non_null. It would have been better to change the original putType and putDType to use references, but I avoided this to avoid changing too many files.

Noremos:
> If you not choose better names for methods you created to not duplicate code, I prefer to have the code duplicated. I understand nothing what the refactored methods do by their names.

I tried to follow the style of the original method names. `putType` and `putDtype` are not the most transparent names either. So it would be nice if you could suggest alternative names for my methods

src/dsql/DsqlCompilerScratch.cpp
dyemanov: I'd make the reference-based `putType()` private to avoid confusion.

And maybe rename `putDtype()` to `putFieldType()` or make it a `putType()`  overload with the `dsql_fld*` argument. Less than a dozen lines changed, I guess.

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -239,299 +239,180 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
+			if (DTYPE_IS_EXACT(type.dtype) || dtype_quad == type.dtype)
+				appendUChar(type.scale);
 			break;
 	}
 }
 
-void DsqlCompilerScratch::putType(const TypeClause* type, bool useSubType)
+
+template<bool THasTableName>
+void DsqlCompilerScratch::putTypeNameBlr(const TypeClause& type, const bool useExplicitCollate)
```
dyemanov: All `put*` methods are already about BLR, so it may be named just `putTypeName()`.

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -239,299 +239,180 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
-
 		return;
 	}
 
-	switch (field->dtype)
+	// Maybe it is possible to use GEN_descriptor here?
+	putDTypeBlr(type, useSubType);
+}
+
+void DsqlCompilerScratch::putDTypeBlr(const TypeClause& type, const bool useSubType)
```
dyemanov: AFAIU, this method is used just once. Does it exist just to reduce the number of lines inside `putType()`? Anyway, it should be private, IMHO.

And if `putDtype()` above is renamed, then I'd suggest to remove the `Blr` suffix here (see the reason in the comment below).

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -239,299 +239,180 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
+		UCHAR name;
+		UCHAR name2;
+		UCHAR name3;
+	};
 
-		sprintf(buffer, "Invalid dtype %d in put_dtype", type->dtype);
-		ERRD_bugcheck(buffer);
-	}
-#endif
+	static constexpr BlrNameSet BLR_COLUMN_SET{
```
dyemanov: Here and below - please add the space before the curly brace or move it into a separate line.

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -239,299 +239,180 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
-		appendUChar(blr_not_nullable);
+	static constexpr BlrNameSet BLR_DOMAIN_SET{
+		blr_domain_name,
+		blr_domain_name2,
+		blr_domain_name3,
+	};
+
+	constexpr BlrNameSet blrSet = []()
+	{
+		if constexpr(THasTableName)
```
dyemanov: A space after `constexpr`, please

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -239,299 +239,180 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
+		{
+			appendMetaString(type.typeOfTable.object.c_str());
+		}
+
+		appendMetaString(type.typeOfName.object.c_str());
+
+		if (useExplicitCollate)
+		{
+			appendUShort(type.textType);
+		}
```
dyemanov: Here and above: one-liner statement bodies without braces, please.

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -239,299 +239,180 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
+		else
+			return BLR_DOMAIN_SET;
+	}();
 
-	if (type->typeOfName.object.hasData())
+	bool differentSchema;
+	if constexpr (THasTableName)
+		differentSchema = type.typeOfTable.schema != ddlSchema;
+	else
+		differentSchema = type.typeOfName.schema != ddlSchema;
```
dyemanov, line 369: Wouldn't something like this possible and maybe better?

```C++
const auto typeSchema = constexpr (THasTableName) ? type.typeOfTable.schema : type.typeOfName.schema;
if (typeSchema != ddlSchema)
```

sim1984: `if constexpr` is a single construct in C++ hat discards one of the execution branches at compile time. I doubt it can be replaced.

Noremos: unfortunately, there is no constexpr ternary operator. This is why I used the lambda function for the `blrSet`

src/dsql/DsqlCompilerScratch.cpp
Noremos: Done

src/dsql/DsqlCompilerScratch.cpp
Noremos: Done

src/dsql/DsqlCompilerScratch.cpp
Noremos: Done

src/dsql/DsqlCompilerScratch.cpp
Noremos: I removed the blr suffix as you suggested.

src/dsql/DsqlCompilerScratch.cpp
Noremos: I renamed the putDtype to putType as you suggested

Noremos: > Does it exist just to reduce the number of lines inside `putType()`

Initially, I took out only this method from putType and putDType, but then I thought that this was a half-measure and it was necessary to do a full-fledged refactoring. 
For now I've left this function separate, as I hope in the future it will make it easier to refactor and use a single function to generate `TypeClause` and `dsc` blr

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -238,300 +238,174 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
-
 		return;
 	}
 
-	switch (field->dtype)
+	// Maybe it is possible to use GEN_descriptor here?
+	putFieldType(type, useSubType);
+}
+
+void DsqlCompilerScratch::putFieldType(const TypeClause& type, const bool useSubType)
```
dyemanov: I suggested to rename this function to `putDType()` after the original `putDtype()` was made an overloaded `putType()`. The reason is simple - it doesn't care about fields, it only generates BLR for the given data type.

src/dsql/DsqlCompilerScratch.h
```diff
@@ -289,6 +295,13 @@ class DsqlCompilerScratch : public BlrDebugWriter
 	bool pass1RelProcIsRecursive(RecordSourceNode* input);
 	BoolExprNode* pass1JoinIsRecursive(RecordSourceNode*& input);
 
+	void putField(const TypeClause& type, bool useSubType, bool useExplicitCollate);
+
+	template<bool THasTableName>
+	void putFieldName(const TypeClause& type, const bool useExplicitCollate);
+
+	void putFieldType(const TypeClause& type, const bool useSubType);
```
dyemanov: The naming looks confusing - we pass `TypeClause` (not `dsql_fld`) but they're named `putField*`.
What's wrong with `putType()` (private overload) / `putTypeName()` / `putDType()` ?

Noremos: I'm just confused by the `TypeClause` containing a reference to a _field_. Anyway, I renamed the methods as suggested

dyemanov: I've also suggested `putField() `-> `putType()`. Or do you think it will confuse users given we already have two public `putType()` overloads?

dyemanov: BTW, conflicts must be resolved before merging.

Noremos: Renamed putField() to putType(). 
@asfernandes was worried about mixing reference and pointer parameters, but I hope that keeping the methods with reference parameter in the private zone will help

src/dsql/DsqlCompilerScratch.cpp
```diff
@@ -238,298 +238,173 @@ void DsqlCompilerScratch::putBlrMarkers(ULONG marks)
...
-				appendUChar(field->scale);
+			appendUChar(blr_dtypes[type.dtype]);
+			if (DTYPE_IS_EXACT(type.dtype) || dtype_quad == type.dtype)
+				appendUChar(type.scale);
 			break;
 	}
 }
 
-void DsqlCompilerScratch::putType(const TypeClause* type, bool useSubType)
+template<bool THasTableName>
```
asfernandes, line 333: What means `THasTableName`?

src/jrd/tests/BlrGenTest.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+BOOST_AUTO_TEST_SUITE(DsqlCompilerScratchTests)
+
+// HELPERS
+// --------
+
+DsqlCompilerScratch* makeScratch()
+{
+	auto& pool = *getDefaultMemoryPool();
+	return FB_NEW DsqlCompilerScratch(pool, nullptr, nullptr);
```
asfernandes, line 20: Why create objects in heap and leak them instead of use local objects and let RAII destruct them?

src/dsql/DsqlCompilerScratch.cpp
Noremos: This indicates the presence of `typeOfTable.*` fields in the `TypeClause& type`. That is, we use blr_column_name (i.e., the table name with typeOfTable) or blr_domain_name (with typeOfName).

dyemanov: I'd question the need of the `T` prefix, as it usually means some _type_ used as the template argument, not the variable. Simple `HasTableName` or maybe `IsTypeOfTable` is IMHO better readable.

Noremos: It would be nice to have some explicit coding style for such cases. [The existing one](https://firebirdsql.org/en/coding-style) suggests using a single letter, which is not the clearest name possible in this case. Personally, I use the T prefix to make the template argument easier to recognize in the code, especially with `if constexpr` statements

asfernandes: > I'd question the need of the `T` prefix, as it usually means some _type_ used as the template argument, not the variable. Simple `HasTableName` or maybe `IsTypeOfTable` is IMHO better readable.

Yes.

src/jrd/tests/BlrGenTest.cpp
Noremos: Oops, I forgot to fix my raw temporary  implementation

Noremos: Oh, this is why I am leaking memory:
```
protected:
	// DsqlCompilerScratch should never be destroyed using delete.
	// It dies together with it's pool.
	~DsqlCompilerScratch()
	{
	}
```

I can create a derived class and use it, but I'm not sure if it's safe to delete it, since the DsqlCompilerScratch has a strict prohibition

PR #8620, opened 2025-06-22
SYSTEM schema should always be in the search path (#8605).

### LINKED ISSUE #8605: Character sets and collations should always search in SYSTEM first, even if it's not on the search path
Currently this fails:

```
SQL> set search_path to public;
SQL> select cast('abc' as varchar(50) character set utf8) from system.rdb$database;
Statement failed, SQLSTATE = HY004
Dynamic SQL Error
-SQL error code = -204
-Data type unknown
-CHARACTER SET "PUBLIC"."UTF8" is not defined
```

And this really should *never* happen, no matter what the search path is.

Following the rules of the SQL standard of section 4.3.1 and section 5.4 syntax rules 12 and 13a, unqualified character sets and collations should *always* lookup the system character sets and collations (which in the standard live in `INFORMATION_SCHEMA`, and for us in `SYSTEM`). We can relax this a bit to also search the search path, but using the character sets/collations from `SYSTEM` *first* should be implemented.

asfernandes:


aafemt:
There is no need to check if system schema is already in the search path: it doesn't matter if an object is found in the first occurrence or in the last.

AlexPeshkoff (approved):

dyemanov (approved):

PR #8626, opened 2025-06-25
Add replication journal cleanup when drop database

XaBbl4:
_Problem: when deleting a DB, log segment files remain, which break replication in the future_

**First fix** - when deleting a DB using DBMS query (drop database), add deleting segments
1. There is a DB with replication, create a backup
2. One of two options:
2.1. Execute the `drop database` command
2.2. Restore via gbak with the `-recreate_database overwrite` or `-replace_database` option, without deleting the DB and without clearing the journal directory
3. Old segments will be deleted as unnecessary

**Second fix** - add a unique GUID of the DB to the segment filename pattern
1. There is a DB with replication, create a backup
2. Delete the DB file via the file system
3. Restore via gbak, without clearing the journal directory
4. Old segments will remain in the file system, but will not interfere with replication, since the names of the new ones will be different


dyemanov (approved):

dyemanov:
Please resolve the existing conflicts.

mrotteveel:
Is this actually a safe change? As in, if a database is dropped unintentionally or maliciously, could the replica miss some final changes before the drop?

dyemanov:
IIRC, existing journal segments are synchronously archived before deletion, so they will be transferred to the replica.

mrotteveel:
Ah, you're right, I see that is done in `Changelog::cleanup()`.

PR #8633, opened 2025-06-27
fix(int128): Use all 16 decimal digits to represent int128 min/max in double number correctly

TreeHunter9:
The value `I128_MIN_dbl` was not represented precisely, as it didn't utilize all 16 decimal digits (52 bits of mantissa) available in a double-precision number. As a result, the value was rounded down to the closest representable number in the double format.

This inaccuracy led to an error in the following query:
```
SELECT CAST(-POWER(2, 127) AS INT128) FROM RDB$DATABASE;
=======================
-numeric value is out of range
```

### DIFF
--- a/src/common/cvt.cpp
+++ b/src/common/cvt.cpp
@@ -3169,10 +3169,10 @@ Int128 CVT_get_int128(const dsc* desc, SSHORT scale, DecimalStatus decSt, ErrorF
 	Decimal128 tmp;
 	double d, eps;
 
-	static const double I128_MIN_dbl = -1.701411834604692e+38;
-	static const double I128_MAX_dbl =  1.701411834604692e+38;
+	static const double I128_MIN_dbl = -1.7014118346046923e+38;
+	static const double I128_MAX_dbl =  1.7014118346046921e+38;
 	static const CDecimal128 I128_MIN_dcft("-1.701411834604692317316873037158841E+38", decSt);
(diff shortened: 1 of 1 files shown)


src/common/cvt.cpp
```diff
@@ -3169,10 +3169,10 @@ Int128 CVT_get_int128(const dsc* desc, SSHORT scale, DecimalStatus decSt, ErrorF
 	Decimal128 tmp;
 	double d, eps;
 
-	static const double I128_MIN_dbl = -1.701411834604692e+38;
-	static const double I128_MAX_dbl =  1.701411834604692e+38;
+	static const double I128_MIN_dbl = -1.7014118346046923e+38;
+	static const double I128_MAX_dbl =  1.7014118346046921e+38;
 	static const CDecimal128 I128_MIN_dcft("-1.701411834604692317316873037158841E+38", decSt);
-	static const CDecimal128 I128_MAX_dcft("1.701411834604692317316873037158841E+38", decSt);
+	static const CDecimal128 I128_MAX_dcft( "1.701411834604692317316873037158841E+38", decSt);
```
AlexPeshkoff, line 3175: Why whitespace?

AlexPeshkoff (changes requested):

src/common/cvt.cpp
TreeHunter9: To look in line with `I128_MIN_dcft`, like it's done with `I128_MIN_dbl` and `I128_MAX_dbl`

PR #8641, opened 2025-07-06
Deprecate array and blob API functions that do not work with schemas and longer object names

asfernandes:
I'm do considering alternative versions a requirement for the deprecations because these functions were already bugged by many time since max identifier lengths were increased from 31 characters.

Also, not deprecating `isc_blob_default_desc` and `isc_blob_gen_bpb` because their usage are not impacted by schema and longer identifiers.


aafemt:
Isn't whole ISC API already documented to be deprecated? Deprecating single functions of a deprecated API doesn't make any sense.

dyemanov (approved):

PR #8642, opened 2025-07-08
Fix conversion warnings in CvtFormat

mrotteveel:
+ adds some convenience methods for std::string_view in StatusVector and string

### DIFF
--- a/src/common/StatusArg.h
+++ b/src/common/StatusArg.h
@@ -221,7 +221,7 @@ public:
 		return *this;
 	}
 
-	StatusVector& operator<<(const std::string_view& text) noexcept
+	StatusVector& operator<<(std::string_view text) noexcept
 	{
 		implementation->shiftLeft(string(text.data(), static_cast<string::size_type>(text.length())));
 		return *this;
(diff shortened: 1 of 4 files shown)


mrotteveel:
@asfernandes My reason for using a PR are: 

1) is it OK to add these methods accepting std::string_view?
2) Should I add more complete support instead of just those two methods I needed to solve my immediate problem?

src/common/StatusArg.h
```diff
@@ -221,6 +221,12 @@ class StatusVector : public Base
 		return *this;
 	}
 
+	StatusVector& operator<<(const std::string_view& text) noexcept
```
asfernandes: It's generally recommended to use `std::string_view` (not const reference) as parameter.

src/common/classes/fb_string.h
```diff
@@ -739,6 +739,10 @@ namespace Firebird
 		{
 			return assign(first, last - first);
 		}
+		StringType& assign(const std::string_view& s)
```
asfernandes: It's generally recommended to use std::string_view (not const reference) as parameter.

mrotteveel:
Also replaced a use of `const string_view&` I did earlier.

asfernandes (approved):

PR #8645, opened 2025-07-09
Fix #8639 - Concurrency problem in remote profiler communication mechanism

### LINKED ISSUE #8639: Concurrency problem in remote profiler communication mechanism
The communication mechanism used in remote profiling is susceptible to race condition if the client abandon the read of a message.

That may happen if a `RDB$PROFILER` query is cancelled.

In that case, it will unlock the shared memory mutex while the server is processing a command. As soon the server is read to write the response, it will write things to the buffer, but as the client already unlocked the mutex, another client may already had taken the mutex and have written a new command to the same buffer.

asfernandes:


dyemanov:
Out of curiosity, how do you recover from the shared spinlock being "locked" but the process holding that spinlock crashed? I'm mostly worried about CS. So far we were using shared mutexes because they get automatically "unlocked" in such scenario.

AlexPeshkoff:
> Out of curiosity, how do you recover from the shared spinlock being "locked" but the process holding that spinlock crashed? 

I also want to know an answer. Despite the merged PR!

asfernandes:
> I also want to know an answer. Despite the merged PR!

I will look at this with more detail.

aafemt:
What will be wrong if the initialization of the shared memory structures (mutex) is done under file lock?

PR #8650, opened 2025-07-14
Fix codechecker warnings

AlomeProg:
Correction of some warnings of the static analyzer clang-tidy and clangsa.


src/common/IntlParametersBlock.h
```diff
@@ -56,13 +58,17 @@ class IntlDpb : public IntlParametersBlock
...
+	~IntlDpb() override = default;
 };
 
 class IntlSpb : public IntlParametersBlock
 {
 public:
 	TagType checkTag(UCHAR tag, const char** tagName);
 	UCHAR getUtf8Tag();
+
+	~IntlSpb() override = default;
```
asfernandes: The destructors with `override = default` should not be needed.

AlomeProg: And it really is. I checked and realized that they are optional. The main thing is to have a virtual destructor in the base class. I will fix this code soon. Thanks for the review.

src/common/ThreadStart.cpp
```diff
@@ -181,7 +181,7 @@ Thread Thread::start(ThreadEntryPoint* routine, void* arg, int priority_arg, Han
 	{
 #ifdef HAVE_PTHREAD_CANCEL
 		int dummy;		// We do not want to know old cancel type
-		state = pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, &dummy);
+		state = pthread_setcanceltype(PTHREAD_CANCEL_DEFERRED, &dummy);
```
asfernandes, line 185: What about this @AlexPeshkoff ?

src/common/IntlParametersBlock.h
```diff
@@ -53,29 +53,30 @@ class IntlParametersBlock
...
 public:
-	IntlSpbStart()
+	IntlSpbStart() noexcept
 		: mode(0)
 	{ }
 
-	TagType checkTag(UCHAR tag, const char** tagName);
-	UCHAR getUtf8Tag();
+	TagType checkTag(UCHAR tag, const char** tagName) noexcept override;
+	UCHAR getUtf8Tag() noexcept final;
```
aafemt, line 78: Whole `IntlSpbStart` is already declared `final` so existence of derived classes is impossible. Why did code checker issue this warning?

PR #8657, opened 2025-07-22
Switch to C++20

asfernandes:


TreeHunter9:
The only thing I can suggest prohibiting in C++20: modules :)

dyemanov (approved):

asfernandes:
> The only thing I can suggest prohibiting in C++20: modules :)

With current state of tooling, I agree.

PR #8664, opened 2025-07-23
Re-add indexes for object name without schema name, removing their uniqueness

asfernandes:


dyemanov:
I suggest to defer this PR until we resolve the known performance regressions in the existing code. Otherwise, new indices may hide these issues (allowing better plans than now), but strike back later when the new "backward compatibility" indices are removed.

asfernandes:
Do we still have these performance problems?

dyemanov:
> Do we still have these performance problems?

AFAIK no, but I will double-check the old cases and report back.

dyemanov:
I've double-checked all regression cases reported by Pavel after schemas were merged and found that all of them are currently fixed. We seem to still have performance issues, as reported [here](http://firebirdtest.com/oltp-emul-fb/), but I don't think they're related to system query plans and should be researched separately.

PR #8666, opened 2025-07-23
Fixed crash after calling incorrectly parametrized request

Ev3nt:
Init table:
```
CREATE TABLE dept (id NUMERIC(2), name VARCHAR(14), location VARCHAR(13));
```

If we execute procedure with incorrectrly paramerized request, we have an error (as expected):
```
EXECUTE BLOCK 
AS 
     DECLARE sql_stmt    VARCHAR(200);
     DECLARE dept_id     NUMERIC(2) = 50;
     DECLARE dept_name   TYPE OF COLUMN dept.name  = 'PERSONNEL';
     DECLARE location    TYPE OF COLUMN dept.location default 'DALLAS';
BEGIN
    sql_stmt = 'INSERT INTO dept VALUES (:A, :B, :C)';
    EXECUTE STATEMENT (:sql_stmt) (:dept_id, :dept_name, :location);
END;
```

> Statement failed, SQLSTATE = 42S22
> Dynamic SQL Error
> -SQL error code = -206
> -Column unknown
> -A
> -At line 1, column 26
> -At block line: 9, col: 5

But when we make it correctly and try incorrectly request again, we have a crash. Correctly request:
```
EXECUTE BLOCK 
AS 
     DECLARE sql_stmt    VARCHAR(200);
     DECLARE dept_id     NUMERIC(2) = 50;
     DECLARE dept_name   TYPE OF COLUMN dept.name  = 'PERSONNEL';
     DECLARE location    TYPE OF COLUMN dept.location default 'DALLAS';
BEGIN
    sql_stmt = 'INSERT INTO dept VALUES (:A, :B, :C)';
    EXECUTE STATEMENT (:sql_stmt) (A := :dept_id, B :=  :dept_name, C :=  :location);
END;
```

I added 2 conditions to check if names is not a nullptr.
Now it throws an exception:

> Statement failed, SQLSTATE = 42000
> Input parameter 'A' have no value set
> -At block line: 9, col: 5

PR #8670, opened 2025-07-24
Fixed crash after calling incorrectly parametrized request

Ev3nt:
Same as: https://github.com/FirebirdSQL/firebird/pull/8666


hvlad:
I've already ported #8666 into v4 and v6

PR #8680, opened 2025-08-02
Explicitly default the expected Windows API to 0x0601 (Windows 7)

mrotteveel:
Explicitly default the expected Windows API to 0x0601 (Windows 7)

Also:

* Reject versions before Windows 7 (0x0601)
* Ensure firebird.h is included first to avoid translation units unexpectedly using a different version or other config than defined in firebird.h
* Also raise libcds expected Windows API to Windows 7 (0x601) instead of Windows 2000 (0x0500)/Windows XP (0x501)
* Removes some inappropriate defines to very old Windows API versions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,7 +204,8 @@ if (MINGW)
     set(CMAKE_SHARED_MODULE_PREFIX)
     set(CMAKE_STATIC_LIBRARY_PREFIX)
 
-    add_definitions(-D_WIN32_WINNT=0x0600)
+    # 0x0A00 is Windows 10
+    add_definitions(-D_WIN32_WINNT=0x0A00)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4 -std=c++20")
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,8 +204,8 @@ if (MINGW)
     set(CMAKE_SHARED_MODULE_PREFIX)
     set(CMAKE_STATIC_LIBRARY_PREFIX)
 
-    # 0x0A00 is Windows 10
-    add_definitions(-D_WIN32_WINNT=0x0A00)
+    # 0x0601 is Windows 7 - see also comments in autoconfig_msvc.h
+    add_definitions(-D_WIN32_WINNT=0x0601)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4 -std=c++20")
 endif()
 
--- a/src/include/gen/autoconfig_msvc.h
+++ b/src/include/gen/autoconfig_msvc.h
@@ -6,8 +6,9 @@
 // use to this version.
 #define FB_WIN32_WINNT_BASELINE 0x0601
 
+// Normally, _WIN32_WINNT is defined in the PreprocessorDefinitions of Firebird.Common.props.
 #ifndef _WIN32_WINNT
-// Set _WIN32_WINNT to the baseline, but allow overriding
+// Fallback to set _WIN32_WINNT to the baseline.
 #define _WIN32_WINNT FB_WIN32_WINNT_BASELINE
 #elif (_WIN32_WINNT < FB_WIN32_WINNT_BASELINE)
 // Consider Windows versions before the baseline really too old.
--- a/src/include/gen/autoconfig_msvc.h
+++ b/src/include/gen/autoconfig_msvc.h
@@ -1,26 +1,17 @@
-/*
- * 2002.02.15 Sean Leyne - Code Cleanup, removed obsolete ports:
- *                          - MAC (MAC, MAC_AUX and "MAC_CP" defines)
- *                          - EPSON, DELTA, IMP, NCR3000, NeXT, M88K, Cray
- *                          - OS/2, Apollo
- *
- * 2002-02-23 Sean Leyne - Code Cleanup, removed old Win3.1 port (Windows_Only)
- *
- * 2002.10.27 Sean Leyne - Code Cleanup, removed obsolete "UNIXWARE" port
- * 2002.10.27 Sean Leyne - Code Cleanup, removed obsolete "Ultrix" port
- *
- * 2002.10.28 Sean Leyne - Completed removal of obsolete "DGUX" port
- * 2002.10.28 Sean Leyne - Code cleanup, removed obsolete "MPEXL" port
- * 2002.10.28 Sean Leyne - Code cleanup, removed obsolete "SGI" port
- *
- * 2002.10.29 Sean Leyne - Removed obsolete "Netware" port
- *
- * 2002.10.30 Sean Leyne - Removed support for obsolete "PC_PLATFORM" define
- *
- */
-
 #ifndef AUTOCONFIG_H
 #define AUTOCONFIG_H
+
+ // Windows 10 (0x0A00) is the baseline version for Firebird 6
+#define FB_WIN32_WINNT_BASELINE 0x0A00
+
+#ifndef _WIN32_WINNT
+// By default set _WIN32_WINNT to the baseline, but allows overriding (maybe for fbclient?)
+#define _WIN32_WINNT FB_WIN32_WINNT_BASELINE
+#elif (_WIN32_WINNT < 0x0601)
+// Consider Windows versions before Windows 7 (0x0601) really too old. This does not mean that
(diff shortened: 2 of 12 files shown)


mrotteveel:
Note that defaulting to Windows 10 (0x0A00) just formalizes and makes explicit what already happens inside the Windows 10 API headers when `_WIN32_WINNT` is _not_ defined, and gives us a point to control it for the future.

I have considered setting the yvalve project (fbclient) to explicitly use a lower `_WIN32_WINNT` value (e.g. 0x0601 for Windows 7), but did not do so at this time.

hvlad:
> Note that defaulting to Windows 10 (0x0A00) just formalizes and makes explicit what already happens inside the Windows 10 API headers when `_WIN32_WINNT` is _not_ defined, and gives us a point to control it for the future.

  I'm not sure it should be done this way. Ideally, `_WIN32_WINNT` should be set to the minimal possible value that allows code to compile. Or even not set, but asserted. This allow anyone to use different (older) Platform SDK to try build Firebird for own purposes (for ex. to ensure compatibility with pre-Win10).
 
> I have considered setting the yvalve project (fbclient) to explicitly use a lower `_WIN32_WINNT` value (e.g. 0x0601 for Windows 7), but did not do so at this time.

  This define will not make anything compatible with Win7. It doesn't affect CRT code, for example. Also, fbclient is not build from the single project.

  I.e. I support removal of current define's (3 or 4 places, iirc) but don't think we must explicitly set `_WIN32_WINNT`.
In some places it could be check'ed for the required minimal value, if really necessary.

hvlad:
Also, I see no need to change LibCDS as any other external library. At least until it builds.

mrotteveel:
> > Note that defaulting to Windows 10 (0x0A00) just formalizes and makes explicit what already happens inside the Windows 10 API headers when `_WIN32_WINNT` is _not_ defined, and gives us a point to control it for the future.
> 
> I'm not sure it should be done this way. Ideally, `_WIN32_WINNT` should be set to the minimal possible value that allows code to compile. Or even not set, but asserted. This allow anyone to use different (older) Platform SDK to try build Firebird for own purposes (for ex. to ensure compatibility with pre-Win10).

I agree it should be set to the minimum value that we expect it to compile, and I think that is the version that we _want_ to support, or want to be able to support.

I asked for input on firebird-devel, and the only response I got was from Paul, so I simply decided to put in the default that we're currently already using by way of how the Windows API decides if you do not explicitly specify it! If that should be a lower value, then so be it, I can change it to a lower value.

> > I have considered setting the yvalve project (fbclient) to explicitly use a lower `_WIN32_WINNT` value (e.g. 0x0601 for Windows 7), but did not do so at this time.
> 
> This define will not make anything compatible with Win7. It doesn't affect CRT code, for example. Also, fbclient is not build from the single project.

As I understand it, yvalve depends on the projects it needs but not on their build outputs, so I expect that will cause those projects to be built as needed using the configuration for the yvalve, but maybe I'm misunderstanding how msbuild works there.

> I.e. I support removal of current define's (3 or 4 places, iirc) but don't think we must explicitly set `_WIN32_WINNT`. In some places it could be check'ed for the required minimal value, if really necessary.

We should explicitly set `_WIN32_WINNT`, so that 1) all builds conform to what we want to use and be able to use and 2) compilation fails if you try to use a Windows API method that does not exist in the Windows versions we expect to support, or try to build using a too old version of the Windows API, and 3) it makes explicit what we want to support instead of the current situation that no one actually seems to know what Windows versions we intend to support (or at least allow Firebird to run, even if we don't support them).

Sure, setting it to Windows 10 might be "too high", but then suggest which version we _should_ target.

mrotteveel:
> Also, I see no need to change LibCDS as any other external library. At least until it builds.

As I mentioned on firebird-devel, LibCDS has some conditional code that does some detection or initialization differently for Windows Vista and higher, and for Windows 7 and higher, so raising it from some targets Windows 2000, some targets Windows XP to all targets Windows 7 seems rational to me.

Let me know if you want me to revert it.

src/include/gen/autoconfig_msvc.h
```diff
@@ -1,27 +1,18 @@
...
- *
- */
-
 #ifndef AUTOCONFIG_H
 #define AUTOCONFIG_H
 
+ // Windows 10 (0x0A00) is the baseline version for Firebird 6
+#define FB_WIN32_WINNT_BASELINE 0x0A00
+
+#ifndef _WIN32_WINNT
```
asfernandes, line 10: Global defines, IMO, should be defined in property sheets, and not depend that everyone includes `firebird.h` first.

mrotteveel: Given the other configuration that autoconfig_msvc.h does, it already needs to be first, otherwise you run the chance that some of it is not actually defined (e.g. the `WIN32_LEAN_AND_MEAN` and `NOxxx` defines) before the Windows API is included.

Do you want me to move those also to, for example, Firebird.Common.props?

hvlad:
> > > I have considered setting the yvalve project (fbclient) to explicitly use a lower `_WIN32_WINNT` value (e.g. 0x0601 for Windows 7), but did not do so at this time.
> > 
> > 
> > This define will not make anything compatible with Win7. It doesn't affect CRT code, for example. Also, fbclient is not build from the single project.
> 
> As I understand it, yvalve depends on the projects it needs but not on their build outputs, so I expect that will cause those projects to be built as needed using the configuration for the yvalve, but maybe I'm misunderstanding how msbuild works there.

Note value of [Link Library Dependencies](https://learn.microsoft.com/en-us/cpp/build/reference/linker-property-pages?view=msvc-170&f1url=%3FappId%3DDev17IDEF1%26l%3DEN-US%26k%3Dk(VC.Project.VCLinkerTool.LinkLibraryDependencies)%26rd%3Dtrue) setting (true).
`yvalve` depends on `common` and `remote` and linked with them.

> > I.e. I support removal of current define's (3 or 4 places, iirc) but don't think we must explicitly set `_WIN32_WINNT`. In some places it could be check'ed for the required minimal value, if really necessary.
> 
> We should explicitly set `_WIN32_WINNT`, so that 1) all builds conform to what we want to use and be able to use and 2) compilation fails if you try to use a Windows API method that does not exist in the Windows versions we expect to support, or try to build using a too old version of the Windows API, and 3) it makes explicit what we want to support instead of the current situation that no one actually seems to know what Windows versions we intend to support (or at least allow Firebird to run, even if we don't support them).
> 
> Sure, setting it to Windows 10 might be "too high", but then suggest which version we _should_ target.

0x0601 (Win7), I think.

hvlad:
> > Also, I see no need to change LibCDS as any other external library. At least until it builds.
> 
> As I mentioned on firebird-devel, LibCDS has some conditional code that does some detection or initialization differently for Windows Vista and higher, and for Windows 7 and higher, so raising it from some targets Windows 2000, some targets Windows XP to all targets Windows 7 seems rational to me.
> 
> Let me know if you want me to revert it.

After more close look at the LibCDS code and tend to agree with you here.

src/include/gen/autoconfig_msvc.h
hvlad: These defines are truly optional and I prefer to not pollute .props with it.

mrotteveel: OK, but should I define `_WIN32_WINNT` in Firebird.Common.props?

hvlad: > OK, but should I define `_WIN32_WINNT` in Firebird.Common.props?

I think - yes, but, please, add comment about it at autoconfig_msvc.h

CMakeLists.txt
```diff
@@ -204,7 +204,8 @@ if (MINGW)
     set(CMAKE_SHARED_MODULE_PREFIX)
     set(CMAKE_STATIC_LIBRARY_PREFIX)
 
-    add_definitions(-D_WIN32_WINNT=0x0600)
+    # 0x0A00 is Windows 10
+    add_definitions(-D_WIN32_WINNT=0x0A00)
```
hvlad: Should it be 0x0601 ?

src/include/gen/autoconfig_msvc.h
mrotteveel: OK, done

mrotteveel:
Can this be merged, @asfernandes @hvlad ?

hvlad (approved):

asfernandes (approved):

Issue #7632, opened 2023-06-16
Support ordered-set (WITHIN GROUP) aggregates.


sim1984:
Need to add support for ordered-set and hypothetical-set aggregate functions, as well as implementations of the instances defined in SQL:2008 (percentile_cont(), percentile_disc(), rank(), dense_rank(), percent_rank(), cume_dist()).  It is good to added mode() though it is not in the spec. In addition, you can gracefully solve the problem of element unordering in the LIST aggregate function (in the SQL standard LISTAGG).

Ordered-Set Aggregate Functions
```
PERCENTILE_CONT( <expr> ) WITHIN GROUP ( ORDER BY <order_expr> )
PERCENTILE_DISC( <expr> ) WITHIN GROUP ( ORDER BY <order_expr> )
LISTAGG( <expr>, <separator> ) WITHIN GROUP ( ORDER BY <order_expr> )
```

It is also possible to introduce the LISTAGG function, and make LIST an alias for compatibility with the standard.

Hypothetical-Set Aggregate Functions
```
RANK( <args> ) WITHIN GROUP ( ORDER BY <order_exprs> )
DENSE_RANK( <args> ) WITHIN GROUP ( ORDER BY <order_exprs> )
PERCENT_RANK( <args> ) WITHIN GROUP ( ORDER BY <order_exprs> )
CUME_DIST( <expr> ) WITHIN GROUP ( ORDER BY <order_exprs> )
```

You can read more about how this is done in PostgreSQL here.
https://www.postgresql.org/docs/current/functions-aggregate.html
https://paquier.xyz/postgresql-2/postgres-9-4-feature-highlight-within-group/

dyemanov:
If `LISTAGG` is added for standard compatibility, should legacy `LIST` also support ordering? If yes, should it follow the standard syntax (so that `LIST` and `LISTAGG` just become synonyms) or, being non-standard, may use its own syntax for ordring, e.g. `LIST(FIELD1 ORDER BY FIELD2)`?

sim1984:
I think it's enough to make it just a synonym. But if they make their own version to simplify, then do not  object.

dyemanov:
The standard `LISTAGG` also has the separator string being mandatory while our `LIST` implementation has it optional, with a comma implied if omitted. And the standard `LISTAGG` defines the separator as a literal only, while we allow arbitrary value expressions. Would it be OK to preserve these differences for the `LISTAGG` too?

[...]

PR #8689, opened 2025-08-06
New function LISTAGG

ChudaykinAlex:
1) Purpose
The current implementation has an aggregate function LIST which concatenates multiple row fields into a blob. The SQL standard has a similar function called LISTAGG. The major difference is that it also supports the ordered concatenation.

2) Syntax and rules
```
<listagg set function> ::=
LISTAGG <left paren> [ <set quantifier> ] <character value expression> <comma> <listagg separator> [ <listagg overflow clause> ] <right paren> <within group specification>

<listagg separator> ::= 
<character string literal>

<listagg overflow clause> ::=
ON OVERFLOW <overflow behavior>

<overflow behavior> ::=
ERROR | TRUNCATE [ <listagg truncation filler> ] <listagg count indication>

<listagg truncation filler> ::=
<character string literal>

<listagg count indication> ::=
WITH COUNT | WITHOUT COUNT

<within group specification> ::=
WITHIN GROUP <left paren> ORDER BY <sort specification list> <right paren>
```
The legacy LIST syntax is preserved for backward compatibility, LISTAGG is added to cover the standard features.

There is a ```<listagg overflow clause>``` rule in the standard, which is intended to output an error when the output value overflows. Since the LIST function always returns a BLOB, it was decided that this rule would be meaningless. It was not implemented and silently ignored if specified.

If DISTINCT is specified for LISTAGG, then ORDER BY ```<sort specification list>``` must fully match ```<character value expression>```

If DISTINCT is specified, the presence of WITHIN GROUP must obey the restriction and will not affect the subsequent code execution.

3) Examples:
```
CREATE TABLE TEST_T
	(COL1 INT, COL2 VARCHAR(2), COL3 VARCHAR(2), COL4 VARCHAR(2), COL5 BOOLEAN, COL6 VARCHAR(2)
	CHARACTER SET WIN1251);
COMMIT;
INSERT INTO TEST_T values(1, 'A', 'A', 'J', false, 'П');
INSERT INTO TEST_T values(2, 'B', 'B', 'I', false, 'Д');
INSERT INTO TEST_T values(3, 'C', 'A', 'L', true,  'Ж');
INSERT INTO TEST_T values(4, 'D', 'B', 'K', true,  'Й');
COMMIT;

SELECT LISTAGG (ALL COL4, ':') AS FROM TEST_T;
=======
J:I:L:K

SELECT LISTAGG (DISTINCT COL4, ':') FROM TEST_T;
========
I:J:K:L

SELECT LISTAGG (DISTINCT COL3, ':') FROM TEST_T;
====
A:B

SELECT LISTAGG (DISTINCT COL3, ':') WITHIN GROUP (ORDER BY COL2) FROM TEST_T;
====
A:B

SELECT LISTAGG (DISTINCT COL3, ':') WITHIN GROUP (ORDER BY COL2 DESCENDING) FROM TEST_T;
====
A:B

SELECT LISTAGG (COL2, ':') WITHIN GROUP (ORDER BY COL2 DESCENDING) FROM TEST_T;
=======
D:C:B:A

SELECT LISTAGG (COL4, ':') WITHIN GROUP (ORDER BY COL3 DESC) FROM TEST_T;
=======
I:K:J:L

SELECT LISTAGG (COL3, ':') WITHIN GROUP (ORDER BY COL5 ASCENDING) FROM TEST_T;
=======
A:B:A:B

SELECT LISTAGG (COL4, ':') WITHIN GROUP (ORDER BY COL3 ASC) FROM TEST_T;
=======
J:L:I:K

SELECT LISTAGG (ALL COL2) WITHIN GROUP (ORDER BY COL4) FROM TEST_T;
=======
B,A,D,C

SELECT LISTAGG (COL2, ':') WITHIN GROUP (ORDER BY COL3 DESC, COL4 ASC) FROM TEST_T;
=======
B:D:A:C

SELECT LISTAGG (COL2, ':') WITHIN GROUP (ORDER BY COL3 DESC, COL4 DESC) FROM TEST_T;
=======
D:B:C:A

SELECT LISTAGG (COL2, ':') WITHIN GROUP (ORDER BY COL3 ASC, COL4 DESC) FROM TEST_T;
=======
C:A:D:B

SELECT LISTAGG (ALL COL6, ':')FROM TEST_T;
=======
П:Д:Ж:Й

SELECT LISTAGG (ALL COL6, ':') WITHIN GROUP (ORDER BY COL2 DESC) FROM TEST_T;
=======
Й:Ж:Д:П

SELECT LISTAGG (ALL COL2, ':') WITHIN GROUP (ORDER BY COL6) FROM TEST_T;
=======
B:C:D:A

INSERT INTO TEST_T values(5, 'E', NULL, NULL, NULL, NULL);
INSERT INTO TEST_T values(6, 'F', 'C', 'N', true, 'К');

SELECT LISTAGG (ALL COL2, ':') WITHIN GROUP (ORDER BY COL3) FROM TEST_T;
===========
E:A:C:B:D:F

SELECT LISTAGG (ALL COL2, ':') WITHIN GROUP (ORDER BY COL3 NULLS LAST) FROM TEST_T;
===========
A:C:B:D:F:E

SELECT LISTAGG (ALL COL2, ':') WITHIN GROUP (ORDER BY COL6 NULLS FIRST) FROM TEST_T;
===========
E:B:C:D:F:A

SELECT LISTAGG (DISTINCT COL3, ':') WITHIN GROUP (ORDER BY COL2) FROM TEST_T;
========
SQL error code = -104
-Invalid command
-Sort-key of the ORDER BY specification must match the argument list
```


sim1984:
This PR implements one of the aggregate functions (LISTAGG), which depends on the order of the input stream records. The full list can be seen here: https://github.com/FirebirdSQL/firebird/issues/7632

dyemanov (approved):

doc/sql.extensions/README.listagg
```diff
@@ -0,0 +1,144 @@
...
+	(COL1 INT, COL2 VARCHAR(2), COL3 VARCHAR(2), COL4 VARCHAR(2), COL5 BOOLEAN, COL6 VARCHAR(2)
+	CHARACTER SET WIN1251);
+COMMIT;
+INSERT INTO TEST_T values(1, 'A', 'A', 'J', false, 'П');
+INSERT INTO TEST_T values(2, 'B', 'B', 'I', false, 'Д');
+INSERT INTO TEST_T values(3, 'C', 'A', 'L', true,  'Ж');
+INSERT INTO TEST_T values(4, 'D', 'B', 'K', true,  'Й');
+COMMIT;
+
+SELECT LISTAGG (ALL COL4, ':') AS FROM TEST_T;
```
asfernandes, line 53: In the syntax format section, `<within group specification>` is mandatory but does not exist in some examples.

doc/sql.extensions/README.listagg
```diff
@@ -0,0 +1,144 @@
...
+
+SELECT LISTAGG (COL2, ':') WITHIN GROUP (ORDER BY COL3 DESC, COL4 DESC) FROM TEST_T;
+=======
+D:B:C:A
+
+SELECT LISTAGG (COL2, ':') WITHIN GROUP (ORDER BY COL3 ASC, COL4 DESC) FROM TEST_T;
+=======
+C:A:D:B
+
+SELECT LISTAGG (ALL COL6, ':')FROM TEST_T;
```
asfernandes:
```suggestion
SELECT LISTAGG (ALL COL6, ':') FROM TEST_T;
```

src/dsql/AggNodes.cpp
```diff
@@ -427,6 +430,44 @@ bool AggNode::aggPass(thread_db* tdbb, Request* request) const
...
+
+			MOVE_CLEAR(data, asb->length);
+
+			auto descOrder = asb->descOrder.begin();
+			auto keyItem = asb->keyItems.begin();
+
+			for (auto& nodeOrder : sort->expressions)
+			{
+				dsc toDesc = *(descOrder++);
+				toDesc.dsc_address = data + (IPTR)toDesc.dsc_address;
```
asfernandes:
```suggestion
				toDesc.dsc_address = data + (IPTR) toDesc.dsc_address;
```

src/dsql/AggNodes.cpp
```diff
@@ -427,6 +430,44 @@ bool AggNode::aggPass(thread_db* tdbb, Request* request) const
...
+				if (const auto fromDsc = EVL_expr(tdbb, request, nodeOrder))
+				{
+					if (IS_INTL_DATA(fromDsc))
+						INTL_string_to_key(tdbb, INTL_TEXT_TO_INDEX(fromDsc->getTextType()),
+							fromDsc, &toDesc, INTL_KEY_UNIQUE);
```
asfernandes, line 457:
```suggestion
					if (IS_INTL_DATA(fromDsc))
					{
						INTL_string_to_key(tdbb, INTL_TEXT_TO_INDEX(fromDsc->getTextType()),
							fromDsc, &toDesc, INTL_KEY_UNIQUE);
					}
```

src/dsql/AggNodes.cpp
```diff
@@ -427,6 +430,44 @@ bool AggNode::aggPass(thread_db* tdbb, Request* request) const
...
+				}
+				else
+					*(data + keyItem->getSkdOffset()) = TRUE;
+
+				// The first key for NULLS FIRST/LAST, the second key for the sorter
+				keyItem += 2;
+			}
+
+			dsc toDesc = asb->desc;
+			toDesc.dsc_address = data + (IPTR)toDesc.dsc_address;
```
asfernandes:
```suggestion
			toDesc.dsc_address = data + (IPTR) toDesc.dsc_address;
```

src/dsql/AggNodes.cpp
```diff
@@ -478,7 +519,10 @@ dsc* AggNode::execute(thread_db* tdbb, Request* request) const
 				break;
 			}
 
-			desc.dsc_address = data + (asb->intl ? asb->keyItems[1].getSkdOffset() : 0);
+			if (distinct)
+				desc.dsc_address = data + (asb->intl ? asb->keyItems[1].getSkdOffset() : 0);
+			else
+				desc.dsc_address = data + (IPTR)asb->desc.dsc_address;
```
asfernandes:
```suggestion
				desc.dsc_address = data + (IPTR) asb->desc.dsc_address;
```

src/dsql/AggNodes.cpp
```diff
@@ -899,6 +944,26 @@ void ListAggNode::make(DsqlCompilerScratch* dsqlScratch, dsc* desc)
...
+	AggNode::genBlr(dsqlScratch);
+	GEN_sort(dsqlScratch, blr_sort, dsqlOrderClause);
+}
+
+AggNode* ListAggNode::pass1(thread_db* tdbb, CompilerScratch* csb)
+{
+	if (sort && distinct)
+	{
+		ValueExprNode* const sortNode = *sort->expressions.begin();
+		if (!arg->sameAs(sortNode, false) || sort->expressions.getCount() > 1)
```
asfernandes: Should they be identical? Why?
I think it should have like GROUP BY rules.

src/jrd/optimizer/Optimizer.cpp
```diff
@@ -1418,7 +1418,120 @@ void Optimizer::generateAggregateDistincts(MapNode* map)
...
+	dsc descriptor;
+	dsc* desc = &descriptor;
+
+	const auto asb = FB_NEW_POOL(getPool()) AggregateSort(getPool());
+
+	sort_key_def* prevKey = nullptr;
+	const auto keyCount = aggNode->sort->expressions.getCount() * 2;
+	sort_key_def* sortKey = asb->keyItems.getBuffer(keyCount);
+
+	auto const* direction = aggNode->sort->direction.begin();
```
asfernandes: `const auto*` for consistency, please.

src/dsql/AggNodes.cpp
```diff
@@ -307,7 +308,7 @@ bool AggNode::dsqlMatch(DsqlCompilerScratch* dsqlScratch, const ExprNode* other,
 	// ASF: We compare name address. That should be ok, as we have only one AggInfo instance
 	// per function.
 	return aggInfo.blr == o->aggInfo.blr && aggInfo.name == o->aggInfo.name &&
-		distinct == o->distinct && dialect1 == o->dialect1;
+		distinct == o->distinct && dialect1 == o->dialect1 && sort == o->sort;;
```
asfernandes: Does not look correct the comparation of pointer address here for `sort`.

dyemanov: @ChudaykinAlex IMO, the sort node should be added to the list by `ListAggNode::getChildren() `-- this way you may remove the sort comparison in `dsqlMatch()` and also remove `doPass2(sort)` in `pass2()`, as the sort node will be processed by the inherited methods automagically.

src/dsql/AggNodes.cpp
dyemanov: Good point. Currently, we sort only once and this is a good bonus. If we follow your suggestion and allow slightly different expressions, then we should either sort twice or ignore the user-specified ordering after `DISTINCT`. BTW, in this PR it also seems to be ignored -- `LISTAGG(DISTINCT COL) WITHIN GROUP (ORDER BY COL DESC)` would produce ASC-ordered output. But I suppose it should work the same way as for plain `SELECT DISTINCT(COL) FROM T ORDER BY COL DESC`, i.e. respect the `ORDER BY ` ordering and optimize the sorts (merge two sorts into one) only if they fully (expressions / directions / NULLs placement) match each other.

ChudaykinAlex:
Thanks for the recommendations, I'll fix it soon.

src/dsql/AggNodes.cpp
aafemt: Why direction and null placement are important for DISTINCT?

dyemanov: Nulls placement is not important, I agree, as NULLs are skipped by all aggregate functions.

dyemanov: By standard, `DISTINCT` eliminates duplicates in the ordered (if specified) result set, it should not change the user-defined ordering. Why do you think `LISTAGG` should behave differently?

aafemt: It shouldn't but I don't quite understand why you said

> i.e. respect the ORDER BY ordering and optimize the sorts (merge two sorts into one) **only if they fully** (expressions / directions / NULLs placement) match each other.


BTW, IMHO, `sort->expressions.getCount() > 1` condition here is not needed as it is fine for distinct to use only first sorting segment.

dyemanov: Ah, sorry, my bad. Surely, direction for the combined sort should be taken from `ORDER BY` -- like we already do in `Optimizer::checkSorts()`.

asfernandes: > Nulls placement is not important, I agree, as NULLs are skipped by all aggregate functions.

With my suggestion to use same existing rules, the DISTINCT LISTAGG expression may be something like `COALESCE(field, 'z') with a `ORDER BY Z`.

doc/sql.extensions/README.listagg
dyemanov: The SQL specification declares `<within group specification>` as mandatory. However, IMHO it's quite restrictive and neither Oracle nor DB2 follows that rule, they have it optional. Given that `LIST` and `LISTAGG` share the same syntax in this PR, we've also made `<within group specification>` optional. So the easiest solution is to fix the README ;-)

Or we may go the standard way and separate the legacy `LIST` (leave it with the current grammar, without ordering) from `LISTAGG` (which is strictly standard-compliant) at the parser level. But IMHO it would be annoying for users to select either of them depending on whether you need ordering or not. So personally I'd keep everything "as is" and just fix the docs.

Other opinions?

sim1984: I agree, `LIST` and `LISTAGG` should be complete synonyms.

I would also remove the mention of `ON OVERFLOW` from the documentation. It's standard, but we don't support it. It could be mentioned if it were simply ignored, but mentioning it leads to errors.

```sql
SELECT
LISTAGG(TRIM(RDB$RELATION_NAME), ';' ON OVERFLOW ERROR) WITHIN GROUP(ORDER BY RDB$RELATION_NAME) AS REL_NAMES
FROM RDB$RELATIONS;
```

```
Invalid token.
Dynamic SQL Error.
SQL error code = -104.
Token unknown - line 2, column 52.
ERROR.
----------------------------------
SQLCODE: -104
SQLSTATE: 42000
GDSCODE: 335544569
```

```sql
SELECT 
LISTAGG(TRIM(RDB$RELATION_NAME), ';' ON OVERFLOW TRUNCATE '...' WITHOUT COUNT) WITHIN GROUP(ORDER BY RDB$RELATION_NAME) AS REL_NAMES
FROM RDB$RELATIONS;
```

```
Invalid token.
Dynamic SQL Error.
SQL error code = -104.
Token unknown - line 2, column 52.
TRUNCATE.
----------------------------------
SQLCODE: -104
SQLSTATE: 42000
GDSCODE: 335544569
```

ChudaykinAlex: Strange. I need to check it out

dyemanov: If we remove `ON OVERFLOW` from the docs, then I believe we should remove it from the parser too.

sim1984: `ON OVERFLOW` can be kept if it will not cause errors.

src/dsql/AggNodes.cpp
dyemanov: Given that we support queries like:

`select distinct emp_no from employee order by last_name`

I'm not sure we need to introduce the `GROUP BY` rules inside `LISTAGG (DISTINCT ...) ... WITHIN GROUP`. I suppose `DISTINCT + ORDER BY` should work the same way as either a standalone construct or within the `LISTAGG` group, i.e. without any restrictions. That said, I see no practical point in the sort behind the `ORDER BY <other expression>` clause because its result is not going to be user-visible anyway. It basically means that if `DISTINCT` is present, we may simply ignore the `ORDER BY` clause -- except when `ORDER BY` mentions the aggregated expression in the very beginning of its `<sort specification list>` -- then its sort direction should be taken into account inside `DISTINCT`. 

Do I miss anything?

src/dsql/AggNodes.cpp
```diff
@@ -877,28 +923,49 @@ AggNode* AvgAggNode::dsqlCopy(DsqlCompilerScratch* dsqlScratch) /*const*/
...
 }
 
 DmlNode* ListAggNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScratch* csb, const UCHAR blrOp)
 {
-	ListAggNode* node = FB_NEW_POOL(pool) ListAggNode(pool,
-		(blrOp == blr_agg_list_distinct));
+	ListAggNode* node = FB_NEW_POOL(pool) ListAggNode(pool,	(blrOp == blr_agg_list_distinct));
 	node->arg = PAR_parse_value(tdbb, csb);
 	node->delimiter = PAR_parse_value(tdbb, csb);
+	node->sort = PAR_sort(tdbb, csb, blr_sort, true);
```
dyemanov: This will fail while parsing the BLR restored from the previous versions, because `blr_sort` will be missing.
One solution is to add `blr_agg_list2` and use it only when the ordering is specified. Another could be something like this:

```cpp
if (csb->csb_blr_reader.peekByte() == blr_sort)
	node->sort = PAR_sort(tdbb, csb, blr_sort, true);
```
In this case, I'd also change `genBlr()` below to call `GEN_sort()` only if the ordering is specified -- this way we keep generating a compatible BLR for the legacy `LIST` syntax, making it possible to downgrade the database if required.

pavel-zotov:
```
echo set list on; select listagg('q','#') from (select 1 i from rdb$types rows 3); | isql /:employee -z
...
WI-T6.0.0.1356 Firebird 6.0 Initial/tcp (PZ)/P20:C

LIST                            0:1
q#q#q
```

Why LISTAGG() function is shown as ```LIST``` ?

pavel-zotov:
::: QA note :::
All related tests see in: 
```$QA_HOME/tests/functional/intfunc/listagg/test_*.py```

PR #8697, opened 2025-08-12
fix(pio): Replace 32-bit offset calculation with 64-bit in `PIO_extend`

TreeHunter9:
Replace 32-bit arguments with 64-bit ones to prevent overflow when handling large files (>4GB).
The original implementation incorrectly calculated offsets as 32-bit values, which made it impossible to extend the file via `fallocate` after it reached 4GB in size.

As I can see, every version with `PIO_extend` is affected.

### DIFF
--- a/src/jrd/os/posix/unix.cpp
+++ b/src/jrd/os/posix/unix.cpp
@@ -320,13 +320,16 @@ void PIO_extend(thread_db* tdbb, jrd_file* file, const ULONG extPages, const USH
 	if (file->fil_flags & FIL_no_fast_extend)
 		return;
 
-	const off_t filePages = PIO_get_number_of_pages(file, pageSize);
-	const off_t extendBy = MIN(MAX_ULONG - filePages, extPages);
+	const ULONG filePages = PIO_get_number_of_pages(file, pageSize);
+	const ULONG extendBy = MIN(MAX_ULONG - filePages, extPages);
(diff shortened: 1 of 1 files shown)


src/jrd/os/posix/unix.cpp
```diff
@@ -320,8 +320,8 @@ void PIO_extend(thread_db* tdbb, jrd_file* file, const ULONG extPages, const USH
 	if (file->fil_flags & FIL_no_fast_extend)
 		return;
 
-	const ULONG filePages = PIO_get_number_of_pages(file, pageSize);
-	const ULONG extendBy = MIN(MAX_ULONG - filePages, extPages);
+	const off_t filePages = PIO_get_number_of_pages(file, pageSize);
+	const off_t extendBy = MIN(MAX_ULONG - filePages, extPages);
```
dyemanov: While technically this is correct, both variables actually represent number of pages and thus cannot exceed `ULONG`. The problem arrives when the byte offset is calculated using multiplication inside the `fallocate()` call. So I'd rather add a cast there, or better add new `off_t` variables for offset (`(off_t) filePages * pageSize`) and length (`(off_t) extendBy * pageSize`) -- similarly to how it's done in the Windows version of `PIO_extend()`.

dyemanov (approved):

PR #8717, opened 2025-08-28
Decouple LockManager from engine

asfernandes:


src/jrd/lck.h
```diff
@@ -86,6 +86,22 @@ enum lck_owner_t {
...
 
+class LockManagerEngineCallbacks final : public LockManager::Callbacks
+{
+public:
+	LockManagerEngineCallbacks(thread_db* aTdbb)
+		: tdbb(aTdbb)
+	{
+	}
+
+	ISC_STATUS getCancelState() const override;
```
Noremos, line 97: If the class is marked final, perhaps it would be better to mark the methods final as well?

aafemt:
Can templates be used here instead of callback parameter to resolve interface at compiler-time instead of run-time?

asfernandes:
> Can templates be used here instead of callback parameter to resolve interface at compiler-time instead of run-time?

First, show numbers of what is the performance penalty introduced.

hvlad:
> Can templates be used here instead of callback parameter to resolve interface at compiler-time instead of run-time?

No.

src/jrd/lck.h
asfernandes: That can be inferred by the compiler.

asfernandes:
I don't know what's happening with the android build. Maybe a compiler issue with the host machine compiler. This first phase seems to work locally.

hvlad:
> I don't know what's happening with the android build. Maybe a compiler issue with the host machine compiler. This first phase seems to work locally.

Is it possible to produce stack backtrace at the compiler host and extract it from there for analysis ?

hvlad:
> Is it possible to produce stack backtrace at the compiler host and extract it from there for analysis ?

Sorry, I wrongly thought it is an engine crashed. Actually, it is clang compiler... weird

asfernandes:
> > Is it possible to produce stack backtrace at the compiler host and extract it from there for analysis ?
> 
> Sorry, I wrongly thought it is an engine crashed. Actually, it is clang compiler... weird

Trying NDK update.

asfernandes:
> > Is it possible to produce stack backtrace at the compiler host and extract it from there for analysis ?
> 
> Sorry, I wrongly thought it is an engine crashed. Actually, it is clang compiler... weird

I think it's not clang (NDK), but GPRE running that is causing the problem. That's why make deletes temp/Release/burp/OdsDetection.cpp.
I'm now trying to update the runners so the host x64 compiler is upgraded.

src/jrd/tests/LockManagerTest.cpp
```diff
@@ -0,0 +1,124 @@
...
+	{
+	public:
+		ISC_STATUS getCancelState() const
+		{
+			return 0;
+		}
+
+		ULONG adjustWait(ULONG wait) const
+		{
+			return 0;
```
hvlad:
```suggestion
			return wait;
```

TreeHunter9:
> Can templates be used here instead of callback parameter to resolve interface at compiler-time instead of run-time?

I guess templates can be used here by converting this whole thing to header only, which is definitely not worth it, and performance increase here will be non-existent due to other heavy work being done in functions, so overhead of virtual calls would be >0.1%.
And Adriano is right, for this macro optimization, measurements need to be done first.

src/jrd/tests/LockManagerTest.cpp
```diff
@@ -0,0 +1,124 @@
...
+	std::vector<std::thread> threads;
+	std::latch latch(THREAD_COUNT);
+	std::mutex mutex;
+
+	for (unsigned threadNum = 0u; threadNum < THREAD_COUNT; ++threadNum)
+	{
+		threads.emplace_back([&]() {
+			const UCHAR LOCK_KEY[] = {'1'};
+			FbLocalStatus statusVector;
+			LOCK_OWNER_T ownerId = threadNum + 1;
```
hvlad: Here many threads could see the same value of `threadNum`, in my case all threads sees value 8.
Using an separate `atomic_int` to generate `ownerId` and test passed.

TreeHunter9: I think the idea was to take `threadNum` from the loop, and this single loop will increment `threadNum`, so no race should be here, but it was captured by reference, so yea, it lead to race condition. I guess capture `threadNum` by value will fix the issue, without atomics.

asfernandes: Yes @TreeHunter9 . Thanks for catching this, @hvlad and sorry for the noise. It seems to work as well in Linux. I will improve this test and make others.

hvlad:
The crash on Android build makes me to ask - why it is different from other platforms ? 
I.e. why it run blocking action in single process ? 
Or are there more than a single process accessing same database ?  
Another (previous) gpre instance ?

asfernandes:
The crash initially didn't happened when I removed parallel make (-j4). I caught the problem in the core dump.
But all platforms are using parallel make.

PR #8704, opened 2025-08-17
Fix #8695 - Increase Android page size to 16KB

asfernandes:

PR #8729, opened 2025-09-04
Add 16KB page size support for Android builds

alexander-zubkov:
### What was done
- Updated build configuration to align segments to 16KB memory pages.
- This ensures compatibility with newer Android devices that use 16KB page size.

### Why
Without this change, the fbclient library fails to load on devices with 16KB memory pages.

### How it was tested
- Verified using `llvm-readelf -l libfbclient.so` on the built artifacts — alignment shows `0x4000` (16KB).
- Deployed on Android arm64 device — the application runs without crashes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -479,6 +479,7 @@ jobs:
 
       - name: Build
         run: |
+          export LDFLAGS="-Wl,-z,max-page-size=16384"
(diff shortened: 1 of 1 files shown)


asfernandes:
Don't you see there is https://github.com/FirebirdSQL/firebird/pull/8704?

alexander-zubkov:
Hi! Thanks for your PR #8704. We see that it almost solves the issue but needs a bit of finishing. We'll close our PR to avoid duplicating work. Could you please complete yours?

We really appreciate your contribution!

PR #8735, opened 2025-09-08
Possible fix for #8734

aafemt:


src/dsql/DsqlRequests.cpp
```diff
@@ -596,19 +596,28 @@ void DsqlDmlRequest::doExecute(thread_db* tdbb, jrd_tra** traHandle,
...
+					{
+						status_exception::raise(Arg::Gds(isc_stream_eof));
+					}
 
-				std::unique_ptr<UCHAR[]> message_buffer(new UCHAR[outMsgLength]);
+					// Create a temp message buffer and try one more receive.
+					// If it succeed then the next record exists.
 
-				JRD_receive(tdbb, request, message->msg_number, outMsgLength, message_buffer.get());
+					std::unique_ptr<UCHAR[]> message_buffer(new UCHAR[outMsgLength]);
```
hvlad: HalfStaticArray, pls

aafemt: Size of the message is known at this point, so usage of `HalfStaticArray` will just waste space on stack or cause stack overflow if this size is used as `InlineCapacity`.

hvlad: Do you think that `outMsgLength` could be used as `InlineCapacity` template argument ?
Looks like you not understand purpose and usage of `HalfStaticArray`.

aafemt: > Do you think that `outMsgLength` could be used as `InlineCapacity` template argument ?

No.



> Looks like you not understand purpose and usage of `HalfStaticArray`.

Yes. Especially in this place which is executed exactly once per query.

hvlad (changes requested):
Use HalfStaticArray to avoid unnecessary pool allocation in most cases

AlexPeshkoff:
> Use HalfStaticArray to avoid unnecessary pool allocation in most cases

+1

src/dsql/DsqlRequests.cpp
```diff
@@ -596,19 +596,28 @@ void DsqlDmlRequest::doExecute(thread_db* tdbb, jrd_tra** traHandle,
...
+					// If it succeed then the next record exists.
 
-				JRD_receive(tdbb, request, message->msg_number, outMsgLength, message_buffer.get());
+					HalfStaticArray<UCHAR, BUFFER_SMALL> message_buffer(getPool(), outMsgLength);
 
-				// Still active request means that second record exists
-				if ((request->req_flags & req_active))
-				{
-					status_exception::raise(Arg::Gds(isc_sing_select_err));
+					JRD_receive(tdbb, request, message->msg_number, outMsgLength, message_buffer.getBuffer(outMsgLength, false));
```
hvlad: Buffer space already reserved in ctor at line 612, no need to do it again.
`message_buffer.begin()` is enough here.

src/dsql/DsqlRequests.cpp
```diff
@@ -596,19 +596,28 @@ void DsqlDmlRequest::doExecute(thread_db* tdbb, jrd_tra** traHandle,
 
 			// if this is a singleton select that return some data, make sure there's in fact one record
 
-			if (singleton && (request->req_flags & req_active) && outMsgLength > 0)
+			if (singleton)
 			{
-				// Create a temp message buffer and try one more receive.
-				// If it succeed then the next record exists.
+				if (outMsgLength > 0)
```
hvlad: I don't understand why there is check for `outMsgLength > 0` - is it possible to be false when `singleton == true` ?
Even if out message with zero length is possible (???), what should happens for the `singleton && outMsgLength == 0` case ?
This place is very unclear for me.
In v5 there was just check for `singleton == true`, btw

aafemt: Just in case. If the function was called with `outMetadata == nullptr` - just assume success. The difference from v5 is that buffers here are provided by application rather than of engine itself.

hvlad: Is it possible to have `singleton == true && outMetadata == nullptr` ?
In any case, why two nested `if` here ?

aafemt: Initially check for request's activity was out of inner `if`. Then I realized that some BLR code can intentionally run to the end without entering `blr_send` and an application aware of it can provide no output buffer.

If you are sure that I'm wrong feel free to change this piece later.

hvlad: I'm sure that this code is not clear. And you changed it this way. 
Formally, 
```
if (a) {
  if (b) {
...
  }
}

```
is the same as 
```
if (a && b) {
...
}
```
thus I ask why you choose first one. If there is no real reason, change it.
If there is a reasons - provide it, and put in comments to avoid similar misunderstanding for the future readers.

hvlad (approved):
Much better now.

dyemanov (approved):

PR #8740, opened 2025-09-12
Replace wrong uses of SQLstate class 08 (connection error) (#8739)

### LINKED ISSUE #8739: Wrong SQLSTATE in case of table alias conflict
Hi *!

SELECT *
FROM rdb$database a
    JOIN rdb$database b ON a.rdb$relation_id = b.rdb$relation_id
    JOIN rdb$database b ON a.rdb$relation_id = b.rdb$relation_id

Undefined name.
Dynamic SQL Error.
SQL error code = -204.
alias B conflicts with an alias in the same statement.

SQLCODE: -204
**SQLSTATE: 08001**
GDSCODE: 335544569

Because of this org.firebirdsql.jaybird.xca.FatalErrorHelper.isFatal() returns true for this error and closes the Connection.
(FATAL_SQL_STATE_CLASSES / SQLSTATE_CLASS_CONNECTION_ERROR)

FB 3.x, 4.x and 5.x are all affected.

mrotteveel:
Fix for #8739

Except for `bad_db_handle`, I've replaced the invalid handle errors that were associated with `08003` with `HY000` (*CLI-specific condition - invalid handle* as defined in ISO/IEC 9075-9), as already used by for example `bad_svc_handle`. We could also opt to define a custom subclass of `HY` for our use (the standard explicitly allows this, as the *invalid handle* error has no subclass, which means it must either be `000`, or implementation-specific, i.e. a subclass starting with 5-9 or I-Z), as HY000 is pretty generic. 

One could also argue that it can be `42000` (*syntax error or access rule violation* as defined in ISO/IEC 9075-2), as already used by `bad_req_handle`, `bad_segstr_handle`.

The error `alias_conflict_err` of the original issues was changed to `42000`, just like its siblings `procedure_conflict_error` and `relation_conflict_error` already used.

The error `batch_open` was changed from `08002` to `HY010` (*CLI-specific condition - function sequence error* as defined in ISO/IEC 9075-9), as that seems to match cases in the standard (book 9), where the error is a result of the user calling API functions out of sequence.  It can be argued it can be `42000` (*syntax error or access rule violation* as defined in ISO/IEC 9075-2) as well.


mrotteveel:
Once we've settled on the SQLstates. I'll also backport this to Firebird 3, 4 and 5.

dyemanov (approved):

asfernandes (approved):

PR #8742, opened 2025-09-17
Possible fix #8730

### LINKED ISSUE #8730: Unnecessary quotation marks in names when using IUTILS_name_to_string
`IUTILS_name_to_string` contains a call to `KEYWORD_stringIsAToken`, which is triggered by any keyword, even if it is not actually needed. For example, when calling `SHOW SYSTEM FUNCTIONs RDB$BLOB_UTIL.SEEK;` we expect the following output:
```
Engine: SYSTEM
Parameters:
OUTPUT (RDB$INTEGER) INTEGER
HANDLE                            INPUT (RDB$BLOB_UTIL_HANDLE) INTEGER
MODE                              INPUT (RDB$INTEGER) INTEGER
OFFSET                            INPUT (RDB$INTEGER) INTEGER
```

However, we end up with the following output:
```
Engine: SYSTEM
Parameters:
OUTPUT (RDB$INTEGER) INTEGER
HANDLE                            INPUT (RDB$BLOB_UTIL_HANDLE) INTEGER
MODE                              INPUT (RDB$INTEGER) INTEGER
"OFFSET"                            INPUT (RDB$INTEGER) INTEGER
```

I have investigated the code, and this is happening because the word `OFFSET` is included in ParserTokens.h, which is causing this issue. Similar triggers occur with other keywords.

For example, if you call `SHOW SYSTEM PROCEDURE RDB$BLOB_UTIL.CANCEL_BLOB;`, you should see:
```
Engine: SYSTEM
Parameters:
BLOB                              INPUT (RDB$BLOB) BLOB CHARACTER SET NONE
=============================================================================
```

However, in the end we get:
```
Engine: SYSTEM
Parameters:
"BLOB"                              INPUT (RDB$BLOB) BLOB CHARACTER SET SYSTEM.NONE
=============================================================================
```

Ev3nt:
Before:
```sql
SQL> show function SYSTEM.RDB$BLOB_UTIL.IS_WRITABLE;
Function: SYSTEM.RDB$BLOB_UTIL.IS_WRITABLE
Engine: SYSTEM
Parameters:
                                  OUTPUT (SYSTEM.RDB$BOOLEAN) BOOLEAN
"BLOB"                            INPUT (SYSTEM.RDB$BLOB) BLOB CHARACTER SET SYSTEM.NONE

```
After:
```sql
SQL> show function SYSTEM.RDB$BLOB_UTIL.IS_WRITABLE;
Function: SYSTEM.RDB$BLOB_UTIL.IS_WRITABLE
Engine: SYSTEM
Parameters:
                                  OUTPUT (SYSTEM.RDB$BOOLEAN) BOOLEAN
BLOB                                                            INPUT (SYSTEM.RDB$BLOB) BLOB CHARACTER SET SYSTEM.NONE
```

The same goes for any other keywords that come up.


sim1984:
What is the benefit of this change? On the contrary, using quotes allows you to copy the parameter for its actual use.

```sql
select RDB$BLOB_UTIL.IS_WRITABLE(BLOB => '') as b
from rdb$database;
```

```
Dynamic SQL Error
-SQL error code = -104
-Token unknown - line 1, column 34
-BLOB
```

vs

```sql
select RDB$BLOB_UTIL.IS_WRITABLE("BLOB" => '') as b
from rdb$database;
```

```
      B
=======
<false>
```

aafemt:
> using quotes allows you to copy the parameter for its actual use.

If you copy an identifier from ISQL output it is simper to quote it always. Anyway ISQL doesn't provide a ready to copy-paste list of fields, you have to insert commas in between.

sim1984:
> > using quotes allows you to copy the parameter for its actual use.
> 
> If you copy an identifier from ISQL output it is simper to quote it always. Anyway ISQL doesn't provide a ready to copy-paste list of fields, you have to insert commas in between.

I don't want to use quotation marks where they're unnecessary. Putting a comma between parameters is easy. Let's not improve what doesn't need improvement.

asfernandes:
In this example, if quotes were not inserted, you would have two functions with the same name.
```
SQL> create schema S1;
SQL> create schema "S1.F1";
SQL> create function S1."F1.X" returns integer as begin end;
SQL> create function "S1.F1".X returns integer as begin end;
SQL> show function;
S1."F1.X"
"S1.F1".X
```

PR #8746, opened 2025-09-21
Allow isc_tpb_read_consistency to imply read committed

dyemanov:
We allow to specify `isc_tpb_read_consistency` alone and at least one driver currently specifies _ReadCommitted ReadConsistency_ mode this way. However, in fact the transaction isolation becomes _Snapshot_ with `TRA_read_consistency` flag being set too. This causes asserts in the code and I'm not really sure the chase/restart logic works properly in all cases.

Given that such a usage seems quite logical and handy, I suggest to not raise an error but rather imply _ReadCommitted_ instead of _Snapshot_ in this case.

Also, the transaction isolation detection in the trace was changed to be more resistant to weird flag combinations like the one described above and not raise unexpected asserts. It's now also in sync with the value of `MON$ISOLATION_MODE`.

### DIFF
--- a/src/jrd/tra.cpp
+++ b/src/jrd/tra.cpp
@@ -2942,17 +2942,6 @@ static void transaction_options(thread_db* tdbb,
 					Arg::Str("isc_tpb_consistency") << Arg::Str("isc_tpb_read_consistency"));
 			}
 
-			if (rec_version.isAssigned())
-			{
-				const auto tpbStr = rec_version.asBool() ?
-					"isc_tpb_rec_version" : "isc_tpb_no_rec_version";
-
-				ERR_post(Arg::Gds(isc_bad_tpb_content) <<
-					// 'Option @1 is not valid if @2 was used previously in TPB'
-					Arg::Gds(isc_tpb_conflicting_options) <<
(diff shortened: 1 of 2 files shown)


src/jrd/tra.cpp
```diff
@@ -3479,24 +3527,12 @@ static void transaction_options(thread_db* tdbb,
...
-			ERR_post(Arg::Gds(isc_bad_tpb_content) <<
-					 Arg::Gds(isc_tpb_option_without_rc) << Arg::Str("isc_tpb_rec_version"));
-		}
-		else
-		{
-			ERR_post(Arg::Gds(isc_bad_tpb_content) <<
-					 Arg::Gds(isc_tpb_option_without_rc) << Arg::Str("isc_tpb_no_rec_version"));
-		}
-	}
-
```
hvlad, line 3495: Is it allowed now to specify `isc_tpb_rec_version` or `isc_tpb_no_rec_version` without `isc_tpb_read_committed` too ?

dyemanov: Yes, but it wasn't intended. I think it's better to restore the original error for this case.

hvlad (approved):

PR #8747, opened 2025-09-21
Refactor RuntimeStatistics to make it more extendable for future counters

dyemanov:
- Improve bounds checking (make it compile-time using `class enum`)
- Simplify retrieval of the counters (using `operator[]`)
- Make it easier to extend internally, especially adding grouped counters (per-tablespace page I/O counters, temp I/O counters, etc)

### DIFF
--- a/src/jrd/RuntimeStatistics.h
+++ b/src/jrd/RuntimeStatistics.h
@@ -28,6 +28,9 @@
 #include "../common/classes/objects_array.h"
 #include "../common/classes/init.h"
 #include "../common/classes/tree.h"
+#include "../common/classes/File.h"
+
+#include <algorithm>
 
 namespace Firebird
 {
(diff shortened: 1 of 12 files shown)


src/jrd/RuntimeStatistics.h
```diff
@@ -48,82 +51,135 @@ class jrd_rel;
 
 typedef Firebird::HalfStaticArray<Firebird::TraceCounts, 5> TraceCountsArray;
 
-// Runtime statistics class
+// Runtime statistics
+
+enum class PageStatType
```
asfernandes, line 56: Should these two enuns be int-based or could use a underlying smaller type?

dyemanov: Given that we don't store it anywhere and it's always passed through stack, I think the default int is OK.

PR #8431, opened 2025-02-05
Make sure only one error will be sent to not-started Service

Noremos:
This is a fix for https://github.com/FirebirdSQL/firebird/issues/8430

Issue #8430, opened 2025-02-05
Unstable error messages in services with trace enabled


Noremos:
Sometimes it reproduced with the following case:

`fbsvcmgr service_mgr -user SYSDBA -action_backup -dbname employee -bkp_file /a/test_db_backup.fbk`

Usually, only one error message appears (cannot open backup file in this case), but with trace enabled, a second error sometimes appears (-Exiting before completion due to errors in this case).

PR #8752, opened 2025-09-22
Postfix for #8430

Ev3nt:
As I see, there is a problem with fbsvcmgr again (https://github.com/FirebirdSQL/firebird/pull/8431#issuecomment-3289842284), so I offer my solution. It is that `ALICE_print` calls `tdgbl->uSvc->started()`, which immediately after the output of information releases the stream, which does not allow further output of information.


src/alice/alice.cpp
```diff
@@ -557,7 +557,7 @@ int alice(Firebird::UtilSvc* uSvc)
 
 			if (any_error)
 			{
-				ALICE_print(24);	// msg 24: Summary of validation errors\n
+				ALICE_print(24, SafeArg(), false);	// msg 24: Summary of validation errors\n
```
dyemanov: Given that we know the service has been already started, I'd consider to do:

```cpp
if (tdgbl->uSvc->isService())
			tdgbl->uSvc->started();
```
explicitly before this `ALICE_print()` call instead of hacking `ALICE_print()` itself for this particular case.

dyemanov (approved):

PR #8758, opened 2025-09-29
refactor(pag): Use `fallocate` as the primary method for extending a file

TreeHunter9:
- This should eliminate double writes of pages when file is extended.
- If fallocate is not supported by filesystem fallback to the old method of writing zeroes to the file.

Restore gets the biggest performance boost from this patch. I tested restore on a database with one table `CREATE TABLE TAB1(V1 VARCHAR(25), V2 VARCHAR(25), V3 VARCHAR(25), V4 VARCHAR(25))`, with a file size of about 16 GB. The testing was done on two different PCs with the same OS (Ubuntu 24.04):
`Intel Core i5-10400 + WD Red SN700 1000 GB (111150WD)`: +10% speedup with `-par 8`;
`Ryzen 7 7700 + KINGSTON SKC3000S1024G (EIFK31.6)`: +18% speedup with `-par 8`;
On Windows, the performance improvement is small (about 3-5%), so the real target of this patch is Linux.

### DIFF
--- a/src/jrd/nbak.cpp
+++ b/src/jrd/nbak.cpp
@@ -389,10 +389,12 @@ bool BackupManager::extendDatabase(thread_db* tdbb)
 	if (maxAllocPage >= maxPage)
 		return true;
 
-	if (!pgSpace->extend(tdbb, maxPage, true))
-		return false;
-
+	const auto extension_result = pgSpace->extend(tdbb, maxPage, true);
 	maxAllocPage = pgSpace->maxAlloc();
+	if (extension_result.success && maxAllocPage > maxPage)
+		return true;
+
+	// Fast file extension not succeeded for some reason, try file extension via direct file writes.
 	while (maxAllocPage < maxPage)
 	{
 		const USHORT ret = PIO_init_data(tdbb, pgSpace->file, tdbb->tdbb_status_vector,
--- a/src/jrd/nbak.cpp
+++ b/src/jrd/nbak.cpp
@@ -389,10 +389,12 @@ bool BackupManager::extendDatabase(thread_db* tdbb)
 	if (maxAllocPage >= maxPage)
 		return true;
 
-	const auto allocatedPages = pgSpace->extend(tdbb, maxPage, true);
-	maxAllocPage = pgSpace->maxAlloc();
-	if (allocatedPages.has_value() && maxAllocPage > maxPage)
-		return true;
+	if (pgSpace->extend(tdbb, maxPage, true))
+	{
+		maxAllocPage = pgSpace->maxAlloc();
+		if (maxAllocPage >= maxPage)
+			return true;
+	}
 
 	// Fast file extension not succeeded for some reason, try file extension via direct file writes.
 	while (maxAllocPage < maxPage)
--- a/src/jrd/nbak.cpp
+++ b/src/jrd/nbak.cpp
@@ -389,9 +389,9 @@ bool BackupManager::extendDatabase(thread_db* tdbb)
 	if (maxAllocPage >= maxPage)
 		return true;
 
-	const auto extension_result = pgSpace->extend(tdbb, maxPage, true);
+	const auto extensionResult = pgSpace->extend(tdbb, maxPage, true);
 	maxAllocPage = pgSpace->maxAlloc();
-	if (extension_result.success && maxAllocPage > maxPage)
+	if (extensionResult.success && maxAllocPage > maxPage)
 		return true;
 
 	// Fast file extension not succeeded for some reason, try file extension via direct file writes.
--- a/src/jrd/nbak.cpp
+++ b/src/jrd/nbak.cpp
@@ -389,9 +389,9 @@ bool BackupManager::extendDatabase(thread_db* tdbb)
 	if (maxAllocPage >= maxPage)
 		return true;
 
-	const auto extensionResult = pgSpace->extend(tdbb, maxPage, true);
+	const auto allocatedPages = pgSpace->extend(tdbb, maxPage, true);
 	maxAllocPage = pgSpace->maxAlloc();
-	if (extensionResult.success && maxAllocPage > maxPage)
+	if (allocatedPages.has_value() && maxAllocPage > maxPage)
 		return true;
 
 	// Fast file extension not succeeded for some reason, try file extension via direct file writes.
--- a/src/jrd/os/posix/unix.cpp
+++ b/src/jrd/os/posix/unix.cpp
@@ -302,9 +302,7 @@ bool PIO_expand(const TEXT* file_name, USHORT file_length, TEXT* expanded_name, 
 bool PIO_fast_extension_is_supported(const Jrd::jrd_file& file) noexcept
 {
 #if defined(HAVE_LINUX_FALLOC_H) && defined(HAVE_FALLOCATE)
-	return file.fil_flags & FIL_no_fast_extend
-		? false
-		: true;
+	return !(file.fil_flags & FIL_no_fast_extend);
 #else
 	return false;
 #endif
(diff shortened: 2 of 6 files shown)


src/jrd/pag.cpp
```diff
@@ -266,99 +266,94 @@ namespace
 		CCH_RELEASE(tdbb, &window);
 	}
 
-	ULONG ensureDiskSpace(thread_db* tdbb, WIN* pip_window, const PageNumber pageNum, ULONG pipUsed)
+	ULONG ensureDiskSpace(thread_db* tdbb, WIN* pip_window, const PageNumber page_num, const ULONG pip_used)
```
AlexPeshkoff: Why change variables' names to old_style from newStyle?

AlexPeshkoff (changes requested):

src/jrd/pag.cpp
TreeHunter9: Sorry, I just prefer snake_case and forgot that we use camelCase.
Changed the variable names to camelCase.

src/jrd/pag.cpp
```diff
@@ -266,99 +266,94 @@ namespace
...
 		const PageManager& pageMgr = dbb->dbb_page_manager;
 		PageSpace* const pageSpace = pageMgr.findPageSpace(pageNum.getPageSpaceID());
 
-		ULONG newUsed = pipUsed;
 		const ULONG sequence = pageNum.getPageNum() / pageMgr.pagesPerPIP;
-		const ULONG relative_bit = pageNum.getPageNum() - sequence * pageMgr.pagesPerPIP;
+		const ULONG relativeBit = pageNum.getPageNum() - sequence * pageMgr.pagesPerPIP;
+
+		if (relativeBit + 1 <= pipUsed)
+			return pipUsed;
```
AlexPeshkoff, line 279: Condition 
`if (relativeBit + 1 <= pipUsed)`
is equivalent to
`if (relativeBit < pipUsed)`
and after it following assertion seems to make no sense

src/jrd/os/posix/unix.cpp
```diff
@@ -299,7 +299,19 @@ bool PIO_expand(const TEXT* file_name, USHORT file_length, TEXT* expanded_name,
 }
 
 
-void PIO_extend(thread_db* tdbb, jrd_file* file, const ULONG extPages, const USHORT pageSize)
+bool PIO_fast_extension_is_supported(const Jrd::jrd_file& file) noexcept
+{
+#if defined(HAVE_LINUX_FALLOC_H) && defined(HAVE_FALLOCATE)
+	return file.fil_flags & FIL_no_fast_extend
+		? false
+		: true;
```
asfernandes:
```suggestion
	return !(file.fil_flags & FIL_no_fast_extend);
```

AlexPeshkoff (approved):

src/jrd/pag.h
```diff
@@ -129,8 +129,13 @@ class PageSpace : public pool_alloc<type_PageSpace>
...
 	static ULONG usedPages(const Database* dbb);
 
+	struct ExtendResult
+	{
+		bool success = false;
+		ULONG pagesAllocated = 0;
+	};
 	// extend page space
-	bool extend(thread_db*, const ULONG, const bool);
+	ExtendResult extend(thread_db* tdbb, ULONG pageNum, bool forceSize);
```
dyemanov: If it's possible to return `{success == true && pagesAllocated == 0}`, then I'd suggest to return `std::optional<ULONG>` instead of the new struct.

src/jrd/nbak.cpp
```diff
@@ -389,10 +389,12 @@ bool BackupManager::extendDatabase(thread_db* tdbb)
...
 
-	if (!pgSpace->extend(tdbb, maxPage, true))
-		return false;
-
+	const auto allocatedPages = pgSpace->extend(tdbb, maxPage, true);
 	maxAllocPage = pgSpace->maxAlloc();
+	if (allocatedPages.has_value() && maxAllocPage > maxPage)
+		return true;
+
+	// Fast file extension not succeeded for some reason, try file extension via direct file writes.
```
dyemanov, line 399: I think something like this would be easier to understand:

```cpp
if (pgSpace->extend(tdbb, maxPage, true))
{
	maxAllocPage = pgSpace->maxAlloc();
	if (maxAllocPage > maxPage)
		return true;
}
```

BTW, why `> maxPage` here but `>= maxPage` in the same check above?

src/jrd/pag.cpp
```diff
@@ -1952,64 +1947,76 @@ bool PageSpace::extend(thread_db* tdbb, const ULONG pageNum, const bool forceSiz
...
-			{
-				PIO_extend(tdbb, file, extPages, dbb->dbb_page_size);
-				break;
-			}
-			catch (const status_exception&)
+			if (!PIO_extend(tdbb, file, extPages, dbb->dbb_page_size))
+				return std::nullopt;
+
+			// File was extended, reset cached value
+			maxPageNumber = 0;
```
dyemanov: I know this was the original code, but still wondering -- why we cannot do `maxPageNumber += reqPages`, given we know the file was extended successfully?

src/jrd/pag.cpp
```diff
@@ -266,99 +266,93 @@ namespace
...
-				const ULONG start = sequence * pageMgr.pagesPerPIP + pipUsed;
+		FbLocalStatus status;
+		const ULONG usedPages = sequence * pageMgr.pagesPerPIP + pipUsed;
 
-				init_pages = PIO_init_data(tdbb, pageSpace->file, &status, start, init_pages);
-			}
+		const bool allocateExactNumberOfPages = dbb->dbb_flags & DBB_no_reserve;
+		const auto allocatedPages = pageSpace->extend(tdbb, usedPages + initPages, allocateExactNumberOfPages);
+		if (allocatedPages.has_value())
+			initPages = allocatedPages.value();
```
dyemanov: Just a suggestion, given that `allocatedPages` is not used elsewhere:

```cpp
if (const auto allocatedPages = pageSpace->extend(tdbb, usedPages + initPages, allocateExactNumberOfPages))
	initPages = allocatedPages.value();
else
```

src/jrd/nbak.cpp
TreeHunter9: > BTW, why > maxPage here but >= maxPage in the same check above?

Fixed, my bad

dyemanov (approved):

src/jrd/pag.cpp
TreeHunter9: The only reason I found is that if our page size is less than OS page size, we may get different value from `PIO_get_number_of_pages()` because it uses `fstat()`, but in all other cases we should get the same value. And as I can see in current code we cannot call `PageSpace::extend()` from multiple threads due to fetching PIP with write lock before calling `PageSpace::extend()`.

dyemanov: Couldn't it be called from `BackupManager::endBackup()` and `ensureDiskSpace()` simultaneously? If not, please add a comment near the increment explaining why it's safe. If there are any doubts, better leave it being reset to zero.

TreeHunter9: If I understand correctly, calls to `extendDatabase()` inside `endBackup()` happens when `backup_state` is `hdr_nbak_stalled`, so calls to `ensureDiskSpace()` will not result in the file being extended. Also there is `BackupManager::StateReadGuard stateGuard(tdbb);` in `ensureDiskSpace()` which provides even more protection against races from nbakup.

TreeHunter9: But it is very bug-prone in future, if this logic will be changed, so I believe it is better to reset the cached value to zero (and we shouldn't see a significant performance improvement, the additional call to `fstat()` to obtain the file size should be fast anyway, as it only uses information from inode (I hope it's fast)).

dyemanov (approved):

Issue #7954, opened 2024-01-12
Shared metadata cache


AlexPeshkoff:
Since FB3 we use per-attachment metadata cache in SS. This slows attach to already opened DB, makes a lot of other caches not enough efficient and causes serious loss of RAM for DBs with big metadata.

vpavlov33:
When to expect this to arrive in main tree? I would like to test it when merged.

AlexPeshkoff:
Beginning of next year

vpavlov33:
Hi @AlexPeshkoff ,
Is there a change of plan about metacache in FB6? Haw can I help about it?

AlexPeshkoff:
No changes.
No idea how to help.

PR #8759, opened 2025-09-30
Implementation of external data management #1158

paradox1307:
Link to RFC - [Distributed queries (Management of External Data) proposal](https://groups.google.com/g/firebird-devel/c/XWoknE2mZvg/m/x75vY8K7BQAJ)


paradox1307:
This is a *draft* implementation of external data management (SQL:MED). It will be improved after the metacache #7954 is merged.

sim1984:
Are there any restrictions for users/roles on using foreign servers? I think there should be. That is, we should have something like:

```
GRANT USAGE ON SERVER <server_name>
TO <grantee_list> [WITH GRANT OPTION]
[{GRANTED BY | AS} [USER] <grantor>];

REVOKE [GRANT OPTION FOR]
USAGE ON SERVER <server_name>
FROM <grantee_list>
[{GRANTED BY | AS} [USER] <grantor>];
```

This is important, especially if the foreign server stores authentication data. Not every authenticated user should be able to use it in an `EXECUTE STATEMENT`.

doc/sql.extensions/README.MED.md
```diff
@@ -0,0 +1,334 @@
...
+
+`RDB$FOREIGN_MAPPING_OPTIONS` - contains a description of all options defined for user mappings to foreign servers in the database:
+
+| Field                      | Type                 | Description                                                |
+|----------------------------|----------------------|------------------------------------------------------------|
+| RDB$USER                   | CHAR \(63\)          | The user name to map                                       |
+| RDB$FOREIGN_SERVER_NAME    | CHAR \(63\)          | The foreign server name                                    |
+| RDB$FOREIGN_OPTION_NAME    | CHAR \(63\)          | Option name                                                |
+| RDB$FOREIGN_OPTION_VALUE   | VARCHAR \(32765\)    | Option value                                               |
+| RDB$OWNER_NAME             | SHORT                | Option type                                                |
```
sim1984: May be `RDB$FOREIGN_OPTION_TYPE `?

src/jrd/ForeignServer.cpp
```diff
@@ -0,0 +1,864 @@
...
+{
+	string buffer;
+	for (const auto& option: options)
+	{
+		// Skip general options
+		if (isGenericOption(option.first))
+			continue;
+
+		buffer.append(option.first.c_str());
+		buffer.append(";");
```
sim1984: Why is there a semicolon `;` here? I would have expected an equals symbol `=`.

paradox1307:
> Are there any restrictions for users/roles on using foreign servers? I think there should be. 

Yes, there should be restrictions on using foreign servers. The MED standard requires `GRANT/REVOKE USAGE ON FOREIGN SERVER`. Implemented in [df133b1](https://github.com/FirebirdSQL/firebird/pull/8759/commits/df133b10483fd4c294b47df987161c3bd762b179) . Please note that when creating foreign tables and user mappings, `USAGE` checks have also been added, as required by the standard.

doc/sql.extensions/README.MED.md
paradox1307: Yes, corrected.

src/jrd/ForeignServer.cpp
paradox1307: The idea was that option names and option values follow each other and there is no sense in adding a special character to separate them, i.e. `option_1_name;option_1_value;option_2_name;option_2_value;...`. If this is confusing, it can be changed to something else.

sim1984: In principle, this approach will work, provided that each option requires a value. In this case, we eliminate the possibility of passing only the option name as a flag.

doc/sql.extensions/README.MED.md
```diff
@@ -0,0 +1,334 @@
...
+The `CREATE SERVER` statement can be executed by administrators or users with the `CREATE SERVER` privilege.
+
+#### ALTER SERVER definition
+
+```sql
+ALTER SERVER <foreign server name>
+        [[DROP] FOREIGN DATA WRAPPER | [DROP] USING PLUGIN <provider(plugin)>] [OPTIONS(<option> [, <option> ...] )]
+    <option> ::= {
+        [DROP] <option name> [= 'value']
+        | [DROP] <option name> [{FILE | ENV}] ['value'] }
```
sim1984: This description confuses people because it's unclear what's required and what's optional. A better explanation would be:

```
ALTER SERVER <foreign server name>
        [[DROP] {FOREIGN DATA WRAPPER | USING PLUGIN} <provider(plugin)>] 
        [OPTIONS (<option> [, <option> ...] )]

<option> ::= {
          <option name> = 'value'
        | <option name> [{FILE | ENV}] 'value' 
        | DROP <option name>
}
```

sim1984:
> > Are there any restrictions for users/roles on using foreign servers? I think there should be.
> 
> Yes, there should be restrictions on using foreign servers. The MED standard requires `GRANT/REVOKE USAGE ON FOREIGN SERVER`. Implemented in [df133b1](https://github.com/FirebirdSQL/firebird/pull/8759/commits/df133b10483fd4c294b47df987161c3bd762b179) . Please note that when creating foreign tables and user mappings, `USAGE` checks have also been added, as required by the standard.

This should be described in the `README.MED.md` documentation.

doc/sql.extensions/README.MED.md
paradox1307: Corrected, thanks.

paradox1307:
> This should be described in the `README.MED.md` documentation.

Added description `GRANT/REVOKE USAGE ON FOREIGN SERVER` to the documentation.

sim1984:
I will join the testing with my ODBCEngine/MySQLEngine providers when the Windows builds are ready.

doc/sql.extensions/README.MED.md
```diff
@@ -0,0 +1,334 @@
...
+* A foreign column cannot be computed \(`COMPUTED [BY] | GENERATED ALWAYS AS`\);
+* Foreign columns can have `NOT NULL` and `CHECK` integrity constraints, but this does not guarantee that the foreign server will check these constraints;
+* Otherwise, the rules for creating a foreign table are the same as for a regular table.
+
+The optional `OPTIONS` clause specifies additional foreign column options and table options. The required foreign column and table options depend on the provider \(plugin\) used to access table data. General table options are:
+* SCHEMA_NAME - the schema name where the table is located on the foreign server;
+* TABLE_NAME - name of the table on the foreign server.
+General column options are:
+* COLUMN_NAME - the name of the column in the foreign table;
+* PRIMARY_KEY - an identifier explicitly specifying that the column is used in the primary key.
```
sim1984: Where READONLY option?

sim1984:
For consistency, it would be nice to add the ability to save user comments like for other database objects.

This is how it should look for foreign servers.

```
COMMENT ON SERVER <foreign server name> IS 'text';
```

I can't imagine how this might look for user mappings, since they don't have simple names.

For foreign tables, it might look like this.

```
COMMENT ON FOREIGN TABLE <table name> IS 'text';
COMMENT ON COLUMN <table name>.<column_name> IS 'text';
```

Perhaps the additional `COMMENT ON FOREIGN TABLE` is unnecessary, and you could simply use

```
COMMENT ON TABLE <table name> IS 'text';
```

paradox1307:
> For consistency, it would be nice to add the ability to save user comments like for other database objects.

Agree. I'll add the ability to save comments for foreign servers and tables.

> I can't imagine how this might look for user mappings, since they don't have simple names.

Maybe:
```
COMMENT ON USER MAPPING FOR <user> SERVER <server_name> IS 'text';
```
?

sim1984:
> 
> Maybe:
> 
> ```
> COMMENT ON USER MAPPING FOR <user> SERVER <server_name> IS 'text';
> ```
> 
> ?

OK

paradox1307:
> I will join the testing with my ODBCEngine/MySQLEngine providers when the Windows builds are ready.

Windows builds have been fixed. You can try.

doc/sql.extensions/README.MED.md
paradox1307: Description added.

PR #8762, opened 2025-10-07
Feature #8761 - Add API method Util::convert

### LINKED ISSUE #8761: Add API method Util::convert
It's very difficult for client libraries to implement conversion of types in the same manner as implemented internally by Firebird (client
or engine). There are many complexities like conversion of floating point types and integral scaled ones, for example.

It's not sufficient to let client change message field types and let Firebird do it.

Usually, libraries have methods like `getInt`, `getString`, etc for individual columns and nobody can anticipate what the developer is going
to call. The same for setters.

Therefore I propose new utility function:

```
interface Util : Versioned
{
	...

	void convert(
		Status status,
		uint sourceType,
		uint sourceScale,
		uint sourceLength,
		const void* source,
		uint targetType,
		uint targetScale,
		uint targetLength,
		void* target);
}
```

It will be very raw, as anyone that can already read/write data to messages can use this function with buffers and read/write data to it.

It will be thin wrapper for a CVT_move call. It should not work for blobs.

asfernandes:


sim1984:
Please add a description of this function to the documentation doc/Using_OO_API.html

PR #8772, opened 2025-10-13
fix(pag): Incorrect calculation of last used page

TreeHunter9:
If `pip_used` has its maximum possible value, we can read over the page buffer.
This can lead to incorrect validation of SCN pages when requesting the last used page to calculate the number of SCNs.

v5 and v4 are also affected.

### DIFF
--- a/src/jrd/pag.cpp
+++ b/src/jrd/pag.cpp
@@ -1824,7 +1824,7 @@ ULONG PageSpace::lastUsedPage()
 
 	const page_inv_page* pip = (page_inv_page*) window.win_buffer;
 
-	int last_bit = pip->pip_used;
+	int last_bit = pip->pip_used > 0 ? pip->pip_used - 1 : 0;
 	int byte_num = last_bit / 8;
 	UCHAR mask = 1 << (last_bit % 8);
 	while (last_bit >= 0 && (pip->pip_bits[byte_num] & mask))
--- a/src/jrd/pag.cpp
+++ b/src/jrd/pag.cpp
@@ -1824,27 +1824,30 @@ ULONG PageSpace::lastUsedPage()
 
 	const page_inv_page* pip = (page_inv_page*) window.win_buffer;
 
-	int last_bit = pip->pip_used;
-	int byte_num = last_bit / 8;
-	UCHAR mask = 1 << (last_bit % 8);
-	while (last_bit >= 0 && (pip->pip_bits[byte_num] & mask))
+	int last_bit = pip->pip_used - 1;
+	if (pip->pip_used > 0)
 	{
-		if (mask == 1)
+		int byte_num = last_bit / 8;
+		UCHAR mask = 1 << (last_bit % 8);
+		while (last_bit >= 0 && (pip->pip_bits[byte_num] & mask))
 		{
-			mask = 0x80;
-			byte_num--;
-			//fb_assert(byte_num > -1); ???
-		}
-		else
-			mask >>= 1;
+			if (mask == 1)
+			{
+				mask = 0x80;
+				byte_num--;
+				//fb_assert(byte_num > -1); ???
+			}
+			else
+				mask >>= 1;
 
-		last_bit--;
+			last_bit--;
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/jrd/pag.cpp
```diff
@@ -1824,7 +1824,7 @@ ULONG PageSpace::lastUsedPage()
 
 	const page_inv_page* pip = (page_inv_page*) window.win_buffer;
 
-	int last_bit = pip->pip_used;
+	int last_bit = pip->pip_used > 0 ? pip->pip_used - 1 : 0;
```
hvlad: This change is correct but not enough: when `pip->pip_used == 0` then `last_bit == -1` after the while loop and returned value will be wrong.

hvlad (changes requested):

src/jrd/pag.cpp
hvlad: Consider this patch

```

 src/jrd/pag.cpp | 29 ++++++++++++++++-------------
 1 file changed, 16 insertions(+), 13 deletions(-)

diff --git a/src/jrd/pag.cpp b/src/jrd/pag.cpp
index 8b4e1d3369..2745fe5fbe 100644
--- a/src/jrd/pag.cpp
+++ b/src/jrd/pag.cpp
@@ -1835,27 +1835,30 @@ ULONG PageSpace::lastUsedPage()
 
 	const page_inv_page* pip = (page_inv_page*) window.win_buffer;
 
-	int last_bit = pip->pip_used;
-	int byte_num = last_bit / 8;
-	UCHAR mask = 1 << (last_bit % 8);
-	while (last_bit >= 0 && (pip->pip_bits[byte_num] & mask))
+	int last_bit = pip->pip_used - 1;
+	if (pip->pip_used > 0)
 	{
-		if (mask == 1)
+		int byte_num = last_bit / 8;
+		UCHAR mask = 1 << (last_bit % 8);
+		while (last_bit >= 0 && (pip->pip_bits[byte_num] & mask))
 		{
-			mask = 0x80;
-			byte_num--;
-			//fb_assert(byte_num > -1); ???
-		}
-		else
-			mask >>= 1;
+			if (mask == 1)
+			{
+				mask = 0x80;
+				byte_num--;
+				//fb_assert(byte_num > -1); ???
+			}
+			else
+				mask >>= 1;
 
-		last_bit--;
+			last_bit--;
+		}
 	}
 
 	CCH_RELEASE(tdbb, &window);
 	pipMaxKnown = pipLast;
 
-	return last_bit + (pipLast == pipFirst ? 0 : pipLast);
+	return last_bit + 1 + (pipLast == pipFirst ? 0 : pipLast);
 }
 
 ULONG PageSpace::lastUsedPage(const Database* dbb)

```

TreeHunter9: Oh, I didn't notice that there is another bug, the returned page number is 1 value off. Then yeah, this patch looks good to me.

hvlad (approved):

TreeHunter9:
It looks like the patch is incorrect when we are working with the first PIP. Because the first PIP is interpreted as 0 page, the `last_bit + 1` logic results in an unneeded offset being added.

I will prepare a patch with the fix.

hvlad:
`last_bit` is the relative number of the last used page. When it reside on first PIP, it become absolute number.
What is wrong ?

TreeHunter9:
> When it reside on first PIP, it become absolute number.

Yes, but we add `1` to this absolute number, so the result page number is wrong.

hvlad:
You right, I (wrongly) considered that it should return **count** of used pages.

aafemt:
Still it is not clear why page zero is different from others and what happen with the gap between the last bit on this page and first bit on the next.

TreeHunter9:
> Still it is not clear why page zero is different from others and what happen with the gap between the last bit on this page and first bit on the next.

I'm not quite getting your question, but Vlad explained how `last_bit` works above.

aafemt:
Ugh, yes. What was missed in this explanation was that `pipLast` is not just `pagesPerPip*pageNumber` which makes whole calculation unnecessarily complicated.

PR #8776, opened 2025-10-14
Prevent past-EOF access to uni-directional cursors

Ev3nt:
This code is more secure than the stub in https://github.com/FirebirdSQL/firebird/blob/master/src/dsql/DsqlRequests.cpp#L383

> ```c++
> // Fetch from already finished request should not produce error
> if (!(request->req_flags & req_active))
> {
> 	if (req_timer)
> 		req_timer->stop();
> 
> 	trace.fetch(true, ITracePlugin::RESULT_SUCCESS);
> 	return false;
> }
> ```

It can be used as a second level of cursor status checking.

### DIFF
--- a/src/dsql/DsqlCursor.cpp
+++ b/src/dsql/DsqlCursor.cpp
@@ -135,6 +135,14 @@ int DsqlCursor::fetchNext(thread_db* tdbb, UCHAR* buffer)
 {
 	if (!(m_flags & IStatement::CURSOR_TYPE_SCROLLABLE))
 	{
+		if (m_state == EOS)
+		{
+			fb_assert(m_eof);
+			return 1;
+		}
+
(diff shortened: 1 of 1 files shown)


dyemanov:
A few notes:

- The original goal is to avoid fetching beyond EOF in cursors, which is currently still possible in the remote protocol prefetch logic. This causes `isc_req_sync` errors being reported in the trace. It can be additionally fixed inside the protocol code, but this patch is more robust from the possible regressions POV.
- While this protection it's not strictly required in master (due to the aforementioned check added by Dimitry Sibiryakov recently), it's IMHO better understandable.
- It unifies EOF handling between uni-directional and scrollable cursors.
- It can be easily cherry-picked into the prior versions, if desired.

PR #8792, opened 2025-10-28
`getPerfCounters()` now uses scope tags

Ev3nt:
The goal of this PR is to utilize new scope tags in isql.  Now when using `SET STATS` we got this table of stats scopes:
| Tag | Scope |
| ---- | ------- |
| isc_info_fetches                | fb_info_counts_scope_att |
| isc_info_marks                  | fb_info_counts_scope_att |
| isc_info_reads                   | fb_info_counts_scope_att |
| isc_info_writes                  | fb_info_counts_scope_att |
| isc_info_current_memory | fb_info_counts_scope_db |
| isc_info_max_memory      | fb_info_counts_scope_db |

This change fixes the old issue with ISQL reporting unstable (depending on the concurrent load) page-level counters in the execution statistics.

### DIFF
--- a/src/yvalve/perf.cpp
+++ b/src/yvalve/perf.cpp
@@ -337,9 +337,6 @@ constexpr KnownCounters knownCounters[TOTAL_COUNTERS] = {
 
 } // anonymous namespace
 
-static Firebird::IAttachment* lastAttachment = nullptr;
-static bool scopeTagsSupported = false;
-
 void Why::UtilInterface::getPerfCounters(Firebird::CheckStatusWrapper* status,
 	Firebird::IAttachment* att, const char* countersSet, ISC_INT64* counters)
 {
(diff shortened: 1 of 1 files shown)


aafemt:
Not synchronized access to a global variable. Completely unnecessary additional round-trip. Not good, IMHO.

dyemanov:
> Completely unnecessary additional round-trip.

Do you suggest a try-except loop (first iteration with scopes, second without)?

aafemt:
> Do you suggest a try-except loop (first iteration with scopes, second without)?

Unknown info items don't throw, there is no need in try-catch. isc_info_error is already handled, so if server doesn't support scopes, they will be silently skipped by existing code and result will be exactly the same as before.

All that is needed is to add scope tags to the request buffer. They are cheap, so even optimization by grouping counters might be unnecessary, just fill the buffer with `scope,tag` pairs from the table.

dyemanov (approved):

mrotteveel:
Given ISQL doesn't display marks (as far as I can tell), is there anything else affected by this change? And if not, why is it asking for marks if it doesn't show it?

dyemanov:
ISQL doesn't ask for marks. But it calls `UtilInterface::getPerfCounters()` which itself requests a predefined set of counters (including marks) and returns only those requested by ISQL.

mrotteveel:
OK, so in the release notes, does this need to be documented in API and ODS as well?

dyemanov:
Yes, this change is related to API rather than to ISQL (although this is where it becomes user-visible).

PR #8795, opened 2025-10-30
GENERATE_SERIES function

sim1984:
# GENERATE_SERIES function

The `GENERATE_SERIES` function creates a series of numbers within a specified interval. 
The interval and the step between series values ​​are defined by the user.

## Syntax

```
<generate_series_function> ::=
    GENERATE_SERIES(<start>, <finish> [, <step>]) [AS] <correlation name> [ ( <derived column name> ) ]
```

## Arguments

* `start` - The first value in the interval. `start` is specified as a variable, a literal, or a scalar expression of type 
`smallint`, `integer`, `bigint`, `int128` or `numeric/decimal`.

* `finish` - The last value in the interval. `finish` is specified as a variable, a literal, or a scalar expression of 
type `smallint`, `integer`, `bigint`, `int128` or `numeric/decimal`. The series stops once the last generated step value exceeds 
the `finish` value.

* `step` - Indicates the number of values to increment or decrement between steps in the series. `step` is an expression 
of type `smallint`, `integer`, `bigint`, `int128` or `numeric/decimal`. `step` can be either negative or positive, but can't be zero (0).  This argument is optional. The default value for `step` is 1.

## Returning type

The function `GENERATE_SERIES` returns a set with `BIGINT`, `INT128` or `NUMERIC/DECIMAL` column, where the scale is 
determined by the maximum of the scales of the function arguments.

## Rules

* If `start > finish` and a negative `step` value is specified, an empty set is returned.

* If `start < finish` and a positive `step` value is specified, an empty set is returned.

* If the `step` argument is zero, an error is thrown.

## Examples

```
SELECT n
FROM GENERATE_SERIES(1, 3) AS S(n);

SELECT n
FROM GENERATE_SERIES(3, 1, -1) AS S(n);

SELECT n
FROM GENERATE_SERIES(0, 9.9, 0.1) AS S(n);

SELECT 
  DATEADD(n MINUTE TO timestamp '2025-01-01 12:00') AS START_TIME,
  DATEADD(n MINUTE TO timestamp '2025-01-01 12:00:59.9999') AS FINISH_TIME
FROM GENERATE_SERIES(0, 59) AS S(n);
```

A similar function exists in PostgreSQL: https://www.postgresql.org/docs/current/functions-srf.html and in MS SQL 2022: https://learn.microsoft.com/en-us/sql/t-sql/functions/generate-series-transact-sql?view=sql-server-ver17

PostgreSQL has the ability to generate date and time sequences, but we don't have interval types, so this isn't implemented. However, by generating a sequence of numbers, we can generate any sequence.


sim1984:
# PSQL procedure vs built-in function performance test

```sql
CREATE EXCEPTION E_INVALID_STEP 'Procedure SP_GENERATE_SERIES required values of parameter STEP_VALUE non equal zero';

SET TERM ^;

CREATE OR ALTER PROCEDURE SP_GENERATE_SERIES (
  START_VALUE  BIGINT,
  FINISH_VALUE BIGINT,
  STEP_VALUE   BIGINT DEFAULT 1)
RETURNS (
  CURRENT_VALUE BIGINT)
AS
BEGIN
  IF (STEP_VALUE = 0) THEN EXCEPTION E_INVALID_STEP;
  IF (STEP_VALUE > 0) THEN
    FOR CURRENT_VALUE = START_VALUE TO FINISH_VALUE BY STEP_VALUE DO
      SUSPEND;
  IF (STEP_VALUE < 0) THEN
    FOR CURRENT_VALUE = START_VALUE DOWNTO FINISH_VALUE BY ABS(STEP_VALUE) DO
      SUSPEND;
END^

SET TERM ;^
```

1000000 records

```sql
SELECT 
  COUNT(*), MIN(CURRENT_VALUE), MAX(CURRENT_VALUE)
FROM SP_GENERATE_SERIES(1, 1000000);
```

```

                COUNT                   MIN                   MAX
===================== ===================== =====================
              1000000                     1               1000000

Current memory = 19793200
Delta memory = 240
Max memory = 19869584
Elapsed time = 0.294 sec
Buffers = 2048
Reads = 0
Writes = 0
Fetches = 0
```

```sql
SELECT 
  COUNT(*), MIN(N), MAX(N)
FROM GENERATE_SERIES(1, 1000000) S(N);
```

```

                COUNT                   MIN                   MAX
===================== ===================== =====================
              1000000                     1               1000000

Current memory = 19809936
Delta memory = 208
Max memory = 19869616
Elapsed time = 0.047 sec
Buffers = 2048
Reads = 0
Writes = 0
Fetches = 0
```

10000000 records

```sql
SELECT 
  COUNT(*), MIN(CURRENT_VALUE), MAX(CURRENT_VALUE)
FROM SP_GENERATE_SERIES(1, 10000000);
```

```

                COUNT                   MIN                   MAX
===================== ===================== =====================
             10000000                     1              10000000

Current memory = 19828352
Delta memory = 240
Max memory = 36589008
Elapsed time = 2.867 sec
Buffers = 2048
Reads = 0
Writes = 0
Fetches = 0
```

```sql
SELECT 
  COUNT(*), MIN(N), MAX(N)
FROM GENERATE_SERIES(1, 10000000) S(N);
```


```
                COUNT                   MIN                   MAX
===================== ===================== =====================
             10000000                     1              10000000

Current memory = 19844848
Delta memory = 208
Max memory = 19869616
Elapsed time = 0.453 sec
Buffers = 2048
Reads = 0
Writes = 0
Fetches = 0
```

aafemt:
It is a system selectable procedure with fixed number of parameters and output. Why did you create a new BLR verb instead of using existing BLR for calling it by name?

sim1984:
> It is a system selectable procedure with fixed number of parameters and output. Why did you create a new BLR verb instead of using existing BLR for calling it by name?

1. We don't have any system stored procedures, except for those in system packages.
2. The output type takes the maximum scale of the arguments.

dyemanov:
> It is a system selectable procedure with fixed number of parameters and output. Why did you create a new BLR verb instead of using existing BLR for calling it by name?

This PR correctly extends `blr_table_value_fun` with a new subcode, it does not introduce a new BLR code. This was added exactly to implement new table-valued functions (and they may have non-procedure-like syntax).

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -346,3 +348,170 @@ bool UnlistFunctionScan::nextBuffer(thread_db* tdbb) const
...
+		assignParameter(tdbb, &fromDesc, toDesc, 0, record);
+		impure->m_recordExists = true;
+	};
+
+	impure->m_result += impure->m_step;
+
+	if (((impure->m_step > 0) && (impure->m_result <= impure->m_finish)) ||
+        ((impure->m_step < 0) && (impure->m_result >= impure->m_finish)))
+	{
+		setIntToRecord(impure->m_result);
```
hvlad: Why to use lambda here ?

aafemt:
> 2\. The output type takes the maximum scale of the arguments.

  It can be handled at DSQL stage.

> This PR correctly extends `blr_table_value_fun` with a new subcode, it does not introduce a new BLR code.

Ugh, I see.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -4378,6 +4386,77 @@ dsql_fld* UnlistFunctionSourceNode::makeField(DsqlCompilerScratch* dsqlScratch)
...
+		newField->fld_id = 0;
+	}
+
+	if (dsqlNameColumns.hasData())
+	{
+		if (dsqlNameColumns.getCount() > 1)
+		{
+			ERRD_post(Arg::Gds(isc_sqlerr) << Arg::Num(-104) << Arg::Gds(isc_dsql_command_err)
+										   << Arg::Gds(isc_dsql_table_value_many_columns)
+										   << Arg::Str(GenSeriesFunctionSourceNode::FUNC_NAME)
```
hvlad: Is it required to specify class name here ?
Or, why not use `getName()` ?

src/jrd/RecordSourceNodes.h
```diff
@@ -1081,6 +1081,24 @@ class UnlistFunctionSourceNode : public TableValueFunctionSourceNode
...
+	explicit GenSeriesFunctionSourceNode(MemoryPool& pool) : TableValueFunctionSourceNode(pool)
+	{
+	}
+
+	RecordSource* compile(thread_db* tdbb, Optimizer* opt, bool innerSubStream) final;
+	dsql_fld* makeField(DsqlCompilerScratch* dsqlScratch) final;
+
+	static constexpr char const* FUNC_NAME = "GENERATE_SERIES";
+
+	const char* getName() const override
```
hvlad, line 1097: Not sure why it is necessary to mark other overridden methods as `final`, but ... why not here ?

hvlad (commented):
Just a few little notes.

src/jrd/recsrc/TableValueFunctionScan.cpp
sim1984: You are absolutely right, lambda is redundant here.

doc/sql.extensions/README.generate_series.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+<generate_series_function> ::=
+    GENERATE_SERIES(<start>, <finish> [, <step>]) [AS] <correlation name> [ ( <derived column name> ) ]
+```
+
+## Arguments
+
+* `start` - The first value in the interval. `start` is specified as a variable, a literal, or a scalar expression of type 
+`smallint`, `integer`, `bigint` or `numeric(18, x)`.
```
asfernandes: What about INT128?

src/dsql/parse.y
```diff
@@ -6833,6 +6836,34 @@ table_value_function_correlation_name
...
+table_value_function_gen_series_arg_list
+	: value ',' value step_opt
+		{
+			$$ = newNode<ValueListNode>($1);
+			$$->add($3);
+			$$->add($4);
+		}
+	;
+
+%type <valueExprNode> step_opt
```
asfernandes: I'd call it `comma_value_opt`.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -4378,6 +4386,77 @@ dsql_fld* UnlistFunctionSourceNode::makeField(DsqlCompilerScratch* dsqlScratch)
...
+	DsqlDescMaker::fromNode(dsqlScratch, &startDesc, startItem, true);
+
+	dsc finishDesc;
+	DsqlDescMaker::fromNode(dsqlScratch, &finishDesc, finishItem, true);
+
+	dsc stepDesc;
+	DsqlDescMaker::fromNode(dsqlScratch, &stepDesc, stepItem, true);
+
+	// common scale
+	const auto scale = MIN(MIN(startDesc.dsc_scale, finishDesc.dsc_scale), stepDesc.dsc_scale);
```
asfernandes, line 4433: It's documented to be the max, but here you use min.

src/jrd/RecordSourceNodes.cpp
```diff
@@ -4378,6 +4386,77 @@ dsql_fld* UnlistFunctionSourceNode::makeField(DsqlCompilerScratch* dsqlScratch)
...
+
+	dsql_fld* field = dsqlField;
+
+	if (!field)
+	{
+		const auto newField = FB_NEW_POOL(dsqlScratch->getPool()) dsql_fld(dsqlScratch->getPool());
```
asfernandes: I'd use field directly without that extra newField.

src/jrd/RecordSourceNodes.h
```diff
@@ -1081,6 +1081,24 @@ class UnlistFunctionSourceNode : public TableValueFunctionSourceNode
 	}
 };
 
+class GenSeriesFunctionSourceNode : public TableValueFunctionSourceNode
```
asfernandes: I'd say it's better to declare the class as `final` instead of each method. Methods should be declared with `override`.

src/jrd/RecordSourceNodes.h
```diff
@@ -1081,6 +1081,24 @@ class UnlistFunctionSourceNode : public TableValueFunctionSourceNode
...
+{
+public:
+	explicit GenSeriesFunctionSourceNode(MemoryPool& pool) : TableValueFunctionSourceNode(pool)
```
asfernandes: It's not the common convention to put `: ...` in the same line as the constructor.

src/jrd/recsrc/RecordSource.h
```diff
@@ -1615,6 +1615,43 @@ namespace Jrd
 		NestConst<ValueListNode> m_inputList;
 	};
 
+	class GenSeriesFunctionScan final : public TableValueFunctionScan
+	{
+		enum GenSeriesTypeItemIndex : unsigned
```
asfernandes: Why unsigned instead of UCHAR?

asfernandes: Isn't `enum class` a much more elegant thing than use common prefix for legacy enums?

src/jrd/recsrc/RecordSource.h
```diff
@@ -1615,6 +1615,43 @@ namespace Jrd
...
+			SINT64 m_result;
+			SCHAR m_scale;
+			bool m_recordExists;
+		};
+
+	public:
+		GenSeriesFunctionScan(CompilerScratch* csb, StreamType stream, const Firebird::string& alias,
+						   ValueListNode* list);
+
+	protected:
```
asfernandes, line 1662: Use override in methods.

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -346,3 +348,166 @@ bool UnlistFunctionScan::nextBuffer(thread_db* tdbb) const
...
+	auto stepItem = m_inputList->items[GEN_SERIES_INDEX_STEP];
+	const auto stepDesc = EVL_expr(tdbb, request, stepItem);
+	if (stepDesc == nullptr)
+	{
+		rpb->rpb_number.setValid(false);
+		return;
+	}
+
+	// common scale
+	const auto scale = MIN(MIN(startDesc->dsc_scale, finishDesc->dsc_scale), stepDesc->dsc_scale);
```
asfernandes: MIN vs MAX

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -346,3 +348,166 @@ bool UnlistFunctionScan::nextBuffer(thread_db* tdbb) const
...
+
+	if (!(impure->irsb_flags & irsb_open))
+	{
+		rpb->rpb_number.setValid(false);
+		return false;
+	}
+
+	rpb->rpb_number.increment();
+
+	do
```
asfernandes: I didn't go to suggest better thing, but for me this code and `m_recordExists` is very confusing.

src/jrd/RecordSourceNodes.cpp
sim1984: In NUMERIC(18, x) types, scale = -x. Therefore, we find the minimum. I could write in the documentation that the result has the minimum scale among the argument scales, but would that be clear to the user?

asfernandes: Sorry, it's ok then.

src/dsql/parse.y
sim1984: Hmm, I would agree if this could be reused in any function, but the default value of 1 prevents this. Any other ideas for a name for this rule?

aafemt: No, default is fine if you pass it as a parameter. I.e. `comma_value_opt(1)` and define it as
```
comma_value_opt($default_value)
	: /* nothing */		{ $$ = MAKE_const_sint64($default_value, 0); }
	| ',' value			{ $$ = $2; }
	;

```

aafemt: Though name "optional_parameter" would be better.

sim1984: Well, you replaced the optional parameter with a predefined default value with an optional parameter with different values, but of a strictly defined type ((

aafemt: I'm not an expert in YACC so perhaps someone else can suggest how to pass node type and value via parameters to clause. May be something like `optional_parameter(LiteralNode, 1)` with `$$ = newNode<$node_type>($value)` inside...

src/jrd/recsrc/TableValueFunctionScan.cpp
sim1984: I think the code is clear now.

doc/sql.extensions/README.generate_series.md
sim1984: Fixed

src/jrd/recsrc/RecordSource.h
sim1984: I don't think this is a good idea. This enumeration is used as an index in a container where the index is an integer. Additionally, there is a comparison of the container's size with the enumeration element. If there is an enum class here, type conversions will be required.

src/jrd/recsrc/RecordSource.h
```diff
@@ -1615,6 +1616,43 @@ namespace Jrd
...
+		{
+			GEN_SERIES_INDEX_START = 0,
+			GEN_SERIES_INDEX_FINISH = 1,
+			GEN_SERIES_INDEX_STEP = 2,
+			GEN_SERIES_INDEX_LAST = 3
+		};
+
+		struct Impure : public TableValueFunctionScan::Impure
+		{
+			std::variant<SINT64, Firebird::Int128> m_start;
```
asfernandes: I think this is going to produce undefined behavior.
Impure are generally initialized with memset and nothing gurantees this is correct for a variant.

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -346,3 +348,211 @@ bool UnlistFunctionScan::nextBuffer(thread_db* tdbb) const
...
+		{
+			rpb->rpb_number.setValid(false);
+			return;
+		}
+
+		impure->m_start = start;
+		impure->m_finish = finish;
+		impure->m_step = step;
+	}
+	else {
```
asfernandes:
```suggestion
	else
	{
```

src/jrd/RecordSourceNodes.cpp
```diff
@@ -4378,6 +4386,92 @@ dsql_fld* UnlistFunctionSourceNode::makeField(DsqlCompilerScratch* dsqlScratch)
...
+	finishItem->setParameterType(
+		dsqlScratch, [](dsc* desc) { desc->makeInt64(0); }, false);
+
+	const auto stepItem = inputList->items[2].getObject();
+	stepItem->setParameterType(
+		dsqlScratch, [](dsc* desc) { desc->makeInt64(0); }, false);
+
+	dsc startDesc;
+	DsqlDescMaker::fromNode(dsqlScratch, &startDesc, startItem, true);
+	if (!startDesc.isExact() && !startDesc.isNull()) {
```
asfernandes:
```suggestion
	if (!startDesc.isExact() && !startDesc.isNull())
	{	
```

src/jrd/RecordSourceNodes.cpp
```diff
@@ -4378,6 +4386,92 @@ dsql_fld* UnlistFunctionSourceNode::makeField(DsqlCompilerScratch* dsqlScratch)
...
+
+	dsc startDesc;
+	DsqlDescMaker::fromNode(dsqlScratch, &startDesc, startItem, true);
+	if (!startDesc.isExact() && !startDesc.isNull()) {	
+		status_exception::raise(Arg::Gds(isc_argmustbe_exact_function) << Arg::Str(getName()));
+	}
+
+	dsc finishDesc;
+	DsqlDescMaker::fromNode(dsqlScratch, &finishDesc, finishItem, true);
+	if (!finishDesc.isExact() && !finishDesc.isNull()) {
```
asfernandes:
```suggestion
	if (!finishDesc.isExact() && !finishDesc.isNull())
	{
```

src/jrd/RecordSourceNodes.cpp
```diff
@@ -4378,6 +4386,92 @@ dsql_fld* UnlistFunctionSourceNode::makeField(DsqlCompilerScratch* dsqlScratch)
...
+		status_exception::raise(Arg::Gds(isc_argmustbe_exact_function) << Arg::Str(getName()));
+	}
+
+	dsc stepDesc;
+	DsqlDescMaker::fromNode(dsqlScratch, &stepDesc, stepItem, true);
+	if (!stepDesc.isExact() && !stepDesc.isNull()) {
```
asfernandes:
```suggestion
	if (!stepDesc.isExact() && !stepDesc.isNull())
	{
```

src/jrd/RecordSourceNodes.cpp
```diff
@@ -4378,6 +4386,92 @@ dsql_fld* UnlistFunctionSourceNode::makeField(DsqlCompilerScratch* dsqlScratch)
...
+	{
+		field = FB_NEW_POOL(dsqlScratch->getPool()) dsql_fld(dsqlScratch->getPool());
+
+		dsc desc;
+		if (dtype == dtype_int128) {
+			desc.makeInt128(scale);
+		}
+		else {
+			desc.makeInt64(scale);
+		}
```
asfernandes:
```suggestion
		if (dtype == dtype_int128)
			desc.makeInt128(scale);
		else
			desc.makeInt64(scale);

```

src/jrd/recsrc/RecordSource.h
sim1984: What are the alternatives? Use `union` or `impure_value`? Move the `m_start`, `m_finish`, `m_step`, and `m_result` values ​​outside of `Impure`?

I considered using `impure_value`, but it seemed redundant.

sim1984: Decided to use union.

src/dsql/parse.y
dyemanov: If the default value of one is function-specific (and thus can be hardly reused elsewhere), then I'd suggest something like `gen_series_step_opt`.

src/jrd/recsrc/RecordSource.h
aafemt: `m_start`, `m_finish` and `m_step` definitely don't belong to impure which is supposed to keep volatile data between node executions. Only current value should be there.

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -346,3 +348,211 @@ bool UnlistFunctionScan::nextBuffer(thread_db* tdbb) const
...
+
+	rpb->rpb_number.setValue(BOF_NUMBER);
+
+	fb_assert(m_inputList->items.getCount() >= GEN_SERIES_INDEX_LAST);
+
+	auto startItem = m_inputList->items[GEN_SERIES_INDEX_START];
+	const auto startDesc = EVL_expr(tdbb, request, startItem);
+	if (startDesc == nullptr)
+	{
+		rpb->rpb_number.setValid(false);
```
dyemanov: Here and below -- it's not necessary to invalidate the record number in this method. It will be invalidated later inside `internalGetRecord()` due to the missing `irsb_open` flag.

src/jrd/recsrc/RecordSource.h
dyemanov: > `m_start`, `m_finish` and `m_step` definitely don't belong to impure which is supposed to keep volatile data between node executions. Only current value should be there.

In this case they will be re-evaluated per every generated row, slowing down the overall performance. Although for literal arguments the difference will be hardly noticeable.

src/dsql/parse.y
sim1984: done

src/jrd/recsrc/TableValueFunctionScan.cpp
sim1984: done

dyemanov (approved):

asfernandes (approved):

asfernandes (changes requested):
A simple try to run this crashed the debug build.

```
SQL> select * FROM GENERATE_SERIES(1, 3) AS S(n);
Assertion (false) failure: /home/asfernandes/fb/dev/firebird-master.git/src/jrd/RecordSourceNodes.cpp 4323
Aborted (core dumped)
```

asfernandes:
This works:
```
select * FROM GENERATE_SERIES(1, 3) x
```

This do not work. Why, as the table alias is not even used in the above case?
```
select * FROM GENERATE_SERIES(1, 3)
```

This returns the GENERATE_SERIES label:
```
SQL> select * FROM GENERATE_SERIES(1, 3) x;

      GENERATE_SERIES 
===================== 
                    1 
                    2 
                    3 
```

This allow to select using the GENERATE_SERIES name created in the internal query:
```
SQL> select generate_series from (select * FROM GENERATE_SERIES(1, 3) x);

      GENERATE_SERIES 
===================== 
                    1 
                    2 
                    3 
```

This do not work. Why?
```
SQL> select generate_series FROM GENERATE_SERIES(1, 3) x;
Statement failed, SQLSTATE = 42S22
Dynamic SQL Error
-SQL error code = -206
-Column unknown
-"GENERATE_SERIES"
-At line 1, column 8
```

dyemanov:
@asfernandes, I can answer the first question. In the SQL spec, `<derived table>, <lateral derived table>, <collection derived table>, <table function derived table>, <JSON table>` -- all require `<correlation name>`. We followed the same rule while implementing the `UNLIST` function. And now the same approach was used for `GENERATE_SERIES` and I think it's OK.

dyemanov:
The second issue seems being a bug, because this works:
```sql
select unlist from (select * from unlist('1,2,3') as a);
```
and I expect this to work for `generate_series` too.

sim1984:
> The second issue seems being a bug, because this works:
> 
> ```sql
> select unlist from (select * from unlist('1,2,3') as a);
> ```
> 
> and I expect this to work for `generate_series` too.

This is exactly how `generate_series` works. See @asfernandes 4th query.

Query 5 doesn't work for unlist either.

```
SQL> select unlist from unlist('1,2,3') as a;
Statement failed, SQLSTATE = 42S22
Dynamic SQL Error
-SQL error code = -206
-Column unknown
-"UNLIST"
-At line 1, column 8
```

Query 5 doesn't work for unlist either. This is mentioned in `README.unlist.txt`, but the reason isn't explained.

> Unacceptable behavior:

asfernandes (approved):

pavel-zotov:
::: QA note :::
See tests in ```$QA_HOME/tests/functional/intfunc/generate_series/``` folder

PR #8800, opened 2025-11-06
refactor(vio): Initialize schema search path statically in `VIO_store()` to remove unnecessary memory allocations

TreeHunter9:
Recently, I have comparing flamegraphs for gbak restore v5 vs v6, found that VIO_store takes longer in v6 than in v5.
About 25% of the execution time of `VIO_store()` was taken up by `MemPool::allocate()` and `~ObjectsArray()`. As I can see these function calls related to new schema code, so they are not present in the v5 flame graph.

I have made `schemaSearchPath` static, and these calls was removed from flamegraphs, and restore speed for my test table (`create table tab1(v1 varchar(25), v2 varchar(25), v3 varchar(25), v4 varchar(25));`) increased by ~5%.

### DIFF
--- a/src/common/classes/init.h
+++ b/src/common/classes/init.h
@@ -147,10 +147,11 @@ public:
 		FB_NEW InstanceControl::InstanceLink<GlobalPtr, P>(this);
 	}
 
-	template <std::invocable TFunc>
+	template <typename TFunc>
+		requires(std::invocable<TFunc, MemoryPool&>)
 	GlobalPtr(TFunc initializationFunc)
 	{
-		instance = initializationFunc();
+		instance = initializationFunc(*getDefaultMemoryPool());
 		FB_NEW InstanceControl::InstanceLink<GlobalPtr, P>(this);
 	}
 
@@ -171,6 +172,7 @@ public:
 	{
 		return instance;
 	}
+
 	const T* get() const noexcept
 	{
 		return instance;
--- a/src/common/classes/init.h
+++ b/src/common/classes/init.h
@@ -147,6 +147,13 @@ public:
 		FB_NEW InstanceControl::InstanceLink<GlobalPtr, P>(this);
 	}
 
+	template <std::invocable TFunc>
+	GlobalPtr(TFunc initializationFunc)
+	{
+		instance = initializationFunc();
+		FB_NEW InstanceControl::InstanceLink<GlobalPtr, P>(this);
+	}
+
 	T* operator->() noexcept
 	{
 		return instance;
@@ -156,6 +163,15 @@ public:
 		return *instance;
 	}
 	T* operator&() noexcept
+	{
+		return instance;
+	}
+
+	T* get() noexcept
+	{
+		return instance;
+	}
+	const T* get() const noexcept
 	{
 		return instance;
 	}
--- a/src/common/classes/init.h
+++ b/src/common/classes/init.h
@@ -303,6 +303,64 @@ public:
 	}
 };
 
+template <typename T>
+class DefaultNonLazyInstanceAllocator
+{
+public:
+	template <typename... TArgs>
+	static T* create(TArgs&&... args)
+	{
+		return FB_NEW_POOL(*getDefaultMemoryPool()) T(std::forward<TArgs>(args)...);
+	}
+
+	static void destroy(T* inst) noexcept
+	{
+		delete inst;
+	}
+};
+
+// Best use case is for static local variable
+template <typename T, class TAllocator = DefaultNonLazyInstanceAllocator<T>, template <class I> class DestroyControl = DeleteInstance >
+class NonLazyInitInstance : private DestroyControl<NonLazyInitInstance<T, TAllocator, DestroyControl> >
+{
+public:
+	template <typename... TArgs>
+	NonLazyInitInstance(TArgs&&... args)
+	{
+		MutexLockGuard guard(*StaticMutex::mutex, "NonLazyInitInstance");
+		m_instance = TAllocator::create(std::forward<TArgs>(args)...);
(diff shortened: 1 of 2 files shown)


TreeHunter9:
I wanna add @asfernandes as a reviewer, but it seems I don't have permissions for that.

aafemt:
Cannot `ObjectArray<MetaString>` be made `constexpr` instead?

I understand that this is a hack for gbak only but still having static variable for schema path here smells.

aafemt:
BTW, what's difference between new `NonLazyInitInstance` and old `GlobalPtr`?

AlexPeshkoff:
If different gbak restore instances may have different schema paths use of static variable for it is very bad idea. 
What about storing appropriate pointer in attachment?

aafemt:
Isn't schema search path already a property of attachment?..

AlexPeshkoff:
Yes, there is one, but looks like gbak needs one more (I did not look deep into this code).

BTW, I do not see NonLazyInitInstance in src.

aafemt:
Perhaps you are looking at single commit of two.

TreeHunter9:
> Cannot `ObjectArray<MetaString>` be made `constexpr` instead?

No, `ObjectArray` use a memory allocation for every element, so it's not possible to make it constexpr. 

> BTW, what's difference between new NonLazyInitInstance and old GlobalPtr?

I need to pass arguments to ctor, but `GlobalPtr` does not have this option, and I don't want to change the old code (maybe I should) to avoid breaking anything.

aafemt:
> No, `ObjectArray` use a memory allocation for every element, so it's not possible to make it constexpr.

Is it unavoidable? std::array is able to be constexpr.

TreeHunter9:
> > No, `ObjectArray` use a memory allocation for every element, so it's not possible to make it constexpr.
> 
> Is it unavoidable? std::array is able to be constexpr.

I know, but `MET_qualify_existing_name()` expects `ObjectArray`, and this is an .epp file, so we can't pass `std::array` easily, we need some kind of `std::span`, but now passing `ObjectArray` will become impossible. It's very clunky, and I don't think it's worth it.

TreeHunter9:
Also, I don't quite understand why some variables and parameters related to the schema are `MetaString`, while others are `MetaName`, and they are converting back and forth wasting time. Shouldn't they all just be`MetaName` @asfernandes?

aafemt:
Perhaps better solution would be to use the schema search path from the attachment here, as in any other place, and set the path explicitly in gbak or implicitly during gbak attach.

asfernandes:
Please don't create another hack for initialize variables. We have more than the enough of them.

It's fine to use a const static, it's const anyway.

You can probably initialize it like IntlManager.cpp's ModulesMap, with a constructor that inserts the elements and a GlobalPtr.

asfernandes:
Or even better, add a new contructor to GlobalPtr to receive a lambda initializer. Something like:

```
GlobalPtr<MyContainer> myContainer(
    [] (auto& container)
    {
        container.push(...);
    }
);
```

src/jrd/vio.cpp
```diff
@@ -4223,7 +4223,10 @@ void VIO_store(thread_db* tdbb, record_param* rpb, jrd_tra* transaction)
 			}},
 		};
 
-		ObjectsArray<MetaString> schemaSearchPath({SYSTEM_SCHEMA, PUBLIC_SCHEMA});
+		static const GlobalPtr<ObjectsArray<MetaString>> schemaSearchPath([]()
+		{
+			return FB_NEW_POOL(*getDefaultMemoryPool()) ObjectsArray<MetaString>(*getDefaultMemoryPool(), {SYSTEM_SCHEMA, PUBLIC_SCHEMA});
```
asfernandes: Since you prefered your lambda to construct the instance, it should receive a pool parameter and use it.
It looks very weird to the lambda use getDefaultMemoryPool() directly.

src/common/classes/init.h
```diff
@@ -160,6 +167,15 @@ class GlobalPtr : private InstanceControl
 		return instance;
 	}
 
+	T* get() noexcept
+	{
+		return instance;
+	}
+	const T* get() const noexcept
```
asfernandes, line 176:
```suggestion

	const T* get() const noexcept
```

asfernandes (changes requested):

TreeHunter9:
> Also, I don't quite understand why some variables and parameters related to the schema are `MetaString`, while others are `MetaName`, and they are converting back and forth wasting time. Shouldn't they all just be`MetaName` @asfernandes?

@asfernandes What about this?

asfernandes:
Must review all cases, but MetaName is engine only (and very invasive, depending on TLS), while MetaString is generic. They are incompatible so any common algorithm must use MetaString.

AlexPeshkoff:
On 11/10/25 18:10, Adriano dos Santos Fernandes wrote:
> *asfernandes* left a comment (FirebirdSQL/firebird#8800) 
> <https://github.com/FirebirdSQL/firebird/pull/8800#issuecomment-3512263300>
>
> Must review all cases, but MetaName is engine only (and very invasive, 
> depending on TLS), while MetaString is generic.
>

Is not VIO part of engine?

> They are incompatible so any common algorithm must use MetaString.
>

Conversion is possible. We have
     MetaName(const Firebird::MetaString& s);
     MetaName& operator=(const Firebird::MetaString& s);
     operator Firebird::MetaString() const noexcept;
and appropriate comparisons.

There is one more historical issue. Descriptors were used in engine and 
qli (and may be gdef, do not remember exactly). Therefore a lot of 
things related with descriptors (including intl conversions) had to 
support use w/o TDBB. Now when qli is long ago gone away -  may be it's 
time to review use of descriptors and move them back to jrd.

PR #8803, opened 2025-11-16
Fix #8799: BUGCHECK "decompression overran buffer (179)" when WITH LOCK clause is used

### LINKED ISSUE #8799: BUGCHECK "decompression overran buffer (179)" when WITH LOCK clause is used
FB3/4/5 are affected.

How to reproduce:
Connection 1 in `isql`:
1. Create a table:
```
create table t1 (id integer, str1 varchar(50), str2 varchar(2000));
commit;
```
2. Insert a record:
```
insert into t1 select 0, NULL, rpad('', 2000, uuid_to_char(gen_uuid())) from rdb$database;
commit;
```
3. Select this record and leave the transaction active to make sure it won't be garbage-collected:
```
select * from t1;
```

Connection 2 in `isql`:
4. Update the record in a way that stores the current version as a delta:
```
update t1 set id = 1;
commit;
```
5. Drop `str1` field:
```
alter table t1 drop str1;
commit;
```
6. Use `WITH LOCK` clause (BUGCHECK may occur here for FB4/5):
```
select * from t1 with lock;
```

Return to connection 1:
7. Select the old record version again:
```
select * from t1;
```
8. Get BUGCHECK "decompression overran buffer (179)".

What happens:
`VIO_writelock` creates `new_rpb` by copying from `org_rpb` which has `rpb_delta` flag set. `new_rpb` is passed to `prepare_update`. It creates a delta but its size exceeds the limit (1024) so the current version is stored as a regular record. The problem is that `rpb_delta` flag remains set for `new_rpb`, and then `replace_record` sets it in the header of the primary version. From this moment any attempt to get data of older versions fails.
The fix I would suggest: let `prepare_update` clear `rpb_delta` flag from `new_rpb` just before an attempt to generate a delta.

The issue #8676 looks kinda similar and the format was changed there as well but I'm not sure if it's related or not.

ilya071294:
`VIO_writelock` creates `new_rpb` by copying from `org_rpb` which has `rpb_delta` flag set. `new_rpb` is passed to `prepare_update`. It creates a delta but its size exceeds the limit (1024) so the current version is stored as a regular record. The problem is that `rpb_delta` flag remains set for `new_rpb`, and then `replace_record` sets it in the header of the primary version. From this moment any attempt to get data of older versions fails.
The fix is to let `prepare_update` clear `rpb_delta` flag from `new_rpb` just before an attempt to generate a delta.

### DIFF
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -6557,6 +6557,8 @@ static PrepareResult prepare_update(thread_db* tdbb, jrd_tra* transaction, TraNu
 
 	if (new_rpb)
 	{
+		new_rpb->rpb_flags &= ~rpb_delta;
+
 		// If both descriptors share the same record, there cannot be any difference.
 		// This trick is used by VIO_writelock(), but can be a regular practice as well.
 		if (new_rpb->rpb_address == temp->rpb_address)


hvlad (commented):
Looks ok. 

To be completely sure: have you considered to change `VIO_writelock()` instead ? 
What about other flags, copied from old record (`rpb_incomplete`, for example) ?

ilya071294:
> To be completely sure: have you considered to change `VIO_writelock()` instead ?

Yes, but that leaves open the possibility of facing a similar bug again, so I see it as a worse solution.

> What about other flags, copied from old record (`rpb_incomplete`, for example) ?

I thought about it and came to this logic: if `prepare_update` determines how the backversion is stored then it should have full control over `rpb_delta` flag. It doesn't do anything else with `new_rpb`, so it should not care about other flags. The caller should. And both `VIO_writelock` and `VIO_modify` handle other flags correctly as I can see.

hvlad (approved):

PR #8808, opened 2025-11-22
Per pagespace I/O statistics and new trace API interfaces to allow extendable statistics

dyemanov:
- Deprecate non-extendable PerformanceInfo struct in favor of the new PerformanceCounters/PerformanceStats interfaces
- Adjust the trace implementation to the new API
- Add internal support for per-pagespace I/O statistics. It's not used now, but will be used by PR #8314. This part is moved here to simplify review of the tablespaces.

### DIFF
--- a/doc/Using_OO_API.md
+++ b/doc/Using_OO_API.md
@@ -2136,9 +2136,9 @@ If `getCounters()` is called with a counter group not supported by the implement
 <a name="PerformanceCounters"></a> PerformanceCounters interface:
 
 - unsigned getObjectCount() - returns number of objects (e.g. tables) containing non-zero performance counters
-- unsigned getCountersCapacity() - returns total number of performance counters supported by the implementation (it's the same for all objects of the same type)
+- unsigned getMaxCounterIndex() - returns maximum index number of the performance counters supported by the implementation (it's the same for all objects of the same group)
 - unsigned getObjectId(unsigned index) - returns ID of the specified object
 - const char* getObjectName(unsigned index) - returns name of the specified object
-- const ISC_INT64* getObjectCounters(unsigned index) - returns pointer to the vector of performance counters (containing getCountersCapacity() elements) of the specified object
+- const ISC_INT64* getObjectCounters(unsigned index) - returns pointer to the vector of performance counters (containing getMaxCounterIndex() + 1 elements) of the specified object
 
 The returned pointer to the vector (as well as the whole instance of `PerformanceStats`) is valid until the object used to obtain the statistics (using the `getPerfStats()` method) is destroyed.
--- a/doc/Using_OO_API.md
+++ b/doc/Using_OO_API.md
@@ -2124,8 +2124,14 @@ The returned pointer may be `nullptr` if the corresponding trace object is not i
 
 - ISC_UINT64 getElapsedTime() - returns the elapsed time, in milliseconds
 - ISC_UINT64 getFetchedRecords() - returns number of records fetched during execution
-- IPerformanceCounters* getPageCounters() - returns per-pagespace performance counters
-- IPerformanceCounters* getTableCounters() - returns per-table performance counters
+- IPerformanceCounters* getCounters(unsigned group) - returns the requested performance counters group
+
+The following groups of performance counters are currently supported:
+
+- COUNTER_GROUP_PAGES - per-pagespace counters
+- COUNTER_GROUP_TABLES - per-table counters
+
+If `getCounters()` is called with a counter group not supported by the implementation, `nullptr` is returned.
 
 <a name="PerformanceCounters"></a> PerformanceCounters interface:
 
--- a/doc/Using_OO_API.md
+++ b/doc/Using_OO_API.md
@@ -2101,3 +2101,38 @@ struct FbVarChar
 
 This document is currently missing 2 types of plugins – ExternalEngine and Trace. Information about them will be made
 available in next release of it.
+
+# Trace plugin
+
+_TODO_
+
+# Trace objects
+
+_TODO_
+
+# Trace performance statistics
+
+Trace plugin may retrieve various performance statistics available using the `getPerfStats()` method of the trace object, which returns a pointer to the `IPerformanceStats` interface.
+
+```cpp
+IPerformanceStats* stats = statement->getPerfStats();
+```
+
+The returned pointer may be `nullptr` if the corresponding trace object is not in the terminate state yet (i.e. still active / being executed).
+
+<a name="PerformanceStats"></a> PerformanceStats interface:
+
+- ISC_UINT64 getElapsedTime() - returns the elapsed time, in milliseconds
+- ISC_UINT64 getFetchedRecords() - returns number of records fetched during execution
+- IPerformanceCounters* getPageCounters() - returns per-pagespace performance counters
+- IPerformanceCounters* getTableCounters() - returns per-table performance counters
+
+<a name="PerformanceCounters"></a> PerformanceCounters interface:
+
+- unsigned getObjectCount() - returns number of objects (e.g. tables) containing non-zero performance counters
+- unsigned getCountersCapacity() - returns total number of performance counters supported by the implementation (it's the same for all objects of the same type)
+- unsigned getObjectId(unsigned index) - returns ID of the specified object
+- const char* getObjectName(unsigned index) - returns name of the specified object
+- const ISC_INT64* getObjectCounters(unsigned index) - returns pointer to the vector of performance counters (containing getCountersCapacity() elements) of the specified object
+
+The returned pointer to the vector (as well as the whole instance of `PerformanceStats`) is valid until the object used to obtain the statistics (using the `getPerfStats()` method) is destroyed.
--- a/src/include/firebird/FirebirdInterface.idl
+++ b/src/include/firebird/FirebirdInterface.idl
@@ -1899,12 +1899,35 @@ interface ProfilerStats : Versioned
 
 interface PerformanceCounters : Versioned
 {
-	uint getCount();
-	uint getVectorCapacity();
+	// Page-level performance counters (grouped per tablespace)
+	const uint PAGE_FETCHES = 0;
+	const uint PAGE_READS = 1;
+	const uint PAGE_MARKS = 2;
+	const uint PAGE_WRITES = 3;
 
-	uint getId(uint index);
-	const string getName(uint index);
-	const int64* getCounterVector(uint index);
+	// Record-level performance counters (grouped per table)
+	const uint RECORD_SEQ_READS = 0;
+	const uint RECORD_IDX_READS = 1;
+	const uint RECORD_UPDATES = 2;
+	const uint RECORD_INSERTS = 3;
+	const uint RECORD_DELETES = 4;
+	const uint RECORD_BACKOUTS = 5;
+	const uint RECORD_PURGES = 6;
+	const uint RECORD_EXPUNGES = 7;
+	const uint RECORD_LOCKS = 8;
+	const uint RECORD_WAITS = 9;
+	const uint RECORD_CONFLICTS = 10;
+	const uint RECORD_BACK_READS = 11;
+	const uint RECORD_FRAGMENT_READS = 12;
+	const uint RECORD_RPT_READS = 13;
+	const uint RECORD_IMGC = 14;
+
+	uint getObjectCount();
+	uint getCountersCapacity();
+
+	uint getObjectId(uint index);
+	const string getObjectName(uint index);
+	const int64* getObjectCounters(uint index);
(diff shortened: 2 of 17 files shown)


aafemt:
Documentation is missing.

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1876,3 +1894,26 @@ interface ProfilerStats : Versioned
...
+	const string getName(uint index);
+	const int64* getCounterVector(uint index);
+}
+
+interface PerformanceStats : Versioned
+{
+	uint64 getElapsedTime();	// in milliseconds
+	uint64 getFetchedRecords();
+
+	PerformanceCounters getGlobalCounters();
```
aafemt: If you claim expandability, this must be one function with enum parameter: `cGlobal`, `cPage`, `cRelation`.

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1876,3 +1894,26 @@ interface ProfilerStats : Versioned
...
+// Extendable replacement for struct PerformanceInfo
+
+interface PerformanceCounters : Versioned
+{
+	uint getCount();
+	uint getVectorCapacity();
+
+	uint getId(uint index);
+	const string getName(uint index);
+	const int64* getCounterVector(uint index);
```
aafemt: TTL of returned value is...?

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1876,3 +1894,26 @@ interface ProfilerStats : Versioned
...
+
+// Extendable replacement for struct PerformanceInfo
+
+interface PerformanceCounters : Versioned
+{
+	uint getCount();
+	uint getVectorCapacity();
+
+	uint getId(uint index);
+	const string getName(uint index);
```
aafemt: ID and name of... what?

src/include/firebird/FirebirdInterface.idl
dyemanov: We can easily add more functions to extend the interface. But I may agree a single function would be simpler.

src/include/firebird/FirebirdInterface.idl
dyemanov: > TTL of returned value is...?

It corresponds to the object TTL.

dyemanov:
> Documentation is missing.

True, to be added.

src/include/firebird/FirebirdInterface.idl
dyemanov: > ID and name of... what?

Object (table, tablespace, whatever) counters belong to. Would you suggest `getObjectId() / getObjectName()`?

src/include/firebird/FirebirdInterface.idl
aafemt: Well... Returning of a plain array is simple and effective, but doesn't quite fit the ideology of OO API. How about returning of another object with methods `getRead`, `getWrites`, `getFetches`, etc? In this case `getVectorCapacity` is not needed (I guess).

src/include/firebird/FirebirdInterface.idl
aafemt: No, I just wonder what these methods shall return for Global and Page counters.

dyemanov: Nothing (zero) for global, valid id/name otherwise.

aafemt: Perhaps, it could be useful to return hardcoded "Global" and "Pages" in these cases...

src/include/firebird/FirebirdInterface.idl
dyemanov: Valid point that I don't have a strong opinion about. Given the performance stats being somewhat low-level thing, I tried to follow the old-style interface with struct and counter vector, just made it extendable for new counter groups. Also, when we add new counters into the engine they automatically become available in the API, even being undocumented as enum members they're still usable without the need to recompile your trace plugin. With the pure OO approach the trace plugin should be recompiled to utilize new interface methods (and they should be added every time we add new counters).

@AlexPeshkoff @hvlad @asfernandes What do you think about this?

AlexPeshkoff: I do not think that adding separate method for each counter is good idea. An additional argument not to follow OO rules too strict here is that calling virtual function once and working with array elements after it should be faster than calling N virtual functions.

aafemt: > when we add new counters into the engine they automatically become available in the API, even being undocumented as enum members they're still usable without the need to recompile your trace plugin

That's true but the problem is that new trace plugins will crash on old servers.


> With the pure OO approach the trace plugin should be recompiled to utilize new interface methods

As well as to utilize new enum constants. But CLOOP makes it safe.

BTW, in any case shouldn't these constants to be moved into the interface declaration?

aafemt: > An additional argument not to follow OO rules too strict here is that calling virtual function once and working with array elements after it should be faster than calling N virtual functions.

That's why functions in CLOOP aren't virtual.

dyemanov: > That's true but the problem is that new trace plugins will crash on old servers.

They won't if they respect `getVectorCapacity()` and don't access the vector beyond its reported limits.

> BTW, in any case shouldn't these constants to be moved into the interface declaration?

Yes, I think it would be better.

aafemt: > They won't if they respect `getVectorCapacity()` and don't access the vector beyond its reported limits.

This comparison is slower than virtual function call and has to be copy-pasted in every implemented plugin. Not a big deal, but...

asfernandes: > > An additional argument not to follow OO rules too strict here is that calling virtual function once and working with array elements after it should be faster than calling N virtual functions.
> 
> That's why functions in CLOOP aren't virtual.

That's NOT the reason.
And the cost is the same, an indirect call.

aafemt: > That's NOT the reason.

In the beginning I voted for pure virtual interfaces. That time non-virtual CLOOP functions **was** argument against them.

asfernandes: > Also, when we add new counters into the engine they automatically become available in the API, even being undocumented as enum

How do a v7 caller of this interface can know the returned array from a v6 implementation does not have its expected items?

asfernandes: > > That's NOT the reason.
> 
> In the beginning I voted for pure virtual interfaces. That time non-virtual CLOOP functions **was** argument against them.

Looks like you missed the whole thing or are insisting on bad things 20 years later. Virtual functions has no standard ABI layout between different compilers.

aafemt: > Virtual functions has no standard ABI layout between different compilers.

You are dead wrong.

hvlad: > I do not think that adding separate method for each counter is good idea. An additional argument not to follow OO rules too strict here is that calling virtual function once and working with array elements after it should be faster than calling N virtual functions.

+1

src/include/firebird/FirebirdInterface.idl
hvlad: > Would you suggest `getObjectId() / getObjectName()`?

Yes, please. Now it is confusing as could be read as `ObjectName` or `CounterName`

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1876,3 +1894,26 @@ interface ProfilerStats : Versioned
 {
 	uint64 getElapsedTicks();
 }
+
+// Extendable replacement for struct PerformanceInfo
+
+interface PerformanceCounters : Versioned
+{
+	uint getCount();
+	uint getVectorCapacity();
```
hvlad: What it purpose of this method ? Why `getCount()` is not enough ?

src/include/firebird/FirebirdInterface.idl
dyemanov: > > Also, when we add new counters into the engine they automatically become available in the API, even being undocumented as enum
> 
> How do a v7 caller of this interface can know the returned array from a v6 implementation does not have its expected items?

If you want to access 9th counter, but `getVectorCapacity()` returns 7, then you're out of luck.

src/include/firebird/FirebirdInterface.idl
dyemanov: `getCount()` returns a number of objects (tables, tablespaces, etc), each of them has a vector of `getVectorCapacity()` counters. `getVectorCapacity()` is intended to ensure compatibility without interface versioning -- you can easily check whether the engine supports the counter you'd like to access.

dyemanov: I suppose `getObjectCount()` would suit better here, given we discuss `getObjectId() / getObjectName()`.

aafemt: I would suggest `maxCounter` because it is count of counters, not objects, and it is constant for every object/counter's set.

src/include/firebird/FirebirdInterface.idl
dyemanov: > No, I just wonder what these methods shall return for Global and Page counters.

Thinking twice, I suppose `getGlobalCounters()` is not needed at all. Currently, it's used only to report page-level stats in the trace and this output will be modified anyway when tablespaces are merged (per-tablespace stats will be printed instead). And any global counter may be aggregated manually from the appropriate `PerformanceCounters` interface.

dyemanov:
> > Documentation is missing.
> 
> True, to be added.

The trace interfaces seem being not documented entirely, see _Using_OO_API.md_:

> This document is currently missing 2 types of plugins – ExternalEngine and Trace. Information about them will be made available in next release of it.

I'm not in position to document the whole bunch of the trace interfaces right now, sorry. So documentation gonna be deferred.

aafemt:
> I'm not in position to document the whole bunch of the trace interfaces right now, sorry.

Limit to your new interfaces and methods.

doc/Using_OO_API.md
```diff
@@ -2101,3 +2101,44 @@ struct FbVarChar
...
+IPerformanceStats* stats = statement->getPerfStats();
+```
+
+The returned pointer may be `nullptr` if the corresponding trace object is not in the terminate state yet (i.e. still active / being executed).
+
+<a name="PerformanceStats"></a> PerformanceStats interface:
+
+- ISC_UINT64 getElapsedTime() - returns the elapsed time, in milliseconds
+- ISC_UINT64 getFetchedRecords() - returns number of records fetched during execution
+- IPerformanceCounters* getCounters(unsigned group) - returns the requested performance counters group
```
hvlad, line 2127: I read it as "returns the group", not "returns the counters".
May be a bit better to change it by "returns the performance counters of the requested group"

doc/Using_OO_API.md
```diff
@@ -2101,3 +2101,44 @@ struct FbVarChar
...
+
+- COUNTER_GROUP_PAGES - per-pagespace counters
+- COUNTER_GROUP_TABLES - per-table counters
+
+If `getCounters()` is called with a counter group not supported by the implementation, `nullptr` is returned.
+
+<a name="PerformanceCounters"></a> PerformanceCounters interface:
+
+- unsigned getObjectCount() - returns number of objects (e.g. tables) containing non-zero performance counters
+- unsigned getCountersCapacity() - returns total number of performance counters supported by the implementation (it's the same for all objects of the same type)
```
hvlad: Sorry, "capacity" doesn't sounds good to me. Why not use "total" as in description ?
For example: `getTotalCounters()`

aafemt: Why not simply "size of the vector returned by getObjectCounters()"?

dyemanov: > Sorry, "capacity" doesn't sounds good to me. Why not use "total" as in description ? For example: `getTotalCounters()`

"capacity" was just the first synonym for "count" that comes to mind and we're used to it in our codebase. I'm OK with any better name. However, I'm not sure "total" is better here, because it's often used with the "sum/aggregate" meaning. If `getObjectCounters()` means vector of per-object counters, then `getTotalCounters()` looks like vector of aggregated counters. I will think more to suggest better naming.

AlexPeshkoff: May be getNumberOfCounters

dyemanov: Maybe Dmitry Sibiryakov's suggestion really suits better here: `getMaxCounterIndex()` or `getMaxCounterId()` or `getMaxCounterType()`.

hvlad: `getCountersCount()` ? :)

PR #8812, opened 2025-11-25
Fixes a loop in the GENERATE_SERIES function on boundary values.

sim1984:
The GENERATE_SERIES function behaves incorrectly at boundary values ​​for the BIGINT and INT128 types. A loop occurs with the BIGINT type, and an exception is thrown with the INT128 type. Examples of incorrect behavior:

```sql
select * from generate_series(-9223372036854775807, -9223372036854775809, -1) as x rows 10;
```

```
INPUT message field count: 0
 
OUTPUT message field count: 1
01: sqltype: 32752 INT128 scale: 0 subtype: 0 len: 16
  :  name: GENERATE_SERIES  alias: GENERATE_SERIES
  : table:   schema:   owner:
 
                              GENERATE_SERIES
=============================================
                         -9223372036854775807
                         -9223372036854775808
                         -9223372036854775809
```

(OK, expected)

```sql
select * from generate_series(-9223372036854775807, -9223372036854775808, -1) as x rows 10;
```

```
INPUT message field count: 0
 
OUTPUT message field count: 1
01: sqltype: 580 INT64 scale: 0 subtype: 0 len: 8
  :  name: GENERATE_SERIES  alias: GENERATE_SERIES
  : table:   schema:   owner:
 
      GENERATE_SERIES
=====================
 -9223372036854775807
 -9223372036854775808
  9223372036854775807 // <-- Error. 
  9223372036854775806
  9223372036854775805
  9223372036854775804
  9223372036854775803
  9223372036854775802
  9223372036854775801
  9223372036854775800
===============
```

For `INT128` example:

```sql
select * from generate_series(bin_shl(cast(2 as int128),126)+1, bin_shl(cast(2 as int128),126), -1) as x rows 10;
```

```
INPUT message field count: 0

OUTPUT message field count: 1
01: sqltype: 32752 INT128 scale: 0 subtype: 0 len: 16
  :  name: GENERATE_SERIES  alias: GENERATE_SERIES
  : table:   schema:   owner:

                              GENERATE_SERIES
=============================================
     -170141183460469231731687303715884105727
-Integer overflow.  The result of an integer operation caused the most significant bit of the result to carry.
```

### DIFF
--- a/src/jrd/recsrc/TableValueFunctionScan.cpp
+++ b/src/jrd/recsrc/TableValueFunctionScan.cpp
@@ -522,7 +522,13 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 			fromDesc.makeInt64(impure->m_scale, &result);
 			assignParameter(tdbb, &fromDesc, toDesc, 0, record);
 
-			result += step;
+			// Fixes freezing at boundary values.
+			if (((step < 0) && (result == MIN_SINT64)) || 
+				((step > 0) && (result == MAX_SINT64)))
+				impure->m_step.vlu_int64 = 0;
+			else
+			    result += step;
+
 			impure->m_result.vlu_int64 = result;
 
 			return true;
@@ -545,7 +551,12 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 			fromDesc.makeInt128(impure->m_scale, &result);
 			assignParameter(tdbb, &fromDesc, toDesc, 0, record);
 
-			result = result.add(step);
+			// Fixes freezing at boundary values.
+			if (((step.sign() < 0) && (result.compare(MIN_Int128) == 0)) ||
+				((step.sign() > 0) && (result.compare(MAX_Int128) == 0)))
+				impure->m_step.vlu_int128.set(0, 0);
+			else
+				result = result.add(step);
 			impure->m_result.vlu_int128 = result;
 
 			return true;
--- a/src/jrd/recsrc/TableValueFunctionScan.cpp
+++ b/src/jrd/recsrc/TableValueFunctionScan.cpp
@@ -441,7 +441,7 @@ void GenSeriesFunctionScan::internalOpen(thread_db* tdbb) const
 	}
 	// result = start
 	MOV_move(tdbb, startDesc, &impure->m_result.vlu_desc);
-	
+
 	impure->irsb_flags |= irsb_open;
 
 	VIO_record(tdbb, rpb, m_format, &pool);
@@ -493,9 +493,9 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 	const auto impure = request->getImpure<Impure>(m_impure);
 
 
-	const auto ñomparison = MOV_compare(tdbb, &impure->m_result.vlu_desc, &impure->m_finish.vlu_desc);
-	if (((impure->m_stepSign > 0) && (ñomparison <= 0)) ||
-		((impure->m_stepSign < 0) && (ñomparison >= 0)))
+	const auto comparison = MOV_compare(tdbb, &impure->m_result.vlu_desc, &impure->m_finish.vlu_desc);
+	if (((impure->m_stepSign > 0) && (comparison <= 0)) ||
+		((impure->m_stepSign < 0) && (comparison >= 0)))
 	{
 		Record* const record = request->req_rpb[m_stream].rpb_record;
 
@@ -504,7 +504,7 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 		assignParameter(tdbb, &impure->m_result.vlu_desc, toDesc, 0, record);
 
 		// evaluate next result
-		try 
+		try
 		{
 			impure_value nextValue;
 
--- a/src/jrd/recsrc/TableValueFunctionScan.cpp
+++ b/src/jrd/recsrc/TableValueFunctionScan.cpp
@@ -429,16 +429,18 @@ void GenSeriesFunctionScan::internalOpen(thread_db* tdbb) const
 	EVL_make_value(tdbb, finishDesc, &impure->m_finish);
 	EVL_make_value(tdbb, stepDesc, &impure->m_step);
 
-	switch (impure->m_result.vlu_desc.dsc_dtype) {
-	case dtype_int64:
-		impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int64);
-		break;
-	case dtype_int128:
-		impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int128);
-		break;
-	default:
-		fb_assert(false);
+	switch (impure->m_result.vlu_desc.dsc_dtype)
+	{
+		case dtype_int64:
+			impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int64);
+			break;
+		case dtype_int128:
+			impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int128);
+			break;
+		default:
+			fb_assert(false);
 	}
+
 	// result = start
 	MOV_move(tdbb, startDesc, &impure->m_result.vlu_desc);
 
@@ -492,7 +494,6 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 	const auto request = tdbb->getRequest();
 	const auto impure = request->getImpure<Impure>(m_impure);
 
-
 	const auto comparison = MOV_compare(tdbb, &impure->m_result.vlu_desc, &impure->m_finish.vlu_desc);
 	if (((impure->m_stepSign > 0) && (comparison <= 0)) ||
 		((impure->m_stepSign < 0) && (comparison >= 0)))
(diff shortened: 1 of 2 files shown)


src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -522,7 +522,13 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 			fromDesc.makeInt64(impure->m_scale, &result);
 			assignParameter(tdbb, &fromDesc, toDesc, 0, record);
 
-			result += step;
+			// Fixes freezing at boundary values.
+			if (((step < 0) && (result == MIN_SINT64)) || 
+				((step > 0) && (result == MAX_SINT64)))
+				impure->m_step.vlu_int64 = 0;
```
dyemanov:
```suggestion
			{
				impure->m_step.vlu_int64 = 0;
			}
```

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -545,7 +551,12 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 			fromDesc.makeInt128(impure->m_scale, &result);
 			assignParameter(tdbb, &fromDesc, toDesc, 0, record);
 
-			result = result.add(step);
+			// Fixes freezing at boundary values.
+			if (((step.sign() < 0) && (result.compare(MIN_Int128) == 0)) ||
+				((step.sign() > 0) && (result.compare(MAX_Int128) == 0)))
+				impure->m_step.vlu_int128.set(0, 0);
```
dyemanov:
```suggestion
			{
				impure->m_step.vlu_int128.set(0, 0);
			}
```

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -522,7 +522,13 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 			fromDesc.makeInt64(impure->m_scale, &result);
 			assignParameter(tdbb, &fromDesc, toDesc, 0, record);
 
-			result += step;
+			// Fixes freezing at boundary values.
```
dyemanov:
```suggestion
			// Prevent freezing at boundary values
```

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -545,7 +551,12 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 			fromDesc.makeInt128(impure->m_scale, &result);
 			assignParameter(tdbb, &fromDesc, toDesc, 0, record);
 
-			result = result.add(step);
+			// Fixes freezing at boundary values.
```
dyemanov:
```suggestion
			// Prevent freezing at boundary values
```

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -522,7 +522,15 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
 			fromDesc.makeInt64(impure->m_scale, &result);
 			assignParameter(tdbb, &fromDesc, toDesc, 0, record);
 
-			result += step;
+			// Prevent freezing at boundary values
+			if (((step < 0) && (result == MIN_SINT64)) ||
+				((step > 0) && (result == MAX_SINT64)))
```
asfernandes: Is this logic correct for scaled numbers?

sim1984: This logic doesn't work if `step` is different from 1 or -1. I'll rework this solution for a more general case.

aafemt: Do these dynamic type switches needed? Why not count in output data type from the beginning?

sim1984: > Do these dynamic type switches needed? Why not count in output data type from the beginning?

The type is already chosen based on the most general of the three parameters. However, calculations with the INT128 type are handled differently than with INT64, so these are separate code paths.

sim1984: > Is this logic correct for scaled numbers?

This is a protection against internal type overflows. It's unlikely that you can specify scaled values ​​at the function parameter level without causing an error sooner. At least these errors will be detected and won't simply lead to a loop.

asfernandes: Can't `ArithmeticNode::add` be used like in `ForRangeNode`?

sim1984: > Can't `ArithmeticNode::add` be used like in `ForRangeNode`?

Are there any significant advantages to this approach?

Testing shows that the for range loop suffers from the same problems, but they manifest themselves in a slightly different way.

```sql
execute block
returns (i bigint)
as
begin
  for i = 9223372036854775806 to 9223372036854775807 by 1 do
    suspend;
end
```

```
                    I
=====================
  9223372036854775806
  9223372036854775807
Integer overflow.  The result of an integer operation caused the most significant bit of the result to carry.
-At block line: 6, col: 5
```

```sql
execute block
returns (i bigint)
as
begin
  for i = -9223372036854775807 downto -9223372036854775808 do
    suspend;
end
```

```
                    I
=====================
 -9223372036854775807
 -9223372036854775808
Integer overflow.  The result of an integer operation caused the most significant bit of the result to carry.
-At block line: 6, col: 5
```

That is, there is no wrap of the signed integer, but the calculations do not stop at the boundaries.

asfernandes: > Are there any significant advantages to this approach?

Not deal with different types every time.

The errors for me are correct, the step is applied and next the condition is checked. The overflow happens when applying the next step.

sim1984: Although I admit that for a FOR RANGE LOOP, this error is perfectly logical. However, `generate_seriers` works a little differently. Such errors are definitely not expected there.

PostgreSQL example:

```sql
select * from generate_series(-9223372036854775807, -9223372036854775808, -1) as x;
```
```
|----------------------|
| x                    |
|----------------------|
| -9223372036854775807 |
| -9223372036854775808 |
```

```sql
select * from generate_series(-9223372036854775806, -9223372036854775807, -3) as x
```
```
|----------------------|
| x                    |
|----------------------|
| -9223372036854775806 |
```

sim1984: In my opinion, using `ArithmeticNode::add` to control computation boundaries is quite difficult. When a user uses the `generate_series` table function, they certainly don't think of it as a loop. And the goal of this PR was precisely to fix the loop/error issue where it shouldn't occur. A refactoring is possible in the future, if anyone deems it useful.

asfernandes: > In my opinion, using `ArithmeticNode::add` to control computation boundaries is quite difficult. When a user uses the `generate_series` table function, they certainly don't think of it as a loop. And the goal of this PR was precisely to fix the loop/error issue where it shouldn't occur. A refactoring is possible in the future, if anyone deems it useful.

You have the option to enclose it in try-catch and swallow the error. It will do only in boundaries case.
Isn't it a good compromise to avoid extra verifications and custom implementations for different data types?

sim1984: Okay, I'll try to do it this way.

sim1984: Done

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -407,53 +407,40 @@ void GenSeriesFunctionScan::internalOpen(thread_db* tdbb) const
...
+	case dtype_int64:
+		impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int64);
+		break;
+	case dtype_int128:
+		impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int128);
+		break;
+	default:
+		fb_assert(false);
 	}
+	// result = start
```
asfernandes:
```suggestion
	switch (impure->m_result.vlu_desc.dsc_dtype)
	{
		case dtype_int64:
			impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int64);
			break;
		case dtype_int128:
			impure->m_result.vlu_desc.dsc_address = reinterpret_cast<UCHAR*>(&impure->m_result.vlu_misc.vlu_int128);
			break;
		default:
			fb_assert(false);
	}

	// result = start
```

src/jrd/recsrc/TableValueFunctionScan.cpp
```diff
@@ -505,51 +492,40 @@ bool GenSeriesFunctionScan::nextBuffer(thread_db* tdbb) const
...
-
-			result = result.add(step);
-			impure->m_result.vlu_int128 = result;
-
-			return true;
+			tdbb->tdbb_status_vector->clearException();
+			// stop evaluate next result
+			impure->m_stepSign = 0;
 		}
+		return true;
```
asfernandes, line 530:
```suggestion

		return true;
```

src/jrd/recsrc/TableValueFunctionScan.cpp
sim1984: Done

src/jrd/recsrc/TableValueFunctionScan.cpp
sim1984: Done

asfernandes (approved):

PR #8815, opened 2025-11-29
Restore the broken record layout optimization by gbak and extend it to the new datatypes

dyemanov:
The order of field ID generation inside DFW is not reliable, because it joins two system tables and generally the retrieval order would depend on the query plan. The added `SORTED BY` clause fixes it for `CREATE/ALTER TABLE`, but the resulting order does not match the order of insertions into `RDB$RELATION_FIELDS` anymore. However, _gbak_ does its best to optimize the record layout and it relies on the ID generation being performed in the storage order. Thus I moved the ID generation to the _gbak_ side to ensure the record format is created with the expected (optimized) layout.

### DIFF
--- a/src/burp/backup.epp
+++ b/src/burp/backup.epp
@@ -1940,6 +1940,7 @@ void put_relation( burp_rel* relation)
 	burp_fld* unaligned = NULL;
 	burp_fld* aligned4 = NULL;
 	burp_fld* aligned8 = NULL;
+	burp_fld* aligned16 = NULL;
 
 	burp_fld* fields = get_fields(relation);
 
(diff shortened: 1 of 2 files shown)


dyemanov:
Before this patch:
```
        Fields:
         id offset type           length sub_type flags
        --- ------ -------------- ------ -------- -----
          0      8 19 BIGINT           8        0  0x00
          1     16  3 VARCHAR         33       52  0x00
          2     56 17 BLOB             8        1  0x00

Format length = 64 bytes
```
After this patch:

```
        Fields:
         id offset type           length sub_type flags
        --- ------ -------------- ------ -------- -----
          0      8 17 BLOB             8        1  0x00
          1     16 19 BIGINT           8        0  0x00
          2     24  3 VARCHAR         33       52  0x00

Format length = 57 bytes
```

PR #8819, opened 2025-11-30
Shared metadata cache 

AlexPeshkoff:
Tasks to be solved.
- Have a single metadata cache that is used by all attachments in both jrd and dsql.
- Provide the ability to mix DLL/DML in a single transaction, including working with new/modified elements.

The old metadata cache is essentially single-threaded. For me, the problem I faced with attempts to introduce synchronization into this structure and get off with relatively little blood was taking into account the use of procedures and functions using a second reference counter, and in order to decide whether to delete an object, a global recalculation of these references is required throughout the cache. dsqlcache is also extremely unsympathetic, which puts objects created and still unmixed in a single transaction for all transactions of the same attach. An additional consideration is that objects stored in the cache are used in a huge number of places in the code, and they can be modified from other places (and threads) during MET_scan_xxx. This could still be dealt with using CopyOnWrite, but how to approach the first two points was unclear.

The new metadata cache is versioned. Each object is divided into 2 parts - a constant and a variable. For relation, for example, the type (table, view, external table, etc.), name, all locks (GCLock, partners lock, rescan lock) will be constant, and format, list of fields, triggers will be variable. The list of versions is linked to a permanent part of the object. As versions become obsolete, the list is cleared (although I won't say that it's very active - only as the cache is accessed, there is no special garbage collector).

In order to access the desired object for reading without locks, it was possible to allocate memory for the entire cache at once. In principle, there are not so many - 64K pointers to each type of cached objects. But I kept in mind the possible increase in the number of objects in the database and therefore made a 2-level array - each lower-level block of N objects is allocated once and for all (until the database is closed), and the upper-level block has the right to grow when the required number of lower-level blocks does not fit there. Therefore, the upper level is implemented using a read-safe array. For reading, we would take a slice of the current state of the array and this slice remains unchanged while we work with it. An external mutex is used to modify the array. This array (SharedReadVector) turned out to be very convenient, I also use it for requests in a statement and a list of formats in a relation. The 2-level arrays (CacheVector) for all objects are the same (template), and all this ultimately lives in DBB.

Ideally, I would like to remove dsqlcache altogether. So far, it has been removed for relations, procedures, and functions. At the same time, I did not abandon the dsql_rel format and the like - they are needed in order to be able to work with it when creating/changing an object (dsql_rel), although there is no such object in the metadata cache yet / or the format is not the same. Temporary objects of this format are created in dsqlScratch and die with it (by pool). 

libcds is used to implement secure pointers (HazardPtr), they are needed for version headers in the list of object versions and for a slice in the SharedReadVector.

In all requests, instead of a pointer directly to objects, the cache stores a structure of 2 elements - a pointer to the permanent part of the object and a index in an array of pointers to versions of objects. These arrays may differ in different query clones. To reduce duplication, a RefCounted array is used. rpb_relation's are registered in rpb when creating a clone. The decision on whether to update a set of versions in a clone of the request is made in GetRequest, that is, when a clone of the request is received. 

At the moment, TCS is fully working. QA makes mistakes - but I think it's time to pour metacache into the master so as not to prevent merge of other branches. And it's probably time to decide with the indexes of the system tables, which ones are needed and which ones are not very necessary. For example, I need to add indexes by ID for relations and other cached tables.


src/burp/restore.epp
```diff
@@ -280,6 +280,7 @@ static inline void collect_missing_privs(BurpGlobals* tdgbl, USHORT type, const
 
 } // namespace
 
+
```
Noremos: I think it's best to avoid unnecessary changes like these to simplify PR. I'll mark them in case you decide to undo them

src/common/classes/BlrReader.h
```diff
@@ -28,8 +28,8 @@
 #include "../common/StatusArg.h"
 #include "../jrd/constants.h"
 
-namespace Firebird {
 
+namespace Firebird {
```
Noremos, line 31: Unnecessary change

src/common/classes/BlrReader.h
```diff
@@ -233,7 +233,6 @@ class BlrReader
 	const UCHAR* pos;
 };
 
-
```
Noremos, line 236: Unnecessary change

src/common/dsc.h
```diff
@@ -244,21 +243,29 @@ typedef struct dsc
 		return 0;
 	}
 
+	SSHORT getBlobSubType() const noexcept
```
Noremos, line 246: Unnecessary change

src/common/dsc.h
```diff
@@ -499,6 +516,16 @@ typedef struct dsc
 		dsc_address = address;
 	}
 
+	bool operator==(const dsc& v) const noexcept
```
Noremos, line 519: The presence of DSC_EQUIV and this overload operator is a bit confusing. While this may not be relevant to this PR, it would be a nice to remove external functions related to dsc, such as DSC_EQUIV and DTYPE_IS_TEXT.

src/dsql/DdlNodes.epp
```diff
@@ -806,6 +832,7 @@ static void saveRelation(thread_db* tdbb, DsqlCompilerScratch* dsqlScratch,
 	dsqlScratch->relation = relation;
 }
 
+
```
Noremos: Unnecessary change

src/dsql/DdlNodes.epp
```diff
@@ -2673,6 +2712,10 @@ void DropFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsqlScratch
 	// run all statements under savepoint control
 	AutoSavePoint savePoint(tdbb, transaction);
 	bool found = false;
+	MetaId id;
+
+	//MetadataCache::oldVersion<Cached::Function>(tdbb, id);	-- missing ID in the node
```
Noremos: Is it supposed to be like this? Did you accidentally forget to return the call?

src/dsql/DdlNodes.epp
```diff
@@ -6223,8 +6306,154 @@ RelationNode::RelationNode(MemoryPool& p, RelationSourceNode* aDsqlNode)
...
+	static const CachedRequestId idDb;
+	AutoCacheRequest requestDb(tdbb, idDb);
+
+	FOR(REQUEST_HANDLE requestDb)
+		D IN RDB$DATABASE
+	{
+		MetaId startId = D.RDB$RELATION_ID;
+		MetaId stopId = MAX_RELATION_ID + 1;
+
+// !!!!!! printf("suggest %d %s...", startId, name.c_str()); fflush(stdout);
```
Noremos, line 6414: This is probably not a line that should be included in the final code

src/dsql/DdlNodes.epp
```diff
@@ -6223,8 +6306,154 @@ RelationNode::RelationNode(MemoryPool& p, RelationSourceNode* aDsqlNode)
...
+
+		for(;;)
+		{
+			if (startId < MIN_RELATION_ID || startId > MAX_RELATION_ID)
+				startId = MIN_RELATION_ID;
+
+			relId = startId;
+			enum {MISS, VALID, FREE} found = FREE;
```
Noremos, line 6422: It would be safer to use the enum class

src/dsql/DdlNodes.epp
```diff
@@ -6223,8 +6306,154 @@ RelationNode::RelationNode(MemoryPool& p, RelationSourceNode* aDsqlNode)
...
+// !!!!!! printf(" %d\n", relId); fflush(stdout);
+				return relId;
+			}
+
+			if (startId == MIN_RELATION_ID)
+				break;
+
+			stopId = startId;
+			startId = MIN_RELATION_ID;
+		}
```
Noremos, line 6461: Maybe add `fb_assert(found == MISS); `?

src/dsql/DdlNodes.epp
```diff
@@ -7804,2289 +8036,3998 @@ void RelationNode::dropFromPublication(thread_db* tdbb,
...
+					raiseTooManyVersionsError(obj_relation, relName);
+			}
+			else
+			{
+				if (REL.RDB$FORMAT == MAX_VIEW_VERSIONS)
+					raiseTooManyVersionsError(obj_view, relName);
+			}
 
-	checkRelationTempScope(tdbb, transaction, name, relationType.value());
+			if (REL.RDB$FORMAT == MAX_TABLE_VERSIONS)
```
Noremos: Isn't this the same check as the upper one?

src/dsql/DdlNodes.epp
```diff
@@ -7804,2289 +8036,3998 @@ void RelationNode::dropFromPublication(thread_db* tdbb,
...
+									RFR.RDB$UPDATE_FLAG = 0;
+								}
+								if (!null_view && REL.RDB$DBKEY_LENGTH > 8)
+								{
+									AutoRequest temp;
+									RFR.RDB$UPDATE_FLAG = 0;
+									FOR(REQUEST_HANDLE temp) X IN RDB$TRIGGERS WITH
+										X.RDB$SCHEMA_NAME EQ relName.schema.c_str() AND
+										X.RDB$RELATION_NAME EQ relName.object.c_str() AND
+											X.RDB$TRIGGER_TYPE EQ 1
```
Noremos, line 8228: It would be nice to use a constant (`TRIG_TYPE_PRE_STORE`) insead of a number

src/dsql/DdlNodes.epp
```diff
@@ -7804,2289 +8036,3998 @@ void RelationNode::dropFromPublication(thread_db* tdbb,
...
 
-		for (NestConst<Clause>* i = clauses.begin(); i != clauses.end(); ++i)
-		{
-			switch ((*i)->type)
-			{
-				case Clause::TYPE_ADD_COLUMN:
-				{
-					const auto addColumnClause = static_cast<AddColumnClause*>(i->getObject());
-					bool createColumn = true;
+								impure_value tempValue;
```
Noremos: It might be safer to add initialization to the variable like this: `impure_value tempValue{};`

src/dsql/DdlNodes.epp
```diff
@@ -7804,2289 +8036,3998 @@ void RelationNode::dropFromPublication(thread_db* tdbb,
...
 
-					newDomainName.schema = name.schema;
+		USHORT i = 0, dflCount = 0;
+		for (Format::fmt_defaults_iterator impure = format->fmt_defaults.begin();
+			 impure != format->fmt_defaults.end(); ++impure, ++i)
+		{
+			if (!impure->vlu_desc.isUnknown())
+			{
+				dsc desc = impure->vlu_desc;
+				desc.dsc_address = NULL;
```
Noremos: I think it's better to use `nullptr`

src/common/dsc.h
```diff
@@ -269,39 +276,49 @@ typedef struct dsc
...
 	{
 		if (isText())
-			return dsc_sub_type;
+			return TTypeId(dsc_sub_type);
 
 		if (isBlob())
 		{
 			if (dsc_sub_type == isc_blob_text)
-				return dsc_scale | (dsc_flags & 0xFF00);
+				return TTypeId(CSetId(dsc_scale), CollId(dsc_flags > 8));
```
TreeHunter9: I guess bitshift was intended here?
Same at 305 line.

src/dsql/DdlNodes.epp
```diff
@@ -7804,2289 +8036,3998 @@ void RelationNode::dropFromPublication(thread_db* tdbb,
...
+
+	deletePrivilegesByRelName(tdbb, transaction, name, obj_relation);
+
+	request.reset(tdbb, drq_e_view_prv, DYN_REQUESTS);
+
+	FOR(REQUEST_HANDLE request TRANSACTION_HANDLE transaction)
+		PRIV IN RDB$USER_PRIVILEGES
+		WITH PRIV.RDB$USER_SCHEMA_NAME EQ name.schema.c_str() AND
+			 PRIV.RDB$USER EQ name.object.c_str() AND
+			 PRIV.RDB$USER_TYPE = obj_view AND
```
Noremos, line 10483: Is there always obj_view here, even if we delete the table?

src/common/CharSet.h
```diff
@@ -32,18 +32,19 @@
 
 #include "CsConvert.h"
 #include "IntlUtil.h"
+#include "../jrd/intl.h"
 
 namespace Firebird {
 
-	template <>
```
Noremos, line 38: Unnecessary change

src/dsql/DsqlBatch.cpp
```diff
@@ -771,7 +771,7 @@ Firebird::IBatchCompletionState* DsqlBatch::execute(thread_db* tdbb)
 	if (m_blobMap.count())
 	{
 		DEB_BATCH(fprintf(stderr, "BLOBs %d were not used in messages\n", m_blobMap.count()));
-		ERR_post_warning(Arg::Warning(isc_random) << "m_blobMap.count() BLOBs were not used in messages");		// !!!!!!! new warning
+		ERR_post_warning(Arg::Warning(isc_random) << "m_blobMap.count() BLOBs were not used in messages");		// @@@@@@ new warning
```
Noremos: Unnecessary change. Or am I misunderstanding this

src/jrd/sort.cpp
```diff
@@ -2188,7 +2188,6 @@ void SortOwner::releaseBuffer(UCHAR* memory)
 	buffers.push(memory);
 }
 
-
```
Noremos, line 2191: Unnecessary change

src/jrd/sort.cpp
```diff
@@ -2442,3 +2441,4 @@ sort_record* PartitionedSort::getMerge()
 
 	return eof ? NULL : record;
 }
+
```
Noremos, line 2444: Unnecessary change

src/dsql/ExprNodes.cpp
```diff
@@ -13338,6 +13314,9 @@ DmlNode* UdfCallNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScratch*
 	return node;
 }
 
+
```
Noremos, line 13313: Unnecessary change

src/dsql/ExprNodes.cpp
```diff
@@ -13487,47 +13472,46 @@ ValueExprNode* UdfCallNode::pass1(thread_db* tdbb, CompilerScratch* csb)
...
-		CMP_post_resource(&csb->csb_resources, function, Resource::rsc_function, function->getId());
 	}
 
 	return this;
 }
 
 ValueExprNode* UdfCallNode::pass2(thread_db* tdbb, CompilerScratch* csb)
 {
-	if (function->fun_deterministic)
+	Function* f = function(tdbb);
```
Noremos, line 13509: It's probably better to be consistent with the names and call it `func`, as in `UdfCallNode::execute` below

src/common/dsc.h
AlexPeshkoff: getBlobSubType & setBlobSubType got close to each other after it

src/dsql/DdlNodes.epp
AlexPeshkoff: Certainly :-)

It's here to not forget to ask Pavel to write such test.

src/dsql/DdlNodes.epp
AlexPeshkoff: Sooner forgot to write good comment

src/dsql/DdlNodes.epp
AlexPeshkoff: In this particular case when enum declaration is local inside function use of enum class, inviting name for it and typing it everywhere in the code hardly make it safer, only longer

src/dsql/DdlNodes.epp
AlexPeshkoff: That code was copied here from jrd/dfw.epp - do not want to touch it as long as it works. May be special cleanup pass needed.

src/dsql/DdlNodes.epp
AlexPeshkoff: For table there is deletePrivilegesByRelName() a few lines earlier.

src/common/CharSet.h
AlexPeshkoff: It seems I've followed our formatting rules. Pretty hard to work with code when same file formatted differently.

src/include/fb_exception.h
```diff
@@ -104,9 +104,9 @@ class status_exception : public Exception
 
 	const ISC_STATUS* value() const noexcept { return m_status_vector; }
 
-	[[noreturn]] static void raise(const ISC_STATUS* status_vector);
-	[[noreturn]] static void raise(const Arg::StatusVector& statusVector);
-	[[noreturn]] static void raise(const IStatus* status);
+	static void raise [[noreturn]] (const ISC_STATUS* status_vector);
```
Noremos: To be honest, I've never seen this style before and didn't know it was a valid format.

src/intl/charsets.h
```diff
@@ -22,82 +22,84 @@ CS_737, CS_775, CS_858, CS_862, CS_864, CS_866, CS_869
...
+#include "../jrd/intl.h"
+
 #define DEFAULT_ATTACHMENT_CHARSET	CS_NONE
 
-#define   CS_NONE                0	/* No Character Set */
-#define   CS_BINARY              1	/* BINARY BYTES     */
-#define   CS_ASCII               2	/* ASCII            */
-#define   CS_UNICODE_FSS         3	/* UNICODE in FSS format */
-#define   CS_UTF8                4	/* UTF-8 */
+#define   CS_NONE                CSetId(0)	/* No Character Set */
```
Noremos, line 29: Is there any reason to keep this consatnts as defines?

src/jrd/Attachment.cpp
```diff
@@ -1133,8 +841,16 @@ void Attachment::checkReplSetLock(thread_db* tdbb)
...
+// Move to database level ? !!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
 void Attachment::invalidateReplSet(thread_db* tdbb, bool broadcast)
 {
+
+	att_flags |= ATT_repl_reset;
+
+	att_database->dbb_mdc->invalidateReplSet(tdbb);
+
+/* !!!!!!!!!!!!!!!!!!!!!!!
```
Noremos, line 853: Some Work In Progress code?

src/jrd/Collation.cpp
```diff
@@ -1125,55 +1126,16 @@ Collation* Collation::createInstance(MemoryPool& pool, TTYPE_ID id, texttype* tt
...
-	if (existenceLock)
-		LCK_release(tdbb, existenceLock);
-
-	useCount = 0;
-}
-
 void Collation::destroy(thread_db* tdbb)
 {
-	fb_assert(useCount == 0);
+	fprintf(stderr, "Collation::destroy(%p) tt=%p\n", this, tt);
```
Noremos: Some debug print?

src/dsql/DsqlBatch.cpp
```diff
@@ -771,7 +771,7 @@ Firebird::IBatchCompletionState* DsqlBatch::execute(thread_db* tdbb)
 	if (m_blobMap.count())
 	{
 		DEB_BATCH(fprintf(stderr, "BLOBs %d were not used in messages\n", m_blobMap.count()));
-		ERR_post_warning(Arg::Warning(isc_random) << "m_blobMap.count() BLOBs were not used in messages");		// !!!!!!! new warning
+		ERR_post_warning(Arg::Warning(isc_random) << "m_blobMap.count() BLOBs were not used in messages");		// !!!!!! new warning
```
Noremos, line 773: Does this line changed on porpuse? Is there something wrong with it?

src/jrd/Database.cpp
```diff
@@ -475,6 +490,75 @@ namespace Jrd
 			dbb_filename, dbb_config));
 	}
 
+	// Find an inactive incarnation of a system request.
+	Request* Database::findSystemRequest(thread_db* tdbb, USHORT id, InternalRequest which)
+	{
+		fb_assert(which == IRQ_REQUESTS || which == DYN_REQUESTS || which == CACHED_REQUESTS);
+
+		//Database::CheckoutLockGuard guard(this, dbb_cmp_clone_mutex);
```
Noremos: Temporaly change?

src/jrd/IntlManager.cpp
```diff
@@ -654,69 +654,83 @@ bool IntlManager::lookupCharSet(const QualifiedMetaString& charSetName, charset*
...
+void IntlManager::lookupCollation(const MetaString& collationName,
+								  const CharsetVariants& charsetVariants,
 								  USHORT attributes, const UCHAR* specificAttributes,
 								  ULONG specificAttributesLen, bool ignoreAttributes,
 								  texttype* tt)
 {
 	ExternalInfo charSetExternalInfo;
 	ExternalInfo collationExternalInfo;
-	char statusBuffer[BUFFER_LARGE] = "";
+	char statusBufferBig[BUFFER_LARGE] = "";
```
Noremos, line 665: Maybe using std::array will be a litle bit handy

src/jrd/RecordSourceNodes.cpp
```diff
@@ -3786,15 +3811,15 @@ void RseNode::planSet(CompilerScratch* csb, PlanNode* plan)
 		{
 			map = mapBase;
 			tail = &csb->csb_rpt[*map];
-			viewRelation = tail->csb_relation;
-			viewProcedure = tail->csb_procedure;
+			viewRelation = tail->csb_relation();
+			viewProcedure = tail->csb_procedure();
 
-			// If the plan references the view itself, make sure that
+			// If the plan references  the view itself, make sure that
```
Noremos: Typo

src/include/fb_exception.h
AlexPeshkoff: Both correct, see https://en.cppreference.com/w/cpp/language/attributes/noreturn.html

But taking into an account that other style is used else where I change such places

src/intl/charsets.h
AlexPeshkoff: Doubt that - but it worked and works

src/jrd/Attachment.cpp
AlexPeshkoff: Yes, looks like replication code required some more care to be taken.

src/jrd/Collation.cpp
AlexPeshkoff: Yes, should be removed long ago. Fixed.

src/dsql/DsqlBatch.cpp
AlexPeshkoff: The only wrong here is that isc_random to be replaced here with some special warning code.

src/jrd/Database.cpp
AlexPeshkoff: Really not needed - cleaned up

src/jrd/IntlManager.cpp
AlexPeshkoff: Not sure - called functions are plain text oriented.

To be precise - everything (well almost everything) related with intl support worth redesign and new code. But as long as existing works I prefer to touch it as minimal as possible.

src/dsql/DdlNodes.epp
Noremos: Got it

src/jrd/MetaName.cpp
```diff
@@ -62,8 +62,8 @@ void MetaName::adjustLength(const char* const s, FB_SIZE_T& len) noexcept
 	if (len > MAX_SQL_IDENTIFIER_LEN)
 	{
 #ifdef DEV_BUILD
-		for (FB_SIZE_T i = MAX_SQL_IDENTIFIER_LEN; i < len; ++i)
-			fb_assert(s[i] == '\0' || s[i] == ' ');
+//		for (FB_SIZE_T i = MAX_SQL_IDENTIFIER_LEN; i < len; ++i)
+//			fb_assert(s[i] == '\0' || s[i] == ' ');
```
Noremos, line 67: Is this a temporaly change?

src/jrd/Relation.cpp
```diff
@@ -28,31 +28,248 @@
...
+
+	if (rel->rel_file)
+	{
+		rel->rel_file->release();
+		delete rel->rel_file;
+	}
+
+	rel->rel_indices.cleanup(tdbb);
+/*
+	// delete by pool is broken, needs enhancements in CacheVector
```
Noremos, line 161: Is it still broken?

src/jrd/Relation.cpp
```diff
@@ -28,31 +28,248 @@
...
+}
+
+bool RelationPermanent::isReplicating(thread_db* tdbb)
 {
 	Database* const dbb = tdbb->getDatabase();
 	if (!dbb->isReplicating(tdbb))
 		return false;
 
-	Attachment* const attachment = tdbb->getAttachment();
+	Attachment* const attachment = tdbb->getAttachment();		// Database? !!!!!!!!!!!!!!!!!!!!!!
```
Noremos, line 263: Is something wrong with the line?

src/include/fb_exception.h
Noremos: OK

src/jrd/IntlManager.cpp
Noremos: Got it

src/jrd/MetaName.cpp
AlexPeshkoff: Yes, and sooner of all some code rework or additional fix is needed here. Assert depends on user data - i.e. looks like some levels higher some length control is missing. But I see no way how it can be related with metadata cache.

I've additionally marked this place in order to return to it later.

src/jrd/Relation.cpp
AlexPeshkoff: Sooner of all yes.

src/jrd/Relation.cpp
AlexPeshkoff: There are 3 known big issues to be fixed:
- RECREATE for various objects
- GTTs  ON COMMIT PRESERVE ROWS
- replication (this case)

src/jrd/Statement.cpp
```diff
@@ -246,6 +202,69 @@ Statement::Statement(thread_db* tdbb, MemoryPool* p, CompilerScratch* csb)
...
+		{
+			fb_assert(req);
+			guard.enter();
+		}
+
+		if (ddl || (!latestVer) || (latestVer->version != frontVersion))
+		{
+			const FB_SIZE_T resourceCount = latestVer ? latestVer->getCapacity() :
+				resources->charSets.getCount() + resources->relations.getCount() + resources->procedures.getCount() +
+				resources->functions.getCount() + resources->triggers.getCount();
```
Noremos, line 228: Should resources->indices.getCount() be here?

AlexPeshkoff: No. Indices are not versioned like the rest of objects. Index version might be not needed at all - but initially I have not understood that current method of assigning IDs to indices is bad-bad-bad. If sometimes it will be replaced with IDs permanent during INACTIVE state versioning of indices will be not needed.

Noremos: Thanks for the explanation. I'm currently adapting package constants to the new meta-cache. I spent some time trying to figure out why the constants cache is missing, and it turns the problem was with these lines (missing `+ resources->constants.getCount()`). May I suggest moving the logic to the Rosource class, something like this:

```
inline FB_SIZE_T countVersionedObjects() const noexcept
{
	return charSets.getCount() + relations.getCount() + procedures.getCount() +
			functions.getCount() + triggers.getCount();
}
```

This will make it easier to add a cache for new objects in the future.

AlexPeshkoff: Certainly

.github/workflows/main.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push]
```
asfernandes, line 3: @AlexPeshkoff should not this be reverted?

AlexPeshkoff: Certainly - done.

PR #8818, opened 2025-11-30
This should fix bug #8817 : Fatal lock manager error: invalid lock id

hvlad:

PR #8826, opened 2025-12-15
Fixed potential endless loop inside MET_scan_relation

hvlad:


hvlad:
See also comments at #8818

dyemanov:
It won't work if a new virtual table is added without raising minor ODS version. This is not the practice we followed so far, but technically it works and it may be used one day. Should we consider this a no-go or relax the checks in this patch?

hvlad:
> It won't work if a new virtual table is added without raising minor ODS version. This is not the practice we followed so far, but technically it works and it may be used one day. Should we consider this a no-go or relax the checks in this patch?

It is hard to implement check for rule when the rule itself is not defined ~at all~ well.
What check do you consider as necessary and enough ?
Should it check for system flag only ? Or it should also check for presence in `relations.h` ?

dyemanov:
I'd check only the system flag (and maybe the virtual table flag). If some persistent system table exists in the metadata cache but missing in the database file, then something is seriously wrong.

hvlad:
> I'd check only the system flag (and maybe the virtual table flag). If some persistent system table exists in the metadata cache but missing in the database file, then something is seriously wrong.

Done.

dyemanov (approved):

PR #8829, opened 2025-12-18
#8828 Fix file name in UTF-8 encoding for gstat

### LINKED ISSUE #8828: Gstat does not open database if the system locale is not UTF-8
Reproduced on Windows with CP1251 locale.

paradox1307:


src/utilities/gstat/dba.epp
```diff
@@ -635,6 +635,9 @@ int gstat(Firebird::UtilSvc* uSvc)
 		}
 	}
 
+	if (uSvc->utf8FileNames())
+		ISC_utf8ToSystem(fileName);
+
```
asfernandes, line 640: Should not use ISC_unescape too?

paradox1307: I see that gbak and nbackup do not use `ISC_unescape`. Also, function body is disabled by the `NOT_USED_OR_REPLACED` macro.

PR #8845, opened 2026-01-05
Add option to EXEC_PATH_DISPLAY to display also execution nodes tree

aafemt:
An improvement for EXEC_PATH_DISPLAY. It is more convenient to see the execution tree in ISQL instead of debug log only.

XML formatting of the nodes wasn't changed from original implementation.

### DIFF
--- a/doc/README.isc_info_xxx
+++ b/doc/README.isc_info_xxx
@@ -96,3 +96,10 @@ New items for isc_transaction_info:
 
 6. isc_info_tra_lock_timeout
 	return lock timeout of current transaction
+
+New items for isc_dsql_sql_info:
+
+1. isc_info_sql_exec_path_nodes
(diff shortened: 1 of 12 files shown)


asfernandes:
The XML node-tree is a development feature with missing things.
I'm against put it in production in the current state.
Also it prints internal details (numeric constants) that we do not want to stabilize.

aafemt:
So does BLR printer. This feature is for debugging, not production.

asfernandes:
BLR is stablished thing, it's API.
Emitted BLR should always be parseable and there were already BLR-to-text filters of stored routines.

aafemt:
Note: this pull request doesn't change the node printer, just add a new way to receive the result. If the node printer is incomplete and messy - it is a separate issue and should be fixed by separate PR(s).

PR #8852, opened 2026-01-07
Fixed bug #8851 : 'Internal error' when calling outer procedure after deleting unused result variable from inner procedure

### LINKED ISSUE #8851: 'Internal error' when calling outer procedure after deleting unused result variable from inner procedure
Firebird 5.0.3 classic on 64 bit windows
Firebird 5.0.2 produces ‘conversion error from string "20-NOV-1858 0:00:00.0003"’ error 
Firebird 2.5.9 works fine
Error happens for this and other connections

Here is test case to reproduce:

```
set term ^;
create or alter procedure sp1
returns (field1 integer, field2 integer)
as begin
  field1 = 1;
  field2 = 2;  
  suspend;
end 
^
create or alter procedure sp2
returns (field1 integer) 
as begin
  field1 = 1;
  if (exists(select 1 from sp1 where field1 = 2)) then
    suspend;
end 
^
set term ;^
commit;

select * from sp2;
commit;

set term ^;
create or alter procedure sp1
returns (field1 integer) 
as begin
  field1 = 1;
  suspend;
end 
^
set term ;^

/*following errors out in this and other connections*/
select * from sp2; 
/*5.0.2.1613:
Statement failed, SQLSTATE = 22018
conversion error from string "20-NOV-1858 0:00:00.0003"
-At procedure 'SP2' line: 5, col: 3*/
/*5.0.3.1683:
Statement failed, SQLSTATE = XX000
internal error
-At procedure 'SP2' line: 5, col: 3*/

drop procedure sp2;
drop procedure sp1;
commit;

```

hvlad:


src/jrd/recsrc/ProcedureScan.cpp
```diff
@@ -72,6 +72,12 @@ void ProcedureScan::internalOpen(thread_db* tdbb) const
 
 	const_cast<jrd_prc*>(m_procedure)->checkReload(tdbb);
 
+	// Procedure could be altered and its record format changed since current instance of
+	// ProcedureScan was created. Here it is not the right place to check if new format is
+	// compatible with caller's expectations, so we just use correct (possible new) format.
+
+	m_format = m_procedure->prc_record_format;
```
dyemanov, line 79: Hmmm. `ALTER` should create a new procedure version inside the metadata cache, but the already compiled procedure `SP2` keeps a reference to the older version of `SP1` (with the old format). AFAIU, this means that `m_procedure->prc_record_format` should never change, but your patch implies that the old version of the procedure somehow gets the new record format. How does it happen?

hvlad: `RoutineManager::modifyRoutine()` marks procedure with `FLAG_OBSOLETE | FLAG_BEING_ALTERED` and calls `Routine::remove()` which doesn't delete such routines, but clears its internals and marks it with `FLAG_CLEARED`.
Then `RoutineManager::getDependencies()` calls `MET_procedure()` which doesn't create new version of procedure marked with `FLAG_OBSOLETE ` but re-uses existing instance instead.

PR #8858, opened 2026-01-14
Ensure sequences are replicated even if changed in de-facto read-only transactions

XaBbl4:
Add a new method to the replication API to ensure generators are always replicated (both on commit and rollback).

Previously, if execute:
```sql
select gen_id(t_gen, 1) from rdb$database;
rollback;
```
The generator on the master and replica stored different values.

If decide to backport this to 5.0, might consider:
```cpp
version: // 5.0 => 5.0.4
    void flushSequences(Status status);

version: // 5.0.4 => 6.0 Alpha1
    void setSequence2(...);
```


dyemanov:
In practice this PR avoids inconsistent replica when applications use an attachment-level sequence cache (IDs are prefetched and distributed among rows being inserted). This does not allow sharing the cached IDs among different attachments, but so far we haven't seen such an use case.

Custom replication plugins are free to choose between synchronous or delayed (conditional or unconditional) sequence replication.

src/jrd/replication/Publisher.cpp
```diff
@@ -452,7 +460,15 @@ void REPL_trans_rollback(thread_db* tdbb, jrd_tra* transaction)
 {
 	const auto replicator = transaction->tra_replicator;
 	if (!replicator)
+	{
+		const auto replicator = getReplicator(tdbb);
+		if (replicator)
+		{
+			FbLocalStatus status;
+			replicator->flushSequences(&status);
```
aafemt: If there is no `replicator` on commit/rollback, it means that `setSequence` was never called and there is nothing to flush. Replication plugin already has an opportunity to push changed sequences to replica on rollback, right?

XaBbl4: > If there is no `replicator` on commit/rollback, it means that `setSequence` was never called and there is nothing to flush.

Yes, it can be change to `attachment->att_replicator`

> Replication plugin already has an opportunity to push changed sequences to replica on rollback, right?

Previously, when a transaction was rollback, sequences were not transferred to the replica, only during the first commit in the current attachment.

aafemt: > Previously, when a transaction was rollback, sequences were not transferred to the replica, only during the first commit in the current attachment.

But it was the choice of the standard replication plugin (only). Why API changes are required if you can limit changes to plugin code?

dyemanov: > > Previously, when a transaction was rollback, sequences were not transferred to the replica, only during the first commit in the current attachment.
> 
> But it was the choice of the standard replication plugin (only). Why API changes are required if you can limit changes to plugin code?

Current API does not allow to replicate generators changed in read-only (de-facto) transactions, because the transaction replicator object is not created without I/U/D ops and thus there's no commit/rollback event at the end. Thus a need for the new API hook.

XaBbl4: > But it was the choice of the standard replication plugin (only). Why API changes are required if you can limit changes to plugin code?

This is an optimization of the current plugin implementation, where generator values ​​are first cached and only transmitted to the replica when commit, for example:

```
select gen_id(t_gen, 1) from rdb$database;
select gen_id(t_gen, 1) from rdb$database;
select gen_id(t_gen, 1) from rdb$database;
commit;
```

As a result, the generator will be transmitted to the replica with a value of +3, instead of three times for each operation.

Furthermore, there are situations where the generator value changes without an actual transaction, as in the example with `select gen_id`. In this case, `ReplicatedTransaction` doesn't yet exist and commit/rollback isn't called. This is why a new method call was added to Publisher.cpp.

aafemt: No `select` can be performed outside of transaction thus no `REPL_gen_id` can be called without transaction. IMHO, it is enough to change call to `getReplicator` inside of `REPL_gen_id` to let standard plugin work. Cleanup of attachment-level replicator is optional.

aafemt: What do I care about is creation of plugin instance for single call that may have nothing to work with. Direct usage of `att_replicator` can solve it though, as you suggested.

dyemanov: > No `select` can be performed outside of transaction thus no `REPL_gen_id` can be called without transaction. IMHO, it is enough to change call to `getReplicator` inside of `REPL_gen_id` to let standard plugin work. Cleanup of attachment-level replicator is optional.

`setSequence()` is a method of `IReplicatedSession`, not `IReplicatedTransaction`. Or do you mean just creating a dummy instance of `IReplicatedTransaction` together with calling `IReplicatedSession::setSequence()`?

aafemt: > `setSequence()` is a method of `IReplicatedSession`, not `IReplicatedTransaction`.

Oops, I forgot that.

> Or do you mean just creating a dummy instance of `IReplicatedTransaction` together with calling `IReplicatedSession::setSequence()`?

Yes. BTW, moving of `setSequence` into `ReplTransaction` is not an option, right?..

dyemanov: > > Or do you mean just creating a dummy instance of `IReplicatedTransaction` together with calling `IReplicatedSession::setSequence()`?
> 
> Yes.

Doable. But as for me, it makes things even slower (`startTransaction()` + `commit/rollback()` instead of just `flushSequences()`). And it expects that the plugin replicates generators either synchronously or inside `rollback()` too.

> BTW, moving of `setSequence` into `ReplTransaction` is not an option, right?..

Nope.

aafemt: > But as for me, it makes things even slower (`startTransaction()` + `commit/rollback()` instead of just `flushSequences()`).

`flushSequence` implementation is the same creation of the buffer, filling it and sending, AFAIK. No performance difference is expected. Replication plugin is free to consider if the buffer contain no DML and avoid sending of such buffer completely. At least my plugin does that.

aafemt: BTW, perhaps I read you wrong. By "dummy instance of `IReplTransation`" did you mean an instance dedicated to replication of generator change that isn't bound to transaction where `gen_id` has been called? I would expected contrary: an instance of `IReplTransaction` bound to `Transaction` as for any other DML. Such instance will have no overhead if `gen_id` is used in DML or DML is executed later in the same transaction.

dyemanov: > BTW, perhaps I read you wrong. By "dummy instance of `IReplTransation`" did you mean an instance dedicated to replication of generator change that isn't bound to transaction where `gen_id` has been called? I would expected contrary: an instance of `IReplTransaction` bound to `Transaction` as for any other DML. Such instance will have no overhead if `gen_id` is used in DML or DML is executed later in the same transaction.

I meant the same. However, unlike I/U/D operations, `DPM_gen_id()` may be theoretically called with a NULL transaction, so `tra_replicator` should be created only if the user transaction exists.

src/jrd/replication/Replicator.h
```diff
@@ -135,6 +143,21 @@ namespace Replication
 			{
 				buffer->add(data, length);
 			}
+
+			void putGenerators(const GeneratorCache* generators)
```
dyemanov: Pass generators by reference, please -- AFAIU we always pass `m_generators` there and it's mandatory.

XaBbl4: Changed, thanks

src/jrd/replication/Publisher.cpp
XaBbl4: Remove the new API and create tra_replicator in REPL_gen_id if there were no previous operations in the transaction.

PR #8866, opened 2026-01-22
Fix incorrect platform detection in setenvvar.bat on VS2022

### LINKED ISSUE #8865: Platform detection in setenvvars.bat always fails and forces x64
### Summary

The Windows build of Firebird 3.0.13, 4.0.6, and 5.0.3 cannot reliably detect
the correct target platform on modern Visual Studio environments due to
incorrect logic in `setenvvars.bat`. This prevents proper Win32/Win64
separation and causes the build system to behave incorrectly.

This issue was observed while using **Microsoft Visual Studio 2022 Build Tools**.

### Problem Details

`setenvvars.bat` determines the target platform using the
`PROCESSOR_ARCHITECTURE` environment variable.  
However, with VS2022 Build Tools, `PROCESSOR_ARCHITECTURE` always reports
`AMD64` even when using the **x86 Native Tools Command Prompt**.

As a result:

- `FB_TARGET_PLATFORM` is always set to `x64`
- `FB_OUTPUT_DIR` always points to `output_x64`
- `clean_all` removes the wrong output directory
- Win32 builds do not behave correctly because the environment is misconfigured

This behavior was verified on:

- Firebird **3.0.13**
- Firebird **4.0.6**
- Firebird **5.0.3**

### Expected Behavior

When using the x86 Native Tools Command Prompt,  
`FB_TARGET_PLATFORM` should be set to `Win32`, and the build system should
produce valid Win32 output directories and behave consistently.

### Proposed Fix

Use the `%Platform%` environment variable provided by Visual Studio instead of
`PROCESSOR_ARCHITECTURE`.  
`%Platform%` correctly reports `x86` or `x64` depending on the toolchain
invoked, and works consistently with **VS2022 Build Tools**.

I will submit a pull request with this fix.

tomneko:
This pull request fixes incorrect platform detection in `setenvvar.bat` when
building Firebird on modern Visual Studio environments (VS2022 Build Tools).

The current script relies on `PROCESSOR_ARCHITECTURE`, which always reports
`AMD64` even when using the x86 Native Tools Command Prompt. As a result,
`FB_TARGET_PLATFORM` is always set to `x64`, and Win32 builds do not behave
correctly.

This patch replaces the detection logic with the `%Platform%` variable, which
correctly reports `x86` or `x64` depending on the toolchain.

Fixes #8865

### DIFF
--- a/builds/win32/setenvvar.bat
+++ b/builds/win32/setenvvar.bat
@@ -16,12 +16,13 @@
 
 @echo off
 
-::set FB_PROCESSOR_ARCHITECTURE=AMD64
+:: 1. Determine target platform
+if "%Platform%"=="x86"  set FB_TARGET_PLATFORM=Win32
+if "%Platform%"=="x64"  set FB_TARGET_PLATFORM=x64
 
-:: Default target CPU architecture is the native environment
-if NOT DEFINED FB_PROCESSOR_ARCHITECTURE (
-set FB_PROCESSOR_ARCHITECTURE=%PROCESSOR_ARCHITECTURE%
-)
+:: 2.If Platform variable is not set, use PROCESSOR_ARCHITECTURE
+if "%FB_TARGET_PLATFORM%"=="Win32" set FB_PROCESSOR_ARCHITECTURE=x86
+if "%FB_TARGET_PLATFORM%"=="x64"   set FB_PROCESSOR_ARCHITECTURE=AMD64
 
 ::===============================
 ::Set up the compiler environment


builds/win32/setenvvar.bat
```diff
@@ -16,12 +16,13 @@
 
 @echo off
 
-::set FB_PROCESSOR_ARCHITECTURE=AMD64
+:: 1. Determine target platform
+if "%Platform%"=="x86"  set FB_TARGET_PLATFORM=Win32
```
hvlad, line 20: I see this variable is defined by VS 2017 and VS 2022 and not defined by VS 2013.
Official compiler for FB3 is VS 2010.
Afraid this patch can be applied to the FB4+ only.

hvlad (changes requested):

dyemanov:
@hvlad Should it be front-ported manually or just closed?

hvlad:
> @hvlad Should it be front-ported manually or just closed?

Front-ported, I think

PR #8879, opened 2026-02-03
Refactor `class Thread` and relatives

hvlad:
Refactor `class Thread` and relatives:
- ensure correct and one-time release of thread-related resources, such as thread handle on Windows,
- hide OS-related internals. 

Thanks to @AlexPeshkoff.


asfernandes:
Why modern std C++ thread library still cannot be used in Firebird code?

hvlad:
> Why modern std C++ thread library still cannot be used in Firebird code?

It still missed some functionality.

PR #8882, opened 2026-02-05
build error :  function  inet_error has 5 parameters but call with 4

Khalyutin:
inet_error(bool releasePort, rem_port* port, const TEXT* function, ISC_STATUS operation, int status)

### DIFF
--- a/src/remote/inet.cpp
+++ b/src/remote/inet.cpp
@@ -3121,7 +3121,9 @@ static bool packet_receive(rem_port* port, UCHAR* buffer, SSHORT buffer_length, 
 			INET_force_error = 1;
 			try
 			{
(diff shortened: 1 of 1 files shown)


src/remote/inet.cpp
```diff
@@ -3121,7 +3121,9 @@ static bool packet_receive(rem_port* port, UCHAR* buffer, SSHORT buffer_length,
 			INET_force_error = 1;
 			try
 			{
-				inet_error(false, port, "simulated error - read", isc_net_read_err);
+				// function  inet_error as 5 parameters
+				// inet_error(bool releasePort, rem_port* port, const TEXT* function, ISC_STATUS operation, int status)
```
asfernandes:
```suggestion
```

asfernandes (changes requested):

PR #8887, opened 2026-02-08
Build process chash when btyacc parse file was created in windows with end of line mark as \r\n

Khalyutin:
if parsing file was created in windows but processing on linux we need  delete windows end of line mark as \r\n (CR + LF) Otherwise, the file will not be found.

	more info: https://github.com/ChrisDodd/btyacc/pull/32

### DIFF
--- a/extern/btyacc/reader.c
+++ b/extern/btyacc/reader.c
@@ -146,13 +146,11 @@ char *get_line() {
 
   /* VM: Process %include line */
   if(strncmp(&line[0], "%include ", 9)==0) {
-
-    char *inc_file_name = line+9;
-    while (isspace(*inc_file_name)) inc_file_name++;
-    i = strlen(inc_file_name);
-    while (i > 0 && isspace(inc_file_name[i-1])) --i;
-    inc_file_name[i] = 0;
-
+    int ii=0;
+    for(i=9; line[i]!='\n' && line[i]!=' ' && !(line[i] =='\r' && line[i+1]=='\n'); i++, ii++) {
+      inc_file_name[ii] = line[i];
+    }
+    inc_file_name[ii] = 0;
     if(inc_file) {
       error(lineno, 0, 0, "Nested include lines are not allowed");
     }
--- a/extern/btyacc/reader.c
+++ b/extern/btyacc/reader.c
@@ -147,10 +147,13 @@ char *get_line() {
   /* VM: Process %include line */
   if(strncmp(&line[0], "%include ", 9)==0) {
     int ii=0;
-    for(i=9; line[i]!='\n' && line[i]!=' '; i++, ii++) {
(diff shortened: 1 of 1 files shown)


aafemt:
Such problems are usually solved by `.gitattributes` file.

Khalyutin:
> Such problems are usually solved by `.gitattributes` file.

Yes, of course. But we tried to eliminate the root cause of the problem so that others wouldn't have to waste resources trying to localize it in the future.  :)

aafemt:
Will you also fix `cmd.exe` that cannot process files with LF-only EoLNs?

Khalyutin:
> Will you also fix `cmd.exe` that cannot process files with LF-only EoLNs?

source code of ms command interpreter not include in FirebirdSQL/firebird/extern/

hvlad:
Why not pickup [commit from original source tree](https://github.com/ChrisDodd/btyacc/commit/fdad391) ?

Khalyutin:
> Why not pickup [commit from original source tree](https://github.com/ChrisDodd/btyacc/commit/fdad391) ?

the source tree is very different from the version of btyacc in extern/btyacc/  start from   "read_from_file()" in original  insted of "fopen()" in extern/btyacc/

Adding read_from_file() may require further changes, up to replacing the entire library :(.

As a result, fixing a known compilation error can addition  unknown parsing errors...

I think the risk of taking a significant block of code from the btyacc original source is unjustified.

extern/btyacc/reader.c
```diff
@@ -146,11 +146,13 @@ char *get_line() {
 
   /* VM: Process %include line */
   if(strncmp(&line[0], "%include ", 9)==0) {
-    int ii=0;
-    for(i=9; line[i]!='\n' && line[i]!=' '; i++, ii++) {
-      inc_file_name[ii] = line[i];
-    }
-    inc_file_name[ii] = 0;
+
+    char *inc_file_name = line+9;
```
hvlad: `inc_file_name` is already declared as global variable at line 26 above.
It is used later in other functions.
This change break it, AFAIU

Khalyutin: Yes, you're right, the consequences of breaking a global variable "inc_file_name", are unpredictable.

The PR to the btyacc original source tree was intended to clarify the initial understanding regarding the handling of "\r\n" characters.

However, an attempt to integrate the original code from the btyacc source was failed...

My original idea was to handle only "\r\n" EndOfLine characters without touching anything else: 
```
-- for(i=9; line[i]!='\n' && line[i]!=' '; i++, ii++) {
++ for(i=9; line[i]!='\n' && line[i]!=' ' && !(line[i] =='\r' && line[i+1]=='\n'); i++, ii++) {

```

Is this acceptable?

PR #8900, opened 2026-02-11
Improvement #8895 : 'invalid request BLR' puts actual error (and procedure where it happened) at the end where it is truncated by ISC API

### LINKED ISSUE #8895: 'invalid request BLR' puts actual error (and procedure where it happened) at the end where it is truncated by ISC API
When preparing to switch application written in Delphi using IBObjects from 2.5 to 5.0 we noticed new 'Missing arg #1 - possibly status vector overflow' errors in multiple places, for example:

SQL: select * from ex_ssccPkg_getList (561161, -1, -1, null, f_now(), 5004, '12/31/2020', 1, 0, 1)
ISC ERROR CODE:335544876
ISC ERROR MESSAGE:
Error while parsing procedure EX_SSCCPKG_GETLIST's BLR
Error while parsing procedure INT_SSCCPKG_GETLIST4's BLR
Error while parsing procedure INT_FCCPLANIDS_VALIDATE's BLR
Error while parsing procedure INT_BROADBANDLABELS's BLR
Error while parsing procedure <Missing arg #1 - possibly status vector overflow>'s BLR

note that procedure list unusually starts with parent, thus I quite likely the error is at the end of the list and was truncated by ISC API

without seeing actual error and where it happened we did not have clues how to reproduce and show the problem

later we bumped into another scenario which is reproducible -- create a few stored procedures, from youngest child call existing UDF function, then rename UDF and call parent procedure

in 2.5 there was error first and then silently truncated procedure list:

invalid request BLR at offset 979
-function F_NOW is not defined
-module name or entrypoint could not be found
-Error while parsing procedure INT_GET_USER's BLR
-Error while parsing procedure INT_USERS_GETCURRENTPK's BLR

in 5.0 procedure list is reversed and error is at the end (visible in isql via OO API):

Error while parsing procedure INT_CONNECT_CONTEXTS's BLR
-Error while parsing procedure INT_USERS_GETCURRENTPK's BLR
-Error while parsing procedure INT_GET_USER's BLR
-invalid request BLR at offset 976
-function F_NOW is not defined
-module name or entrypoint could not be found

but when using IBO which uses ISC API only first 4 lines are shown, followed by 'Missing arg #1 - possibly status vector overflow':

Error while parsing procedure INT_CONNECT_CONTEXTS's BLR                                            
Error while parsing procedure INT_USERS_GETCURRENTPK's BLR                                          
Error while parsing procedure INT_GET_USER's BLR                                                    
invalid request BLR at offset 976                                                                   
function <Missing arg #1 - possibly status vector overflow> is not defined     

actual error that UDF function entry point was not found is missing. If there were couple more procedures involved, then we would not see in which procedure error happened

I hope that original errors we bumped into are also 'invalid request BLR' and putting error at the top and reverting procedure list like it used to be in 2.5 would allow us to see what's happening there

Thanks,
Paulius

hvlad:
Revert order of entires in status-vector so that original error is put on the top, not at the bottom

### DIFF
--- a/src/jrd/Function.epp
+++ b/src/jrd/Function.epp
@@ -470,8 +470,8 @@ Function* Function::loadMetadata(thread_db* tdbb, USHORT id, bool noscan, USHORT
 						{
 							StaticStatusVector temp_status;
 							ex.stuffException(temp_status);
-							(Arg::Gds(isc_bad_fun_BLR) << Arg::Str(name)
-								<< Arg::StatusVector(temp_status.begin())).raise();
+							(Arg::StatusVector(temp_status.begin()) <<
+								Arg::Gds(isc_bad_fun_BLR) << Arg::Str(name)).raise();
 						}
 					}
 				}
@@ -609,8 +609,8 @@ bool Function::reload(thread_db* tdbb)
 				ex.stuffException(temp_status);
 
 				const string name = this->getName().toString();
-				(Arg::Gds(isc_bad_fun_BLR) << Arg::Str(name)
-					<< Arg::StatusVector(temp_status.begin())).raise();
+				(Arg::StatusVector(temp_status.begin()) <<
+					Arg::Gds(isc_bad_fun_BLR) << Arg::Str(name)).raise();
 			}
 		}
 		catch (const Exception&)
--- a/src/jrd/met.epp
+++ b/src/jrd/met.epp
@@ -3732,8 +3732,8 @@ jrd_prc* MET_procedure(thread_db* tdbb, USHORT id, bool noscan, USHORT flags)
 						{
 							StaticStatusVector temp_status;
 							ex.stuffException(temp_status);
-							(Arg::Gds(isc_bad_proc_BLR) << Arg::Str(name)
-								<< Arg::StatusVector(temp_status.begin())).raise();
+							(Arg::StatusVector(temp_status.begin()) <<
+								Arg::Gds(isc_bad_proc_BLR) << Arg::Str(name)).raise();
 						}
 					}
 				}
@@ -3842,8 +3842,8 @@ bool jrd_prc::reload(thread_db* tdbb)
 				ex.stuffException(temp_status);
 
 				const string name = this->getName().toString();
-				(Arg::Gds(isc_bad_proc_BLR) << Arg::Str(name)
-					<< Arg::StatusVector(temp_status.begin())).raise();
+				(Arg::StatusVector(temp_status.begin()) <<
+					Arg::Gds(isc_bad_proc_BLR) << Arg::Str(name)).raise();
 			}
 		}
 		catch (const Exception&)


dyemanov:
Could you please show the error stack before and after the fix? I'm afraid this gonna break all programmer checks like `status[1] == isc_bad_proc_BLR` to check whether BLR is known to be invalid.

hvlad:
> Could you please show the error stack before and after the fix? 

Before:
```
SQL> select fn_test3() from rdb$database;
Statement failed, SQLSTATE = 2F000
Error while parsing function FN_TEST3's BLR
-Error while parsing function FN_TEST2's BLR
-Error while parsing function FN_TEST1's BLR
-invalid request BLR at offset 32
-function UDF_SRAND is not defined
-module name or entrypoint could not be found
```

After:
```
SQL> select fn_test3() from rdb$database;
Statement failed, SQLSTATE = 39000
invalid request BLR at offset 32
-function UDF_SRAND is not defined
-module name or entrypoint could not be found
-Error while parsing function FN_TEST1's BLR
-Error while parsing function FN_TEST2's BLR
-Error while parsing function FN_TEST3's BLR
```

Also, note, now this order is the same as in execution error:
```
SQL> select fn_test3() from rdb$database;

    FN_TEST3
============
Statement failed, SQLSTATE = 39000
function UDF_SRAND is not defined
-module name or entrypoint could not be found
-At function 'FN_TEST1' line: 5, col: 3
At function 'FN_TEST2' line: 5, col: 3
At function 'FN_TEST3' line: 5, col: 3
```

> I'm afraid this gonna break all programmer checks like `status[1] == isc_bad_proc_BLR` to check whether BLR is known to be invalid.

I would be very surprized if someone look for error code in `status[1]` only.

dyemanov:
> I would be very surprized if someone look for error code in `status[1]` only.

It's a common practice in the Firebird codebase :) In RDB, we check for `isc_bad_proc_BLR` and its friends in `gbak` in order to restore databases with invalid BLR. Of course, this check can be improved to search through the whole status vector, but it would look inconsistent with another error handlers across the codebase. With the new code, perhaps we could rely on `isc_invalid_blr` instead, but old checks are still needed for backward compatibility. Sigh.

hvlad:
> > I would be very surprized if someone look for error code in `status[1]` only.
> 
> It's a common practice in the Firebird codebase :) 

Not always. Usually it depends on exact knowledge of what error code is primary.

> In RDB, we check for `isc_bad_proc_BLR` and its friends in `gbak` in order to restore databases with invalid BLR. 

`isc_bad_proc_BLR` is not primary error. It means that something wrong happens when BLR was parsed and specify unlucky routine name. But this is not necessary "bad BLR" as in the case above.

> Of course, this check can be improved to search through the whole status vector, but it would look inconsistent with another error handlers across the codebase. With the new code, perhaps we could rely on `isc_invalid_blr` instead, but old checks are still needed for backward compatibility. Sigh.

BTW, this PR restored how error looks like before v3.
So, do you object to this PR or have I convinced you ?

dyemanov:
I have no objections, just some headache to figure out the best way to fix things back ;)

PR #8908, opened 2026-02-16
Fixed #8749 : Computed index on RDB$RECORD_VERSION don't work as expected.

### LINKED ISSUE #8749: Computed index on RDB$RECORD_VERSION don't work as expected
When computed index is created on RDB$RECORD_VERSION pseudo-column:
```
ALTER TABLE COUNTRY ADD TRA_ID BigInt COMPUTED BY (RDB$RECORD_VERSION);
```
Then query like this that use it:
```
SELECT r.RDB$RECORD_VERSION,r.COUNTRY, r.CURRENCY
FROM COUNTRY r
where r.RDB$RECORD_VERSION >= 100
```
will return empty set.

Similar problem could be observed with RDB$RECORD_VERSION  in `BEFORE` triggers. For example:
```
ALTER TABLE COUNTRY ADD 
TRA_ID BigInt ;

CREATE TRIGGER COUNTRY_BIU FOR COUNTRY
ACTIVE BEFORE INSERT OR UPDATE POSITION 0
AS
BEGIN
    NEW.TRA_ID = NEW.RDB$RECORD_VERSION ;
END^
```
This trigger will always store `0` into `TRA_ID`.

hvlad:


src/jrd/exe.cpp
```diff
@@ -1196,6 +1196,9 @@ void EXE_execute_triggers(thread_db* tdbb,
 					{
 						trigger->req_rpb[0].rpb_number = old_rpb->rpb_number;
 						trigger->req_rpb[0].rpb_number.setValid(true);
+
+						// This allows to use OLD.RDB$RECORD_VERSION in triggers
+						trigger->req_rpb[0].rpb_transaction_nr = old_rec->getTransactionNumber();
```
dyemanov, line 1201: Given that `old_rec / new_rec` are taken from `old_rpb / new_rpb`, wouldn't it be more reliable to get the transaction number directly from their corresponding `rpb's`?

hvlad: `new_rpb` have not set transaction number in before triggers, it is set when new record is stored.
Yes, it could be set in advance, same as done for `new_rec`, but is it really needed ?

dyemanov (approved):

PR #8909, opened 2026-02-17
Support VARCHAR/VARBINARY without explicit length

mrotteveel:
Currently, Firebird doesn't have a default length for VARCHAR/CHARACTER VARYING, { NCHAR | NATIONAL {CHAR|CHARACTER} } VARYING, and VARBINARY/BINARY VARYING.

This PR adds a default length of 255 for all these types.

Why 255:

- Big enough for common use cases
- Big enough to cast from other (non-string) scalar types without truncation
- Small enough to not be excessive in storage use, and fits in an index on page size 8192 with UTF8, with some room to spare (e.g. for compound indices)
- 0xFF ;)

Why at all:

- Required by the SQL standard (see below)
- Allows things like `CAST(something as VARCHAR)` (see also second point under "Why 255"), which can be a simplification for query generators
- Can result in simpler code for ad-hoc queries and such 

For reference, the SQL:2023 standard requires such a default, as 6.1 \<data type> defines:

> ```
> <character string type> ::=
>     [...]
>   | CHARACTER VARYING [ <left paren> <character maximum length> <right paren> ]
>   | CHAR VARYING [ <left paren> <character maximum length> <right paren> ]
>   | VARCHAR [ <left paren> <character maximum length> <right paren> ]
>   | [...]
> 
> [...]
> 
> <binary string type> ::=
>     [..]
>   | BINARY VARYING [ <left paren> <maximum length> <right paren> ]
>   | VARBINARY [ <left paren> <maximum length> <right paren> ]
>   | [...]
> 
> [...]
> 
> <national character string type> ::=
>     [...]
>   | NATIONAL CHARACTER VARYING [ <left paren> <character maximum length> <right paren> ]
>   | NATIONAL CHAR VARYING [ <left paren> <character maximum length> <right paren> ]
>   | NCHAR VARYING [ <left paren> <character maximum length> <right paren> ]
>   | [...]
> ```

With syntax rule 6 (for BINARY) and 7 (for VARCHAR/NCHAR VARYING) specifying:

> 6\) If \<maximum length> is omitted, then an implementation-defined (IL007) \<maximum length> is implicit.
> 
> 7\) If \<character maximum length> is omitted, then an implementation-defined (ID069) \<character maximum length> is implicit.

See also thread [Introducing a default VARCHAR length](https://groups.google.com/g/firebird-devel/c/4ewVpoWqA4I) on firebird-devel.

### DIFF
--- a/src/dsql/parse.y
+++ b/src/dsql/parse.y
@@ -5452,7 +5452,7 @@ national_character_type
 		{
 			$$ = newNode<dsql_fld>();
 			$$->dtype = dtype_text;
-			$$->charLength = 1;
+			$$->charLength = DEFAULT_CHAR_LENGTH;
 			$$->flags |= FLD_national;
 		}
 	| national_character_keyword VARYING '(' pos_short_integer ')'
@@ -5466,7 +5466,7 @@ national_character_type
 		{
 			$$ = newNode<dsql_fld>();
 			$$->dtype = dtype_varying;
-			$$->charLength = 255;
+			$$->charLength = DEFAULT_VARCHAR_LENGTH;
 			$$->flags |= FLD_national;
 		}
 	;
@@ -5488,8 +5488,8 @@ binary_character_type
 		{
 			$$ = newNode<dsql_fld>();
 			$$->dtype = dtype_text;
-			$$->charLength = 1;
-			$$->length = 1;
+			$$->charLength = DEFAULT_BINARY_LENGTH;
+			$$->length = DEFAULT_BINARY_LENGTH;
 			$$->textType = ttype_binary;
 			$$->charSetId = CS_BINARY;
 			$$->subType = fb_text_subtype_binary;
@@ -5510,8 +5510,8 @@ binary_character_type
 		{
 			$$ = newNode<dsql_fld>();
 			$$->dtype = dtype_varying;
-			$$->charLength = 255;
-			$$->length = 255 + sizeof(USHORT);
+			$$->charLength = DEFAULT_VARBINARY_LENGTH;
+			$$->length = DEFAULT_VARBINARY_LENGTH + sizeof(USHORT);
 			$$->textType = ttype_binary;
 			$$->charSetId = CS_BINARY;
 			$$->subType = fb_text_subtype_binary;
@@ -5532,7 +5532,7 @@ character_type
 		{
 			$$ = newNode<dsql_fld>();
 			$$->dtype = dtype_text;
-			$$->charLength = 1;
+			$$->charLength = DEFAULT_CHAR_LENGTH;
 		}
 	| varying_keyword '(' pos_short_integer ')'
 		{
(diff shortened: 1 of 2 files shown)


aafemt:
Instead of doubling branches I would introduce `opt_char_length` clause that returns 255 on absence and replace with it `'(' pos_short_integer ')'` in existing code. Did you try it?

src/dsql/parse.y
```diff
@@ -5523,6 +5541,12 @@ character_type
 			$$->charLength = (USHORT) $3;
 			$$->flags |= FLD_has_len;
 		}
+	| varying_keyword
+		{
+			$$ = newNode<dsql_fld>();
+			$$->dtype = dtype_varying;
+			$$->charLength = 255;
```
dyemanov: Please no magic numbers. Add a named constant (e.g. in `constants.h`) and use it inside the parser.

mrotteveel:
> Instead of doubling branches I would introduce `opt_char_length` clause that returns 255 on absence and replace with it `'(' pos_short_integer ')'` in existing code. Did you try it?

I also have to track `FLD_has_len`, which is needed for `SET BIND` (and maybe other things), and this follows the pattern that is also applied for BINARY, CHAR and NCHAR.

src/dsql/parse.y
```diff
@@ -5911,20 +5935,9 @@ set_bind_from
...
-bind_type
-	: non_array_type
-	| varying_keyword
-		{
-			$$ = newNode<dsql_fld>();
-			$$->dtype = dtype_varying;
-			$$->charLength = 0;
-		}
-	;
-
```
dyemanov, line 5924: This changes `charLength` from 0 to 255. Is it OK from the `BIND` command POV?

dyemanov (changes requested):

src/dsql/parse.y
mrotteveel: Yes, it's only set to `0`, because it isn't needed (SET BIND uses the `FLD_has_len` flag to determine if it needs the length or not)

src/dsql/parse.y
mrotteveel: OK, should I also do the same for CHAR/BINARY which currently use `$$->charLength = 1`?

dyemanov: > OK, should I also do the same for CHAR/BINARY which currently use `$$->charLength = 1`?

Well, `1` looks more understandable, but a named constant wouldn't hurt here too, I believe.

mrotteveel: Done in https://github.com/FirebirdSQL/firebird/pull/8909/commits/17282ab6ca8a066ed5d953a8671ef9e2bebbc840

dyemanov (approved):

mrotteveel:
@dyemanov Any objections if I merge? (The build failure is a transient failure that occurs regularly on macOS.)

PR #8917, opened 2026-02-25
Libtommath update to 1.2.0

Khalyutin:
- update *.c and *.h files in source tree extern\libtommath 
to to https://github.com/libtom/libtommath/releases/tag/v1.2.0

- cross-platform compilation error patch (lilke win only patch  https://github.com/microsoft/vcpkg/blob/master/ports/libtommath/msvc-dce.patch)

- update  *.c and *.h  file list in MSVC15.vcxproj  (need for build-windows)

- update  *.o file list in makefiles  (need for build-apline)


extern/btyacc/reader.c
```diff
@@ -147,10 +147,13 @@ char *get_line() {
   /* VM: Process %include line */
   if(strncmp(&line[0], "%include ", 9)==0) {
     int ii=0;
-    for(i=9; line[i]!='\n' && line[i]!=' '; i++, ii++) {
-      inc_file_name[ii] = line[i];
-    }
-    inc_file_name[ii] = 0;
+
```
AlexPeshkoff, line 150: How is it related with libtommath update?  :)

src/remote/inet.cpp
```diff
@@ -3121,6 +3121,8 @@ static bool packet_receive(rem_port* port, UCHAR* buffer, SSHORT buffer_length,
 			INET_force_error = 1;
 			try
 			{
+				// function  inet_error as 5 parameters
```
AlexPeshkoff, line 3124: Please do not put all stuff into single PR.

AlexPeshkoff (changes requested):

hvlad:
Why 1.2.0 ? AFAIU, latest stable release is 1.3.0

extern/btyacc/reader.c
Khalyutin: it is already deleted from this PR 
https://github.com/FirebirdSQL/firebird/pull/8917/changes/87cb3693861fcf818b4a61a1345f16f4e81d7bfc/

src/remote/inet.cpp
Khalyutin: This PR was approved earlier.
https://github.com/FirebirdSQL/firebird/pull/8882..
I don't understand why this change was repeated here...
and how to remove it from here :(

Khalyutin:
> Why 1.2.0 ? AFAIU, latest stable release is 1.3.0

 in the Linux dockerfile we noticed downloads  and compilations on the fly libtommath v1.2.0. 
- We didn't want to add different versions of the library.
- Besides, my experience tells me it's safer to switch to the second-to-last version when a new one is released. :)

Khalyutin (commented):
-

AlexPeshkoff:
> > Why 1.2.0 ? AFAIU, latest stable release is 1.3.0
> 
> in the Linux dockerfile we noticed downloads and compilations on the fly libtommath v1.2.0.

Prebuilt linux binaries can use any version of library which is installed in host OS. Yes, docker is using 1.2, but it makes no difference for the rest of platforms.

> 
>     * Besides, my experience tells me it's safer to switch to the second-to-last version when a new one is released. :)

Gentoo stable (known with it's careful approach of choosing really stable versions) is using 1.3.0. I agree with Vlad - let's upgrade to it.

src/remote/inet.cpp
AlexPeshkoff: For example create new fresh branch and cherry-pick required commits from this branch.

Khalyutin:
Okay, but we need to test again for 1.3 ...

extern/btyacc/reader.c
Khalyutin: If possible, of course, please pay attention for:

https://github.com/FirebirdSQL/firebird/pull/8887
https://github.com/FirebirdSQL/firebird/pull/8893

are pending. We are preparing about ten more corrections... (not fount #include files, c++20 standard error etc.)
It would be much easier for us to debug and testing...

Khalyutin:
Libtommath update to 1.3.0
https://github.com/FirebirdSQL/firebird/pull/8919

PR #8937, opened 2026-03-10
Get rid of the unnecessary permissions for implicit domains

### LINKED ISSUE #8881: Large amount of unnecessary privileges in RDB$USER_PRIVILEGES for SYSDBA
In a small database with 40 tables in Firebird 3:

```sql
SELECT 
    RDB$PRIVILEGE,
    RDB$GRANT_OPTION,
    COUNT(*) as cnt
FROM RDB$USER_PRIVILEGES 
WHERE RDB$USER = 'SYSDBA'
  AND RDB$OBJECT_TYPE = 9
GROUP BY RDB$PRIVILEGE, RDB$GRANT_OPTION
ORDER BY cnt DESC;
```

Result:

```
G   1   327,000 records.
```

What is this and how to fix it? Has this been addressed somewhere? What is this?

The G privilege with GRANT_OPTION = 1 means "WITH GRANT OPTION" - permission to grant these privileges to other users. Having 327,000 such records for SYSDBA is completely abnormal and redundant because:

- SYSDBA already has all privileges by default
- SYSDBA can grant any privileges anyway
- This is system-level bloat that slows down backups

dyemanov:
This PR addresses some issues spotted in #8881:

1) Dropping a field / parameter which is based on implicit domain leaves orphan records inside `RDB$USER_PRIVILEGES` and `RDB$SECURITY_CLASSES`
2) No privileges should be assigned to implicit domains created under the hood

### DIFF
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -1937,7 +1937,7 @@ void CreateAlterFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsql
 
 		// second pass
 		if (alterIndividualParameters)
-			executeAlterIndividualParameters(tdbb, dsqlScratch, transaction, !altered, true, false);
+			executeAlterIndividualParameters(tdbb, dsqlScratch, transaction, false, true, false);
 		else
 			executeAlter(tdbb, dsqlScratch, transaction, !altered, true, false);
 
@@ -3012,7 +3012,7 @@ void CreateAlterProcedureNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsq
 
 		// second pass
 		if (alterIndividualParameters)
-			executeAlterIndividualParameters(tdbb, dsqlScratch, transaction, !altered, true, false);
+			executeAlterIndividualParameters(tdbb, dsqlScratch, transaction, false, true, false);
 		else
 			executeAlter(tdbb, dsqlScratch, transaction, !altered, true, false);
 
--- a/src/dsql/DdlNodes.epp
+++ b/src/dsql/DdlNodes.epp
@@ -1111,8 +1111,16 @@ void DdlNode::storeGlobalField(thread_db* tdbb, jrd_tra* transaction, QualifiedN
 			Arg::Gds(isc_dsql_max_arr_dim_exceeded));
 	}
 
+	bool implicitDomain = false;
+
 	if (name.object.isEmpty())
+	{
 		DYN_UTIL_generate_field_name(tdbb, name);
+		implicitDomain = true;
+	}
+
+	// Avoid the security class being assigned for implicit domains
+	AutoSetRestoreFlag<ULONG> noSecurityClass(&tdbb->tdbb_flags, implicitDomain ? TDBB_no_security_class : 0, true);
 
 	AutoCacheRequest requestHandle(tdbb, drq_s_fld_src, DYN_REQUESTS);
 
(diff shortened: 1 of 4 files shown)


aafemt:
While you are at it, could you also mark implicit domains as system, please?..

dyemanov:
> While you are at it, could you also mark implicit domains as system, please?..

I'm not sure this is correct (and that it will not cause backward compatibility issues).

aafemt:
But it would allow to clean out the hack with `RDB$` prefix comparison from Firebird code and give users more freedom in domain naming.

sim1984:
In my mind, a system domain is a domain that is used in system tables. You can add the automatically generated domain flag to ods 14.

aafemt:
I always read it as "created and maintained by system" in contrast to "created by user".

src/dsql/DdlNodes.epp
```diff
@@ -1937,7 +1937,7 @@ void CreateAlterFunctionNode::execute(thread_db* tdbb, DsqlCompilerScratch* dsql
 
 		// second pass
 		if (alterIndividualParameters)
-			executeAlterIndividualParameters(tdbb, dsqlScratch, transaction, !altered, true, false);
+			executeAlterIndividualParameters(tdbb, dsqlScratch, transaction, false, true, false);
```
AlexPeshkoff: Sorry that I did not mention it at once - in that case no reasons to have `create` parameter in `executeAlterIndividualParameters`, false is always passed to it.

dyemanov:
> I always read it as "created and maintained by system" in contrast to "created by user".

As for me, "system" is something automagically created along with the database and that cannot be altered/dropped by user.

That said, I don't mind adding an explicit/implicit flag for domains to avoid those dumb `RDB$` prefix checks. However, this is unlikely to be really useful provided that all restored databases will have domains migrated with the `<unknown>` status in this flag.

sim1984:
> That said, I don't mind adding an explicit/implicit flag for domains to avoid those dumb `RDB$` prefix checks. However, this is unlikely to be really useful provided that all restored databases will have domains migrated with the `<unknown>` status in this flag.

You can restore RDB$XXXX domains from previous ODSs, where X is a number with the correct flag.

I see that the user cannot create domains with such names.

```sql
CREATE DOMAIN RDB$100000000 AS INTEGER;
```

```
duplicate specification of  - not supported.
unsuccessful metadata update.
CREATE DOMAIN RDB$100000000 failed.
SQL error code = -637.
Implicit domain name RDB$100000000 not allowed in user created domain.
----------------------------------------------------------------------
SQLCODE: -637
SQLSTATE: 42000
GDSCODE: 335544351
```

aafemt:
> I don't mind adding an explicit/implicit flag for domains to avoid those dumb `RDB$` prefix checks. However, this is unlikely to be really useful provided that all restored databases will have domains migrated with the `<unknown>` status in this flag.

There is no need for a separate flag, value `2`, for example, in `RDB$SYSTEM_FLAG` should be enough. In this case in the restored databases this field will normally get value `0` with meaning "user-created" which is acceptable and harmless. Taken into account that code used to compare value of this field with zero, no old code should be affected.

AlexPeshkoff (approved):

dyemanov:
> There is no need for a separate flag, value `2`, for example, in `RDB$SYSTEM_FLAG` should be enough. In this case in the restored databases this field will normally get value `0` with meaning "user-created" which is acceptable and harmless. Taken into account that code used to compare value of this field with zero, no old code should be affected.

I tend to agree, we use this approach for implicit generators created for identity columns, so it would be consistent. The value is gonna be `7` though, as the system flag domain is shared among all system tables:

```c++
enum fb_sysflag {
	fb_sysflag_user = 0,
	fb_sysflag_system = 1,
	fb_sysflag_qli = 2,
	fb_sysflag_check_constraint = 3,
	fb_sysflag_referential_constraint = 4,
	fb_sysflag_view_check = 5,
	fb_sysflag_identity_generator = 6
};
```

And I will probably commit it in a separate PR, because this one is likely to be backported.

PR #8938, opened 2026-03-11
Various fixes for CodeChecker warnings

craftmaster1231:
This PR contains warning fixes contributed by several students.
Each student was assigned a specific set of warnings to resolve.
The warnings were generated using CodeChecker tool on a Release build.
Most fixes are self-explanatory.

See:
https://codechecker.readthedocs.io/en/latest/


src/common/StatusArg.cpp
```diff
@@ -143,7 +143,7 @@ void StatusVector::ImplStatusVector::assign(const StatusVector& v) noexcept
 
 void StatusVector::ImplStatusVector::assign(const Exception& ex) noexcept
 {
-	clear();
+	ImplStatusVector::clear();
```
aafemt: Using of class name inside of the class' methods is completely pointless.

src/burp/restore.epp
```diff
@@ -9401,7 +9401,7 @@ bool get_trigger_old (BurpGlobals* tdgbl, burp_rel* relation)
  *	Get a trigger definition for a relation.
  *
  **************************************/
-	enum trig_t	type;
+	enum trig_t	type = trig_none;
```
aafemt, line 9404: This value is better to be named `trig_invalid` and a check for valid type was assigned should be added.

src/common/classes/NoThrowTimeStamp.cpp
```diff
@@ -341,51 +341,36 @@ int NoThrowTimeStamp::convertGregorianDateToWeekDate(const struct tm& times) noe
...
 
 	// Find the jan1Weekday for y (Monday=1, Sunday=7)
-	const int yy = (y - 1) % 100;
-	const int c = (y - 1) - yy;
-	const int g = yy + yy / 4;
-	const int jan1Weekday = 1 + (((((c / 100) % 4) * 5) + g) % 7);
+	const int y_1 = y - 1;
+	const int yy = y_1 % 100;
+	const int c = y_1 - yy;
+	const int g = yy + (yy >> 2);
```
aafemt, line 351: Modern compilers are clever enough to reuse pieces of calculation and replace division by 4 with shifts, but division is better readable by human.

src/dsql/StmtNodes.cpp
```diff
@@ -523,7 +523,7 @@ BlockNode* BlockNode::pass2(thread_db* tdbb, CompilerScratch* csb)
 const StmtNode* BlockNode::execute(thread_db* tdbb, Request* request, ExeState* exeState) const
 {
 	jrd_tra* transaction = request->req_transaction;
-	SavNumber savNumber;
+	SavNumber savNumber = 0;
```
aafemt, line 526: It is better to move this declaration to the points of real use to make sure that savepoint zero is never used by any mistake.

src/gpre/c_cxx.cpp
```diff
@@ -2668,6 +2668,8 @@ static void gen_put_segment( const act* action, int column)
 	args.pat_ident1 = blob->blb_len_ident;
 	args.pat_ident2 = blob->blb_buff_ident;
 	args.pat_string1 = global_status_name;
+	args.pat_long1 = NULL;
```
aafemt, line 2671: Here and in other places where variable type is `SLONG`: assign zero instead of `NULL`, please.

src/isql/isql.epp
```diff
@@ -7219,7 +7253,7 @@ static unsigned print_item(TEXT** s, const IsqlVar* var, const unsigned length)
 				// one digit more than the specified precision when the value is 0
 				// The bug appears in TCS DSQL_DOMAIN_12 and 13
 				//
-				const double value = *var->value.asFloat;
+				const double value = static_cast<double>(*var->value.asFloat);
```
aafemt, line 7256: Assignment of float to double doesn't lose precision. What is really fixed here?

src/jrd/IntlManager.cpp
```diff
@@ -458,11 +458,11 @@ bool IntlManager::initialize()
 				string configInfo;
 
 				const ConfigFile::Parameter* module = ch->sub->findParameter("intl_module");
-				const ConfigFile::Parameter* objModule;
+				const ConfigFile::Parameter* objModule = NULL;
```
aafemt, line 461: `nullptr` instead of `NULL`, please.

src/jrd/vio.cpp
```diff
@@ -7264,8 +7264,9 @@ void VIO_update_in_place(thread_db* tdbb,
 	{
 		stack = &new_rpb->rpb_record->getPrecedence();
 	}
-	else if (org_rpb->rpb_record) // we apply update to delete stub
+	else // we apply update to delete stub
 	{
+	    fb_assert(org_rpb->rpb_record);
```
aafemt, line 7269: After this change `fb_assert` below lost purpose and can be removed. The comment may be moved inside and preserved, but, please, erase reference to me from it.

src/jrd/btr.h
```diff
@@ -180,10 +180,10 @@ inline constexpr int key_empty		= 1;	// Key contains empty data / empty string
 
 struct temporary_mini_key
 {
-	USHORT key_length;
+	USHORT key_length = 0;
```
aafemt, line 183: Converting POD type to constructible one should be very well considered. Especially in this ancient and dark piece of code.

src/common/isc_sync.cpp
```diff
@@ -1473,7 +1473,7 @@ SharedMemoryBase::SharedMemoryBase(const TEXT* filename, ULONG length, IpcObject
 #endif
 #endif
 
-				memset(sh_mem_mutex->mtx_mutex, 0, sizeof(*(sh_mem_mutex->mtx_mutex)));
+				memset(sh_mem_mutex->mtx_mutex, 0, sizeof(sh_mem_mutex->mtx_mutex));
```
aafemt, line 1476: This change is definitely wrong.

sim1984:
In my opinion, such huge PRs without a clear goal have zero chance of being accepted. It's possible to fix some warnings in a separate subsystem or a small number of files so they can be easily reviewed.

AlexPeshkoff:
> In my opinion, such huge PRs without a clear goal have zero chance of being accepted. It's possible to fix some warnings in a separate subsystem or a small number of files so they can be easily reviewed.

Agreed. Moreover, it contains errors like was shown by @aafemt . I suggest to reject PR.

aafemt:
Errors can be fixed, that's a normal process of learning for students, but I agree: several PRs aimed to separate issues would be better.

craftmaster1231:
Errors discovered by @aafemt will be fixed by students. Thank you for review!
If this PR is going to be rejected due to unclear goals and size even after fixes, would it be OK to open 8-12 different PRs grouped by the purpose of the fixes? Changes were already grouped, but I decided not to create PR flood.

AlexPeshkoff:
On 3/17/26 12:20, Alex wrote:
> *craftmaster1231* left a comment (FirebirdSQL/firebird#8938) 
> <https://github.com/FirebirdSQL/firebird/pull/8938#issuecomment-4073481990>
>
> Errors discovered by @aafemt <https://github.com/aafemt> will be fixed 
> by students. Thank you for your review!
> If this PR is going to be rejected due to unclear goals and size even 
> after fixes, would it be OK to open 8-12 different PRs grouped by the 
> purpose of the fixes? Changes were already grouped, but I decided to 
> not create PR flood.
>

Please do.

src/common/StatusArg.cpp
craftmaster1231: Virtual method calls bypass virtual dispatch during construction and deconstruction.

I propose to keep names in constructors and destructor to make called functions obvious. Names in other methods should be deleted. I agree, they are pointless and can harm logic with future overrides.

aafemt: > Virtual method calls bypass virtual dispatch during construction and deconstruction.

Yes, and this is well-known and established behavior. Perhaps it is even engraved in standard.

aafemt: But if your code analyzer insists on fixing it, the right fix would be to **remove** calls to virtual methods from constructor and destructor, leaving comment why these methods should not be called here, and replacing them with calls to ordinary methods.

src/common/isc_sync.cpp
craftmaster1231: `mtx_mutex` is:
`pthread_mutex_t mtx_mutex[1];`

Sizeof will return same value.
After this change `memset` still nulls out whole array, but without accessing pthread_mutex_t as an object (even in non-evaluated context).

aafemt: Indeed, I was wrong overlooking the declaration. Whole this `memset` in this case looks unneeded because of immediately following `pthread_mutex_init` call.

craftmaster1231: Here is what said with `BUGGY_LINUX_MUTEX` macro enabling. This nulling out gets executed both with and without this macro.

```
#if (defined(HAVE_PTHREAD_MUTEXATTR_SETPROTOCOL) || defined(USE_ROBUST_MUTEX)) && defined(LINUX)
// glibc in linux does not conform to the posix standard. When there is no RT kernel,
// ENOTSUP is returned not by pthread_mutexattr_setprotocol(), but by
// pthread_mutex_init(). Use a hack to deal with this broken error reporting.
#define BUGGY_LINUX_MUTEX
#endif
```

aafemt: Does it matter? Value returned by `pthread_mutex_init` is handled correctly and [cleaning of the variable](https://codebrowser.dev/glibc/glibc/nptl/pthread_mutex_init.c.html#89) is the first thing that happen inside after attributes validation.

craftmaster1231: There are cases ENOTSUP is returned [here](https://codebrowser.dev/glibc/glibc/nptl/pthread_mutex_init.c.html#78) and [here](https://codebrowser.dev/glibc/glibc/nptl/pthread_mutex_init.c.html#84) before this cleaning, which is expected [here](https://github.com/FirebirdSQL/firebird/blob/c90dd14715580628bf7dc6667f9bc7b09838a69a/src/common/isc_sync.cpp#L1482) and [here](https://github.com/FirebirdSQL/firebird/blob/c90dd14715580628bf7dc6667f9bc7b09838a69a/src/common/isc_sync.cpp#L1493) as return from `pthread_mutex_init` with `BUGGY_LINUX_MUTEX`  after firebird's cleaning. 


In my understanding, if ENOTSUP is returned, mtx_mutex will not be cleaned after call, so it is easier to do it unconditionally.

AlexPeshkoff:
Closed in favor of set of smaller PRs.

PR #8949, opened 2026-03-23
postfix(pag): Incorrect calculation of last used page when database contains only one PIP

TreeHunter9:
Postfix after https://github.com/FirebirdSQL/firebird/pull/8772.


aafemt:
Because this PR is for master and master has C++20 enabled, consider using [std::countl_zero](https://en.cppreference.com/w/cpp/numeric/countl_zero.html) in the loop above.

TreeHunter9:
> Because this PR is for master and master has C++20 enabled, consider using [std::countl_zero](https://en.cppreference.com/w/cpp/numeric/countl_zero.html) in the loop above.

I don't like the idea of adding some changes that doesn't belong to initial patch intent, especially when it comes to postfix patches, besides, this patch should be backported to v5.

But the idea of using `std::countl_zero` is good overall, however, it seems that we won't see any difference in performance as it is not a "hot" function.

PR #8954, opened 2026-03-25
Implement #8928 : Add limit for max value of -PARALLEL switch value when restoring a database

### LINKED ISSUE #8928: Add limit for max value of -PARALLEL switch value when restoring a database
Currently gbak knows no limits when reading -par parameter and when a user tries to restore a database with gigantic value, (on purpose or by mistake - for example 16384) this will cause OOM killer on Linux. On Windows this caused whole OS to stop functioning and hard restart was required.
My proposition - if gbak is able to attach to firebird process (which has access to firebird.conf) just by using services manager, then why not to use it exactly to read MaxParallelWorkers? It seems an ideal parameter to restrict max parallel workers. Why does it have to be limited for parallel operations apart from gbak?
Best regards.

hvlad:


src/burp/burpswi.h
```diff
@@ -239,6 +239,7 @@ static const char* burp_repl_mode_sw_table[] =
 	BURP_SW_MODE_NONE, BURP_SW_MODE_RO, BURP_SW_MODE_RW
 };
 
+const int BURP_MAX_PARALLEL_WORKERS = 64;
```
aafemt, line 242: `constexpr` may be?..

hvlad: I see no reason to use 1 `constexpr` while other 55 are `const int`.
And - no - I'm not going to replace all other `const int` in this PR.

dyemanov (approved):

PR #8955, opened 2026-03-26
Charset encoding fix

Ev3nt:
## Problem
When processing text data without an explicitly set encoding, non-ASCII characters were unconditionally replaced with `'?'`. This caused irreversible data loss — for example, usernames containing Cyrillic or other multibyte characters appeared corrupted in trace output.

### Request
```sql
create user user1 password 'user1' firstname 'Абоба';
```

### firebird.conf
```properties
AuditTraceConfigFile = fbtrace.conf
TracePlugin = fbtrace
```
### fbtrace.conf
```properties
database
{
  enabled = true
  log_filename = fbtrace.log
  log_statement_finish = true
  time_threshold = 0
}

```

## Before
```console
create user user1 password 'user1' firstname '??????????'
...
param3 = varchar(128), "??????????"
```

## After
```console
create user user1 password 'user1' firstname 'Абоба'
...
param3 = varchar(128), "Абоба"
```

## Fix
Instead of unconditionally replacing unrecognized characters with `'?'`, the code now first attempts to convert the text to UTF-8. The `'?'` replacement is used only as a fallback if the conversion fails.

### DIFF
--- a/src/dsql/dsql.cpp
+++ b/src/dsql/dsql.cpp
@@ -594,8 +594,6 @@ static RefPtr<DsqlStatement> prepareStatement(thread_db* tdbb, dsql_dbb* databas
 
 		const auto charSetId = database->dbb_attachment->att_charset;
 
-		string transformedText;
-
 		{	// scope to delete parser before the scratch pool is gone
 			Jrd::ContextPoolHolder scratchContext(tdbb, scratchPool);
 
@@ -618,41 +616,16 @@ static RefPtr<DsqlStatement> prepareStatement(thread_db* tdbb, dsql_dbb* databas
 			if (parser.isStmtAmbiguous())
 				scratch->flags |= DsqlCompilerScratch::FLAG_AMBIGUOUS_STMT;
 
-			transformedText = parser.getTransformedString();
+			const string& source = parser.getTransformedString();
+			string transformedText(*statementPool);
+
+			// If the attachment charset is NONE, we first try to convert data to UTF8;
+			// and if that fails, replace non-ASCII characters by question marks.
+			static_assert(CS_METADATA == CS_UTF8);
+			const bool isConverted = DataTypeUtil::convertToUTF8(source, transformedText, charSetId, ERRD_post);
+			dsqlStatement->setSqlText(FB_NEW_POOL(*statementPool) RefString(*statementPool,
+				isConverted ? transformedText : source));
 		}
-
-		// If the attachment charset is NONE, replace non-ASCII characters by question marks, so
-		// that engine internals doesn't receive non-mappeable data to UTF8. If an attachment
-		// charset is used, validate the string.
-		if (charSetId == CS_NONE)
-		{
-			for (char* p = transformedText.begin(), *end = transformedText.end(); p < end; ++p)
-			{
-				if (UCHAR(*p) > 0x7F)
-					*p = '?';
-			}
-		}
-		else
-		{
-			CharSet* charSet = INTL_charset_lookup(tdbb, charSetId);
-
-			if (!charSet->wellFormed(transformedText.length(),
-					(const UCHAR*) transformedText.begin(), NULL))
-			{
-				ERRD_post(Arg::Gds(isc_sqlerr) << Arg::Num(-104) <<
-						  Arg::Gds(isc_malformed_string));
-			}
-
-			UCharBuffer temp;
-
-			CsConvert conversor(charSet->getStruct(),
-				INTL_charset_lookup(tdbb, CS_METADATA)->getStruct());
-			conversor.convert(transformedText.length(), (const UCHAR*) transformedText.c_str(), temp);
-
-			transformedText.assign(temp.begin(), temp.getCount());
-		}
-
-		dsqlStatement->setSqlText(FB_NEW_POOL(*statementPool) RefString(*statementPool, transformedText));
 
 		// allocate the send and receive messages
 
--- a/src/dsql/dsql.cpp
+++ b/src/dsql/dsql.cpp
@@ -617,7 +617,7 @@ static RefPtr<DsqlStatement> prepareStatement(thread_db* tdbb, dsql_dbb* databas
 				scratch->flags |= DsqlCompilerScratch::FLAG_AMBIGUOUS_STMT;
 
 			const string& source = parser.getTransformedString();
-			string transformedText(*statementPool);
+			string transformedText(*scratchPool);
 
 			// If the attachment charset is NONE, we first try to convert data to UTF8;
 			// and if that fails, replace non-ASCII characters by question marks.
--- a/src/jrd/DataTypeUtil.cpp
+++ b/src/jrd/DataTypeUtil.cpp
@@ -219,7 +219,10 @@ ULONG DataTypeUtilBase::convertLength(ULONG len, CSetId srcCharSet, CSetId dstCh
 	if (dstCharSet == CS_NONE || dstCharSet == CS_BINARY)
 		return len;
 
-	return (len / maxBytesPerChar(srcCharSet)) * maxBytesPerChar(dstCharSet);
+	const ULONG srcBPC = maxBytesPerChar(srcCharSet);
+	const ULONG dstBPC = maxBytesPerChar(dstCharSet);
+
+	return (ROUNDUP(len, srcBPC) / srcBPC) * dstBPC;
 }
 
 
@@ -390,9 +393,9 @@ bool DataTypeUtil::convertToUTF8(const string& src, string& dst, CSetId charset,
 
 		dst.resize(length);
 	}
-	catch (const status_exception& e)
+	catch (const status_exception& ex)
 	{
-		const Arg::StatusVector v(e);
+		const Arg::StatusVector v(ex);
 
 		if (charset == CS_NONE)
 		{
--- a/src/jrd/DataTypeUtil.cpp
+++ b/src/jrd/DataTypeUtil.cpp
@@ -376,17 +376,9 @@ bool DataTypeUtil::convertToUTF8(const string& src, string& dst, CSetId charset,
 	if (charset == CS_UTF8 || charset == CS_UNICODE_FSS)
 		return false;
 
-	if (charset == CS_NONE)
-	{
-		const FB_SIZE_T length = src.length();
-
-		const char* s = src.c_str();
-		char* p = dst.getBuffer(length);
-
-		for (const char* end = src.end(); s < end; ++p, ++s)
-			*p = (*s < 0 ? '?' : *s);
-	}
-	else // charset != CS_UTF8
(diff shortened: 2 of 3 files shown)


dyemanov:
Just to clarify the description - we optimistically assume that the input is already in UTF8 and validate whether it's well-formed. If so, it's used "as is", otherwise the original replacement happens.

src/dsql/dsql.cpp
```diff
@@ -618,41 +616,16 @@ static RefPtr<DsqlStatement> prepareStatement(thread_db* tdbb, dsql_dbb* databas
...
-			{
-				if (UCHAR(*p) > 0x7F)
-					*p = '?';
-			}
+			// If the attachment charset is NONE, we first try to convert data to UTF8;
+			// and if that fails, replace non-ASCII characters by question marks.
+			static_assert(CS_METADATA == CS_UTF8);
+			const bool isConverted = DataTypeUtil::convertToUTF8(source, transformedText, charSetId, ERRD_post);
+			dsqlStatement->setSqlText(FB_NEW_POOL(*statementPool) RefString(*statementPool,
+				isConverted ? transformedText : source));
```
hvlad, line 627: In which pool the `source` was allocated ?

aafemt:
Application that uses charset NONE rather send queries in ANSI on Windows and system locale on Linux (which nowadays is UTF-8 indeed), IMHO.

src/jrd/DataTypeUtil.cpp
```diff
@@ -376,28 +376,37 @@ bool DataTypeUtil::convertToUTF8(const string& src, string& dst, CSetId charset,
...
 
 		length = INTL_convert_bytes(tdbb,
 			CS_UTF8, (UCHAR*) dst.getBuffer(length), length,
 			charset, (const BYTE*) src.begin(), src.length(),
-			err);
+			status_exception::raise);
 
 		dst.resize(length);
 	}
+	catch (const status_exception& e)
```
hvlad: `ex`, please

src/jrd/intl.h
```diff
@@ -35,8 +35,8 @@ struct IdStorage
 	constexpr explicit IdStorage(USHORT id) : val(id) { }
 
 	constexpr operator USHORT() const { return val; }
-	bool operator==(const IdStorage& id) const { return val == id.val; }
-	bool operator!=(const IdStorage& id) const { return val != id.val; }
+	constexpr bool operator==(const IdStorage& id) const { return val == id.val; }
+	constexpr bool operator!=(const IdStorage& id) const { return val != id.val; }
```
hvlad, line 39: Is it related change ?

dyemanov:
> Application that uses charset NONE rather send queries in ANSI on Windows

And nothing will change for them.

> and system locale on Linux (which nowadays is UTF-8 indeed)

And they will be printed correctly, instead of being replaced with question marks.

src/jrd/intl.h
Ev3nt: Yes, this is a necessary change, as it is necessary to verify that CS_METADATA is CS_UTF8.
https://github.com/Ev3nt/firebird/blob/charset_encoding_fix/src/dsql/dsql.cpp#L624

src/jrd/DataTypeUtil.cpp
Ev3nt: Ok, i'll rename it in next commit.

src/dsql/dsql.cpp
Ev3nt: As I understand from the code, in `defaultMemoryPool`.

aafemt:
Why not just replace the loop with `ISC_SystemToUtf8` call? It would please users of both platforms.

src/dsql/dsql.cpp
hvlad: I see that `Parser` and its `transformedString` allocated from `scratchPool`

aafemt: Line 626 explicitly allocates the resulting string in `statementPool`, no?

Ev3nt:
> Why not just replace the loop with `ISC_SystemToUtf8` call? It would please users of both platforms.

Isn't it too expensive to lock a mutex every time?
```cpp
MutexLockGuard g(mtx, FB_FUNCTION);
```

src/dsql/dsql.cpp
Ev3nt: `transformedString` uses a default constructor that calls `getDefaultMemoryPool()`, no?

hvlad: > `transformedString` uses a default constructor that calls `getDefaultMemoryPool()`, no?

Look at `Parser::Parser()`. Or should I show it here ?

aafemt: `RefString` is not "a reference to string", it is "a reference-counted string". It is explicitly allocated in the statement pool and is given with the statement pool for copying content from `transformedText` or `source`, so it doesn't matter where `Parser` and `transformedString` is allocated, the text will end up in the statement pool.

hvlad: Finally something meaningful...

aafemt: We could save some time if type names in Firebird were meaningful.

hvlad: > We could save some time if type names in Firebird were meaningful.

Not in this case, for sure

hvlad: BTW, I would reconsider needs to copy parser's `transformedString` into separate temporary string (`transformedText`) or, at least, allocate `transformedText` from short-lived `scratchPool` to not waste space in `statementPool`

asfernandes:
Why the hell we should assume not explicitly defined charset as UTF8?
This makes no sense IMO.
This is also not used only for NONE charset. There is "introducer" syntax that could mix data of different encodings in the text.
This PR should be rejected IMO.

dyemanov:
> Why the hell we should assume not explicitly defined charset as UTF8?

Because this is the charset we're converting the SQL text into. And because this is user-friendly.

> There is "introducer" syntax that could mix data of different encodings in the text.

Hmmm. And how do we currently convert WIN1251 statement with embedded _WIN1250 literals into UTF8? I guess this fails anyway?

asfernandes:
> > Why the hell we should assume not explicitly defined charset as UTF8?
> 
> Because this is the charset we're converting the SQL text into.

"Into" is different than "from". It's why a conversion exists.

> And because this is user-friendly.

It's not, we would masquerading problems in one plataform that appears in another. We should assume nothing about what character set user is using if he does not inform it in the way it should. NONE character never was there to assume the client charset.

> 
> > There is "introducer" syntax that could mix data of different encodings in the text.
> 
> Hmmm. And how do we currently convert WIN1251 statement with embedded _WIN1250 literals into UTF8? I guess this fails anyway?

This case actually seems to use hex strings (see `introducedMarks`.

dyemanov:
> > Because this is the charset we're converting the SQL text into.
> 
> "Into" is different than "from". It's why a conversion exists.

Sure. And of course it's pointless to find out what was the original charset. But we may at least check whether it looks like UTF8 (the one we need).

> We should assume nothing about what character set user is using if he does not inform it in the way it should. NONE character never was there to assume the client charset.

Inside the engine, `CS_NONE / CS_BINARY` strings are converted (into whatever else) using a simple byte copy. We don't explicitly assume anything. But if the destination charset is UTF8, we assume the result will be in UTF8. And `INTL_convert_bytes()` we use for transliteration explicitly calls `toCharSet->wellFormed()` to ensure that. This means we implicitly assume the source string being in UTF8 too. I see zero difference with this PR.

aafemt:
> Isn't it too expensive to lock a mutex every time?

Perhaps it is cheaper than `iconv_open` but I'm not sure.

asfernandes:
> > > Because this is the charset we're converting the SQL text into.
> > 
> > 
> > "Into" is different than "from". It's why a conversion exists.
> 
> Sure. And of course it's pointless to find out what was the original charset. But we may at least check whether it looks like UTF8 (the one we need).
> 

Every thing that looks like UTF-8 looks like NONE, as NONE have no limitations on its own.
It may looks like as others charsets too.

> > We should assume nothing about what character set user is using if he does not inform it in the way it should. NONE character never was there to assume the client charset.
> 
> Inside the engine, `CS_NONE / CS_BINARY` strings are converted (into whatever else) using a simple byte copy. We don't explicitly assume anything. But if the destination charset is UTF8, we assume the result will be in UTF8. And `INTL_convert_bytes()` we use for transliteration explicitly calls `toCharSet->wellFormed()` to ensure that. This means we implicitly assume the source string being in UTF8 too. I see zero difference with this PR.

This is different. It generally happens by individual things (a move from one place to another) and generally to an user defined thing (with user decided to create the target as UTF-8).

Now we have clear thing: you used non-ASCII characters and didn't defined the connection charset, you have question marks as output.

In this PR is a generalization for a whole statement text to different *internal* things (metadata text, trace output, monitoring) is done. And it does it using assumptions that may be wrong. A user in Windows using WIN1252 will think this is a bug and that we should convert it from WIN1252 to UTF8 instead of this.

dyemanov:
> Now we have clear thing: you used non-ASCII characters and didn't defined the connection charset, you have question marks as output.

Yes, it's clear. But in one particular case (input is in fact in UTF8) we could handle it better, but decided to not. This is what I call being not so friendly.

As for me, this clear statement won't become much harder if re-stated as "you used non-ASCII characters and didn't defined the connection charset and your actual encoding wasn't UTF8". But maybe this is just me.

asfernandes:
> Yes, it's clear. But in one particular case (input is in fact in UTF8) we could handle it better, but decided to not. This is what I call being not so friendly.

"It's wrong these question marks. Let me discover what is wrong. Ah, I'm not using the important connection charset. Now I have this problem and many others fixed."

aafemt:
I would suggest to close this PR and try a different approach: in absence of explicit `isc_dpb_lc_ctype` in DPB, client library should add it with the value corresponding to current locale (ANSI codepage). This way the solution would be more general and handle cases of clients with different locales.

dyemanov:
> "It's wrong these question marks. Let me discover what is wrong. Ah, I'm not using the important connection charset. Now I have this problem and many others fixed."

Or they may not notice the problem immediately. And PSQL sources are stored as unrecoverably broken. And DBA cannot get anything meaningful from the trace logs. While it could be avoided.

All these use cases (stored PSQL text, monitoring, trace) are IMO expected to preserve as much original information as possible. Giving up at something "possibly unknown" without checking whether it could be actually processed is easy but somewhat underdone.

And honestly, I don't buy the Linux/Windows argument at all. It's not about the system locale per se, it's about the missing attachment charset. Data could be sent in UTF8 even on Windows (e.g. from a Java application), it's not necessarily comes from CLI tools like ISQL.

livius2:
Wouldn't it be better to throw an error that no charset was provided during the connection? And add compatibility flag in the firebird.conf to support NONE?

PR #8957, opened 2026-03-26
New setting AllowUpdateOverwrite : defines how UPDATE should handle the case when a record was updated by trigger.

hvlad:
After #8538 there was long private discussion with customer (big sponsor of Firebird). 
The root of the issue is the way Firebird implements per-row triggers - it is far from SQL standard in regards of consistency. 
There is well-known historical reasons for it and I don't want to repeat it here.
By my opinion, the best way to fix the issue would be to re-implement per-row triggers according to SQL standard - to fire before and after whole statement processing. But this is not quick nor easy task.
Finally this patch was implemented and customer uses it to prepare own migration to v5.
Now it is time to decide if it could be accepted into codebase.

### DIFF
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -523,6 +523,23 @@
 #SubQueryConversion = false
 
 
+# ----------------------------
+# Defines how UPDATE should handle the case when a record was updated by trigger.
+#
+# It is possible that BEFORE or AFTER UPDATE trigger was fired previously by the
+# current UPDATE statement for the other record and has already changed the
+# current record being updated. Due to cursor stability, UPDATE should not see
+# such changes, and could silently overwrite them.
+#
+# If set to 0 (default), trigger changes will be overwritten by the	UPDATE.
+# If set to 1, error "UPDATE will overwrite changes made by trigger" will be raised.
+#
+# Per-database configurable.
+#
+# Type: integer
+#
+#UpdateOverwriteMode = 0
+
 # ============================
 # Plugin settings
 # ============================
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -534,6 +534,10 @@
 # If set to 0 (default), trigger changes will be overwritten by the	UPDATE.
 # If set to 1, error "UPDATE will overwrite changes made by trigger" will be raised.
 #
+# CAUTION!
+# There is no guarantee that this setting will be available in future Firebird
+# versions.
+#
 # Per-database configurable.
 #
 # Type: integer
--- a/builds/install/misc/firebird.conf
+++ b/builds/install/misc/firebird.conf
@@ -524,15 +524,18 @@
 
 
 # ----------------------------
-# Defines how UPDATE should handle the case when a record was updated by trigger.
+# Defines how UPDATE should handle the case when a record was already updated by 
+# trigger or by the another positional UPDATE in the same cursor.
 #
 # It is possible that BEFORE or AFTER UPDATE trigger was fired previously by the
 # current UPDATE statement for the other record and has already changed the
 # current record being updated. Due to cursor stability, UPDATE should not see
 # such changes, and could silently overwrite them.
 #
-# If set to 0 (default), trigger changes will be overwritten by the	UPDATE.
-# If set to 1, error "UPDATE will overwrite changes made by trigger" will be raised.
+# Setting AllowUpdateOverwrite manage how engine should handle such cases:
+# - if set to true (default), prior changes will be overwritten,
+# - if set to false, error "UPDATE will overwrite changes made by the trigger or 
+#	by the another UPDATE in the same cursor" will be raised.
 #
 # CAUTION!
 # There is no guarantee that this setting will be available in future Firebird
@@ -542,7 +545,8 @@
 #
 # Type: integer
 #
-#UpdateOverwriteMode = 0
+#AllowUpdateOverwrite = true
+
 
 # ============================
 # Plugin settings
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -6617,31 +6617,31 @@ static void refresh_fk_fields(thread_db* tdbb, Record* old_rec, record_param* cu
  * Functional description
  *	Update new_rpb with foreign key fields values changed by cascade triggers.
  *  Consider self-referenced foreign keys only.
+ *  Also, if UpdateOverwriteMode is set to 1, raise error when foreign key fields
+ *  were changed by user triggers.
  *
  *  old_rec - old record before modify
- *  cur_rpb - just read record with possibly changed FK fields
+ *  cur_rpb - just read record with possibly changed fields
  *  new_rpb - new record evaluated by modify statement and before-triggers
  *
  **************************************/
+	const Database* dbb = tdbb->getDatabase();
+	const auto overwriteMode = dbb->dbb_config->getUpdateOverwriteMode();
+
 	jrd_rel* relation = cur_rpb->rpb_relation;
 
 	MET_scan_partners(tdbb, relation);
 
-	if (!(relation->rel_foreign_refs.frgn_relations))
-		return;
-
-	const FB_SIZE_T frgnCount = relation->rel_foreign_refs.frgn_relations->count();
-	if (!frgnCount)
-		return;
+	const FB_SIZE_T frgnCount = relation->rel_foreign_refs.frgn_relations ?
+		relation->rel_foreign_refs.frgn_relations->count() : 0;
 
 	RelationPages* relPages = cur_rpb->rpb_relation->getPages(tdbb);
 
-	// Collect all fields of all foreign keys
+	// Collect all fields of self-referenced foreign keys
 	SortedArray<int, InlineStorage<int, 16> > fields;
 
 	for (FB_SIZE_T i = 0; i < frgnCount; i++)
 	{
-		// We need self-referenced FK's only
 		if ((*relation->rel_foreign_refs.frgn_relations)[i] == relation->rel_id)
 		{
 			index_desc idx;
@@ -6663,26 +6663,63 @@ static void refresh_fk_fields(thread_db* tdbb, Record* old_rec, record_param* cu
 	}
 
 	if (fields.isEmpty())
-		return;
-
-	DSC desc1, desc2;
-	for (FB_SIZE_T idx = 0; idx < fields.getCount(); idx++)
 	{
-		// Detect if user changed FK field by himself.
-		const int fld = fields[idx];
-		const bool flag_old = EVL_field(relation, old_rec, fld, &desc1);
-		const bool flag_new = EVL_field(relation, new_rpb->rpb_record, fld, &desc2);
+		if (overwriteMode == 0)
+			return;
 
-		// If field was not changed by user - pick up possible modification by
-		// system cascade trigger
-		if (flag_old == flag_new &&
-			(!flag_old || (flag_old && !MOV_compare(tdbb, &desc1, &desc2))))
+		if (cur_rpb->rpb_record->getFormat() == old_rec->getFormat())
 		{
-			const bool flag_tmp = EVL_field(relation, cur_rpb->rpb_record, fld, &desc1);
-			if (flag_tmp)
-				MOV_move(tdbb, &desc1, &desc2);
-			else
-				new_rpb->rpb_record->setNull(fld);
+			if (memcmp(cur_rpb->rpb_address, old_rec->getData(), cur_rpb->rpb_length) == 0)
+				return;
+
+			fb_assert(overwriteMode == 1);
+			ERR_post(Arg::Gds(isc_random) << "UPDATE will overwrite changes made by trigger");
+		}
+		// Else compare field-by-field
+	}
+
+	for (FB_SIZE_T fld = 0, frn = 0; fld < relation->rel_current_format->fmt_count; fld++)
+	{
+		dsc dsc_old;
+		const bool flag_old = EVL_field(relation, old_rec, fld, &dsc_old);
+
+		const bool is_fk = (frn < fields.getCount() && fields[frn] == fld);
+		if (!is_fk)
+		{
+			if (overwriteMode == 0)
+				continue;
+
+			dsc dsc_cur;
+			const bool flag_cur = EVL_field(relation, cur_rpb->rpb_record, fld, &dsc_cur);
+
+			// Check if current record differs from old record
+			if ((flag_cur != flag_old) ||
+				(flag_cur && flag_old && MOV_compare(tdbb, &dsc_old, &dsc_cur) != 0))
+			{
+				// Record was modified by trigger.
+				fb_assert(overwriteMode == 1);
+				ERR_post(Arg::Gds(isc_random) << "UPDATE will overwrite changes made by trigger");
+			}
+		}
+		else
+		{
+			dsc dsc_new;
+			const bool flag_new = EVL_field(relation, new_rpb->rpb_record, fld, &dsc_new);
+
+			// If field was not changed by user - pick up possible modification by
+			// system cascade trigger
+			if (flag_old == flag_new &&
+				(!flag_old || (flag_old && !MOV_compare(tdbb, &dsc_old, &dsc_new))))
+			{
+				dsc dsc_cur;
+				const bool flag_cur = EVL_field(relation, cur_rpb->rpb_record, fld, &dsc_cur);
+				if (flag_cur)
+					MOV_move(tdbb, &dsc_cur, &dsc_new);
+				else
+					new_rpb->rpb_record->setNull(fld);
+			}
+
+			frn++;
 		}
 	}
 }
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -6617,8 +6617,8 @@ static void refresh_changed_fields(thread_db* tdbb, Record* old_rec, record_para
  * Functional description
  *	Update new_rpb with foreign key fields values changed by cascade triggers.
  *  Consider self-referenced foreign keys only.
- *  Also, if UpdateOverwriteMode is set to 1, raise error when non self-referenced
- *  foreign key fields were changed by user triggers.
+ *  Also, if AllowUpdateOverwrite is set to false, raise error when non
+ *  self-referenced foreign key fields were changed by user triggers.
  *
  *  old_rec - old record before modify
  *  cur_rpb - just read record with possibly changed fields
@@ -6626,7 +6626,7 @@ static void refresh_changed_fields(thread_db* tdbb, Record* old_rec, record_para
  *
  **************************************/
 	const Database* dbb = tdbb->getDatabase();
-	const auto overwriteMode = dbb->dbb_config->getUpdateOverwriteMode();
+	const auto allowOverwrite = dbb->dbb_config->getAllowUpdateOverwrite();
 
 	jrd_rel* relation = cur_rpb->rpb_relation;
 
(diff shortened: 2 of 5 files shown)


AlexPeshkoff:
On 3/26/26 12:04, Vlad Khorsun wrote:
>
> After #8538 <https://github.com/FirebirdSQL/firebird/issues/8538> 
> there was long private discussion with customer (big sponsor of Firebird).
> The root of the issue is the way Firebird implements per-row triggers 
> - it is far from SQL standard in regards of consistency.
> There is well-known historical reasons for it and I don't want to 
> repeat it here.
> By my opinion, the best way to fix the issue would be to re-implement 
> per-row triggers according to SQL standard - to fire before and after 
> whole statement procesing. But this is not quick nor easy task.
> Finally this patch was implemented and customer uses it to prepare own 
> migration to v5.
>

Vlad, am I missing something or this patch works only with system 
triggers that implement cascade FK?

hvlad:
> Vlad, am I missing something or this patch works only with system triggers that implement cascade FK?

The routine, I modified in the patch, was initially used to workaroud issue with self-referenced FK's

builds/install/misc/firebird.conf
```diff
@@ -523,6 +523,23 @@
...
+# current UPDATE statement for the other record and has already changed the
+# current record being updated. Due to cursor stability, UPDATE should not see
+# such changes, and could silently overwrite them.
+#
+# If set to 0 (default), trigger changes will be overwritten by the	UPDATE.
+# If set to 1, error "UPDATE will overwrite changes made by trigger" will be raised.
+#
+# Per-database configurable.
+#
+# Type: integer
```
dyemanov, line 546: Why `integer` and `0/1` instead of `boolean` and `false/true`? And rename to something like `UpdateOverwriteStrictMode` or `RestrictUpdateOverwrite`?

src/jrd/vio.cpp
dyemanov: Maybe worth renaming to `refresh_fields()` or `refresh_changed_fields()`, given it has a non-FK branch now.

src/jrd/vio.cpp
```diff
@@ -6663,26 +6663,63 @@ static void refresh_fk_fields(thread_db* tdbb, Record* old_rec, record_param* cu
...
 
 	if (fields.isEmpty())
-		return;
+	{
+		if (overwriteMode == 0)
+			return;
 
-	DSC desc1, desc2;
-	for (FB_SIZE_T idx = 0; idx < fields.getCount(); idx++)
+		if (cur_rpb->rpb_record->getFormat() == old_rec->getFormat())
```
dyemanov:
```suggestion
		if (cur_rpb->rpb_record->getFormat()->fmt_version == old_rec->getFormat()->fmt_version)
```

src/jrd/vio.cpp
```diff
@@ -6663,26 +6663,63 @@ static void refresh_fk_fields(thread_db* tdbb, Record* old_rec, record_param* cu
...
+		{
+			if (memcmp(cur_rpb->rpb_address, old_rec->getData(), cur_rpb->rpb_length) == 0)
+				return;
+
+			fb_assert(overwriteMode == 1);
+			ERR_post(Arg::Gds(isc_random) << "UPDATE will overwrite changes made by trigger");
```
dyemanov: Please add a new ISC code, unless this gonna be reimplemented in v6.

src/jrd/vio.cpp
```diff
@@ -6663,26 +6663,63 @@ static void refresh_fk_fields(thread_db* tdbb, Record* old_rec, record_param* cu
...
-			else
-				new_rpb->rpb_record->setNull(fld);
+			if (overwriteMode == 0)
+				continue;
+
+			dsc dsc_cur;
+			const bool flag_cur = EVL_field(relation, cur_rpb->rpb_record, fld, &dsc_cur);
+
+			// Check if current record differs from old record
+			if ((flag_cur != flag_old) ||
```
dyemanov, line 6695: If the field didn't exist in the old format but exists in the new format and gets some value updated by a trigger (e.g. because the new field is `NOT NULL`), an error will be raised. Is it intended?

dyemanov (commented):
Honestly, I expected this PR to implement the same logic as seems to exist for FKs:

> If field was not changed by user - pick up possible modification by trigger

instead of raising an error unconditionally.

However, AFAIU, we cannot detect this reliably at the VIO layer, because cannot distinguish between `new.field` explicitly assigned to the old value or copied from `org_record`. This would need accessing the `ModifyNode`'s assignment list and checking for modified fields there. Although I'm not sure this is going to be as simple.

Did you consider this approach? Because the new mode reminds me the famous "table is mutating" error in Oracle, if raised even for non-conflicting changes.

hvlad:
> Honestly, I expected this PR to implement the same logic as seems to exist for FKs:
> 
> > If field was not changed by user - pick up possible modification by trigger
> 
> instead of raising an error unconditionally.

I consider this as wrong way because it let user triggers to implicitly change original UPDATE semantic. 
Also, it is unpredictable and depends on internal code path and physical records order.
This is definitely not consistent behavior. 

Case with self-referenced FK's is more predictable and consistent (unless same fields not changed by user trigger).

> However, AFAIU, we cannot detect this reliably at the VIO layer, because cannot distinguish between `new.field` explicitly assigned to the old value or copied from `org_record`. This would need accessing the `ModifyNode`'s assignment list and checking for modified fields there. Although I'm not sure this is going to be as simple.
> 
> Did you consider this approach? 

  No. By the reasons as explained above.

> Because the new mode reminds me the famous "table is mutating" error in Oracle, if raised even for non-conflicting changes.

The more I think on this issue, the more I agree with Oracle - it should not be allowed. And, actually, I've implemented such check when works on the issue past year. But - not a surprise - it was too restrictive and customer does not accepted it. Finally, the setting UpdateOverwriteMode was introduced and agreed as compromise solution.

Ideally, we should reimplement the way per-row triggers works to provide consistensy and comply with SQL standard.

builds/install/misc/firebird.conf
hvlad: Initially I thought about 3rd value - to issue warning, not error. But it was not implemented as I found it unpractical, iirc.

As for the setting name - I can agree with anything reasonable ;) Lets decide if we need such changes.

src/jrd/vio.cpp
hvlad: Ok

src/jrd/vio.cpp
hvlad: If any field of **current** record was changed by trigger, that fired for **another** record - yes, this is error. 
Despite of formats.

src/jrd/vio.cpp
hvlad: Ok, will do

builds/install/misc/firebird.conf
hvlad: Also, I going to add forgoten note that this setting is temporary and could be removed in future versions.

src/jrd/vio.cpp
hvlad: Done

dyemanov:
Well, provided that it's disabled by default, I can accept this patch. Did you try running the QA pass with `UpdateOverwriteMode = 1`?

builds/install/misc/firebird.conf
dyemanov: > As for the setting name - I can agree with anything reasonable ;)

I suppose we need more opinions here. @mrotteveel , can you suggest any better naming for a boolean version of the switch?

aafemt: How about "DisableTableMutation"?

mrotteveel: So, for my understanding:

1. an update modifies a row -> trigger fires
2. that trigger modifies a different row in the same table
3. the update from (1) continues and then modifies the row modified by (2)

If this setting is false, it will allow (3), causing the change of (2) to be undone. If the setting is true, (3) will fail with an error, and that will then cause all changes to be rolled back.

What about the reverse scenario? What happens if the update modified row A, and then row B, and the trigger fired for row B then modifies row A? Will that update be rejected as well if this setting is enabled?

hvlad: > So, for my understanding:
> 
>     1. an update modifies a row -> trigger fires
>     2. that trigger modifies a different row in the same table
>     3. the update from (1) continues and then modifies the row modified by (2)
> 
> If this setting is false, it will allow (3), causing the change of (2) to be undone. If the setting is true, (3) will fail with an error, and that will then cause all changes to be rolled back.

Correct
 
> What about the reverse scenario? What happens if the update modified row A, and then row B, and the trigger fired for row B then modifies row A? Will that update be rejected as well if this setting is enabled?

No, it will proceed.

hvlad:
> Did you try running the QA pass with `UpdateOverwriteMode = 1`?

Yes, but I don't remember the results, sorry. Will try once more time.

builds/install/misc/firebird.conf
mrotteveel: OK, what about `RejectTriggerUpdateConflict`?

aafemt: > No, it will proceed.

Won't this inconsistency make the result of `UPDATE` plan-dependent and unpredictable in common case?

hvlad: > OK, what about `RejectTriggerUpdateConflict`?

There no what is generally accepted as "update conflict".

hvlad: > > No, it will proceed.
> 
> Won't this inconsistency make the result of `UPDATE` plan-dependent and unpredictable in common case?

It can't be made fully consistent without re-implementing of per-row triggers mechanics.

dyemanov: > There no what is generally accepted as "update conflict".

But there's something quite similar to the commonly known "lost update", so maybe `[Reject|Prevent|Disable]TriggerLostUpdate` ?

hvlad: > > What about the reverse scenario? What happens if the update modified row A, and then row B, and the trigger fired for row B then modifies row A? Will that update be rejected as well if this setting is enabled?
> 
> No, it will proceed.

Note, in this scenario trigger sees and updates new version of the row A and there is no "silent overwrite" of previous (main) update as in opposite scenario.

aafemt: > It can't be made fully consistent without re-implementing of per-row triggers mechanics.

Can the logic from `MERGE` statement that prevents double update of the same record be used here? It would at least make throwing of the error consistent.

hvlad: > > There no what is generally accepted as "update conflict".
> 
> But there's something quite similar to the commonly known "lost update", so maybe `[Reject|Prevent|Disable]TriggerLostUpdate` ?

I read it as "trigger that lost some update" which is obviously not correct ;)

Can you explain, please, what is bad with UpdateOverwriteMode ? Don't you like Mode ? 
Here we have UPDATE that really can OVERWRITE other change and we introduce MODE that allows or not allows it.

hvlad: > > It can't be made fully consistent without re-implementing of per-row triggers mechanics.
> 
> Can the logic from `MERGE` statement that prevents double update of the same record be used here? It would at least make throwing of the error consistent.

I'm not sure we should prevent this as it can disable perfectly valid actions of AFTER UPDATE triggers.

aafemt: One would argue that these actions are not valid (and were known to cause problems in the past, including already mentioned "lost update"). As a workaround the check can be disarmed for `AFTER` triggers.

Advantage of this solution would be that the implementation is simple: just go through savepoints and throw the error if any (except transaction's one) already have the bit for the record set.

PS: The setting in this case can be named like "PreventDoubleUpdate" with values "Always", "Never" and "NotInAfterTriggers" with "Always" as default. ;-)

dyemanov: > Can you explain, please, what is bad with UpdateOverwriteMode ? Don't you like Mode ?

Exactly. `Mode` suggests more options than just `true/false`, or maybe two options but explicitly named -- like `enum UpdateOverwriteMode = { Ignore , Restrict }`. This is why I initially suggested `StrictMode` or alike, because it converts `Mode` into a boolean.

hvlad:
> > Did you try running the QA pass with `UpdateOverwriteMode = 1`?
> 
> Yes, but I don't remember the results, sorry. Will try once more time.

Found no differences. Same number and list of failed and passed tests in two runs.

builds/install/misc/firebird.conf
hvlad: > > Can you explain, please, what is bad with UpdateOverwriteMode ? Don't you like Mode ?
> 
> Exactly. `Mode` suggests more options than just `true/false`, or maybe two options but explicitly named -- like `enum UpdateOverwriteMode = { Ignore , Restrict }`. This is why I initially suggested `StrictMode` or alike, because it converts `Mode` into a boolean.

`AllowUpdateOverwriteTrigger` or simple `AllowUpdateOverwrite` ?

mrotteveel: > `AllowUpdateOverwriteTrigger` or simple `AllowUpdateOverwrite` ?

I assume you mean these in combination with an enum as Dmitry suggested. Then I'd suggest something like:

```
TriggerLostUpdateBehaviour = { Allow | Reject }
```

hvlad: > > `AllowUpdateOverwriteTrigger` or simple `AllowUpdateOverwrite` ?
> 
> I assume you mean these in combination with an enum as Dmitry suggested. 

I meant it should be a `boolean`. I should have state it explicitly, sorry.

> Then I'd suggest something like:
> 
> ```
> TriggerLostUpdateBehaviour = { Allow | Reject }
> ```

No, please. "Lost update" is well known term from another area - transaction's isolation and concurrency. 
And there is no "lost update" in classical meaning here.

dyemanov: > `AllowUpdateOverwriteTrigger` or simple `AllowUpdateOverwrite` ?

Either `AllowUpdateOverwrite[Trigger]` (with default = `true`) or `RejectUpdateOverwrite[Trigger]` (with default = `false`) are fine with me.

hvlad: Ok, I'll change it by `AllowUpdateOverwriteTrigger` with default = `true`.

hvlad:
> > > Did you try running the QA pass with `UpdateOverwriteMode = 1`?
> > 
> > 
> > Yes, but I don't remember the results, sorry. Will try once more time.
> 
> Found no differences. Same number and list of failed and passed tests in two runs.

I'm sorry, missed one failed test: tests\bugs\core_3362_complex_test.py

It shows one more (forgotten) case when update could silently overwrite prior update - when second update is positioned on explicit (stable) cursor:

```
-- positioned, positioned
for select ... from table <table> as cursor c
  do begin
    update table set ... where current of c;  -- 1
    update table set ... where current of c;  -- 2
  end

-- or

-- searched, positioned
for select ... from table <table> as cursor c
  do begin
    update table set ... where id = c.id;  -- 1
    update table set ... where current of c;  -- 2
  end
```
Here: update (2) silently overwrites changes by update (1) and error is raised when `AllowUpdateOverwrite = false`

I consider this as correct behavior that should be properly documented. 
Therefore I'll use 'AllowUpdateOverwrite' name for the setting (i.e. not mention trigger) and correct setting description and error message text.
 
Test case `core_3362_complex_test` should be fixed accordingly.

aafemt:
I don't think that raising an error on two separate updates that operate on the same record is a correct behavior.

hvlad:
> I don't think that raising an error on two separate updates that operate on the same record is a correct behavior.

You missed the point - updates are not separate.

Specially for you:
```
insert into t vaues (1, '', '');
for select * from t as cursor c
do begin
  update t set a = 'a' where currect of c;
  update t set b = 'b' where currect of c;
end
select * from t;
```
Most users expects to see (1, 'a', 'b') here but actuall result is (1, '', 'b')

aafemt:
Yes, this point I missed.

I would also expect `(1, 'a', 'b')`, so seeing `(1, '', 'b')` is a definite bug, but throwing an error is not better.

hvlad:
> Yes, this point I missed.
> 
> I would also expect `(1, 'a', 'b')`, so seeing `(1, '', 'b')` is a definite bug, but throwing an error is not better.

Not a bug but implementation detail. Maybe not so obvious.
Error allows users to detect such code and fix it to not produce such behavior.

Also, nobody forces you to set `AllowUpdateOverwrite = false`.

I'm not happy to introduce such settings, but see no better alternative. At least for the stable branch.

aafemt:
> Error allows users to detect such code and fix it to not produce such behavior.

> I'm not happy to introduce such settings, but see no better alternative. At least for the stable branch.

How about [my suggestion](https://github.com/FirebirdSQL/firebird/pull/8957#discussion_r3035553104) above? Result is the same - the error is thrown, but at least it is thrown consistently for all cases "two updates in one routine", "trigger then main request" and "main request then trigger". All these cases can be fixed by changing of user code.

hvlad:
> > Error allows users to detect such code and fix it to not produce such behavior.
> 
> > I'm not happy to introduce such settings, but see no better alternative. At least for the stable branch.
> 
> How about [my suggestion](https://github.com/FirebirdSQL/firebird/pull/8957#discussion_r3035553104) above? Result is the same - the error is thrown, but at least it is thrown consistently for all cases "two updates in one routine", "trigger then main request" and "main request then trigger". All these cases can be fixed by changing of user code.

As far as I understand your suggestion (to reuse MERGE logic):
a) it can't be directly applied to the positioned updates, 
b) it can't directly handle the case when UPDATE run not by the trigger itself but by deeper statement in the call tree,

Also, the patch I offer is in use by the customer for a few months already and got intensive testing. 
I don't want to allow new risky code into stable branch, especially before point release.

aafemt:
> As far as I understand your suggestion (to reuse MERGE logic):

Don't pay so much attention to the comparison. From "MERGE logic" in the suggestion is only throwing of an error on double update of the same record inside of the same statement.

> a) it can't be directly applied to the positioned updates,
> b) it can't directly handle the case when UPDATE run not by the trigger itself but by deeper statement in the call tree,

Perhaps you read it wrong. The proposed loop through savepoints can detect double update on any (desirable) level. No matter how deep is current update, the update on any higher level is reliable detected. Positioned update is not different to any other update in regard of marking a record as "touched" in `VerbAction::vct_records`.

aafemt:
> Also, the patch I offer is in use by the customer for a few months already and got intensive testing.
> I don't want to allow new risky code into stable branch, especially before point release.

BTW, you don't need to throw whole your work away, just change record modification detection from fields' comparison to looking at backlog.

hvlad:
How your "loop through savepoints" will handle this case ?
```
execute block
as
begin
  insert into t1 (id, ...) values (1, ...);
  update t1 set ... where id = 1;
end
```

aafemt:
> How your "loop through savepoints" will handle this case ?

By throwing an error that "allows users to detect such code and fix it to not produce such behavior". And only in the case if they purposefully "set `AllowUpdateOverwrite = false`".

hvlad:
> > How your "loop through savepoints" will handle this case ?
> 
> By throwing an error that "allows users to detect such code and fix it to not produce such behavior". And only in the case if they purposefully "set `AllowUpdateOverwrite = false`".

Wasted time, again. Don't write to my topics anymore.

aafemt:
Do you care to explain how this example is different from core_3362_complex_test?

hvlad:
> Do you care to explain how this example is different from core_3362_complex_test?

Find the stable cursor and positioned update, if you know what these words means.

dyemanov (approved):

elekktron:
Considero que este enfoque es incorrecto, ya que permite que los activadores del usuario modifiquen implícitamente la semántica de ACTUALIZACIÓN original.
Además, es impredecible y depende de la ruta del código interno y del orden de los registros físicos.
Sin duda, este comportamiento no es consistente.

El caso con claves foráneas autorreferenciadas es más predecible y consistente (a menos que los mismos campos no sean modificados por el usuario).

mrotteveel:
@elekktron Please post in English.

PR #8962, opened 2026-03-29
Added multiple NULL and nullptr checks

Zakk0n:
This PR is a part of https://github.com/FirebirdSQL/firebird/pull/8938

### DIFF
--- a/src/common/DynamicStrings.cpp
+++ b/src/common/DynamicStrings.cpp
@@ -87,6 +87,8 @@ unsigned makeDynamicStrings(unsigned length, ISC_STATUS* const dst, const ISC_ST
 		case isc_arg_cstring:
 			fb_assert(string);
 			*to++ = (ISC_STATUS)(IPTR) string;
+			if (!string)
+				break;
 			memcpy(string, reinterpret_cast<const char*>(from[1]), from[0]);
 			string += *from++;
 			*string++ = '\0';
@@ -97,6 +99,8 @@ unsigned makeDynamicStrings(unsigned length, ISC_STATUS* const dst, const ISC_ST
 		case isc_arg_sql_state:
 			fb_assert(string);
 			*to++ = (ISC_STATUS)(IPTR) string;
+			if (!string)
+				break;
 			strcpy(string, reinterpret_cast<const char*>(*from));
 			string += strlen(string);
 			string++;
--- a/src/gpre/obj_cxx.cpp
+++ b/src/gpre/obj_cxx.cpp
@@ -3368,7 +3368,8 @@ static void gen_t_start( const act* action, int column)
 
 	if (trans->tra_db_count == 1)
 	{
-		printa(column, "%s = %s->startTransaction(%s, %d, fb_tpb_%d);",
+		if (trans->tra_tpb) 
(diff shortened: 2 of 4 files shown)


src/common/DynamicStrings.cpp
```diff
@@ -87,6 +87,8 @@ unsigned makeDynamicStrings(unsigned length, ISC_STATUS* const dst, const ISC_ST
 		case isc_arg_cstring:
 			fb_assert(string);
 			*to++ = (ISC_STATUS)(IPTR) string;
+			if (!string)
```
aafemt, line 90: This is a dead code. The loop above makes sure that `string` is not `nullptr`.

src/gpre/obj_cxx.cpp
```diff
@@ -3367,7 +3368,8 @@ static void gen_t_start( const act* action, int column)
 
 	if (trans->tra_db_count == 1)
 	{
-		printa(column, "%s = %s->startTransaction(%s, %d, fb_tpb_%d);",
+		if (trans->tra_tpb)
```
aafemt, line 3371: And what's else? Don't start the transaction at all?..

src/jrd/replication/Applier.cpp
```diff
@@ -990,15 +998,15 @@ void Applier::storeBlob(thread_db* tdbb, TraNumber traNum, bid* blobId,
 	fb_assert(blob->blb_flags & BLB_temporary);
 	fb_assert(!(blob->blb_flags & BLB_closed));
 
-	if (length)
+	if (blob && length)
```
aafemt, line 1001: ...and if `blob` is `nullptr` - still get AV on `BLB_close` call. Instead `else` branch must be added to `if` in line 983 with bugcheck logging and bail.

src/jrd/lck.cpp
```diff
@@ -1069,6 +1049,9 @@ static Lock* hash_get_lock(Lock* lock, USHORT* hash_slot, Lock*** prior)
 
 	// if no collisions found, we're done
 
+	if (!att->att_compatibility_table)
```
aafemt, line 1052: `att_compatibility_table` is surely initialized in line 1042. Another dead code.

src/jrd/lck.cpp
```diff
@@ -1353,10 +1336,13 @@ static USHORT internal_downgrade(thread_db* tdbb, CheckStatusWrapper* statusVect
 
 	// if we can convert to that level, set all identical locks as having that level
 
+	if (!first || !first->lck_physical)
+		return NULL;
```
aafemt, line 1340: Not `NULL` but a proper `LCK_*` constant here. Which one is "proper"? That's the question. Investigation also can find out if cases when `first` is `nullptr` are possible at all. BTW, `lck_physical` is **not** a pointer.

aafemt (changes requested):
Mechanical "fixing" of AI-produced warnings, no code analyze performed. Too bad even for students.

sim1984:
How is this PR different from https://github.com/FirebirdSQL/firebird/pull/8964 ?

dyemanov:
No follow-ups to the raised questions, thus rejected.

PR #8964, opened 2026-03-29
Added default initialization to avoid garbage values + extra NULL checks

Zakk0n:
This PR is a part of https://github.com/FirebirdSQL/firebird/pull/8938

### DIFF
--- a/src/common/DynamicStrings.cpp
+++ b/src/common/DynamicStrings.cpp
@@ -87,6 +87,8 @@ unsigned makeDynamicStrings(unsigned length, ISC_STATUS* const dst, const ISC_ST
 		case isc_arg_cstring:
 			fb_assert(string);
 			*to++ = (ISC_STATUS)(IPTR) string;
+			if (!string)
+				break;
 			memcpy(string, reinterpret_cast<const char*>(from[1]), from[0]);
 			string += *from++;
 			*string++ = '\0';
@@ -97,6 +99,8 @@ unsigned makeDynamicStrings(unsigned length, ISC_STATUS* const dst, const ISC_ST
 		case isc_arg_sql_state:
 			fb_assert(string);
 			*to++ = (ISC_STATUS)(IPTR) string;
+			if (!string)
+				break;
 			strcpy(string, reinterpret_cast<const char*>(*from));
 			string += strlen(string);
 			string++;
--- a/src/gpre/c_cxx.cpp
+++ b/src/gpre/c_cxx.cpp
@@ -2668,6 +2668,8 @@ static void gen_put_segment( const act* action, int column)
 	args.pat_ident1 = blob->blb_len_ident;
 	args.pat_ident2 = blob->blb_buff_ident;
 	args.pat_string1 = global_status_name;
+	args.pat_long1 = NULL;
+	args.pat_long2 = NULL;
 	PATTERN_expand((USHORT) column, pattern1, &args);
 
 	set_sqlcode(action, column);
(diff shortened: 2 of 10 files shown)


src/common/DynamicStrings.cpp
```diff
@@ -87,6 +87,8 @@ unsigned makeDynamicStrings(unsigned length, ISC_STATUS* const dst, const ISC_ST
 		case isc_arg_cstring:
 			fb_assert(string);
 			*to++ = (ISC_STATUS)(IPTR) string;
+			if (!string)
```
aafemt, line 90: This is actually a dead code. The loop above make sure that string is never nullptr if status vector contains strings.

src/gpre/c_cxx.cpp
```diff
@@ -2668,6 +2668,8 @@ static void gen_put_segment( const act* action, int column)
 	args.pat_ident1 = blob->blb_len_ident;
 	args.pat_ident2 = blob->blb_buff_ident;
 	args.pat_string1 = global_status_name;
+	args.pat_long1 = NULL;
+	args.pat_long2 = NULL;
```
aafemt, line 2672: Assign `NULL` to integral types is a bad idea because some compilers define `NULL` as a `(void*)0` and issue a warning on such assignment.

src/gpre/obj_cxx.cpp
```diff
@@ -3367,7 +3368,8 @@ static void gen_t_start( const act* action, int column)
 
 	if (trans->tra_db_count == 1)
 	{
-		printa(column, "%s = %s->startTransaction(%s, %d, fb_tpb_%d);",
+		if (trans->tra_tpb)
```
aafemt, line 3371: But what's else?

src/gpre/pat.cpp
```diff
@@ -355,12 +364,14 @@ void PATTERN_expand( USHORT column, const TEXT* pattern, PAT* args)
 			break;
 
 		case L1:
-			long_value = args->pat_long1;
+			if (args && args->pat_long1)
```
aafemt, line 367: Again, what's else and why only these two cases? Did you look for cases when `args` **can** be nullptr?

src/jrd/replication/Applier.cpp
```diff
@@ -990,15 +998,17 @@ void Applier::storeBlob(thread_db* tdbb, TraNumber traNum, bid* blobId,
 	fb_assert(blob->blb_flags & BLB_temporary);
 	fb_assert(!(blob->blb_flags & BLB_closed));
 
-	if (length)
-		blob->BLB_put_segment(tdbb, data, length);
-	else
-		blob->BLB_close(tdbb);
+	if (blob){
```
aafemt, line 1001: Bracket on a separate line, please, and writing of a diagnostic to the log in `blob == nullptr` branch to prevent the bug being unnoticed.

src/jrd/IntlManager.cpp
```diff
@@ -458,11 +458,11 @@ bool IntlManager::initialize()
 				string configInfo;
 
 				const ConfigFile::Parameter* module = ch->sub->findParameter("intl_module");
-				const ConfigFile::Parameter* objModule;
+				const ConfigFile::Parameter* objModule = NULL;
```
aafemt, line 461: `nullptr`, please.

src/jrd/lck.cpp
```diff
@@ -1353,10 +1336,13 @@ static USHORT internal_downgrade(thread_db* tdbb, CheckStatusWrapper* statusVect
 
 	// if we can convert to that level, set all identical locks as having that level
 
+	if (!first || !first->lck_physical)
+		return NULL;
```
aafemt, line 1340: Meaningful for this case `LCK_*` constant here, please. Which value would be a meaningful? That's the question.

aafemt (changes requested):

dyemanov:
No follow-ups to the raised questions, thus rejected.

Issue #8965, opened 2026-03-30
Wrong PERCENT_RANK computation in Firebird 5


lukaseder:
Run this query:

```sql
select
  a,
  percent_rank() over (order by a desc) "p1"
from (
  select 1 a from rdb$database union all
  select 2 a from rdb$database union all
  select 3 a from rdb$database union all
  select 4 a from rdb$database
) t
order by a asc
```

The correct result as produced by Firebird 4 was:

```
1	1
2	0.66666666666667
3	0.33333333333333
4	0
```

Firebird 5.0.3 now produces:

```
1	0.1111111111111111
2	0.16666666666666666
3	0.3333333333333333
4	0.0
```

I'm using Firebird from here: https://hub.docker.com/r/firebirdsql/firebird.

A workaround, I can calculate `percent_rank() over (order by a desc)` myself using `rank() over (order by a desc) - 1.0) / (count(*) over () - 1.0)`

sim1984:
Fixed in
https://github.com/FirebirdSQL/firebird/pull/8970
https://github.com/FirebirdSQL/firebird/pull/8971
https://github.com/FirebirdSQL/firebird/pull/8972

dyemanov:
@sim1984 I see you've fixed it in the v4 branch too, but the author says it works correctly in v4. Please clarify.

sim1984:
> [@sim1984](https://github.com/sim1984) I see you've fixed it in the v4 branch too, but the author says it works correctly in v4. Please clarify.

Perhaps the author was testing an older version of Firebird 4.0. This bug was introduced by another fix https://github.com/FirebirdSQL/firebird/issues/8589, which is present everywhere PERCENT_RANK appears.

lukaseder:
Indeed, I was testing 4.0.1 as provided here:
https://dbfiddle.uk/_dyqncBf

mrotteveel:
This also seems to indicate a lack of test coverage for PERCENT_RANK.

lukaseder:
> This also seems to indicate a lack of test coverage for PERCENT_RANK.

Probably, a smaller data set is being used? Here, the result is correct:

```sql
select
  a,
  percent_rank() over (order by a desc) "p1"
from (
  select 1 a from rdb$database union all
  select 2 a from rdb$database
) t
order by a asc
```

sim1984:
To test the fix for this bug, it's best to use this set:

```sql
select
  a,
  percent_rank() over (order by a desc) "p1"
[...]

PR #8970, opened 2026-03-31
Fixed PERCENT_RANK computation #8965

sim1984:


sim1984:
@asfernandes, please review this code

PR #8973, opened 2026-03-31
Added default initialization to avoid garbage values

3qupo:
This PR adds default initialization for variables that were previously used without being explicitly initialized, potentially leading to undefined behavior due to garbage values.

The changes include:
- Initializing local variables at declaration
- Using `nullptr` for pointer initializations 

This PR is part of the closed parent PR #8938, which was split into smaller, focused pull requests.


src/burp/restore.epp
```diff
@@ -1608,7 +1608,7 @@ bool get_acl(BurpGlobals* tdgbl, const TEXT* owner_nm, ISC_QUAD* blob_id, ISC_QU
 
 	ULONG length = 0;
 	UCHAR item;
-	USHORT max_segment;
+	USHORT max_segment = 0;
```
aafemt: Wrong initialization. In this case `USHRT_MAX` is a safe assumption.

src/burp/restore.epp
```diff
@@ -1783,7 +1783,7 @@ void get_array(BurpGlobals* tdgbl, burp_rel* relation, UCHAR* record_buffer)
 	SLONG		fld_ranges[2 * MAX_DIMENSION];
 	SLONG       slice_length = 0;
 	SLONG       *range;
-	const SLONG* end_ranges;
+	const SLONG* end_ranges = nullptr;
```
aafemt: `fld_ranges` is the right default here.

src/burp/restore.epp
```diff
@@ -1851,7 +1851,7 @@ void get_array(BurpGlobals* tdgbl, burp_rel* relation, UCHAR* record_buffer)
 	SLONG last_element_dim[MAX_DIMENSION];
 	if (return_length != slice_length)
 	{
-		int upper, lower;
+		int upper = 0, lower = 0;
```
aafemt: This is remains of old C code. Move this declaration to usage instead.

src/burp/restore.epp
```diff
@@ -9097,7 +9097,7 @@ bool get_db_creator(BurpGlobals* tdgbl)
 	att_type		attribute;
 	scan_attr_t		scan_next_attr;
 	TEXT			usr[GDS_NAME_LEN];
-	SSHORT			uType;
+	SSHORT			uType = 0;
```
aafemt, line 9102: Pointless because of initialization of `typeSet` below. But if your AI insist - initialize `usr` as well for symmetry.

src/burp/restore.epp
```diff
@@ -9422,7 +9422,7 @@ bool get_trigger_old (BurpGlobals* tdgbl, burp_rel* relation)
  *	Get a trigger definition for a relation.
  *
  **************************************/
-	enum trig_t	type;
+	enum trig_t	type = trig_none;
```
aafemt, line 9428: For completeness you should also add a check that a meaningful value was assigned below and throw an error if `type` is still `trig_none`.

src/dsql/DdlNodes.epp
```diff
@@ -227,7 +227,8 @@ static void checkForeignKeyTempScope(thread_db* tdbb, jrd_tra* transaction,
 {
 	AutoCacheRequest request(tdbb, drq_l_rel_info, DYN_REQUESTS);
 	QualifiedName masterRelName;
-	rel_t masterType, childType;
+	rel_t masterType = rel_persistent;
```
aafemt, line 230: This function is basically wrong and will fail in the case if the table has several foreign keys. It must be fixed, not polished.

src/jrd/cvt2.cpp
```diff
@@ -973,7 +973,7 @@ USHORT CVT2_make_string2(const dsc* desc, TTypeId to_interp, UCHAR** address, Mo
  **************************************/
 	UCHAR* from_buf;
 	USHORT from_len;
-	TTypeId from_interp;
+	USHORT from_interp = 0;
```
aafemt: First, replacing a specific type with a generic one is a bad idea.
Second, right initial value would be `ttype_none`, but I would suggest to refactor this routine instead.

src/jrd/cvt2.cpp
```diff
@@ -1004,7 +1004,7 @@ USHORT CVT2_make_string2(const dsc* desc, TTypeId to_interp, UCHAR** address, Mo
 
 	if (desc->isText())
 	{
-		if (from_interp == to_interp || to_interp == ttype_none || to_interp == ttype_binary)
+		if (from_interp == ttype_none || from_interp == to_interp || to_interp == ttype_none || to_interp == ttype_binary)
```
aafemt: This may be a wrong change: NONE is the most permissive charset, so conversion from NONE to something else requires validation of data which can be skipped in the other direction.

src/burp/restore.epp
dyemanov: > Wrong initialization. In this case `USHRT_MAX` is a safe assumption.

In _backup.epp_, `max_segment` is initialized with zero in the very same piece of code. If we have no reply for `isc_info_blob_max_segment` and blob total length is not zero, why should we imply length being `USHRT_MAX` as you suggest?

aafemt: Comment "Rdb sometimes gets the length messed up" suggests that total BLOB length can be evaluated wrongly. In this case using buffer of maximum allowed segment size is better.

I suspect that this piece of code exists only to avoid allocation of 64k memory unconditionally. It had sense in the previous century but not today.

src/burp/restore.epp
dyemanov: I'd better use `std::optional` rather than introduce `trig_none` for nothing useful.

src/burp/restore.epp
hvlad: Its all makes no sense and is a waste of time. Vars gets its values from server response. There can be exactly one reason for any item to have no response - some fatal error. This is API contract. If it become broken, many places in code will fail. 

Note, the title say: "...to avoid garbage values" - this code doesn't prevent any kind of garbage values.

Also, why `max_segment` gets default, while `num_segments` - not ? Looks like "slightly-pregnant", sorry...

aafemt: Because `num_segments` is not used at all. This routine still can handle only ACLs of max size 64k. Perhaps, @AlexPeshkoff missed it when fixed #2651.

PR #8977, opened 2026-04-02
Port initialization fixes to prevent uninitialized memory access

laBobberto:
This PR adds default initialization (= 0, = nullptr) to local variables in several core modules (Common, JRD, GPRE, Utilities). This fixes potential garbage value issues and uninitialized memory access warnings.

This PR is part of the closed parent PR https://github.com/FirebirdSQL/firebird/pull/8938, which was split into smaller, focused pull requests.

### DIFF
--- a/src/common/cvt.cpp
+++ b/src/common/cvt.cpp
@@ -1406,7 +1406,7 @@ double CVT_get_double(const dsc* desc, DecimalStatus decSt, ErrorFunction err, b
  *      Convert something arbitrary to a double precision number
  *
  **************************************/
-	double value;
+	double value = 0.0;
 
 	switch (desc->dsc_dtype)
 	{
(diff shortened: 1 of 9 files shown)


asfernandes:
This is not a good path IMO, unless you really found a uninitialized path.
Double initialization in some paths to ensure every path initializes is worse than what we had.
Compilers can verify uninitialized paths and emit warnings.

laBobberto:
> This is not a good path IMO, unless you really found a uninitialized path. Double initialization in some paths to ensure every path initializes is worse than what we had. Compilers can verify uninitialized paths and emit warnings.

I agree. You're right about relying on compiler warnings rather than double initialization. I've added it as a precaution. I'll close this PR to make sure everything is in order. Thanks for the feedback.

PR #8982, opened 2026-04-07
Prohibit inheritance of interface with versioned methods

### LINKED ISSUE #8980: Trace plugin compatibility is broken
My tracing plugin works fine with the FB 6.0 interface when running FB6, but it throws a segmentation fault when running FB5 when calling `getExplainPlan()`. I think the problem is in the new `getPerfStats` method (added in https://github.com/FirebirdSQL/firebird/commit/62a2a1511c2a85842685cd6063fa788e454193aa) 
because it's in the [middle](https://github.com/FirebirdSQL/firebird/blame/master/src/include/firebird/IdlFbInterfaces.h#L18145) of other functions that mess up the offsets (but I could be wrong).

Reproduction:
1) Put the Firebird 6 version of the libfbtrace.so/fbtrace.dll plugin to Firebird 5.
2) Setup trace:

Firebird.conf:
```ini
AuditTraceConfigFile = fbtrace.conf
```

fbtace.conf
```ini
database
{
 enabled = true
 print_plan = true
 explain_plan = true
}
```
3) Attach to any DB and execute a select SQL:
```SQL
isql -user SYSDBA -p masterkey employee
select * from rdb$database;
```

Noremos:
This is a fix for: https://github.com/FirebirdSQL/firebird/issues/8980

### DIFF
--- a/extern/cloop/src/cloop/Parser.cpp
+++ b/extern/cloop/src/cloop/Parser.cpp
@@ -116,6 +116,12 @@ void Parser::parse()
 	}
 }
 
+static bool hasVersionedMethods(const Interface& interface)
+{
+	auto& methods = interface.methods;
+	return methods.size() >= 2 && methods.front()->version != methods.back()->version;
+}
+
 void Parser::parseInterface(bool exception)
 {
 	auto newInterface = std::make_shared<Interface>();
@@ -140,6 +146,9 @@ void Parser::parseInterface(bool exception)
 			error(token, string("Super interface '") + superName + "' not found.");
 
 		interface->super = static_cast<Interface*>(superIt->second.get());
+		if (hasVersionedMethods(*interface->super))
+			error(token, string("Inheriting from the interface '" + superName + "' is not possible because it contains versioned methods."));
+
 		interface->version = interface->super->version + 1;
 	}
 	else
--- a/extern/cloop/src/cloop/Parser.cpp
+++ b/extern/cloop/src/cloop/Parser.cpp
@@ -118,7 +118,7 @@ void Parser::parse()
 
 static bool hasVersionedMethods(const Interface& interface)
 {
-	auto& methods = interface.methods;
+	const auto& methods = interface.methods;
 	return methods.size() >= 2 && methods.front()->version != methods.back()->version;
 }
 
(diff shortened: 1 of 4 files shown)


extern/cloop/src/cloop/Parser.cpp
```diff
@@ -116,6 +116,12 @@ void Parser::parse()
 	}
 }
 
+static bool hasVersionedMethods(const Interface& interface)
+{
+	auto& methods = interface.methods;
```
dyemanov: `const auto&`, please

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1394,13 +1391,17 @@ interface TraceSQLStatement : TraceStatement
 	TraceParams getInputs();
 	const string getTextUTF8();
 	const string getExplainedPlan();
+version: // 5.0 -> 6.0
```
dyemanov, line 1395: Empty line before `version:`, please

src/include/firebird/FirebirdInterface.idl
```diff
@@ -1394,13 +1391,17 @@ interface TraceSQLStatement : TraceStatement
...
+	PerformanceStats getPerfStats();
 }
 
 interface TraceBLRStatement : TraceStatement
 {
 	const uchar* getData();
 	uint getDataLength();
 	const string getText();
+version: // 5.0 -> 6.0
```
dyemanov, line 1405: Empty line before `version:`, please

extern/cloop/src/cloop/Parser.cpp
Noremos: Fixed

src/include/firebird/FirebirdInterface.idl
Noremos: Fixed

src/include/firebird/FirebirdInterface.idl
Noremos: Fixed

aafemt:
Because public API has C++11 enabled, `static_assert` on size of VMT for every class wouldn't hurt in FirebirdInterface.idl as a prevention against the same problem in the future. Changes in any interface that kick the assert for other classes must be prohibited.

dyemanov (approved):

AlexPeshkoff (approved):
I think this patch is enough for our needs on practice. This is the first time for >10 years when need to add method to base interface took place, therefore nothing except added check is really needed.

asfernandes (approved):
Is the PR title correct or should it be 'inheritance'?

Noremos:
> Is the PR title correct or should it be 'inheritance'?

My bad. Now the title is correct

PR #8983, opened 2026-04-07
Feature #8974 - Temporary Tables in Packages

asfernandes:


src/isql/FrontendParser.cpp
```diff
@@ -633,8 +633,14 @@ FrontendParser::AnyShowNode FrontendParser::parseShow()
 
 				return node;
 			}
-			else if (const auto parsed = parseShowOptQualifiedName<ShowTablesNode>(text, TOKEN_TABLES, 5))
-				return parsed.value();
+			else if (text.length() >= 5 && TOKEN_TABLES.find(text) == 0)
```
Noremos: Perhaps it would be better to pass the  `allowPackage` flag inside the `parseShowOptQualifiedName`?

src/isql/show.epp
```diff
@@ -6200,6 +6219,9 @@ static processing_state show_table(const QualifiedMetaString& name, bool isView)
 		isqlGlob.printf(NEWLINE);
 	}
 
+	if (relationName.package.isEmpty())
+	{	// FIXME: indent
```
Noremos, line 6241: Forgot to correct? Or is this a reminder for the future?

src/jrd/idx.h
```diff
@@ -314,9 +314,10 @@ static inline constexpr struct ini_idx_t indices[] =
 	INDEX(58, rel_schemas, idx_unique, 1, ODS_14_0)
 		SEGMENT(f_sch_schema, idx_metadata)		// schema name
 	}},
-	// define index RDB$INDEX_59 for RDB$RELATIONS unique RDB$SCHEMA_NAME, RDB$RELATION_NAME;
-	INDEX(59, rel_relations, idx_unique, 2, ODS_14_0)
+	// define index RDB$INDEX_59 for RDB$RELATIONS unique RDB$SCHEMA_NAME, RDB$PACKAGE_NAME, RDB$RELATION_NAME;
+	INDEX(59, rel_relations, idx_unique, 3, ODS_14_0)
```
Noremos, line 318: Will not this lead to performance degradation? If I rememebr correctly, the similar problem appears after schemas adding (witch lead to readding a lot of indexes in https://github.com/FirebirdSQL/firebird/pull/8664)

src/jrd/Relation.cpp
```diff
@@ -1174,4 +1174,3 @@ GCLock::State GCLock::isGCEnabled() const
 	return State::unknown;
 }
 #endif //DEV_BUILD
-
```
Noremos, line 1258: Why remove the last blank line here and elsewhere?

src/jrd/dfw.epp
```diff
@@ -2545,7 +2547,11 @@ void DFW_reset_icu(thread_db* tdbb)
 			           coalesce(rfl.RDB$COLLATION_ID, fld.RDB$COLLATION_ID) = coll.RDB$COLLATION_ID
 			    join SYSTEM.RDB$CHARACTER_SETS cs
 			        on coll.RDB$CHARACTER_SET_ID = cs.RDB$CHARACTER_SET_ID
-			    where coll.RDB$SPECIFIC_ATTRIBUTES like '%COLL-VERSION=%' and
+			    where ind.RDB$PACKAGE_NAME is null and
+					  rel.RDB$PACKAGE_NAME is null and
+					  seg.RDB$PACKAGE_NAME is null and
+					  rfl.RDB$PACKAGE_NAME is null and
+					  coll.RDB$SPECIFIC_ATTRIBUTES like '%COLL-VERSION=%' and
```
AlexPeshkoff, line 2722: Can wrong collation version appear in declared LTT's index?

asfernandes: The data is temporary, so the index should be empty.

src/jrd/Relation.cpp
asfernandes: Because there were two empty lines.

src/isql/show.epp
asfernandes: It's to be corrected just before the merge or after it.
The indentation makes review difficult.

src/jrd/idx.h
asfernandes: This changed index is already the new, with schema.
The old one with only relation name was maintained.

src/jrd/Relation.cpp
Noremos: Got it. Looks like GitHub isn't displaying the last blank line.

src/jrd/Relation.h
```diff
@@ -988,6 +990,7 @@ class RelationPermanent : public Firebird::PermanentStorage
 
 	MetaName		rel_owner_name;		// ascii owner
 	QualifiedName	rel_security_name;	// security class name for relation
+	bool			rel_private = false;
```
dyemanov: Why not use a flag `REL_private`?

src/dsql/parse.y
```diff
@@ -3175,6 +3214,8 @@ package_item
 		{ $$ = CreateAlterPackageNode::Item::create($2); }
 	| PROCEDURE procedure_clause_start ';'
 		{ $$ = CreateAlterPackageNode::Item::create($2); }
+	| DECLARE LOCAL TEMPORARY TABLE package_ltt_table_clause ';'
```
dyemanov: We define functions/procedures (and constants) without any prefix word, so maybe it should be avoided for LTTs too? `DECLARE TABLE` would be OK inside PSQL modules, as we already have `DECLARE {VARIABLE | CURSOR | PROCEDURE | FUNCTION}`, but at the package level it looks unnecessary, or at least inconsistent.

src/dsql/parse-conflicts.txt
```diff
@@ -1 +1 @@
-137 shift/reduce conflicts, 7 reduce/reduce conflicts.
+138 shift/reduce conflicts, 10 reduce/reduce conflicts.
```
dyemanov: I'm worried about new reduce/reduce conflicts, did you analyse the reason?

src/dsql/parse.y
asfernandes: You want it without `DECLARE` or without `DECLARE LOCAL`?

src/dsql/parse-conflicts.txt
asfernandes: The conflicts happens because index are optional and can start with optional `UNIQUE`, `ASC` and `DESC`.

src/dsql/parse.y
dyemanov: I was thinking about `DECLARE` being unnecessary, but maybe `LOCAL` is not needed too, given we don't have `GLOBAL` packaged temporary tables.

src/dsql/parse-conflicts.txt
sim1984: Maybe change the syntax to
```
WITH [{[UNIQUE] [ASC | DESC] INDEX <index_name> (<column_name>)}...]
```

dyemanov: Yes, `[WITH package_ltt_indexes]` or `[USING package_ltt_indexes]` could be a good workaround, IMHO.

asfernandes: > Yes, `[WITH package_ltt_indexes]` or `[USING package_ltt_indexes]` could be a good workaround, IMHO.

This does not resolve the conflicts. It only resolves puting the WITH before each index when there is multiple indexes:

```
%type package_ltt_index(<createRelationNode>)
package_ltt_index($createRelationNode)
	: WITH unique_opt order_direction INDEX valid_symbol_name column_parens
		{
			const auto node = newNode<CreateIndexNode>(QualifiedName(*$5));
			node->unique = $2;
			node->descending = $3;
			node->columns = $6;

			auto clause = newNode<RelationNode::AddPackageLttIndexClause>(node);
			$createRelationNode->clauses.add(clause);
		}
	;
```

IMO it should not be done. We use btyacc to allow such type of controlled conflicts.

src/dsql/parse.y
asfernandes: I understand. That is going to create a narrative problem to fix. DECLARE is redundant: ok. LOCAL is misleading for declarations in header too.

Then this feature is not "Declared LTT" anymore. It's just packaged temporary tables.

When the similar feature exists local to routines, we will need another narrative to join the similar features.

sim1984: The name of this feature is not entirely correct.

Here's how I see it.

LOCAL/GLOBAL are about the lifetime of metadata.
Temporary is about the lifetime of data.

Temporary table types:

* Global Temporary Table - Metadata is persistent, data is temporary. Metadata is stored at the schema level. Data scope is connection or transaction.
* Local Temporary Table - Metadata is temporary, data is temporary. Metadata is stored per connection (in memory). Data scope is connection or transaction.
* Packaged Temporary Table - Metadata is persistent, data is temporary. Metadata is stored at the package level. Data visibility is connection or transaction.
* Declared Table - Metadata is persistent, data is temporary. Data visibility is the current subroutine call. Metadata is visible only within the subroutine.

sim1984:
I didn't see in the documentation whether comments are supported for these tables and their columns?

```
COMMENT ON TABLE [<schema-name> . ] <package-name> . <table-name> IS 'text';
COMMENT ON COLUMN [<schema-name> . ] <package-name> . <table-name> . <column-name> IS 'text';
```

What permissions should a user/role have on a package to perform operations on a public table?  
Is `EXECUTE` sufficient? 
Or is `USAGE` the same as for constants?

asfernandes:
> What permissions should a user/role have on a package to perform operations on a public table?
> Is `EXECUTE` sufficient?
> Or is `USAGE` the same as for constants?

Probably the table ones (SELECT, INSERT, UPDATE, DELETE) would be more correct.

sim1984:
> > What permissions should a user/role have on a package to perform operations on a public table?
> > Is `EXECUTE` sufficient?
> > Or is `USAGE` the same as for constants?
> 
> Probably the table ones (SELECT, INSERT, UPDATE, DELETE) would be more correct.

Okay. And are these particular privileges being verified now?

dyemanov (approved):

PR #8986, opened 2026-04-14
Race condition between VIO_update_in_place and the garbage collector

MochalovAlexey:
Hello,
I have discovered a race condition between the garbage collector and `VIO_update_in_place`, which is invoked in the system transaction during restore.

A race condition occurs when the record has a delta and `VIO_update_in_place` needs to read the back version to unpack it. However, if the garbage collector is active, it may delete that back version before it is fetched for reading, because it is not protected by anything at that point. This leads to `BUGCHECK(291)`.

My proposal is to acquire the write lock on the record before fetching the back version. This would prevent the background garbage collector from deleting its back versions within `VIO_chase_record_version`.

The issue is intermittent and was reproduced only on our fork using the firebird-qa test `./tests/functional/tabloid/test_eqc_136030.py`. I was unable to create a reproducible case for Firebird itself because the background garbage collector is triggered less frequently there. However, even from analysis alone, this potential problem is evident.

Thanks,
Alexey Mochalov.

### DIFF
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -7321,6 +7321,9 @@ void VIO_update_in_place(thread_db* tdbb,
 
 	AutoTempRecord gc_rec;
 
+	if (!DPM_get(tdbb, org_rpb, LCK_write))
+		BUGCHECK(186);	// msg 186 record disappeared
+
 	record_param temp2;
 	const Record* prior = org_rpb->rpb_prior;
 	if (prior)
@@ -7346,9 +7349,6 @@ void VIO_update_in_place(thread_db* tdbb,
 		const USHORT pageSpaceID = temp2.getWindow(tdbb).win_page.getPageSpaceID();
 		stack->push(PageNumber(pageSpaceID, temp2.rpb_page));
 	}
-
-	if (!DPM_get(tdbb, org_rpb, LCK_write))
-		BUGCHECK(186);	// msg 186 record disappeared
 
 	if (prior)
 	{
--- a/src/jrd/vio.cpp
+++ b/src/jrd/vio.cpp
@@ -7330,10 +7330,8 @@ void VIO_update_in_place(thread_db* tdbb,
 	{
 		temp2 = *org_rpb;
 		temp2.rpb_record = gc_rec = relation->getGCRecord(tdbb);
-		temp2.rpb_page = org_rpb->rpb_b_page;
-		temp2.rpb_line = org_rpb->rpb_b_line;
 
-		if (!DPM_fetch(tdbb, &temp2, LCK_read))
+		if (!DPM_fetch_back(tdbb, &temp2, LCK_read, -1))
 			BUGCHECK(291);	 // msg 291 cannot find record back version
 
 		VIO_data(tdbb, &temp2, relation->rel_pool);
@@ -7343,11 +7341,13 @@ void VIO_update_in_place(thread_db* tdbb,
 		if (temp2.rpb_prior)
 			temp2.rpb_flags |= rpb_delta;
 
-		temp2.rpb_number = org_rpb->rpb_number;
 		DPM_store(tdbb, &temp2, *stack, DPM_secondary);
 
 		const USHORT pageSpaceID = temp2.getWindow(tdbb).win_page.getPageSpaceID();
 		stack->push(PageNumber(pageSpaceID, temp2.rpb_page));
+
+		if (!DPM_get(tdbb, org_rpb, LCK_write))
+			BUGCHECK(186);	// msg 186 record disappeared
 	}
 
 	if (prior)


src/jrd/vio.cpp
```diff
@@ -7321,6 +7321,9 @@ void VIO_update_in_place(thread_db* tdbb,
 
 	AutoTempRecord gc_rec;
 
+	if (!DPM_get(tdbb, org_rpb, LCK_write))
+		BUGCHECK(186);	// msg 186 record disappeared
+
```
hvlad, line 7326: Your change makes possible that two pages will be locked at the same time, this is not good.
When one have primary record version locked and need to get seconary version, he should call `DPM_fetch_back()`.
Of course, after `VIO_data()`, primary record should be re-acquired again.

hvlad (changes requested):
I've got this bugcheck during restore also but didn't investigated it yet as it is not reproducible.
Also, interesting - why it happens on v6 only ?

AlexPeshkoff:
> Also, interesting - why it happens on v6 only ?

Guys, can you take a look - what does GCLock do when the bug happens? I've tested it and found no problems but - it was seriously modified...

hvlad:
> > Also, interesting - why it happens on v6 only ?
> 
> Guys, can you take a look - what does GCLock do when the bug happens? I've tested it and found no problems but - it was seriously modified...

I'll try, thanks for the hint

hvlad:
> > > Also, interesting - why it happens on v6 only ?
> > 
> > 
> > Guys, can you take a look - what does GCLock do when the bug happens? I've tested it and found no problems but - it was seriously modified...
> 
> I'll try, thanks for the hint

GCLock doesn't matters here, especially during restore.

hvlad:
@MochalovAlexey do you have call stack when bugcheck happens ? 
Also, do you know which relation was affected ?

MochalovAlexey:
> @MochalovAlexey do you have call stack when bugcheck happens ? Also, do you know which relation was affected ?

Yes, the table affected is `RDB$DATABASE`.
I've attached two stack traces from our fork: one showing a `BUGCHECK(291)` that occurred, and the other from a few milliseconds earlier, at the moment when GC and `VIO_update_in_place` are competing for the same record (see threads 18 and 22). During the race condition, `rpb` in `VIO_chase_record_version` before the `purge` function contains a reference to the same record as `org_rpb` in `VIO_update_in_place`. Purge deletes the backversion with the delta right before `DPM_fetch` is called on that delta in `VIO_update_in_place`.
[vio_update_in_place_bugcheck.txt](https://github.com/user-attachments/files/26710405/vio_update_in_place_bugcheck.txt)
[vio_update_in_place_vs_gc.txt](https://github.com/user-attachments/files/26710406/vio_update_in_place_vs_gc.txt)

I reproduce the stack trace leading up to the BUGCHECK by setting a breakpoint on these two lines:
1) https://github.com/FirebirdSQL/firebird/blob/812500a116f6a8b9eeb862ac9871dfc85f9572bb/src/jrd/vio.cpp#L1690
2) https://github.com/FirebirdSQL/firebird/blob/812500a116f6a8b9eeb862ac9871dfc85f9572bb/src/jrd/vio.cpp#L7330

hvlad:
Thanks, now I think I see why it happens with v6 only.

Before v6, RDB$DATABASE was modified using user transaction, therefore GC see record as "active" and don't attempt to change it. In v6 system transaction is used to modify RDB$DATABASE and concurrent reader see record as "committed" and thus call purge\expunge for it.

I think, decision to use system transaction to modify records is not very good thought. At least from VIO perspective.

AlexPeshkoff:
On 4/14/26 18:40, Vlad Khorsun wrote:
> *hvlad* left a comment (FirebirdSQL/firebird#8986) 
> <https://github.com/FirebirdSQL/firebird/pull/8986#issuecomment-4245215326>
>
> Thanks, now I think I see why it happens with v6 only.
>
> Before v6, RDB$DATABASE was modified using user transaction, therefore 
> GC see record as "active" and don't attempt to change it. In v6 system 
> transaction is used to modify RDB$DATABASE and concurrent reader see 
> record as "committed" and thus call purge\expunge for it.
>
> I think, decision to use system transaction to modify records is not 
> very good thought. At least from VIO perspective.
>

RDB$DATABASE is modified is system transaction in order to make relation 
ID generation better fit new needs. With user transaction it we will be 
for example unable to create 2 independent relations in 2 transactions 
in parallel.
On the other hand (after reviewing code right now) I do not understand 
what prevents from use of DYN_UTIL_gen_unique_id() like we do for 
procedures and functions also for relations? This will solve both 
problems. (With system transaction behavior of generator, needed here, 
was more or less emulated.) The only side effect I see is that field 
RDB$RELATION_ID will remain unused.

MochalovAlexey:
> I think, decision to use system transaction to modify records is not very good thought. At least from VIO perspective.

I tend to agree.

My experiments showed that `VIO_update_in_place` in system transactions is not safe in terms of preserving data consistency for changes committed by user transactions.

For example, if a `CREATE TABLE `statement is running in the system transaction and thread is paused at `DPM_get` inside `VIO_update_in_place`, then a parallel user transaction changes the database default character set, and the system thread is resumed afterwards, the primary record version gets overwritten. As a result, the data committed by the user transaction (`RDB$CHARACTER_SET_NAME`) is lost without any conflict being detected. The probability of such a race seems low in practice, but it can be reproduced under a debugger.

Therefore, the proper fix seems to be to avoid calling `VIO_update_in_place` from `VIO_modify` solely because the transaction has the system flag.

hvlad:
Afraid, there could be more bad cases and change in `VIO_update_in_place` is not correct way to fix the issue.

First, system transaction can't create backversions as its number is always 0 and it is impossible to find correct record version in a chain where all versions labeled by tx0. Thus, system transaction uses update in place for UPDATE and backout for DELETE.

Next, system transaction changes are "immediately committed" thus garbage collector should take it into account and explicitly avoid to handle record chain when see version by tx0. Before v6 there was no such requirement, because of:

there must be strong rule that user and system transactions can't modify the same record.
This is the real reason of the issue and it must be fixed.

hvlad:
>> On 4/14/26 18:40, Vlad Khorsun wrote: *hvlad* left a comment ([FirebirdSQL/firebird#8986](https://github.com/FirebirdSQL/firebird/pull/8986)) <[#8986 (comment)](https://github.com/FirebirdSQL/firebird/pull/8986#issuecomment-4245215326)
>> Thanks, now I think I see why it happens with v6 only. 
>> 
>> Before v6, RDB$DATABASE was modified using user transaction, therefore GC see record as "active" and don't attempt to change it. In v6 system transaction is used to modify RDB$DATABASE and concurrent reader see record as "committed" and thus call purge\expunge for it. 
>> 
>> I think, decision to use system transaction to modify records is not very good thought. At least from VIO perspective.
>
> RDB$DATABASE is modified is system transaction in order to make relation ID generation better fit new needs. With user transaction it we will be for example unable to create 2 independent relations in 2 transactions in parallel. 

We can use separate (autonomous) transaction to modify RDB$DATABASE and immediately commit, in this case

> 
> On the other hand (after reviewing code right now) I do not understand what prevents from use of DYN_UTIL_gen_unique_id() like we do for procedures and functions also for relations? This will solve both problems. (With system transaction behavior of generator, needed here, was more or less emulated.) The only side effect I see is that field RDB$RELATION_ID will remain unused.

AFAIU, change of `RDB$DATABASE` in user transaction allows to automatically undo increment of `RDB$RELATION_ID` in case of rollback. Approach with generator doesn't allow to do it with such easy.
Of course, "modify in separate transaction and commit" also have this drawback.

AlexPeshkoff:
On 4/15/26 14:28, Vlad Khorsun wrote:
> *hvlad* left a comment (FirebirdSQL/firebird#8986) 
> <https://github.com/FirebirdSQL/firebird/pull/8986#issuecomment-4251595193>
>
> Afraid, there could be more bad cases and change in 
> |VIO_update_in_place| is not correct way to fix the issue.
>
> First, system transaction can't create backversions as its number is 
> always 0 and it is impossible to find correct record version in a 
> chain where all versions labeled by tx0. Thus, system transaction uses 
> update in place for UPDATE and backout for DELETE.
>
> Next, system transaction changes are "immediately committed" thus 
> garbage collector should take it into account and explicitly avoid to 
> handle record chain when see version by tx0. Before v6 there was no 
> such requirement, because of:
>
> there must be strong rule that user and system transactions can't 
> modify the same record.
> This is the real reason of the issue and it must be fixed.
>

I agree with Vlad.

AlexPeshkoff:
On 4/15/26 14:36, Vlad Khorsun wrote:
> *hvlad* left a comment (FirebirdSQL/firebird#8986) 
> <https://github.com/FirebirdSQL/firebird/pull/8986#issuecomment-4251641352>
>
>         On 4/14/26 18:40, Vlad Khorsun wrote: /hvlad/ left a comment
>         (FirebirdSQL/firebird#8986
>         <https://github.com/FirebirdSQL/firebird/pull/8986>) <#8986
>         (comment)
>         <https://github.com/FirebirdSQL/firebird/pull/8986#issuecomment-4245215326>
>         Thanks, now I think I see why it happens with v6 only.
>
>         Before v6, RDB$DATABASE was modified using user transaction,
>         therefore GC see record as "active" and don't attempt to
>         change it. In v6 system transaction is used to modify
>         RDB$DATABASE and concurrent reader see record as "committed"
>         and thus call purge\expunge for it.
>
>         I think, decision to use system transaction to modify records
>         is not very good thought. At least from VIO perspective.
>
>     RDB$DATABASE is modified is system transaction in order to make
>     relation ID generation better fit new needs. With user transaction
>     it we will be for example unable to create 2 independent relations
>     in 2 transactions in parallel.
>
> We can use separate (autonomous) transaction to modify RDB$DATABASE 
> and immediately commit, in this case
>

May be this is the best possible solution. In addition it solves 
backward compatibility issue with next RELATION_ID stored in 
rdb$database. If no serious arguments against it I will make a change 
tomorrow.

>     On the other hand (after reviewing code right now) I do not
>     understand what prevents from use of DYN_UTIL_gen_unique_id() like
>     we do for procedures and functions also for relations? This will
>     solve both problems. (With system transaction behavior of
>     generator, needed here, was more or less emulated.) The only side
>     effect I see is that field RDB$RELATION_ID will remain unused.
>
> AFAIU, change of |RDB$DATABASE| in user transaction allows to 
> automatically undo increment of |RDB$RELATION_ID| in case of rollback. 
> Approach with generator doesn't allow to do it with such easy.
> Of course, "modify in separate transaction and commit" also have this 
> drawback.
>

Yes. But I've looked at it as 'nothing bad happens if IDs are not 
exactly sequential'.

hvlad:
>> AFAIU, change of `RDB$DATABASE` in user transaction allows to automatically undo increment of `RDB$RELATION_ID` in case of rollback. Approach with generator doesn't allow to do it with such easy.
>> Of course, "modify in separate transaction and commit" also have this drawback.
>
> Yes. But I've looked at it as 'nothing bad happens if IDs are not exactly sequential'.

Originally, there was no code that allows to reuse unused relation IDs. It was added not so long ago, in 2.5, IIRC
I.e. it is not about strong sequentiality, but more about usage of limited number space.

AlexPeshkoff:
On 4/15/26 16:21, Vlad Khorsun wrote:
> *hvlad* left a comment (FirebirdSQL/firebird#8986) 
> <https://github.com/FirebirdSQL/firebird/pull/8986#issuecomment-4252417438>
>
>         AFAIU, change of |RDB$DATABASE| in user transaction allows to
>         automatically undo increment of |RDB$RELATION_ID| in case of
>         rollback. Approach with generator doesn't allow to do it with
>         such easy.
>         Of course, "modify in separate transaction and commit" also
>         have this drawback.
>
>     Yes. But I've looked at it as 'nothing bad happens if IDs are not
>     exactly sequential'.
>
> Originally, there was no code that allows to reuse unused relation IDs.
>

Not sure about pre-FB, all FB versions are able to search for the holes 
in ID's space. BTW, in FB1 related code is executed in system 
transaction, like all dfw (if GPRE rule regarding transaction used by 
default in engine mode did not change).

> I.e. it is not about strong sequentiality, but more about usage of 
> limited number space.
>

Yes, and this is not a problem here.

hvlad:
>> Originally, there was no code that allows to reuse unused relation IDs.
>
> Not sure about pre-FB, all FB versions are able to search for the holes in ID's space. BTW, in FB1 related code is executed in system transaction, like all dfw (if GPRE rule regarding transaction used by default in engine mode did not change).

Yes, indeed. My memory was wrong here. 

>> I.e. it is not about strong sequentiality, but more about usage of limited number space.
>
> Yes, and this is not a problem here.

Good. One problem less ;)

PR #8990, opened 2026-04-15
New implementation of bulk insert

hvlad:
In Firebird 5 parallel restore was introduced. It contains a "bulk insert" code that allows to lower contention on PP by concurrent writers. Also it makes each writer to use own dedicated DP to fill with records. 

With shared metadata cache in v6 that code became broken and parallel restore creates a lot of unused data pages. Instead of fixing that first attempt to have bulk insert ability, I offer a new approach that fixes the issue, have better performance and could be used more widely.

Note, patch doesn't remove old code, it could be done a bit later - after agreement on the new code.


dyemanov:
Some details about how the new approach is different from the old one would be appreciated.

hvlad:
The insert code path now avoids a lot of not necessary steps, such as triggers, validations (except of NOT NULL on field level) index maintenance - all of this is abcent during restore. Also, all records inserted into dedicated in-memory buffer to avoid endless latches on data page buffers. When in-memory buffer is full, its contents is copied into actual DB buffers and go to the disk in usual way. In-memory buffer size start from 1 page and resized to the 8 pages when first 8 pages of relation is filled. Also, blob contens are put into separate in-memory buffer that works in the same way.

The code is put into two main classes: 
- `BulkInsert` - implements in-memory buffers and code that works with records (mostly analog/copy of some DPM parts) , and 
- `BulkInsertNode` - replaces `StoreNode` and contains some further optimizations, such as pre-calculated target descriptors.

AlexPeshkoff:
On 4/15/26 17:55, Vlad Khorsun wrote:
> *hvlad* left a comment (FirebirdSQL/firebird#8990) 
> <https://github.com/FirebirdSQL/firebird/pull/8990#issuecomment-4253109155>
>
> The insert code path now avoids a lot of not necessary steps, such as 
> triggers, validations (except of NOT NULL on field level) index 
> maintenance - all of this is abcent during restore. Also, all records 
> inserted into dedicated in-memory buffer to avoid endless latches on 
> data page buffers. When in-memory buffer is full, its contents is 
> copied into actual DB buffers and go to the disk in usual way. 
> In-memory buffer size start from 1 page and resized to the 8 pages 
> when first 8 pages of relation is filled. Also, blob contens are put 
> into separate in-memory buffer that works in the same way.
>
> The code is put into two main classes:
>
>   * |BulkInsert| - implements in-memory buffers and code that works
>     with records (mostly analog/copy of some DPM parts) , and
>   * |BulkInsertNode| - replaces |StoreNode| and contains some further
>     optimizations, such as pre-calculated target descriptors.
>

But what about |relPages->rel_*_data_space| stored at database, not 
attachment level? From new code it seems that they should be moved back 
to attachment from database? Am I missing something?

hvlad:
> But what about `relPages->rel_*_data_space` stored at database, not attachment level? From new code it seems that they should be moved back to attachment from database? Am I missing something?

  These fields not used in bulk insert code and thus was not affected in this patch. 

At next step I'm going to:

leave as is:
```
	ULONG rel_index_root;		// index root page number
	USHORT rel_pg_space_id;
```
make atomic:
```
	ULONG rel_data_pages;		// count of relation data pages
	ULONG rel_slot_space;		// lowest pointer page with slot space
	ULONG rel_pri_data_space;	// lowest pointer page with primary data page space
	ULONG rel_sec_data_space;	// lowest pointer page with secondary data page space
```
remove:
```
	ULONG rel_last_free_pri_dp;	// last primary data page found with space
	ULONG rel_last_free_blb_dp;	// last blob data page found with space
```

aafemt:
I think term "bulk insert" is a little misleading here. Oracle uses "direct-path insert" for the path where data go straight into data pages.

hvlad:
@dyemanov: did you have a chance to look at it ?

dyemanov:
> @dyemanov: did you have a chance to look at it ?

Halfway done, hopefully to be completed during the weekend.

src/dsql/StmtNodes.cpp
```diff
@@ -779,6 +780,349 @@ bool BlockNode::testAndFixupError(thread_db* tdbb, Request* request, const Excep
...
 
 
+static RegisterNode<BulkInsertNode> regBulkInsertNode({blr_bulk_insert});
+
+DmlNode* BulkInsertNode::parse(thread_db* tdbb, MemoryPool& pool, CompilerScratch* csb, const UCHAR blrOp)
+{
+	auto node = FB_NEW_POOL(pool) BulkInsertNode(pool);
+
+	if (csb->csb_blr_reader.peekByte() == blr_marks)
+		/*node->marks |= */PAR_marks(csb);
```
dyemanov, line 790: If marks are skipped, why `blr_marks` can be generated after `blr_bulk_insert`? Or is it just reserved for possible future use?

src/dsql/StmtNodes.cpp
```diff
@@ -779,6 +780,349 @@ bool BlockNode::testAndFixupError(thread_db* tdbb, Request* request, const Excep
...
+
+	preprocessAssignments(tdbb, csb, target->getStream(), nodeAs<CompoundStmtNode>(statement), nullptr);
+	doPass1(tdbb, csb, statement.getAddress());
+
+	const StreamType stream = target->getStream();
+	auto& tail = csb->csb_rpt[stream];
+	tail.csb_flags |= csb_store;
+
+	jrd_rel* const relation = tail.csb_relation(tdbb);
+	if (relation)
```
dyemanov: How could a relation be missing here, given that the target stream is always `RelationSourceNode` or `LocalTableSourceNode`? Maybe an assert instead? Or at least add an `fb_assert(false)` to the `else` part?

src/dsql/StmtNodes.cpp
```diff
@@ -779,6 +780,349 @@ bool BlockNode::testAndFixupError(thread_db* tdbb, Request* request, const Excep
...
+	});
+
+	while (cursor->fetchNext(tdbb))
+	{
+		rpb->rpb_number.setValue(BOF_NUMBER);
+		record->nullify();
+		record->setTransactionNumber(transaction->tra_number);
+
+		assignValues(tdbb, request, relation, record, toDescs.begin());
+		cleanupRpb(tdbb, rpb);
```
dyemanov: I suppose `cleanupRpb()` makes no sense after `record->nullify()` which not only set NULL flags but also zeroes the record data.

src/dsql/StmtNodes.cpp
```diff
@@ -779,6 +780,349 @@ bool BlockNode::testAndFixupError(thread_db* tdbb, Request* request, const Excep
...
+
+		bulk->putRecord(tdbb, rpb, transaction);
+		REPL_store(tdbb, rpb, transaction);
+
+		JRD_reschedule(tdbb);
```
dyemanov: `JRD_reschedule()` is usually called per every record inside `cursor->fetchNext() -> RecordSource::internalGetRecord()`, is it really necessary calling it again after store?

src/dsql/StmtNodes.cpp
```diff
@@ -779,6 +780,349 @@ bool BlockNode::testAndFixupError(thread_db* tdbb, Request* request, const Excep
...
+		const auto count = compound->statements.getCount();
+
+		impure->flags = INIT_DONE;
+		impure->descs = FB_NEW_POOL(*request->req_pool) Array<dsc>(*request->req_pool, count);
+
+		prepareTarget(tdbb, request, impure->descs->getBuffer(count));
+	}
+	auto bulk = transaction->getBulkInsert(tdbb, relation, true);
+	assignValues(tdbb, request, relation, record, impure->descs->begin());
+	cleanupRpb(tdbb, rpb);
```
dyemanov: Same as above re. `cleanupRpb()`

src/dsql/StmtNodes.cpp
```diff
@@ -779,6 +780,349 @@ bool BlockNode::testAndFixupError(thread_db* tdbb, Request* request, const Excep
...
+					name = relField->fld_name.toQuotedString();
+
+				if (name.isEmpty())
+					name = UNKNOWN_STRING_MARK;
+
+				// validation error for column @1, value \"@2\"
+				ERR_post(Arg::Gds(isc_not_valid) << Arg::Str(name) << Arg::Str(NULL_STRING_MARK));
+			}
+
+			record->setNull(toField->fieldId);
```
dyemanov: `assignValues()` seems to be always called after `record->nullify()`, so all fields are already initialized to NULLs and it shouldn't be necessary to do that again.

src/dsql/StmtNodes.cpp
```diff
@@ -779,6 +780,349 @@ bool BlockNode::testAndFixupError(thread_db* tdbb, Request* request, const Excep
...
+
+	rpb->rpb_address = record->getData();
+	rpb->rpb_length = format->fmt_length;
+	rpb->rpb_format_number = format->fmt_version;
+	rpb->rpb_number.setValue(BOF_NUMBER);
+
+	record->nullify();
+	record->setTransactionNumber(transaction->tra_number);
+
+	if (!(impure->flags & INIT_DONE))
```
dyemanov: It looks like you could check for `!impure->descs` and thus avoid `impure->flags` at all?

src/jrd/ods.h
```diff
@@ -845,6 +845,7 @@ inline constexpr UCHAR ppg_dp_large			= 0x02;		// Large object is on data page
 inline constexpr UCHAR ppg_dp_swept			= 0x04;		// Sweep has nothing to do on data page
 inline constexpr UCHAR ppg_dp_secondary		= 0x08;		// Primary record versions not stored on data page
 inline constexpr UCHAR ppg_dp_empty			= 0x10;		// Data page is empty
+inline constexpr UCHAR ppg_dp_reserved		= 0x20;		// Slot is reserved for bulk insert
```
dyemanov, line 848: If the slot was reserved, but the server crashed before the slot has been actually used, can it be reused for regular (non-bulk) operations? If not, shouldn't `gfix` clear this flag if such a case is detected?

src/jrd/BulkInsert.cpp
```diff
@@ -0,0 +1,493 @@
...
+		CCH_MARK(tdbb, &dpWindow);
+
+		// don't overwrite pag_scn
+		m_current->dpg_header.pag_scn = dpage->pag_scn;
+		memcpy(dpage, m_current, m_pageSize);
+
+		CCH_RELEASE(tdbb, &dpWindow);
+
+		if (m_isPrimary)
+			tdbb->bumpStats(RecordStatType::INSERTS, m_relation->getId(), m_current->dpg_count);
```
dyemanov, line 436: Why is this record-level counter increased per page rather than per record (inside `putRecord()`, I guess)?

dyemanov (changes requested):
The only thing (besides my other comments) that I don't really like is a code duplication between dpm.epp and BulkInsert.cpp. For example, storage of a big record seems to share about 80% of code. Do you think it could be improved (i.e. by adding separate routines to dpm.epp that deal with a preallocated page buffer)?

src/dsql/StmtNodes.cpp
hvlad: Yes, just in case it will be used later

src/dsql/StmtNodes.cpp
hvlad: Will add assert, thanks.

src/dsql/StmtNodes.cpp
hvlad: Agree, will remove it.

src/dsql/StmtNodes.cpp
hvlad: Ok, will remove

src/dsql/StmtNodes.cpp
hvlad: Ok.

src/dsql/StmtNodes.cpp
hvlad: Ok.

src/jrd/ods.h
hvlad: > If the slot was reserved, but the server crashed before the slot has been actually used, can it be reused for regular (non-bulk) operations? 

No, it can't

> If not, shouldn't `gfix` clear this flag if such a case is detected?

Yes, and validation always clears it, see:
- `Validation::walk_data_page()` never set `ppg_dp_reserved` at `pp_bits`
- thus, `Validation::walk_pointer_page()` always clear this bit

src/jrd/BulkInsert.cpp
hvlad: Until `flush()` records actually not yet inserted into relation. 
Also, this is a kind of small optimization.

hvlad:
> The only thing (besides my other comments) that I don't really like is a code duplication between dpm.epp and BulkInsert.cpp. For example, storage of a big record seems to share about 80% of code. 

Yes, this is not an easy question and I considered it, of course. The current code is best compromiіe I can think of.
I think the amount of duplicated code is relatively small.
Do you have clear idea how to not duplicate code and not create an additional branches in existing one ?

> Do you think it could be improved (i.e. by adding separate routines to dpm.epp that deal with a preallocated page buffer)?

You mean - move (half of) BulkInsert into DPM ? :)

hvlad:
I'll implement suggested changes and re-test, then commit its.

aafemt:
Will this feature be documented for usage or remain internal only?

src/dsql/StmtNodes.cpp
hvlad: Done

hvlad:
About cleanup of old "bulk insert" code, opinions, please: to do it in this PR or later, after the merge ?

dyemanov:
> About cleanup of old "bulk insert" code, opinions, please: to do it in this PR or later, after the merge ?

I'd say we need support for bulk insert into indices, so that the new implementation could be utilized by the Batch API, for example. Only then the old code should be wiped out.

hvlad:
> > About cleanup of old "bulk insert" code, opinions, please: to do it in this PR or later, after the merge ?
> 
> I'd say we need support for bulk insert into indices, so that the new implementation could be utilized by the Batch API, for example. Only then the old code should be wiped out.

Old code have no relation with indices. I don't get your point here.

dyemanov:
> Old code have no relation with indices. I don't get your point here.

But it allows Batch API to work faster than regular inserts, right? I mean the logic behind `RPB_s_bulk` which affects `DPM`. If it's removed, Batch API gonna become slower unless it could use the new implementation instead. But it cannot because the new implementation does not support indices. Do I miss anything?

hvlad:
> > Old code have no relation with indices. I don't get your point here.
> 
> But it allows Batch API to work faster than regular inserts, right? 

No.

> I mean the logic behind `RPB_s_bulk` which affects `DPM`. If it's removed, Batch API gonna become slower unless it could use the new implementation instead. But it cannot because the new implementation does not support indices. Do I miss anything?

Old code was introduced to make parallel inserts faster. Single thread inserts was not affected.
Also, old code doens't work currently.

dyemanov:
Then I'd perform the cleanup in this PR.

dyemanov (approved):

hvlad:
> > But what about `relPages->rel_*_data_space` stored at database, not attachment level? From new code it seems that they should be moved back to attachment from database? Am I missing something?
> 
> These fields not used in bulk insert code and thus was not affected in this patch.
> 
> At next step I'm going to:

Not in this PR, sorry.

PR #8991, opened 2026-04-16
Fix a possible use-after-free in EDS transaction rollback

MochalovAlexey:
`EDS::Transaction::rollback()` used to call `m_connection.deleteTransaction()` before reporting rollback errors. This cleanup may release or delete the owning `EDS::Connection`.

If rollback returned an error, the code then called `conn.raise()`, which uses `getDataSourceName()` and reads `m_dbName` from the potentially deleted connection. This could lead to a crash while building the error message.

The fix raises the rollback error while the connection is still alive and uses a catch/rethrow block to guarantee external transaction cleanup before propagating the original exception.

### DIFF
--- a/src/jrd/extds/ExtDS.cpp
+++ b/src/jrd/extds/ExtDS.cpp
@@ -1639,15 +1639,30 @@ void Transaction::rollback(thread_db* tdbb, bool retain)
 	doRollback(&status, tdbb, retain);
 
 	Connection& conn = m_connection;
-	if (!retain)
+	const bool hasErrors = status->getState() & IStatus::STATE_ERRORS;
+
+	const auto cleanup = [&]()
 	{
-		detachFromJrdTran();
-		m_connection.deleteTransaction(tdbb, this);
+		if (!retain)
+		{
+			detachFromJrdTran();
+			m_connection.deleteTransaction(tdbb, this);
+		}
+	};
+
+	try
+	{
+		if (hasErrors) {
+			conn.raise(&status, tdbb, "transaction rollback");
+		}
+	}
+	catch (const Exception&)
+	{
+		cleanup();
+		throw;
 	}
 
-	if (status->getState() & IStatus::STATE_ERRORS) {
-		conn.raise(&status, tdbb, "transaction rollback");
-	}
+	cleanup();
 }
 
 Transaction* Transaction::getTransaction(thread_db* tdbb, Connection* conn, TraScope tra_scope)


aafemt:
I would suggest to use `Cleanup` class instead:
```
Cleanup cleanup = [&]()
{
	if (!retain)
	{
		detachFromJrdTran();
		m_connection.deleteTransaction(tdbb, this);
	}
};
```

src/jrd/extds/ExtDS.cpp
```diff
@@ -1639,15 +1639,30 @@ void Transaction::rollback(thread_db* tdbb, bool retain)
...
 	{
-		detachFromJrdTran();
-		m_connection.deleteTransaction(tdbb, this);
-	}
+		if (!retain)
+		{
+			detachFromJrdTran();
+			m_connection.deleteTransaction(tdbb, this);
+		}
+	};
```
hvlad, line 1651: Why not use `class Cleanup` here and avoid re-throw and explicit cleanup ?

MochalovAlexey:
> I would suggest to use `Cleanup` class instead:
> 
> ```
> Cleanup cleanup = [&]()
> {
> 	if (!retain)
> 	{
> 		detachFromJrdTran();
> 		m_connection.deleteTransaction(tdbb, this);
> 	}
> };
> ```

That would be risky here because the cleanup destructor may throw from deleteTransaction().
If that happens while another exception is already being unwound, C++ will call std::terminate().

src/jrd/extds/ExtDS.cpp
MochalovAlexey: Answered below https://github.com/FirebirdSQL/firebird/pull/8991#issuecomment-4259420330

aafemt:
> That would be risky here because the cleanup destructor may throw from deleteTransaction().

Because it is a cleanup routine, I would make sure that it cannot throw. Throwing from `rollback()` can create an unrecoverable state of code.

dyemanov:
Strictly speaking Alexey is correct. I'd say all usages of `class Cleanup` should be revisited, because `CCH_release`, `TRA_rollback`, whatever -- all they can throw (even if unlikely). Moreover, our code in /common/classes often throws from RAII destructors (usually this is `system_call_failed`). The question is whether we're going to solve this problem globally before v6 is out.

aafemt:
> I'd say all usages of `class Cleanup` should be revisited,

They say clang-tidy can track cases when noexcept functions call non-noexcept functions. May be it can handle destructors and lambdas. In this case it could be enough to mark `Cleanup` parameter as `noexcept`.

AlexPeshkoff:
On 4/16/26 20:38, Dmitry Yemanov wrote:
> *dyemanov* left a comment (FirebirdSQL/firebird#8991) 
> <https://github.com/FirebirdSQL/firebird/pull/8991#issuecomment-4262183825>
>
> Strictly speaking Alexey is correct. I'd say all usages of |class 
> Cleanup| should be revisited, because |CCH_release|, |TRA_rollback|, 
> whatever -- all they can throw (even if unlikely). Moreover, our code 
> in /common/classes often throws from RAII destructors (usually this is 
> |system_call_failed|). The question is whether we're going to solve 
> this problem globally before v6 is out.
>

Yes, Cleanup is very small part of bigger problem. We should solve the 
problem in general. May be ideal way is to merge exceptions according to 
ERR_punt() rules but right now I see no easy way to do it, 
std::uncaught_exceptions() returns only unwinding state and anyway we 
can't modify active exception. In absence of better solutions we may 
silently ignore errors in dtor, in most cases they are caused by 
problems related with initial exception.

aafemt:
> Yes, Cleanup is very small part of bigger problem. We should solve the problem in general.

Still elephants can be eaten by pieces. We can add `static_assert(std::is_nothrow_invocable<F>::value == true, "Cleanup function must not throw");` to `Cleanup` right now and let students with static analyzer to track full stack of called functions later.

dyemanov (approved):

hvlad (approved):

hvlad:
I'll backport it into older versions

PR #8994, opened 2026-04-20
Fixes a gbak restore crash for backups created with `-e`

MochalovAlexey:
The bug was reproduced on a real production database backup (5.0 ver.). The restore reader now reserves one extra byte in `IOBuffer` for the next `rec_*` marker after a data record. A record could exactly fill the buffer and trip `fb_assert(space > 0)` before writing `rec_data` or `rec_relation_end`.


hvlad (approved):
Good catch, thanks

PR #9002, opened 2026-04-22
Use correct input type for example UDR multi

Ev3nt:
### Summary
Invalid types conversion.

### Details
When calling ExampleBCD::load, we pass a pointer to the FB_DEC34 data type, but in fact the memory referenced by the pointer is an ordinary int. This causes the crash.

### PoC
A simple script that causes an error:
```py
import os
import subprocess
import sys
from pathlib import Path

ISQL = Path("/home/user/repositories/firebird-6/gen/Debug/firebird/bin/isql")

def build_sql() -> str:
  return f"""
      connect localhost:employee user 'SYSDBA' password 'masterkey';
      set term ^;
      create function mult(a integer, b integer)
      returns integer
      external name 'udrcpp_example!mult'
      engine udr^
      set term ;^
      commit;
      select mult(2.49, -8) from rdb$database;
      quit;
      """

def run_isql(sql: str) -> int:
  proc = subprocess.run(
    [str(ISQL), "-q"],
    input=sql.encode()
  )
  print("returncode:", proc.returncode)
  if proc.returncode < 0:
    print(f"terminated by signal: {-proc.returncode}")
  return proc.returncode

def main():
  run_isql(build_sql())

if __name__ == "__main__":
  main()

### DIFF
--- a/examples/udr/Functions.cpp
+++ b/examples/udr/Functions.cpp
@@ -173,7 +173,8 @@ FB_UDR_BEGIN_FUNCTION(mult)
 		outOffset = outMetadata->getOffset(status, 0);
 		outNullOffset = outMetadata->getNullOffset(status, 0);
 
-		df34 = master->getUtilInterface()->getDecFloat34(status);
+		util = master->getUtilInterface();
+		df34 = util->getDecFloat34(status);
 	}
 
 	// This function requires the INTEGER parameters and return value, otherwise it will crash.
@@ -185,20 +186,26 @@ FB_UDR_BEGIN_FUNCTION(mult)
 			unsigned char bcd[IDecFloat34::BCD_SIZE];
 			int sign, exp;
 
-			void load(void* from, IDecFloat34* df34)
+			void load(ThrowStatusWrapper* status, int* from, IUtil* util, IDecFloat34* df34)
 			{
-				df34->toBcd((FB_DEC34*) from, &sign, bcd, &exp);
+				FB_DEC34 temp;
+				util->convert(status, SQL_LONG,  0, sizeof(int)     , from,
+									  SQL_DEC34, 0, sizeof(FB_DEC34), &temp);
+				df34->toBcd(&temp, &sign, bcd, &exp);
 			}
 
-			void store(void* to, IDecFloat34* df34) const
+			void store(ThrowStatusWrapper* status, int* to, IUtil* util, IDecFloat34* df34) const
 			{
-				df34->fromBcd(sign, bcd, exp, (FB_DEC34*) to);
+				FB_DEC34 temp;
+				df34->fromBcd(sign, bcd, exp, &temp);
+				util->convert(status, SQL_DEC34, 0, sizeof(FB_DEC34), &temp,
+									  SQL_LONG,  0, sizeof(int)     , to);
 			}
 		};
 
 		ExampleBCD a, b, rc;
-		a.load(in + aOffset, df34);
-		b.load(in + bOffset, df34);
+		a.load(status, reinterpret_cast<int*>(in + aOffset), util, df34);
+		b.load(status, reinterpret_cast<int*>(in + bOffset), util, df34);
 
 		// multiply (trivial example - a lot of features are missing)
 		rc.sign = a.sign ^ b.sign;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by asfernandes** [comment]
The example has clear comment:
```
create function mult (
    a decfloat(34) not null,
    b decfloat(34) not null
) returns decfloat(34) not null
    external name 'udrcpp_example!mult'
    engine udr;
```

And you declare the parameters using integer?

Of course it will crash.

Your bug is invalid as well the correction.

Functions that code may work with different declarations should be done different, for example wait_event with its FB_UDR_MESSAGE.

**Comment by dyemanov** [comment]
I'd argue that an ability to crash the server using plain SQL is not something to be considered "by design".

**Comment by aafemt** [comment]
IIRC in Berlin Alex didn't answer the question "why UDR cannot provide own definition itself", so this is indeed a designed vulnerability.

**Comment by Ev3nt** [comment]
As I understand it, this is a user issue. In this case we can use `FB_UDR_MESSAGE`, but specifically in this example there is a deliberate demonstration without it.

**Comment by asfernandes** [comment]
> I'd argue that an ability to crash the server using plain SQL is not something to be considered "by design".

So you are against any time of native code integration with Firebird?

**Comment by dyemanov** [comment]
> > I'd argue that an ability to crash the server using plain SQL is not something to be considered "by design".
> 
> So you are against any time of native code integration with Firebird?

The crash due to bug inside the native code is OK, unless this is *our* code. If we provide examples that can be used to attack the server, then maybe this is not so clever idea.

**Comment by hvlad** [comment]
I'm confused:

> The example has clear comment:
> 
> ```
> create function mult (
>     a decfloat(34) not null,
>     b decfloat(34) not null
> ) returns decfloat(34) not null
>     external name 'udrcpp_example!mult'
>     engine udr;
> ```
> 
> And you declare the parameters using integer?
> 
> Of course it will crash.

But there is another clear comment:

```
// This function requires the INTEGER parameters and return value, otherwise it will crash.
```

so, what is correct - SQL declaration or second comment ? I suppose - SQL declaration.
Also, that comment contunies with 

```
// Metadata is inspected dynamically (in execute). This is not the fastest method.
```

but I failed to see where metadata is inspected. Am I wrong ?
If not - may be the example could be improved by checking data types passed in (if ever possible, I didn't look deeper) ?

dyemanov:
> If not - may be the example could be improved by checking data types passed in (if ever possible, I didn't look deeper) ?

AFAIU, this is what this PR was about -- ensure the passed value is casted to the data type expected by the UDR implementation. So the function could work even if declared incorrectly.

hvlad:
> > If not - may be the example could be improved by checking data types passed in (if ever possible, I didn't look deeper) ?
> 
> AFAIU, this is what this PR was about -- ensure the passed value is casted to the data type expected by the UDR implementation. So the function could work even if declared incorrectly.

Would this patch work if function were declared with VARCHAR args ? I doubt it.

asfernandes:
1) This is example.
2) There is others examples with another safer approach.
3) It's perfect valid to have simple function that crashes when used incorrectly. UDR is a binary that should be put in the server. If you put a virus in your server, it will also have problems.
4) This example had a copied unmmodified comment from the other example, and of course, when it say INTEGER it means DECFLOAT(34).
5) There isn't (apparently) any problem in the *code* of that example.

AlexPeshkoff:
On 4/28/26 03:10, Adriano dos Santos Fernandes wrote:
> *asfernandes* left a comment (FirebirdSQL/firebird#9002) 
> <https://github.com/FirebirdSQL/firebird/pull/9002#issuecomment-4331380507>
>
>  1. This is example.
>  2. There is others examples with another safer approach.
>  3. It's perfect valid to have simple function that crashes when used
>     incorrectly. UDR is a binary that should be put in the server. If
>     you put a virus in your server, it will also have problems.
>  4. This example had a copied unmmodified comment from the other
>     example, and of course, when it say INTEGER it means DECFLOAT(34).
>

Fixed my old copy/paste bug.

>  1. There isn't (apparently) any problem in the /code/ of that example.
>
>

Agree with all said here. Can add that example was oriented on use of 
DECFLOAT(34) <=> BCD conversions.

May be there was invalid text in docs - it's possible to have UDR that 
will not crash (unlike UDF) when declared incorrectly, but certainly 
it's quite possible to write code that will crash in that case. At least 
it's just c++ :-)

PR #9005, opened 2026-04-23
Add missing pointer deletion in AutoPtr move assignment operator

Noremos:


sim1984:
Move assignment can be implemented more simply:

```cpp
		AutoPtr& operator=(AutoPtr&& r) noexcept
		{
			std::swap(ptr, r.ptr);

			return *this;
		}
```

In this case, you don't need to call any cleanup functions. The destructor of the moved object will do the cleanup. You can also safely remove the self-assignment check.

aafemt:
Is self-swap a defined behavior? I don't see mention of it on https://en.cppreference.com/cpp/algorithm/swap

The problem may be that lifetime of the old content become unpredictable in common case. Moving from global static object will let it live forever that may be unexpected. Of course, in most cases the target pointer is empty, that's why this code lived for years unnoticed.

sim1984:
> Is self-swap a defined behavior? I don't see mention of it on https://en.cppreference.com/cpp/algorithm/swap
> 

For trivial types, which pointers are, yes.

> The problem may be that lifetime of the old content become unpredictable in common case. Moving from global static object will let it live forever that may be unexpected. Of course, in most cases the target pointer is empty, that's why this code lived for years unnoticed.

Do you seriously think that global static objects will be wrapped in `AutoPtr`? And even if they are, who would think of moving them?

hvlad:
> Of course, in most cases the target pointer is empty, that's why this code lived for years unnoticed.

The real reason is that this code never used. 
I've just commented out both move ctor and move assignment operator and succesfully build whole solution.

hvlad:
> Move assignment can be implemented more simply:

Don't replace move assigment by swap. This is different things for different purposes.

Noremos:
> The real reason is that this code never used. I've just commented out both move ctor and move assignment operator and succesfully build whole solution.


I use this in my JSON implementation. And that's how I found the bug.

dyemanov (approved):

PR #9013, opened 2026-04-28
Buffer limit for example UDR replicate

Ev3nt:
### Summary
Buffer with a static size of 65536 may be overflowing.

### Details
The `replicate` implementation uses unsafe `strcat` and `snprintf` because there are no size checks. If the buffer overflows, a crash occurs.

### DIFF
--- a/examples/udr/Triggers.cpp
+++ b/examples/udr/Triggers.cpp
@@ -21,9 +21,69 @@
  */
 
 #include "UdrCppExample.h"
+#include <stdarg.h>
 
 using namespace Firebird;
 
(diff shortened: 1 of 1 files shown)


AlexPeshkoff:
Must say that in EXAMPLE, which is targeted to show users how to work with particular firebird feature, adding so much implementation details to avoid BOF is (on my mind) not good idea.

asfernandes:
Agree with Alex.
I could see some value in using std::string or std::vector and avoid Firebird common classes or too much implementation for what the standard already provides.

PR #9019, opened 2026-05-06
stack-buffer-overflow in decNumber

ChudaykinAlex:
## decNumber — follow-up fuzzing fixes (OOB reads + UB), part 1: patch

> **This is an update** to the previous decNumber fix (the three
> `stack-buffer-overflow` cases in `decFloatAdd` / `decFinalize`). It extends that
> work with a new batch of defects found by continued ASan+UBSan fuzzing of the same
> routines. The earlier patch stays in place; this PR only adds the fixes below.

### Background

Follow-up fuzz testing of the decimal floating-point routines in the decNumber
library — this time with **ASan and UBSan** together — surfaced a further batch of
defects beyond the three `stack-buffer-overflow` cases fixed earlier. They fall in
two classes:

- **Out-of-bounds stack reads** (`stack-buffer-overflow`, READ of size 4) in the
  NaN-payload comparison and in the FMA leading-zero strip.
- **Undefined behaviour** — signed integer overflow and signed left shift — while
  parsing over-long exponents and in the power/ln bit-walk loops.

Eight distinct defects are addressed here; a ninth reproduced scenario
(`decQuadAdd`, `decBasic.c:1401`) is already fixed by the previous PR and is kept
only as a regression case.

Potentially on versions 3, 4, 5, and master.

---

### Affected files

| File | Line | Function | Class |
|---|---|---|---|
| `lib/decBasic.c` | 1190 | `decFloatAdd` / `decQuadAdd` | signed left shift (UB) |
| `lib/decBasic.c` | 1762 | `decQuadCompareTotal` | OOB stack read |
| `lib/decBasic.c` | 2245 (+1536, 2080, 2082) | `decQuadFMA` / strip-leading-zeros | OOB stack read |
| `lib/decCommon.c` | 845 | `decFloatFromString` / `decQuadFromString` | signed int overflow |
| `lib/decNumber.c` | 598 | `decNumberFromString` | signed overflow / left shift (UB) |
| `lib/decNumber.c` | 2202 | `decNumberPower` | signed left shift (UB) |
| `lib/decNumber.c` | 5469 | `decNumberLn` | signed left shift (UB) |

---

### How to reproduce

The reproducer `new_poc.c` is a deterministic dispatcher: `seed_bugN[0]` selects the
rounding mode, `seed_bugN[1] % 16` selects the operation, and the remaining bytes are
split into two strings fed to the `decQuad*` / `decNumber*` functions.

**GCC**

```bash
gcc -g -O0 -fsanitize=address,undefined -fno-sanitize-recover=undefined -I./ \
    new_poc.c decContext.c decNumber.c decDouble.c decQuad.c decPacked.c -o poc_runner
./poc_runner 1   # 1..9
```

**Clang**

```bash
clang -g -O0 -fsanitize=address,undefined -fno-sanitize-recover=undefined -I./ \
    new_poc.c decContext.c decNumber.c decDouble.c decQuad.c decPacked.c -o poc_runner
./poc_runner 1   # 1..9
```

All cases fire consistently on the vulnerable code under ASan/UBSan with both compilers.

**Previous reproducer (`poc.c`, regression check)**

Since this is an update, the earlier three cases (`decQuadAdd`, `decDoubleReduce`,
`decQuadReduce`) should still pass. Build and run the original PoC the same way:

```bash
gcc -g -O0 -fsanitize=address,undefined -fno-sanitize-recover=undefined -I./ \
    decContext.c decNumber.c decDouble.c decQuad.c \
    decimal32.c decimal64.c decimal128.c decPacked.c poc.c -o poc_runner
./poc_runner 1   # 1..3

# clang: replace gcc with clang, same flags
```
[new_poc.c](https://github.com/user-attachments/files/29883530/new_poc.c)
[poc.c](https://github.com/user-attachments/files/29883532/poc.c)

---

### Sanitizer output

<details>
<summary>Bug #1 — decQuadAdd (UBSan)</summary>

```
decBasic.c:1190:30: runtime error: left shift of 25 by 30 places cannot be represented in type 'int'
```
</details>

<details>
<summary>Bug #3 — decQuadCompareTotal (ASan)</summary>

```
ERROR: AddressSanitizer: stack-buffer-overflow ... READ of size 4
    #0 decQuadCompareTotal   decBasic.c:1762
[64, 102) 'bufl' (line 1745) <== Memory access at offset 99 partially overflows this variable
```
</details>

<details>
<summary>Bug #4 — decQuadFMA (ASan)</summary>

```
ERROR: AddressSanitizer: stack-buffer-overflow ... READ of size 4
    #0 decQuadFMA   decBasic.c:2245
[304, 416) 'acc' (line 2004) <== Memory access at offset 413 partially overflows this variable
```
</details>

<details>
<summary>Bug #5 — decQuadFromString (UBSan)</summary>

```
decCommon.c:845:20: runtime error: signed integer overflow: 999999999 * 10 cannot be represented in type 'int'
```
</details>

<details>
<summary>Bug #6 / #7 — decNumberFromString (UBSan)</summary>

```
decNumber.c:598:18: runtime error: left shift of 999999999 by 3 places cannot be represented in type 'int'
decNumber.c:598:18: runtime error: signed integer overflow: 444444444 + 1777777776 cannot be represented in type 'int'
```
</details>

<details>
<summary>Bug #8 — decNumberPower (UBSan)</summary>

```
decNumber.c:2202:12: runtime error: left shift of 1660944384 by 1 places cannot be represented in type 'int'
```
</details>

<details>
<summary>Bug #9 — decNumberLn (UBSan)</summary>

```
decNumber.c:5469:12: runtime error: left shift of 1342177280 by 1 places cannot be represented in type 'int'
```
</details>

---

### Changes

**`lib/decBasic.c` — bug #1** — `decFloatAdd`

`BIN2DPD[]` is `uShort`, promoted to signed `int`; a DPD value (up to 12 bits) shifted
by 20/26/28/30 places reaches the sign bit — UB. `encode` is `uInt` and the high-bit
truncation is intended, so shift in unsigned arithmetic:

```c
// before
encode|=BIN2DPD[tac[2]]<<20;
encode|=BIN2DPD[tac[3]]<<30;
encode|=BIN2DPD[tac[6]]<<28;
encode|=BIN2DPD[tac[9]]<<26;
// after
encode|=(uInt)BIN2DPD[tac[2]]<<20;
encode|=(uInt)BIN2DPD[tac[3]]<<30;
encode|=(uInt)BIN2DPD[tac[6]]<<28;
encode|=(uInt)BIN2DPD[tac[9]]<<26;
```

**`lib/decBasic.c` — bug #3** — `decQuadCompareTotal`

When the byte-wise inner loop found a differing NaN-payload byte it broke only out of
the inner loop; the outer 4-byte loop then continued from a misaligned pointer and read
past `bufl`. It was also a logic bug (the winner must be the first/most-significant
difference). Break out of the outer loop once a winner is found:

```c
    for (;; ub++, uc++) {
      if (*ub==*uc) continue;
      if (*ub>*uc) comp=sigl;
       else comp=-sigl;
       break;
      }
    break;                               // winner found; stop scanning
    }
```

**`lib/decBasic.c` — bug #4** — strip-leading-zeros loops (lines 1536, 2080, 2082, 2245)

`&&` evaluates the 4-byte `UBTOUI()` read before the bounds check, reading past the
coefficient near its end. Test the bound first (matching `decCommon.c:267`):

```c
// before
for (; UBTOUI(x)==0 && x+3<lsd;) x+=4;
// after
for (; x+3<lsd && UBTOUI(x)==0;) x+=4;
```

**`lib/decCommon.c` — bug #5** — `decFloatFromString`

The "exponent too long" clip runs after the accumulation loop, so `exp*10` overflows
`int` inside the loop. Accumulate unsigned (the over-large value is still clipped):

```c
// before
exp=exp*10+edig;
// after
exp=(Int)((uInt)exp*10+edig);
```

**`lib/decNumber.c` — bugs #6, #7** — `decNumberFromString`

`X10(exponent)` on a long exponent both left-shifts into the sign bit and overflows the
add. Legitimate 10-digit exponents (up to 1999999999) must be preserved, so compute
`X10` unsigned; genuinely oversized exponents are still cut by the existing length/first
-digit clip:

```c
// before
exponent=X10(exponent)+(Int)*c-(Int)'0';
// after
exponent=(Int)(X10((uInt)exponent)+(uInt)((Int)*c-(Int)'0'));
```

**`lib/decNumber.c` — bugs #8, #9** — `decNumberPower` (2202) and `decNumberLn` (5469)

The exponent bit-walk shifts a signed `Int` left while testing its sign — UB when a bit
reaches the sign position. Shift in unsigned and reinterpret (the `n<0` sign test is
preserved):

```c
// before
n=n<<1;
// after
n=(Int)((uInt)n<<1);
```

---

### Verification

Rebuilt `poc_runner` with the same command and ran cases 1–9 under ASan+UBSan
(`-fno-sanitize-recover=undefined`):

```
=== BUG 1 === clean        === BUG 6 === clean
=== BUG 2 === clean        === BUG 7 === clean
=== BUG 3 === clean        === BUG 8 === clean
=== BUG 4 === clean        === BUG 9 === clean
=== BUG 5 === clean
```

All nine exit cleanly with no ASan/UBSan diagnostic, on both clang and gcc. The previous
reproducer `poc.c` (cases 1–3) also stays clean, confirming no regression of the earlier
fix:

```
=== poc.c CASE 1 === clean
=== poc.c CASE 2 === clean
=== poc.c CASE 3 === clean
```

Functional sanity checks pass unchanged (`decQuadAdd`, `decQuadCompareTotal`,
`decQuadFMA`, `decNumberFromString`, `decNumberPower`, `decNumberLn`), confirming the
fixes preserve arithmetic behaviour.


AlexPeshkoff (approved):

ChudaykinAlex:
@AlexPeshkoff Good day.  Should I write a unit test?

AlexPeshkoff:
> Should I write a unit test?

If you wish - please do. On the one hand result appears to be obvious, but additional test makes no harm.

ChudaykinAlex:
Question: How do I run the decNumber regression tests so that they actually catch bugs?

I added regression test cases for the decNumber defect batch in src/common/tests/DecNumberTest.cpp (on POSIX, these are automatically picked up by the common_test target; on MSVC, you need to manually add them to common_test.vcxproj). The problem is that most of these tests only catch regressions when built with sanitizers, and I don’t see any options for such a build in the project tree.

The defects fall into two categories:

- Out-of-bounds stack reads (decQuadCompareTotal, decQuadFMA) — these consistently fail only under ASan. Without it, they usually read a few bytes beyond the buffer boundary and continue running silently.
- Undefined behavior—signed integer overflow / signed left shift (decFloatAdd, decFloatFromString, decNumberFromString, decNumberPower, decNumberLn)—is diagnosed only under UBSan (-fsanitize=undefined -fno-sanitize-recover=undefined). On a standard two’s-complement build, a “wrapped” value is exactly what the code would clip anyway, so the result is the same with or without the fix: the test passes in either case and catches nothing.

So, with a regular `make run_tests`, these cases effectively only verify that the program “doesn’t crash,” which is meaningless for UB cases.

There are two options, but in my opinion, they are unsatisfactory. 

1. Introduce a `configure` flag (for example, --enable-asan), which adds -fsanitize=address,undefined -fno-sanitize-recover=undefined to CFLAGS/CXXFLAGS/LDFLAGS in builds/posix/make.defaults, and set up a separate CI job that builds and runs common_test under this profile. In a regular build, the tests remain simple “it works/the result is correct” checks, while the sanitizer run provides actual protection against bugs.
2. To allow the sanitizers to detect a bug within the library itself, pass the same flags to the nested `make extern/decNumber`—pass them to `CFLAGS` when calling `$ (MAKE) -C .../extern/decNumber in builds/posix/Makefile.in (currently, the library is hard-coded to build with -O3 in its Makefile). Then libdecFloat.a is compiled with instrumentation under the sanitizer profile, and UBSan/ASan trigger in decNumber.

extern/decNumber/decBasic.c
```diff
@@ -1768,6 +1768,7 @@ decFloat * decFloatCompareTotal(decFloat *result,
            else comp=-sigl;                  // ..
            break;
           }
+        break;                               // winner found; stop scanning
```
AlexPeshkoff, line 1771: Unlike the rest of changes this one modifies behavior. Are you sure it's OK?

ChudaykinAlex: Yes — it changes behaviour deliberately, because the current behaviour is
incorrect. **_But I understand the concerns and am not insisting on this change_**

Let me lay out the evidence.

## What the standard requires

IEEE 754-2008, clause 5.10, which `decFloatCompareTotal` implements:

> totalOrder(x, y) **imposes a total ordering** on canonical members of the
> format of x and y:
> […]
> d) If x and y are unordered numerically because x or y is NaN:
> […]
> 3) If x and y are both NaNs, then totalOrder reflects a total ordering
> based on:
> i) negative sign orders below positive sign
> ii) signaling orders below quiet for +NaN, reverse for −NaN
> iii) **lesser payload, when regarded as an integer**, orders below greater
> payload for +NaN, reverse for −NaN.

The decisive phrase is "when regarded as an integer": integer comparison is
decided by the *most significant* difference. decNumber's own documentation
says the same — "Compares x and y using the IEEE 754 total ordering", "NaNs
are ordered by payload" — as does the Decimal Arithmetic Specification:
"Quiet NaNs and signaling NaNs are ordered according to their payload; a
larger payload is higher in the ordering."

## What the code actually does

Without the `break`, the outer loop keeps scanning after a winner has been
found, and `comp` is overwritten by each subsequent difference. The result
therefore reflects the *last* differing digit rather than the first.
Concretely:

```
decQuadCompareTotal(NaN900001, NaN100009)  ->  -1     (correct: +1)
```

The first difference is `9 > 1`, so the LHS is greater. The inner byte loop
finds it, breaks out of itself only, and the outer loop then reaches the
later `1 < 9` and flips the answer.

## This is visible without any oracle

The function contradicts itself:

```
x = -NaN7928498310302199781717
y = -NaN132562703635197
  compareTotal(x, y) = +1      // x > y
  compareTotal(y, x) = +1      // y > x, simultaneously
```

Over 200,000 random operand pairs the current code produces 530 such
self-contradictions. Whatever else one calls that, it is not a total
ordering, so clause 5.10 is not satisfied.

## Cross-check against an independent implementation

Using `decNumberCompareTotal` as the oracle (separate code path, payloads
compared via `decUnitCompare`), same 200,000 pairs, same tree, the added
`break` being the only difference between the two builds:

| Build | Mismatches vs oracle | Self-contradictions |
|---|---|---|
| without the change | 8555 | 530 |
| with the change | 0 | 0 |

The inputs are synthetic, not a fuzzer corpus: a fixed-seed generator builds
both operands independently — random sign; `NaN`/`sNaN` with a random 0–33
digit payload; `Infinity`; finite values of 1–34 digits with an optional
exponent.

## The rest of the patch is behaviour-neutral, and I checked that separately

Since only this hunk should change anything observable, I diffed 16
operations over 40,000 random operand pairs — 640,000 result lines — between
the unpatched and patched builds. The only differing lines are
`compareTotal` (310) and `compareTotalMag` (584, which just calls
`compareTotal`). `add`, `subtract`, `multiply`, `divide`, `FMA`, `reduce`,
`quantize`, `compare`, `fromString`, `power`, `ln`, `log10` and `exp` are
bit-identical. So the integer casts and the `&&` reordering only take the
computation out of undefined behaviour; the `break` is the single
intentional behaviour change in the series.

## Why the out-of-bounds read is the same defect, not a separate one

The inner byte loop leaves `ub` misaligned by 1–3 bytes relative to the
group boundary. The outer `ub+=4` preserves that misalignment, so
`UBTOUI(ub)` reads from unaligned positions and eventually past the end of
`bufl[DECPMAX+4]`. Re-aligning the pointer would silence ASan while leaving
the wrong results in place; the `break` removes the cause of both.

The existing comment one line above states the original intent —
`// about to find a winner`. Once the winner is found, scanning should stop.
The `break` restores that, rather than introducing new semantics.

## Scope

Confined to the "both operands are NaN of the same kind and sign" branch.
Finite values, infinities, differing signs and differing NaN kinds are
untouched. Payloads that differ in only one 4-byte group are unaffected —
which is why this went unnoticed; it needs two differences, in opposite
directions, in separate groups.

## Regression Tests

I ran the tests on our CI system. No regressions were found.

---

References:
- IEEE Std 754-2008, clause 5.10 "Details of totalOrder predicate"
- Decimal Arithmetic Specification, `compare-total`
  (https://speleotrove.com/decimal/damisc.html)
- decNumber User's Guide (`decnumber.pdf`), `decNumberCompareTotal` /
  `decDoubleCompareTotal`
</content>

PR #9022, opened 2026-05-12
Fix found flaws of package constants 

Noremos:
Fix incorrect search in SCHEMAs list
Fix missing RDB$CONSTANT_SOURCE setup
Add USAGE privilage to SYSTEM packages

### DIFF
--- a/src/dsql/PackageNodes.epp
+++ b/src/dsql/PackageNodes.epp
@@ -679,7 +679,7 @@ void CreatePackageConstantNode::executeCreate(thread_db* tdbb, DsqlCompilerScrat
 
 		CONST.RDB$PRIVATE_FLAG = m_isPrivate;
 
-		attachment->storeBinaryBlob(tdbb, transaction, &CONST.RDB$CONSTANT_SOURCE, source);
+		attachment->storeMetaDataBlob(tdbb, transaction, &CONST.RDB$CONSTANT_SOURCE, source);
 	}
 	END_STORE
 }
--- a/src/dsql/PackageNodes.epp
+++ b/src/dsql/PackageNodes.epp
@@ -576,6 +576,7 @@ string CreatePackageConstantNode::internalPrint(NodePrinter& printer) const
 	DdlNode::internalPrint(printer);
 
 	NODE_PRINT(printer, name);
+	NODE_PRINT(printer, source);
 	NODE_PRINT(printer, m_type);
 	NODE_PRINT(printer, m_expr);
 	NODE_PRINT(printer, m_isPrivate);
@@ -586,6 +587,8 @@ string CreatePackageConstantNode::internalPrint(NodePrinter& printer) const
 DdlNode* CreatePackageConstantNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
 {
 	dsqlScratch->qualifyNewName(name);
+
+	source.ltrim("\n\r\t ");
 
 	m_expr = m_expr->dsqlPass(dsqlScratch);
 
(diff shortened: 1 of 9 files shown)


src/jrd/ini.epp
```diff
@@ -421,8 +421,9 @@ namespace
 					PAD(users[i], PRIV.RDB$USER);
 					PAD(SYSTEM_SCHEMA, PRIV.RDB$RELATION_SCHEMA_NAME);
 					PAD(objName, PRIV.RDB$RELATION_NAME);
-					PRIV.RDB$PRIVILEGE[0] = *privileges;
-					PRIV.RDB$PRIVILEGE[1] = 0;
+					PRIV.RDB$PRIVILEGE[0] = privileges[0];
+					PRIV.RDB$PRIVILEGE[1] = privileges[1];
+					PRIV.RDB$PRIVILEGE[2] = 0;
```
aafemt: Why not loop here?

Noremos: I followed the original style. Is loop worth it?

Noremos: Or do you mean storing each privilege in a separate record?

aafemt: No, I mean that the check and the way one string is assigned to another is strange, but if it was an original code... Probably, it still must be replaced with `strncpy`.

asfernandes (approved):

src/jrd/ini.epp
Noremos: Hm, as I can see, storing each privilege in a separate record is indead a correct way to do it

Noremos:
I found an issue related to metadata. When selecting a constant from a procedure, an error may occur:
`Statement format outdated, need to be reprepared.` I hope I can fix this soon.

Code to reproduce:
 ```SQL
set term ^;

    CREATE PACKAGE TEST1
    AS
    BEGIN
        PROCEDURE P1(I INT) RETURNS (PROCEDURE_OUTPUT INT); -- public procedure
        CONSTANT TEST1_PACKAGE_CONSTANT INTEGER = 10;
    END^

    CREATE PACKAGE BODY TEST1
    AS
    BEGIN
        CONSTANT CP2 INTEGER = 15;
        CONSTANT CP3 INTEGER = 30;
        PROCEDURE P1(I INT) RETURNS (PROCEDURE_OUTPUT INT)
        AS
        BEGIN
            PROCEDURE_OUTPUT = TEST1.TEST1_PACKAGE_CONSTANT + TEST1.CP2;
            SUSPEND;
        END
    END ^

    set term ;^
    commit;

    -- Test Public
    select test1.TEST1_PACKAGE_CONSTANT from rdb$database ;
    select * from test1.p1(0);
    commit;
```

File: src/jrd/ini.epp (hunk 1 above)
Comment (aafemt) [reply src/jrd/ini.epp]: In this case `CHAR(6)` field makes no sense.

File: src/dsql/PackageNodes.epp [hunk 2]
```diff
@@ -676,7 +679,7 @@ void CreatePackageConstantNode::executeCreate(thread_db* tdbb, DsqlCompilerScrat
 
 		CONST.RDB$PRIVATE_FLAG = m_isPrivate;
 
-		CONST.RDB$CONSTANT_SOURCE.NULL = TRUE;
+		attachment->storeBinaryBlob(tdbb, transaction, &CONST.RDB$CONSTANT_SOURCE, source);
```
hvlad: `storeMetaDataBlob()` maybe ?

Noremos: Thanks for the advice. Changed as suggested

src/jrd/ini.epp
```diff
@@ -330,16 +330,20 @@ namespace
 			const size_t ownerNameLength = ownerName.length();
 			fb_assert(ownerNameLength <= MAX_UCHAR);
 
-			const UCHAR privilege = package ? priv_execute : priv_usage;
+			Firebird::Vector<UCHAR, 8> NON_REL_OWNER_ACL
```
asfernandes: Why use `Firebird::` qualifier in a file that uses this namespace?

Noremos: Fixed.  I'm not used to the presence of `using namespace` in the code.

pavel-zotov:
::: QA note :::
See functional/package/test_e588f9ea.py

PR #9025, opened 2026-05-13
UDF compact div overflow fix

Ev3nt:
If you create a function that uses `udf_compat!UC_div`, then when it is called with parameters `-2147483648` and `-1`, it will crash due to overflow. The problem is that the upper bound of int is `2147483647`, so changing the sign causes a crash. `n1` and `n2` in `in` are `int`, so as a solution during the function call, you can cast them to `long long`.

### DIFF
--- a/src/extlib/UdfBackwardCompatibility.cpp
+++ b/src/extlib/UdfBackwardCompatibility.cpp
@@ -96,7 +96,7 @@ FB_UDR_BEGIN_FUNCTION(UC_div)
 		{
 			out->resultNull = FB_FALSE;
 			if (in->n2)
-				out->result = div((long long)in->n1, (long long)in->n2).quot;
+				out->result = div((long long) in->n1, (long long) in->n2).quot;
 			else
 			{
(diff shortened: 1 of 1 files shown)


dyemanov (approved):

src/extlib/UdfBackwardCompatibility.cpp
```diff
@@ -96,7 +96,7 @@ FB_UDR_BEGIN_FUNCTION(UC_div)
 		{
 			out->resultNull = FB_FALSE;
 			if (in->n2)
-				out->result = div(in->n1, in->n2).quot;
+				out->result = div((long long)in->n1, (long long)in->n2).quot;
```
asfernandes: Please follow stablished format rule:

```suggestion
				out->result = div((long long) in->n1, (long long) in->n2).quot;
```

aafemt: In this case I would use `static_cast` for better readability.

PR #9029, opened 2026-05-17
Implement SQL grouping extensions: ROLLUP, CUBE, GROUPING SETS, GROUPING(), GROUP BY DISTINCT

livius2:
Hi

A few years ago I made an earlier attempt at implementing ROLLUP / CUBE support in Firebird. 
That prototype was based on joining the input stream with a procedure producing a sequence of numbers/masks. 
It worked for simple cases, but the approach turned out to be hard to generalize, difficult to integrate cleanly with the optimizer/executor, and not a good long-term fit for Firebird internals.

After several years of following Firebird development and repository changes, I decided to revisit this feature with a cleaner implementation strategy.

**This pull request implements SQL extended grouping support, including:**

- GROUP BY ROLLUP (...)
- GROUP BY CUBE (...)
- GROUP BY GROUPING SETS (...)
- mixed grouping elements, for example GROUP BY a, ROLLUP(b, c)
- empty grouping set ()
- GROUPING() with multiple arguments, SQL feature T433
- GROUP BY DISTINCT, SQL feature T434
- GROUPING_ID as a compatibility extension

The implementation lowers advanced grouping to existing aggregate and UNION ALL machinery, without adding new BLR opcodes, ODS changes.

This is a flattened commit from my local repo, there were too many commits and experiments there.
Please note that all English texts come from a translator.
_I will add tests to Firebird-qa soon._
Tests added now:
https://github.com/FirebirdSQL/firebird-qa/pull/38


sim1984:
>  
> The implementation lowers advanced grouping to existing aggregate and UNION ALL machinery, without adding new BLR opcodes, ODS changes.
> 

This is unfortunate; at least aggregate functions containing only `WITH ROLLUP` can be executed in a single pass (without re-executing the query). This is unlikely to be possible with `CUBE` and `GROUPING SETS`, but there are options there, such as repeating groupings based on records stored in the "Record Buffer".

livius2:
I agree. This implementation intentionally uses lowering to the existing aggregate + UNION ALL infrastructure as a first correctness-focused step, avoiding new BLR, ODS, record source, or executor changes.

This is not meant to be the final optimal execution strategy. ROLLUP is a good candidate for a future single-pass implementation, as its grouping levels are hierarchical and can be produced from one ordered/grouped stream. CUBE and arbitrary GROUPING SETS are less straightforward, but there are still possible optimizations, for example sharing/materializing the base stream or reusing buffered records instead of re-executing the full input for every grouping set.

For this PR, the main goal is SQL semantics and integration with existing DSQL/BLR paths. Native execution and optimizer improvements can be developed as a follow-up without changing the SQL surface introduced here.

sim1984:
I don't see any reason to change the ODS. The introduction of new BLR verbs may be necessary to improve implementation efficiency, but it's not necessary.

The main question: do you plan to implement support in the optimizer yourself or wait for someone else to do it?

livius2:
Yes, if I understand correctly, by optimizer support you mean a future native execution strategy, for example single-pass execution for pure ROLLUP cases.

I do plan to work on that, but I would prefer to do it after this PR is accepted and merged. This PR has already required a significant amount of work, and I would like to avoid investing in a larger implementation before knowing whether this approach is acceptable.

doc/sql.extensions/README.grouping_sets.md
```diff
@@ -0,0 +1,237 @@
...
+  | ROLLUP(<grouping element> [, <grouping element> ...])
+  | CUBE(<grouping element> [, <grouping element> ...])
+  | GROUPING SETS(<grouping set> [, <grouping set> ...])
+
+<grouping set> ::=
+    <value expression>
+  | (<value expression> [, <value expression> ...])
+  | ()
+  | ROLLUP(<grouping element> [, <grouping element> ...])
+  | CUBE(<grouping element> [, <grouping element> ...])
```
dyemanov, line 32: As per SQL specification (and your changes in the parser), grouping sets may be nested, i.e. `<grouping set>` may include a `GROUPING SETS` clause too. This is missing from the docs.

src/dsql/parse.y
```diff
@@ -7245,16 +7253,147 @@ gen_series_step_opt
...
+				GroupingClause::DuplicateMode::DISTINCT : GroupingClause::DuplicateMode::ALL;
+			$$->duplicateModeExplicit = $3 != 0;
+		}
+	;
+
+%type <intVal> group_by_duplicate_opt
+group_by_duplicate_opt
+	: /* nothing */	{ $$ = 0; }
+	| ALL			{ $$ = 1; }
+	| DISTINCT		{ $$ = 2; }
```
dyemanov: I'd suggest to replace `group_by_duplicate_opt` with existing `quantifier_opt`.

src/dsql/parse.y
```diff
@@ -7245,16 +7253,147 @@ gen_series_step_opt
...
+					$3->legacyGroup->items.hasData())
+			{
+				for (auto item = $3->legacyGroup->items.begin();
+					 item != $3->legacyGroup->items.end(); ++item)
+				{
+					$$->legacyGroup->add(*item);
+				}
+			}
+			else
+				$$->legacyGroup = NULL;
```
dyemanov: I have a feeling that fixup of `legacyGroup` does not belong to the parser and should better be done inside `dsqlPass()`.

src/jrd/RecordSourceNodes.h
```diff
@@ -120,6 +121,243 @@ class MapNode : public Firebird::PermanentStorage, public Printable
...
+		{
+			SIMPLE,
+			EMPTY,
+			ROLLUP,
+			CUBE,
+			GROUPING_SETS
+		};
+
+		explicit Element(Type aType = Type::SIMPLE, ValueListNode* aItems = NULL,
+				GroupingClause* aGroupingSets = NULL)
```
dyemanov: `nullptr` instead of `NULL`, please

src/jrd/RecordSourceNodes.h
```diff
@@ -120,6 +121,243 @@ class MapNode : public Firebird::PermanentStorage, public Printable
...
+		Type type;
+		NestConst<ValueListNode> items;
+		NestConst<GroupingClause> groupingSets;
+	};
+
+	enum class DuplicateMode : UCHAR
+	{
+		ALL,
+		DISTINCT
+	};
```
dyemanov: This could be just a `bool`, I suppose.

src/jrd/RecordSourceNodes.h
```diff
@@ -120,6 +121,243 @@ class MapNode : public Firebird::PermanentStorage, public Printable
...
+
+
+class GroupingSpec final : public Firebird::PermanentStorage, public Printable
+{
+public:
+	struct Dimension final : public Printable
+	{
+		explicit Dimension(MemoryPool&)
+			: expr(NULL),
+			  index(0)
```
dyemanov: Here and in other places -- constant-based initialization should better be done in the member declaration, e.g. `unsigned index = 0;`. And `NestConst` already initializes to `nullptr` by default, no need to assign it explicitly.

src/dsql/ExprNodes.h
```diff
@@ -1012,6 +1012,83 @@ class InternalInfoNode final : public TypedNode<ValueExprNode, ExprNode::TYPE_IN
...
+public:
+	const char* getDsqlFunctionName() const
+	{
+		return dsqlGroupingFunction ? "GROUPING" : "GROUPING_ID";
+	}
+
+public:
+	NestConst<ValueListNode> args;
+	bool dsqlGroupingFunction;
+};
```
dyemanov, line 1068: IMHO, `GroupingNode` and `GroupingIdNode` are worth merging into a single class. It could return `INT` or `BIGINT` depending on `args` count.

src/dsql/ExprNodes.cpp
```diff
@@ -7662,6 +7677,257 @@ ValueExprNode* InternalInfoNode::dsqlPass(DsqlCompilerScratch* dsqlScratch)
...
+
+ValueExprNode* GroupingNode::pass2(thread_db* tdbb, CompilerScratch* csb)
+{
+	ValueExprNode::pass2(tdbb, csb);
+	return this;
+}
+
+dsc* GroupingNode::execute(thread_db* /*tdbb*/, Request* /*request*/) const
+{
+	fb_assert(false);
```
dyemanov: Does it mean `<group operation>` is temporarily disallowed (returns `NULL`) until proper BLR codes are added in another PR?

dyemanov (changes requested):
This is the first part of the review, the hardest changes inside `pass1.cpp` are still to be reviewed a bit later (the next week, hopefully).

src/dsql/ExprNodes.cpp
livius2: No, GROUPING/GROUPING_ID are not temporarily disallowed and they do not return NULL.

The node exists only as a DSQL-level representation parsed from SQL. For supported contexts, advanced grouping is lowered before BLR generation, and GROUPING/GROUPING_ID are replaced with branch-specific integer constants: 0/1 for GROUPING and a bit mask for GROUPING_ID.

So GroupingNode::execute() should never be reached in a valid compiled request. The fb_assert(false) is a defensive guard for an internal lowering failure. Similarly, genBlr() raises an error if a GROUPING expression reaches BLR generation without being lowered.

This PR intentionally avoids new BLR opcodes. Proper native BLR/executor support could be added later as an optimization path, but it is not required for the SQL feature to work in this implementation.

doc/sql.extensions/README.grouping_sets.md
livius2: Updated doc in [0e50834](https://github.com/FirebirdSQL/firebird/pull/9029/commits/0e508340923c1305dae43e7555111768e7761c77)

src/dsql/ExprNodes.h
livius2: Done in [d0b6e7c](https://github.com/FirebirdSQL/firebird/pull/9029/commits/d0b6e7cbcc4b8c09c7f3876f393fa3238dc94045)

src/dsql/parse.y
livius2: Done in [8a16dd9](https://github.com/FirebirdSQL/firebird/pull/9029/commits/8a16dd943f8cfb61263713479a2d049e1412764e)

src/jrd/RecordSourceNodes.h
livius2: Done in [7fede78](https://github.com/FirebirdSQL/firebird/pull/9029/commits/7fede78fa8d9eaa4ebd6ebe06dabadafc1a05f64)

src/jrd/RecordSourceNodes.h
livius2: Done in [cfc8434](https://github.com/FirebirdSQL/firebird/pull/9029/commits/cfc84347b60c67db72479677fd42d2ff8448cf3e)

src/jrd/RecordSourceNodes.h
livius2: Done in [3282fff](https://github.com/FirebirdSQL/firebird/pull/9029/commits/3282fffae56de64915b334f7c12bd969189461be)

src/dsql/parse.y
livius2: Done in [25eb2a9](https://github.com/FirebirdSQL/firebird/pull/9029/commits/25eb2a95a34417d48dcc40ecc5307d44f8f2dd3e)

PR #9030, opened 2026-05-18
Extend tempspace in safe way

Noremos:
There are a FIXME  in the code:
https://github.com/FirebirdSQL/firebird/blob/8eb7bba99c09ed5b4937a12912af23e002ade824/src/jrd/TempSpace.cpp#L402-L404

And seem like it caused a segfault in a prodaction DB due to `TempSpace::head` is nullptr or `TempSpace::tail` is nullptr.

I reproduced this using a synteny test

1) Simulate small storage
```
sudo mount -t tmpfs -o size=1K tmpfs /mnt/ramdisk
sudo mount -t tmpfs -o size=1M tmpfs /mnt/ramdisk 
```
2) Firebird.conf:
```
TempDirectories = /mnt/ramdisk
TempCacheLimit = 1M
```

3) Run the code

First case:
```cpp
BOOST_CHECK_THROW(space.allocateSpace(tdbb, MORE_THEN_FILE_STORAGE), Firebird::Exception); // <- exception 
pos = space.allocateSpace(tdbb, 100); <- tail is nullptr
space.write(tdbb, pos, "123", 3); // <- segfault
```

Second case:
```cpp
const auto MORE_THEN_FILE_STORAGE = 10000024;
BOOST_CHECK_THROW(space.allocateSpace(MORE_THEN_FILE_STORAGE), Firebird::Exception); // <- exception 

pos = space.write(pos, "123", 3); // <- segfault
```

### DIFF
--- a/src/jrd/TempSpace.cpp
+++ b/src/jrd/TempSpace.cpp
@@ -261,7 +261,7 @@ FB_SIZE_T TempSpace::read(offset_t offset, void* buffer, FB_SIZE_T length)
 	if (offset + length > logicalSize)
 	{
 		status_exception::raise(Arg::Gds(isc_temp_space_invalid_pos) <<
-			Arg::Num(offset + length) << Arg::Num(logicalSize));
+			Arg::Int64(offset + length) << Arg::Int64(logicalSize));
 	}
 
 	if (length)
@@ -297,7 +297,7 @@ FB_SIZE_T TempSpace::write(offset_t offset, const void* buffer, FB_SIZE_T length
 	if (offset > logicalSize)
 	{
 		status_exception::raise(Arg::Gds(isc_temp_space_invalid_pos) <<
-			Arg::Num(offset) << Arg::Num(logicalSize));
+			Arg::Int64(offset) << Arg::Int64(logicalSize));
 	}
 
 	if (offset + length > logicalSize)
--- a/src/jrd/TempSpace.cpp
+++ b/src/jrd/TempSpace.cpp
@@ -328,9 +328,14 @@ FB_SIZE_T TempSpace::write(offset_t offset, const void* buffer, FB_SIZE_T length
 
 void TempSpace::extend(FB_SIZE_T size)
 {
-	logicalSize += size;
+	const auto originalLogicalSize = logicalSize;
+	const auto originalPhysicalSize = physicalSize;
 
-	if (logicalSize > physicalSize)
+	logicalSize += size;
+	if (logicalSize <= physicalSize)
+		return;
+
+	try
 	{
 		const FB_SIZE_T initialSize = initialBuffer.getCount();
 
@@ -399,12 +404,10 @@ void TempSpace::extend(FB_SIZE_T size)
 			}
 		}
 
-		// NS 2014-07-31: FIXME: missing exception handling.
-		// error thrown in block of code below will leave TempSpace in inconsistent state:
-		// logical/physical size already increased while allocation has in fact failed.
 		if (!block)
 		{
 			// allocate block in the temp file
+			// Possible error thrown when not enough physical memory
 			TempFile* const file = setupFile(size);
 			fb_assert(file);
 			if (tail && tail->sameFile(file))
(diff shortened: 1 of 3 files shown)


src/jrd/TempSpace.cpp
```diff
@@ -430,6 +433,13 @@ void TempSpace::extend(FB_SIZE_T size)
 		}
 		tail = block;
 	}
+	catch (...)
+	{
+		// Restore original state
+		logicalSize = originalLogicalSize;
+		physicalSize = originalPhysicalSize;
+		throw;
```
hvlad, line 454: At line 375 the only block is deleted, how this recovery handle this case ?

Noremos: After an exception occurs, the next call of the write/allocate function will trigger execution of this function. The missing `head` and `tail` will be set correctly.

hvlad: But the object internal state is inconsistent: it have non-zero size but it have no memory block allocated.
Yes, next `write` or `extend` will fix it, but what if next call is `read` ?

Noremos: Indeed, I didn't think about combining `read` and `write` operations. I'll improve the fix.

Noremos: Now, I delay pointer deliting and restore it in case of an error. But there's a problem with non-dynamic TempSpace. By default, `head` and `tail` are NULL. Therefore, if `extend` fails and the logic ignores the exception, reading can lead to unexpected behavior. But TempSpace::read doesn't dereference a null pointer, so we are safe in any case.

src/jrd/TempSpace.cpp
```diff
@@ -328,9 +328,17 @@ FB_SIZE_T TempSpace::write(offset_t offset, const void* buffer, FB_SIZE_T length
 
 void TempSpace::extend(FB_SIZE_T size)
 {
+	const auto originalLogicalSize = logicalSize;
+	const auto originalPhysicalSize = physicalSize;
+
+	AutoPtr<Block> delayDelete; // Restore in case of error
```
hvlad: `originalHead` maybe ? For consistency with other saved fields.

Noremos: Good idea. Changed as sugested

src/jrd/TempSpace.cpp
hvlad: Looks good, thanks. 

The only concern left is - should we convert asserts at the start of `read()` and `write()` into exceptions ? 
I don't think it adds measurable perf penalty, but adds robustness, imho.

hvlad (approved):

src/jrd/TempSpace.cpp
Noremos: I think this is a good idea. Especially in the case of new code using TempSpace. I can add this as part of this PR or a new one.

dyemanov: I think it fits this PR nicely.

hvlad: Me too

Noremos: Added

src/jrd/TempSpace.cpp
```diff
@@ -258,7 +258,11 @@ TempSpace::~TempSpace()
 
 FB_SIZE_T TempSpace::read(offset_t offset, void* buffer, FB_SIZE_T length)
 {
-	fb_assert(offset + length <= logicalSize);
+	if (offset + length > logicalSize)
+	{
+		status_exception::raise(Arg::Gds(isc_temp_space_invalid_pos) <<
+			Arg::Num(offset + length) << Arg::Num(logicalSize));
```
hvlad: Use `Arg::Int64` as `offset_t` is 8-byte integer

aafemt: Do you think that integer overflow is possible here?

Noremos: > Do you think that integer overflow is possible here?

Avoiding data truncation is a good practice. At the very least, we won't get an annoying warning from the compiler.

PR #9031, opened 2026-05-18
Fix segfault after package body drop

Noremos:
Fix for the problem reported in: https://groups.google.com/g/firebird-devel/c/Hr9Y9bnGTqY

### DIFF
--- a/src/jrd/dfw.epp
+++ b/src/jrd/dfw.epp
@@ -2409,27 +2409,9 @@ static bool drop_package_body(thread_db* tdbb, SSHORT phase, DeferredWork* work,
 	Cached::Package* routine;
 	switch (phase)
 	{
-	case 0:
-		routine = MetadataCache::getPerm<Cached::Package>(tdbb, work->dfw_id, CacheFlag::NOSCAN);
-		if (routine)
-			routine->rollback(tdbb);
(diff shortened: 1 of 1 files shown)


src/jrd/dfw.epp
```diff
@@ -2409,27 +2409,9 @@ static bool drop_package_body(thread_db* tdbb, SSHORT phase, DeferredWork* work,
...
-	case 2:
-	case 3:
-	case 4:
-	case 5:
-	case 6:
-		return true;
-	case 7:
-		routine = MetadataCache::getPerm<Cached::Package>(tdbb, work->dfw_id, CacheFlag::ERASED | CacheFlag::NOSCAN);
-		fb_assert(routine);
-		if (routine)
```
AlexPeshkoff, line 2431: I see  MetadataCache::newVersion<Cached::Package>() in CreatePackageBodyNode::execute(). It must have (as a pair) call to commit or rollback cache. Why did you decide to remove it?

Noremos: For a moment, I thought the `commit` and `rollback` commands were only necessary for the `erase` and `create` operations. But as I see it now, the real problem is the invalid flag.

PR #9037, opened 2026-05-25
Fix collate and null processing in package constants

Noremos:
Fix of regressions described in https://groups.google.com/g/firebird-devel/c/5ehuR18Wemk and https://groups.google.com/g/firebird-devel/c/kze2LTWaeco

### DIFF
--- a/src/dsql/PackageNodes.epp
+++ b/src/dsql/PackageNodes.epp
@@ -474,7 +474,6 @@ void PackageReferenceNode::make(DsqlCompilerScratch* dsqlScratch, dsc* desc)
 	jrd_tra* transaction = dsqlScratch->getTransaction();
 	thread_db* tdbb = JRD_get_thread_data();
 	*desc = ConstantValue::getDesc(tdbb, transaction, m_fullName);
-	desc->setNullable(true);
 }
 
 bool PackageReferenceNode::constantExists(thread_db* tdbb, Jrd::jrd_tra* transaction,
@@ -504,7 +503,6 @@ bool PackageReferenceNode::constantExists(thread_db* tdbb, Jrd::jrd_tra* transac
 void PackageReferenceNode::getDesc(thread_db* tdbb, CompilerScratch*, dsc* desc)
 {
 	*desc = ConstantValue::getDesc(tdbb, tdbb->getTransaction(), m_fullName);
-	desc->setNullable(true);
 }
 
 ValueExprNode* PackageReferenceNode::copy(thread_db* tdbb, NodeCopier& copier) const
--- a/src/dsql/PackageNodes.epp
+++ b/src/dsql/PackageNodes.epp
@@ -474,6 +474,7 @@ void PackageReferenceNode::make(DsqlCompilerScratch* dsqlScratch, dsc* desc)
 	jrd_tra* transaction = dsqlScratch->getTransaction();
 	thread_db* tdbb = JRD_get_thread_data();
 	*desc = ConstantValue::getDesc(tdbb, transaction, m_fullName);
+	desc->setNullable(true);
 }
 
 bool PackageReferenceNode::constantExists(thread_db* tdbb, Jrd::jrd_tra* transaction,
@@ -503,6 +504,7 @@ bool PackageReferenceNode::constantExists(thread_db* tdbb, Jrd::jrd_tra* transac
 void PackageReferenceNode::getDesc(thread_db* tdbb, CompilerScratch*, dsc* desc)
 {
 	*desc = ConstantValue::getDesc(tdbb, tdbb->getTransaction(), m_fullName);
+	desc->setNullable(true);
 }
 
 ValueExprNode* PackageReferenceNode::copy(thread_db* tdbb, NodeCopier& copier) const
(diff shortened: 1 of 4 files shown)


src/dsql/PackageNodes.epp
```diff
@@ -474,6 +474,7 @@ void PackageReferenceNode::make(DsqlCompilerScratch* dsqlScratch, dsc* desc)
 	jrd_tra* transaction = dsqlScratch->getTransaction();
 	thread_db* tdbb = JRD_get_thread_data();
 	*desc = ConstantValue::getDesc(tdbb, transaction, m_fullName);
+	desc->setNullable(true);
```
asfernandes: Why is constant references always marked as nullable?

Noremos: To return `NULL` from the `ReferenceNode`, the `nullable` flag is requered. It would be nice to explicitly check the `not null` state of the field, but as I can see, it does not stored in RDB$FIELD. So I decied to put the nullable flag withou extra checks.

The `not null` clasue can only be used when the type is a domain:
```SQL
create domain my_domain int not null;

create or alter package my_package as
begin
    constant my_const my_domain = null;
end;
```
 
But in this case an error will appear instanly:

```
Statement failed, SQLSTATE = 2F000
validation error for CAST, value "*** null ***"
-Error while parsing BLR value of the constant "PUBLIC"."MY_PACKAGE"."MY_CONST"
```
So, putting nullable flag in any case is safe

asfernandes: Every other kind of expressions, include composed ones, do return nullable flag correctly. I see no reason for constants to flag nullable incorrectly.

dyemanov: > It would be nice to explicitly check the `not null` state of the field, but as I can see, it does not stored in RDB$FIELD

`RDB$FIELDS.RDB$NULL_FLAG` ?

Noremos: > > It would be nice to explicitly check the `not null` state of the field, but as I can see, it does not stored in RDB$FIELD
> 
> `RDB$FIELDS.RDB$NULL_FLAG` ?



I missed that somehow. Fixed

src/jrd/Package.epp
```diff
@@ -156,12 +158,12 @@ dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
...
 	{
 		ReadLockGuard guard(m_makeValueLock, FB_FUNCTION);
-		if (m_value.vlu_desc.dsc_address != nullptr)
+		if (m_value.vlu_desc.dsc_dtype != 0)
 			return &m_value.vlu_desc;
 	}
 
 	WriteLockGuard guard(m_makeValueLock, FB_FUNCTION);
-	if (m_value.vlu_desc.dsc_address != nullptr) // Extra check in case of two callers waiting
+	if (m_value.vlu_desc.dsc_dtype != 0) // Extra check in case of two callers waiting
```
asfernandes:
```suggestion
	if (m_value.vlu_desc.dsc_dtype != dtype_unknown) // Extra check in case of two callers waiting
```

src/jrd/Package.epp
```diff
@@ -196,14 +198,11 @@ dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
 
 			TRA_attach_request(tdbb->getTransaction(), request);
 
-			const dsc* temp = EVL_expr(tdbb, request, static_cast<ValueExprNode*>(csb->csb_node));
-			EVL_make_value(tdbb, temp, &m_value, &getPool());
+			//! Execute it here, while AutoSetRestore2 is active
```
asfernandes:
```suggestion
			// Execute it here, while AutoSetRestore2 is active
```

src/jrd/Package.epp
```diff
@@ -226,6 +225,23 @@ dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
 	return &m_value.vlu_desc;
 }
 
+void ConstantValue::executeCsbNode(thread_db* tdbb, CompilerScratch* csb)
+{
+	ValueExprNode* exprNode = static_cast<ValueExprNode*>(csb->csb_node);
```
asfernandes, line 231: Assert `csb->csb_node && csb->csb_node->getKind() == DmlNode::KIND_VALUE`.

src/jrd/Package.epp
Noremos: I used an exclamation point to highlight the warning. Is this style inappropriate for Firebird?

asfernandes: Exclamation point in English is in the end of the sentence :)
Such punctutation after comment remembers doc-style comments.

Noremos: Oh, sorry for the confusion. I'm just using an [extension](https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments) that highlights comments like this.

src/jrd/Package.epp
Noremos: Added as suggested

pavel-zotov:
::: QA note :::
See test $QA_HOME/tests/functional/package/test_ee3e78b8.py

PR #9039, opened 2026-05-26
Fix constant status check for BoolExprNode

Noremos:
Fixes the issue described in https://groups.google.com/g/firebird-devel/c/mtPIIHhd95c


pavel-zotov:
::: QA note :::
See $QA_HOME/tests/functional/package/test_5d04fe75.py

Issue #2058, opened 2016-08-24
Window Function: frame exclude clause [CORE5338]


firebird-automations:
Submitted by: @asfernandes

Jira\_subtask\_inward [CORE1688](https://github.com/FirebirdSQL/firebird/issues?q=CORE1688+in%3Atitle)

Votes: 3

frame clause is useful, especially for first\_value,last\_value, and nth\_value\| that consider only the
rows within the "window frame"

syntax :
<window frame clause\> ::=
  <window frame units\> <window frame extent\>
        \[ <window frame exclusion\> \]

 <window frame units\> ::=
       ROWS
     \| RANGE

<window frame extent\> ::=
      <window frame start\>
    \| <window frame between\>

<window frame start\> ::=
      UNBOUNDED PRECEDING
    \| <window frame preceding\>
    \| CURRENT ROW

<window frame preceding\> ::=
     <unsigned value specification\> PRECEDING

<window frame between\> ::=
    BETWEEN <window frame bound 1\> AND <window frame bound 2\>

<window frame bound 1\> ::=
    <window frame bound\>

<window frame bound 2\> ::=
    <window frame bound\>

<window frame bound\> ::=
      <window frame start\>
    \| UNBOUNDED FOLLOWING
    \| <window frame following\>

<window frame following\> ::=
   <unsigned value specification\> FOLLOWING

<window frame exclusion\> ::=
     EXCLUDE CURRENT ROW
   \| EXCLUDE GROUP
   \| EXCLUDE TIES
   \| EXCLUDE NO OTHERS

mrotteveel:
The currently implemented syntax in parse.y makes `window_frame_exclusion_opt` independent from the `window_frame_extent`, which is not correct according to the SQL:2016-2 syntax, the exclusion can only be specified when specifying a frame.

PR #9041, opened 2026-05-30
Feature #2058 - Window Function: frame exclude clause

asfernandes:


sim1984:
This is a topic for another PR, but I would like to point out that to fully complete the window theme, it is necessary to add frames with `GROUPS` units.

```
<window frame> ::=
  {RANGE | ROWS | GROUPS} <window frame extent> [<window frame exclusion>]
```

```sql
select hire_date
    -- include current & three previous rows
  , count(*) over ( order by hire_date rows 3 preceding ) prev3_rows
    -- include all rows between hire_date - 3 and hire_date for the current row
  , count(*) over ( order by hire_date range 3 preceding ) prev3_days
    -- include all rows with the any of the previous three and current hire_dates
  , count(*) over ( order by hire_date groups 3 preceding ) prev3_values
from   hr.employees
where  hire_date >= date'2015-03-03'
fetch first 5 rows only;
```

```
HIRE_DATE   PREV3_ROWS PREV3_DAYS PREV3_VALUES
----------- ---------- ---------- ------------
03-MAR-2015          1          1            1
10-MAR-2015          2          2            3
10-MAR-2015          3          2            3
11-MAR-2015          4          3            4
19-MAR-2015          4          1            5
```

PR #9045, opened 2026-06-01
Fix calling reload twice for package constants and improve hash calculation

Noremos:
Fixes for the issue described in https://groups.google.com/g/firebird-devel/c/xIxhL_QAjEg : wierd error message when using a consatnt.

To reproduce:
```SQL
set term ^;
CREATE PACKAGE TEST
AS
BEGIN
    CONSTANT C1 INTEGER = 10;
END^

CREATE PROCEDURE P1(I INT) RETURNS (PROCEDURE_OUTPUT INT)
AS
BEGIN
    PROCEDURE_OUTPUT = TEST.C1;
    SUSPEND;
END^

set term ;^
commit;

select TEST.C1 from rdb$database ;
select * from P1(0);
commit;
```

### DIFF
--- a/src/jrd/Package.epp
+++ b/src/jrd/Package.epp
@@ -49,11 +49,19 @@ DATABASE DB = FILENAME "ODS.RDB";
 //----------------------
 
 
+// Exclude address from the hash calculation due to it possible change
+constexpr inline ULONG DSC_SIZE_TO_HASH = sizeof(dsc) - offsetof(dsc, dsc_address);
 
 bool ConstantValue::hash(thread_db* tdbb, Firebird::sha512& digest) const
 {
-	fb_assert(m_value.vlu_desc.dsc_dtype != 0);
-	digest.process(sizeof(m_value.vlu_desc), &m_value.vlu_desc);
+	fb_assert(m_value.vlu_desc.dsc_dtype != dtype_unknown);
+	digest.process(DSC_SIZE_TO_HASH, &m_value.vlu_desc);
+	if (!m_value.vlu_desc.isNull())
+	{
+		fb_assert(m_value.vlu_desc.dsc_length > 0);
+		fb_assert(m_value.vlu_desc.dsc_address != nullptr);
+		digest.process(m_value.vlu_desc.dsc_length, m_value.vlu_desc.dsc_address);
+	}
 
 	return true;
 }
@@ -328,10 +336,7 @@ ScanResult Package::scan(thread_db* tdbb, ObjectBase::Flag flags)
 void Package::checkReload(thread_db* tdbb)
 {
 	if (m_callReload)
-	{
 		reload(tdbb, 0);
-		m_callReload = false;
-	}
 }
 
 ScanResult Package::reload(thread_db* tdbb, ObjectBase::Flag fl)
@@ -359,6 +364,7 @@ ScanResult Package::reload(thread_db* tdbb, ObjectBase::Flag fl)
 	}
 	END_FOR
 
+	m_callReload = false;
 	return ScanResult::COMPLETE;
 }
 
--- a/src/jrd/Package.epp
+++ b/src/jrd/Package.epp
@@ -50,7 +50,7 @@ DATABASE DB = FILENAME "ODS.RDB";
 
 
 // Exclude address from the hash calculation due to it possible change
-constexpr inline ULONG DSC_SIZE_TO_HASH = sizeof(dsc) - offsetof(dsc, dsc_address);
+constexpr inline ULONG DSC_SIZE_TO_HASH = offsetof(dsc, dsc_address);
 
 bool ConstantValue::hash(thread_db* tdbb, Firebird::sha512& digest) const
 {
--- a/src/jrd/Package.epp
+++ b/src/jrd/Package.epp
@@ -56,12 +56,17 @@ bool ConstantValue::hash(thread_db* tdbb, Firebird::sha512& digest) const
 {
 	fb_assert(m_value.vlu_desc.dsc_dtype != dtype_unknown);
 	digest.process(DSC_SIZE_TO_HASH, &m_value.vlu_desc);
-	if (!m_value.vlu_desc.isNull())
+
+	// The value hashing is unneccecery as Alex Peshkoff explained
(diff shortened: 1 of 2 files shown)


src/jrd/Package.epp
```diff
@@ -49,11 +49,24 @@ DATABASE DB = FILENAME "ODS.RDB";
 //----------------------
 
 
+// Exclude address from the hash calculation due to it possible change
+constexpr inline ULONG DSC_SIZE_TO_HASH = sizeof(dsc) - offsetof(dsc, dsc_address);
```
aafemt: Size of dsc is 12, offset of dsc_address is 8. 12 - 8 = 4. Are you sure that this is a right value?
Perhaps either `offset(dsc_offset)` or `offset(dsc_flags) + sizeof(dsc_flags)` should be here.

AlexPeshkoff: DSC_SIZE_TO_HASH = offsetof(dsc, dsc_address);

Noremos: > Size of dsc is 12, offset of dsc_address is 8. 12 - 8 = 4. Are you sure that this is a right value?
> Perhaps either `offset(dsc_offset)` or `offset(dsc_flags) + sizeof(dsc_flags)` should be here.

`dsc` size is 16 bytes:
```cpp
UCHAR	dsc_dtype = 0;       // Size: 1 byte,   Offset: 0 bytes
SCHAR	dsc_scale = 0;       // Size: 1 byte,   Offset: 1 byte
USHORT	dsc_length = 0;      // Size: 2 bytes,  Offset: 2 bytes
SSHORT	dsc_sub_type = 0;    // Size: 2 bytes,  Offset: 4 bytes
USHORT	dsc_flags = 0;       // Size: 2 bytes,  Offset: 6 bytes
UCHAR*	dsc_address = nullptr;  // Size: 8 bytes  (4 for x32), Offset: 8 bytes
```


> DSC_SIZE_TO_HASH = offsetof(dsc, dsc_address);

Thanks, fixed as suggested

aafemt: > `dsc` size is 16 bytes:

Ah, right, I looked at `Descriptor`, sorry.

AlexPeshkoff (approved):
Fix is OK & and I approve it but one question. Why does dsc_address change with same value? Sorry but looks strange for constant.

Noremos:
> Fix is OK & and I approve it but one question. Why does dsc_address change with same value? Sorry but looks strange for constant.

This was a bug. First, the Package::reload method was called, creating a constant. Then, in the Package::checkReload method, m_callReload remained true, so the Package::reload method was called again. Constants were added and recreated. That's why I moved `m_callReload = false` to Package::reload in this PR.

aafemt:
> m_callReload remained true

Cannot `dsc_address` to be used instead of the flag to indicate that the constant is loaded? (With value NULL indicated by `dsc_flags`.)

AlexPeshkoff:
On 6/1/26 18:47, Artyom Abakumov wrote:
> *Noremos* left a comment (FirebirdSQL/firebird#9045) 
> <https://github.com/FirebirdSQL/firebird/pull/9045#issuecomment-4594193895>
>
>     Why does dsc_address change with same value? Sorry but looks
>     strange for constant.
>
> This was a bug. First, the Package::reload method was called, creating 
> a constant. Then, in the Package::checkReload method, m_callReload 
> remained true, so the Package::reload method was called again. 
> Constants were added and recreated. That's why I moved |m_callReload = 
> false| to Package::reload in this PR.
>

Wonderful !!

Noremos:
> > m_callReload remained true
> 
> Cannot `dsc_address` to be used instead of the flag to indicate that the constant is loaded? (With value NULL indicated by `dsc_flags`.)

The `m_callReload` is located inside the package class. The dsc_address is a field of the `ConstantValue` class, so it is not a correct optimization

src/jrd/Package.epp
```diff
@@ -49,11 +49,24 @@ DATABASE DB = FILENAME "ODS.RDB";
 //----------------------
 
 
+// Exclude address from the hash calculation due to it possible change
+constexpr inline ULONG DSC_SIZE_TO_HASH = offsetof(dsc, dsc_address);
```
dyemanov, line 53: What about a static assert which ensures there's no padding before `dsc_address`? Otherwise, if `dsc` layout is changed, we may start hashing garbage bytes.

Noremos: Hope this is enough:
`static_assert(DSC_SIZE_TO_HASH - offsetof(dsc, dsc_flags) == sizeof(dsc::dsc_flags), "There should be no padding");`

dyemanov (approved):

src/jrd/Package.epp
aafemt: It tests only no padding after `dsc_flags` but padding after any other field will pass unnoticed. I would use sum of `sizeof` of every field that must be hashed.

AlexPeshkoff: Checking in all details layout of dsc is definitely out of scope for this fix - hashes for dsc as whole entity were already used before. On the other hand it will be good to have such check in Format::hash. @aafemt can you prepare PR?

aafemt: It doesn't matter who and where put following one-liner:
`static_assert(std::has_unique_object_representations_v<dsc>);`

AlexPeshkoff:
On 6/3/26 22:25, Dimitry Sibiryakov wrote:
>
>
>
> In src/jrd/Package.epp 
> <https://github.com/FirebirdSQL/firebird/pull/9045#discussion_r3351287880>:
>
> > @@ -49,11 +49,24 @@ DATABASE DB = FILENAME "ODS.RDB";
>   //----------------------
>   
>   
> +// Exclude address from the hash calculation due to it possible change
> +constexpr inline ULONG DSC_SIZE_TO_HASH = offsetof(dsc, dsc_address);
>
> It doesn't matter who and where put following one-liner:
> |static_assert(std::has_unique_object_representations_v<dsc>);|
>

Doesn't matter for compiler - but on my mind matters for human. Added in 
Format::hash.

pavel-zotov:
::: QA note :::
COvered by ```$QA_HOME/tests/functional/package/test_e588f9ea.py```

PR #9051, opened 2026-06-03
refactor(path_utils): Reuse existing buffer on each `PosixDirIterator` iteration

TreeHunter9:
I just noticed in our code that we're allocating memory for the `PathName` variable on every iteration of `PosixDirIterator`, so here's a small optimization for that section.

### DIFF
--- a/src/common/os/posix/path_utils.cpp
+++ b/src/common/os/posix/path_utils.cpp
@@ -102,18 +102,18 @@ const PosixDirIterator& PosixDirIterator::operator++()
 	{
 		while ( (ent = os_utils::readdir(dir)) )
 		{
-			PathName entryname;
-			PathUtils::concatPath(entryname, dirPrefix, ent->d_name);
+			PathUtils::concatPath(file, dirPrefix, ent->d_name);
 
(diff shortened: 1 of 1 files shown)


aafemt:
Whole this class must be wiped for the sake of `ScanDir` in src/common/ScanDir. Or even `std::filesystem::directory_iterator`.

TreeHunter9:
> Whole this class must be wiped for the sake of ScanDir in src/common/ScanDir.

Didn't know we have this :)

> Or even std::filesystem::directory_iterator.

Integrating `std::filesystem` into our codebase isn't that simple, especially when it comes to helper classes like `DirIterator`, since this would result in overhead due to type conversion (`std::filesystem::path` <-> `Firebird::PathName`), which we certainly want to avoid.

dyemanov:
Before using `ScanDir`, it should be reviewed to have all the needed functionality and error handling for both Linux and Windows. I remember some fixes landed into the `*DirIterator` code, they should present in `ScanDir` too.

dyemanov (approved):

PR #9052, opened 2026-06-05
fix(cast format): Correct behavior of `YEAR` pattern

TreeHunter9:
- This pattern was taken from Oracle and is supposed to display `spelled out year (TWENTY TWENTY-SIX)`. In the current implementation, it was equivalent to `YYYY` (i.e., the year as a number (2026)), which is incorrect;
- Update existing tests and add new ones for the reimplemented `YEAR` pattern;
- Fix incorrect variable names for `TO_STRING_PATTERNS` and `TO_DATETIME_PATTERNS` (usage was correct);


dyemanov (approved):

pavel-zotov:
::: QA note :::
See test for #8501

PR #9055, opened 2026-06-08
Fix server crash when read a malformed replication segment

XaBbl4:
If a segment has some damage and it falls on the length of a string or the position of an atom, then the server crash with SIGSEGV, instead of output an correct error

### DIFF
--- a/src/jrd/replication/Applier.cpp
+++ b/src/jrd/replication/Applier.cpp
@@ -127,13 +127,21 @@ namespace
 
 		const string& getAtomString()
 		{
-			const auto pos = getInt32();
+			const ULONG pos = static_cast<ULONG>(getInt32());
+
+			if (pos >= m_atoms.getCount())
+				malformed();
+
 			return m_atoms[pos];
 		}
 
 		const MetaString getAtomMetaName()
 		{
-			const auto pos = getInt32();
+			const ULONG pos = static_cast<ULONG>(getInt32());
+
+			if (pos >= m_atoms.getCount())
+				malformed();
+
 			return m_atoms[pos];
 		}
 
@@ -152,7 +160,7 @@ namespace
 
 		string getString()
 		{
-			const auto length = getInt32();
+			const ULONG length = static_cast<ULONG>(getInt32());
 
 			if (m_data + length > m_end)
 				malformed();
--- a/src/jrd/replication/Applier.cpp
+++ b/src/jrd/replication/Applier.cpp
@@ -127,9 +127,9 @@ namespace
 
 		const string& getAtomString()
 		{
-			const ULONG pos = static_cast<ULONG>(getInt32());
+			const auto pos = getInt32();
 
-			if (pos >= m_atoms.getCount())
+			if (pos < 0 || pos >= m_atoms.getCount())
 				malformed();
 
 			return m_atoms[pos];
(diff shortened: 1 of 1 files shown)


src/jrd/replication/Applier.cpp
```diff
@@ -152,7 +160,7 @@ namespace
 
 		string getString()
 		{
-			const auto length = getInt32();
+			const ULONG length = static_cast<ULONG>(getInt32());
 
 			if (m_data + length > m_end)
```
aafemt: It is better to leave length as is and use a correct condition instead: `(length <= 0 || m_end - m_data < length)`.

XaBbl4: I thought about this, but in the case of an unsigned string, the length might be longer. In what case might it be useful to have a negative string length?

XaBbl4: Ok, back length as is in getString.
Also add check length in defineAtom.

aafemt: We should ask @dyemanov why he chose signed result for getIntXX functions.

May be you are right about maximum string length. In this case the right condition avoiding overflow on the end of address space still would be `(m_end - m_data < length)`. Only the last gigabyte is reserved by OSes IIRC so if the buffer by change is just below it, `m_data + length` may be a little above zero causing false negative.

aafemt: Maybe zero-sized string is allowed. @dyemanov ?

dyemanov: > We should ask @dyemanov why he chose signed result for getIntXX functions.

I tried to make getters generic, as I couldn't predict whether signed numbers could ever appear inside the replication protocol or not. Given that numbers > 2^31 are not expected this looked safe.

src/jrd/replication/Applier.cpp
```diff
@@ -127,13 +127,21 @@ namespace
 
 		const string& getAtomString()
 		{
-			const auto pos = getInt32();
+			const ULONG pos = static_cast<ULONG>(getInt32());
+
+			if (pos >= m_atoms.getCount())
```
dyemanov: I'd prefer explicit checks with a signed `pos`:

`if (pos < 0 || pos >= m_atoms.getCount())`

src/jrd/replication/Applier.cpp
```diff
@@ -154,7 +162,7 @@ namespace
 		{
 			const auto length = getInt32();
 
-			if (m_data + length > m_end)
+			if (length <= 0 || m_end - m_data < length)
```
dyemanov: I'd allow zero-length strings even if it looks impossible now.

src/jrd/replication/Applier.cpp
```diff
@@ -127,13 +127,21 @@ namespace
...
+
 			return m_atoms[pos];
 		}
 
 		const MetaString getAtomMetaName()
 		{
-			const auto pos = getInt32();
+			const ULONG pos = static_cast<ULONG>(getInt32());
+
+			if (pos >= m_atoms.getCount())
```
dyemanov: Same as above.

src/jrd/replication/Applier.cpp
```diff
@@ -185,6 +193,9 @@ namespace
 		void defineAtom()
 		{
 			const auto length = getByte();
+			if (length <= 0)
```
dyemanov: Same as above.

dyemanov (changes requested):

src/jrd/replication/Applier.cpp
XaBbl4: Removed this check, because length is unsigned number and there is no need to check for negative numbers.

dyemanov (approved):

PR #9058, opened 2026-06-11
Declared Local Temporary Tables in PSQL

asfernandes:


AlexBekhtin:
Is infinite execution correct by design?

```sql
execute block returns (n integer)
as
    declare local temporary table t (
        id integer not null
    );
begin
     insert into t(id)    
     values(1);

     insert into t(id) 
     select id from t;

    select count(*) from t into n;
    suspend;
end!
```

asfernandes:
> Is infinite execution correct by design?

No, current local table blr (as used for RETURNING) is fragile.
I'm reviewing the implementation details.

sim1984:
Please add documentation about declared ltt restrictions, specifically the number of LTTs per operator.

Could you add the ability to declare indexes for such tables?

asfernandes:
> Could you add the ability to declare indexes for such tables?

I'm working on it. Will be added in another PR.

doc/sql.extensions/README.declared_local_temporary_tables.md
```diff
@@ -0,0 +1,255 @@
...
+  DDL objects.
+- Packaged temporary tables (`TEMPORARY TABLE ...` inside packages), whose definitions are persistent package-owned
+  metadata.
+
+## Syntax
+
+Declared Local Temporary Tables are declared in the same declaration section as variables, cursors and subroutines.
+
+```sql
+DECLARE LOCAL TEMPORARY TABLE <table_name>
```
dyemanov: I'd make `LOCAL` optional. IMHO, the fact that it's declared inside the PSQL module already implies it being local to that module.

src/dsql/dsql.h
```diff
@@ -169,7 +170,8 @@ enum rel_flags_vals {
 	REL_view			= 4, // relation is a view
 	REL_external		= 8, // relation is an external table
 	REL_creating		= 16,	// we are creating the bare relation in memory
-	REL_ltt_created		= 32	// relation is created local temporary table
+	REL_ltt_created		= 32,	// relation is created local temporary table
+	REL_local_table		= 64	// relation is a PSQL declared local table
```
dyemanov: Maybe `REL_ltt_declared` or `REL_ltt_psql`? `REL_local_table` name looks like more generic and including `REL_ltt_created` too.

src/jrd/Statement.cpp
```diff
@@ -1036,6 +1058,50 @@ bool Request::isRoot() const
...
 }
 
+Request* Request::getLocalTableRequest(bool outerDecl)
+{
+	Request* request = this;
+
+	if (outerDecl)
+	{
+		while (request->getStatement()->parentStatement)
+			request = request->req_caller;
```
dyemanov, line 1068: The statement looks weird. I'd expect the condition to be `while (request->req_caller)`. Please elaborate if I'm missing something.

src/jrd/Statement.cpp
```diff
@@ -1036,6 +1058,50 @@ bool Request::isRoot() const
...
+
+jrd_tra* Request::getLocalTableTransaction() const
+{
+	Stack<AutoTranCtx>::const_iterator autoTran(req_auto_trans);
+
+	if (autoTran.hasData())
+		return autoTran.object().m_transaction;
+
+	return req_transaction;
+}
```
dyemanov, line 1079: I'd go for a single line `return autoTran.hasData() ? autoTran.object().m_transaction : req_transaction;` but leave it up to you.

src/jrd/Statement.cpp
```diff
@@ -1036,6 +1058,50 @@ bool Request::isRoot() const
...
+
+FB_UINT64 Request::getLocalTableInstanceId(thread_db* tdbb) const
+{
+	if (!req_local_table_instance_id)
+		req_local_table_instance_id = tdbb->getDatabase()->generateStatementId();
```
dyemanov: Maybe worth adding a dedicated counter to the `TipCache`?

src/jrd/Savepoint.h
```diff
@@ -81,7 +81,8 @@ namespace Jrd
 	{
 	public:
 		VerbAction()
-			: vct_next(NULL), vct_records(NULL), vct_undo(NULL)
+			: vct_next(NULL), vct_relation(NULL), vct_records(NULL), vct_undo(NULL),
+			  vct_temp_instance_id(0)
```
dyemanov: Please move all initialization to member declarations.

src/jrd/Savepoint.cpp
```diff
@@ -385,11 +394,16 @@ void VerbAction::release(jrd_tra* transaction)
...
-VerbAction* Savepoint::createAction(jrd_rel* relation)
+VerbAction* Savepoint::createAction(thread_db* tdbb, jrd_rel* relation, FB_UINT64 tempInstanceId)
 {
 	// Create action for the given relation. If it already exists, just return.
 
-	VerbAction* action = getAction(relation);
+	if (!(relation->getPermanent()->rel_flags & REL_temp_frame))
+		tempInstanceId = 0;
+	else if (!tempInstanceId)
+		tempInstanceId = tdbb->tdbb_temp_frame_id;
```
dyemanov: May it be replaced with:

```c++
if (!tempInstanceId)
    tempInstanceId = relation->getTempInstanceId(tdbb);
```
?

src/jrd/Relation.h
```diff
@@ -903,7 +905,7 @@ class RelationPermanent : public Firebird::PermanentStorage
 	RelationPages* getPages(thread_db* tdbb, TraNumber tran = MAX_TRA_NUMBER, bool allocPages = true);
 	void	fillPages(thread_db* tdbb);
 	RelationPages* getAttPages(thread_db* tdbb, RelationPages::InstanceId inst_id);
-	bool	delPages(thread_db* tdbb, TraNumber tran = MAX_TRA_NUMBER, RelationPages* aPages = NULL);
+	bool	delPages(thread_db* tdbb, RelationPages::InstanceId inst_id = MAX_TRA_NUMBER, RelationPages* aPages = NULL);
```
dyemanov, line 910: Given that `delPages()` calls `getPages()` with the same `inst_id` argument, shouldn't `getPages()` to be also changed from `TraNumber` to `RelationPages::InstanceId` ?

src/dsql/StmtNodes.cpp
```diff
@@ -8756,15 +9135,28 @@ const StmtNode* ModifyNode::modify(thread_db* tdbb, Request* request, WhichTrigg
...
+				if (localTable)
+					DeclareLocalTableNode::validateRecord(localTable, newRpb->rpb_record);
+
+				if (!relation)
+				{
+					fb_assert(localTable && !localTable->useLtt);
+					fb_assert(false);
+					ERR_post(Arg::Gds(isc_wish_list));
+				}
+				else if (auto* extFile = relation->getExtFile())
```
dyemanov: No need for `else` after `ERR_post()`.

src/dsql/StmtNodes.cpp
```diff
@@ -3074,8 +3406,11 @@ const StmtNode* EraseNode::erase(thread_db* tdbb, Request* request, WhichTrigger
 			if (!statement)
 				break;
 
-			const Format* format = rpb->rpb_relation->currentFormat(tdbb);
-			Record* record = VIO_record(tdbb, rpb, format, tdbb->getDefaultPool());
+			const Format* format = relation ? rpb->rpb_relation->currentFormat(tdbb) : localTable->format.getObject();
```
dyemanov, line 3409: Please explain why `relation` may be missing here (and other similar methods)? This seems to seriously complicate the logic.

src/dsql/StmtNodes.cpp
```diff
@@ -3133,12 +3470,19 @@ const StmtNode* EraseNode::erase(thread_db* tdbb, Request* request, WhichTrigger
...
+		preModifyEraseTriggers(tdbb, relation->rel_triggers[TRIGGER_PRE_ERASE], whichTrig, rpb, NULL, TRIGGER_DELETE);
 
-	if (auto* extFile = relation->getExtFile())
+	if (!relation)
+	{
+		fb_assert(localTable);
+		fb_assert(false);
+		ERR_post(Arg::Gds(isc_wish_list));
+	}
+	else if (auto* extFile = relation->getExtFile())
```
dyemanov: No need for `else` after `ERR_post()`.

src/dsql/StmtNodes.cpp
asfernandes: Multi-row RETURNING uses an old implementation of "local table" based in RecordBuffer.

doc/sql.extensions/README.declared_local_temporary_tables.md
asfernandes: It created additional reduce/reduce conflict.

asfernandes:
@dyemanov do you have any more things before this can be merged?

src/dsql/StmtNodes.cpp
dyemanov: But multi-row `RETURNING` was introduced before this PR and `rpb_relation` was mandatory. Also, how can we `INSERT/UPDATE/DELETE` with a result set generated by `RETURNING`? I seem to miss something here.

asfernandes: > But multi-row `RETURNING` was introduced before this PR and `rpb_relation` was mandatory. Also, how can we `INSERT/UPDATE/DELETE` with a result set generated by `RETURNING`? I seem to miss something here.

The old RecordBuffer implementation (blr_local_table_id) should be faster for its purpose, i.e., it does not need record visibility logic, so it continue to be used for multi-row RETURNING. And it did not used "relations".

dyemanov (approved):

pavel-zotov:
::: QA note :::
Tests for this feature are in dedicated [folder](https://github.com/FirebirdSQL/firebird-qa/blob/master/tests/functional/dtt/)

PR #9065, opened 2026-06-16
Package Constant: Check missing dsc_address instead of dsc_dtype and restore impure after using existing request

Noremos:
Original report: https://groups.google.com/g/firebird-devel/c/uPpRhaAKT0M

In the `makeValue` function, the value is compiled only once and then cached. Previously, the `dsc_address`  was always present, but now it is not required for calculating the hash, so it may be missing. This leads to null pointer read/write.

The second issue is related with `request impure space`. First, the calls to `pass1` and `pass2` were missing, leading to invalid addresses being used. Second, using an existing request, its impure will be corrupt. So, it should be restored after the node is executed.

SQL to reproduce:
```sql
shell rm -rf /tmp/test.fdb;
create database ' /tmp/test.fdb';

set bail on;
set list on;
set autoterm on;
set autoddl off;
set keep_tran on;


recreate package pg_test as
begin
    procedure sp_test() returns(o smallint);
end;

recreate package body pg_test as
begin
    constant PG_BODY_CONST smallint = -10;
    procedure sp_test() returns(o smallint) as
    begin
        o = pg_body_const;
        suspend;
    end
end
;
commit;

select p.o as get_pkg_head_const_1  from pg_test.sp_test as p rows 1; -- segfault
```

### DIFF
--- a/src/jrd/Package.epp
+++ b/src/jrd/Package.epp
@@ -195,10 +195,18 @@ dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
 		bid constantBid = getBlobId(tdbb);
 
 		MET_parse_blob(tdbb, &name.schema, nullptr, &constantBid, &csb, nullptr, false, false);
+		if (request != nullptr)
+		{
+			// Allocate new impure data in a correct position
+			csb->csb_impure = request->impureArea.getCount();
+		}
+
 		fb_assert(csb->csb_node != nullptr);
 		csb->csb_node->pass1(tdbb, csb);
 		csb->csb_node->pass2(tdbb, csb);
 
+		// request is null when this function called inside Package::scan (without MINISCAN)
+		// request is not null when this function called in PackageReferenceNode::execute
 		if (request == nullptr)
 		{
 			// Came from Reload
@@ -219,25 +227,13 @@ dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
 		}
 		else
 		{
-			// A csb->csb_node will use impureArea from the incoming request
-			// but it contains other data
-			// Backup the current data and restore it later
-
-			// Reusing a small chunk is faster instead of makeing a new request
-			// And using existing space for a temporally value is better instead of adding this to request->impureArea
-			// and carrying it for whole request live
-
-			const auto nodeImpureSize = csb->csb_impure;
-			if (request->impureArea.getCount() < nodeImpureSize)
-				request->impureArea.grow(nodeImpureSize - request->impureArea.getCount());
-
-			const Array<UCHAR> backup(request->impureArea.begin(), nodeImpureSize);
-
+			if (csb->csb_impure > request->impureArea.getCount())
+			{
+				// impureArea does not automatically extend. Do it manually
+				request->impureArea.grow(csb->csb_impure);
+			}
 			fb_assert(tdbb->getRequest() == request);
 			executeCsbNode(tdbb, csb);
-
-			// Restore
-			memcpy(request->impureArea.begin(), backup.begin(), nodeImpureSize);
 		}
 	}
 	catch (const Exception& ex)
--- a/src/jrd/Package.epp
+++ b/src/jrd/Package.epp
@@ -197,7 +197,6 @@ dsc* ConstantValue::makeValue(thread_db* tdbb)
 
 		MET_parse_blob(tdbb, &name.schema, nullptr, &constantBid, &csb, nullptr, false, false);
 
-		// Came from Reload
 		handmadeStmt = Statement::makeStatement(tdbb, csb, true);
 		request = handmadeStmt->makeRootRequest(tdbb);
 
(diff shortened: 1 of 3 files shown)


aafemt:
Execution nodes are supposed to reserve space in impure area using `CompilerScratch::allocImpure` during compilation, remember the returned offset and during execution get the address for usage with `Request::getImpure`.

src/jrd/Package.epp
```diff
@@ -196,6 +196,8 @@ dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
 
 		MET_parse_blob(tdbb, &name.schema, nullptr, &constantBid, &csb, nullptr, false, false);
 		fb_assert(csb->csb_node != nullptr);
+		csb->csb_node->pass1(tdbb, csb);
+		csb->csb_node->pass2(tdbb, csb);
 
 		if (request == nullptr)
```
asfernandes: When is this function called with null request and when is it called with not null request?

src/jrd/Package.epp
```diff
@@ -216,7 +218,27 @@ dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
...
+				request->impureArea.grow(nodeImpureSize - request->impureArea.getCount());
+
+			const Array<UCHAR> backup(request->impureArea.begin(), nodeImpureSize);
+
+			fb_assert(tdbb->getRequest() == request);
 			executeCsbNode(tdbb, csb);
+
+			// Restore
+			memcpy(request->impureArea.begin(), backup.begin(), nodeImpureSize);
+		}
```
asfernandes: This whole thing is looking as a complete hack for me.

src/jrd/Package.epp
Noremos: With request: when executing ReferenceNode.
Without request: when performing a packet scan (without minican) in metacache.
I will add a comment to clearify it

src/jrd/Package.epp
Noremos: Got it. I just wanted to save time. I'll do it the right way

src/jrd/Package.epp
AlexPeshkoff: BTW - why not save descriptor for getDesc() in a cache? I.e. move almost all code from getDesc() to scan().

src/jrd/Package.epp
Noremos: Fixed

src/jrd/Package.epp
Noremos: Added a comment about null request

> BTW - why not save descriptor for getDesc() in a cache? I.e. move almost all code from getDesc() to scan().

Good point: getDesc queries the descriptor by name, but there's nothing stopping one from getting it via `package->findConstant`. Especially considering the function is already called when checking for the constant's existence. I'll create a separate PR with this optimization.

Noremos: The PR with the improveemnt: https://github.com/FirebirdSQL/firebird/pull/9069

src/jrd/Package.epp
asfernandes: Why are you not building a statement/request which is standard way to execute code?
There are many things allowed in constants. How you do ensure that all that code and new code is prepared to take that shortcut?

Noremos: > There are many things allowed in constants. How you do ensure that all that code and new code is prepared to take that shortcut?

You are right. Now the handmade requset will be used

src/jrd/Package.epp
```diff
@@ -168,22 +168,23 @@ bid ConstantValue::getBlobId(thread_db* tdbb)
 	return m_blrBlobId.isEmpty() ? getConstantBid(tdbb, name) : m_blrBlobId;
 }
 
-dsc* ConstantValue::makeValue(thread_db* tdbb, Request* request)
+dsc* ConstantValue::makeValue(thread_db* tdbb)
 {
 	{
 		ReadLockGuard guard(m_makeValueLock, FB_FUNCTION);
-		if (m_value.vlu_desc.dsc_dtype != dtype_unknown)
+		if (m_value.vlu_desc.dsc_address != nullptr)
```
asfernandes, line 175:
```suggestion
		if (m_value.vlu_desc.dsc_address)
```

src/jrd/Package.epp
```diff
@@ -168,22 +168,23 @@ bid ConstantValue::getBlobId(thread_db* tdbb)
...
 			return &m_value.vlu_desc;
 	}
 
 	WriteLockGuard guard(m_makeValueLock, FB_FUNCTION);
-	if (m_value.vlu_desc.dsc_dtype != dtype_unknown) // Extra check in case of two callers waiting
+	if (m_value.vlu_desc.dsc_address != nullptr) // Extra check in case of two callers waiting
```
asfernandes, line 180:
```suggestion
	if (m_value.vlu_desc.dsc_address) // Extra check in case of two callers waiting
```

asfernandes (approved):

src/jrd/Package.epp
aafemt: Using C-style implicit cast is generally bad idea in C++.

src/jrd/Package.epp
Noremos: I'm sorry, I haven't been able to get online for the past week and didn't see your suggestions

pavel-zotov:
::: QA note :::
[test_d197ce49.py](https://github.com/FirebirdSQL/firebird-qa/blob/master/tests/functional/package/test_d197ce49.py)

PR #9066, opened 2026-06-17
Add BLB_write function to allow modification of data in a BLOB

Noremos:
This PR is part of the JSON implementation. Efficiently writing binary JSON requires modifying a blob. Otherwise, TempSpace is required to use it and convert the data to a blob, significantly increasing overhead.

Additions:

1. New blob functions: BLB_write. It works the same as TempSpace::write
2. New blob functions: BLB_read. It works the same as BLB_seek + BLB_get_data
3. BLB_lseek now works with BLB_put_data (to modify and append data) 
4. BLB_lseek now works with BLB_put_segment (to modify and append data)


hvlad:
Whats wrong or not enough with seek + putSegment ?

Noremos:
> Whats wrong or not enough with seek + putSegment ?

`BLB_lseek` works only for `BLB_get_segment`

hvlad:
> > Whats wrong or not enough with seek + putSegment ?
> 
> `BLB_lseek` works only for `BLB_get_segment`

So why not make it work with `put` too? I suppose it's been done in your PR anyway.

Noremos:
> > > Whats wrong or not enough with seek + putSegment ?
> > 
> > 
> > `BLB_lseek` works only for `BLB_get_segment`
> 
> So why not make it work with `put` too? I suppose it's been done in your PR anyway.

Initially, I wanted to implement the same read/write interface as for TempSpace. But I can do this within the `BLB_get_segment` function.

hvlad:
> > > > Whats wrong or not enough with seek + putSegment ?
> > > 
> > > 
> > > `BLB_lseek` works only for `BLB_get_segment`
> > 
> > 
> > So why not make it work with `put` too? I suppose it's been done in your PR anyway.
> 
> Initially, I wanted to implement the same read/write interface as for TempSpace. 

You still speak about blobs here ?

> But I can do this within the `BLB_get_segment` function.

I'm lost your point here, sorry. How it answers on my question ?

hvlad:
BTW, code in blb.cpp (at least) requires comments. It have ~none~ too small useful comments, imo.

sim1984:
Why is `JsonTempParsingTests.cpp` needed in this PR?

Noremos:
> > > > > Whats wrong or not enough with seek + putSegment ?
> > > > 
> > > > 
> > > > `BLB_lseek` works only for `BLB_get_segment`
> > > 
> > > 
> > > So why not make it work with `put` too? I suppose it's been done in your PR anyway.
> > 
> > 
> > Initially, I wanted to implement the same read/write interface as for TempSpace.
> 
> You still speak about blobs here ?
> 
> > But I can do this within the `BLB_get_segment` function.
> 
> I'm lost your point here, sorry. How it answers on my question ?

I meant that the TempSpace class has convenient read/write functions with the ability to specify a position, so it would be great to have the same function for blobs at some point.

As for `BLB_put_data` and `BLB_put_segment`, I initially didn't want to change them, as the code in them is quite complex and modifying it would be quite difficult. But now I see that the integration is quite simple, so I did it.

So now `BLB_write works` the same as `BLB_lseek` + `BLB_put_data` (or `BLB_lseek` + `BLB_put_segment`).

Noremos:
> Why is `JsonTempParsingTests.cpp` needed in this PR?

Sorry, forgot to delete this file after the branch checkout.

dyemanov:
If you say `BLB_write()` is convenient, I'd also expect `BLB_read()` to be added for symmetry.

aafemt:
Do new methods of `blb` really need this ancient `BLB_` prefix?..

Noremos:
> BTW, code in blb.cpp (at least) requires comments. It have ~none~ too small useful comments, imo.

Added description of modification algorithm:
https://github.com/FirebirdSQL/firebird/pull/9066/changes#diff-b38a2d18c3f2097dc3697386a2d4d4cb6c708408ac2cdad4169aca72bdf2479dR2051-R2079

Noremos:
> If you say `BLB_write()` is convenient, I'd also expect `BLB_read()` to be added for symmetry.

Added as suggested

src/jrd/blb.h
```diff
@@ -133,6 +134,14 @@ class blb : public pool_alloc<type_blb>
 		return destination;
 	}
 
+	// Read form specified position
+	FB_SIZE_T BLB_read(thread_db* tdbb, const offset_t position, void* buffer, const ULONG length);
+
+	// Write data at any position in a temporally (new) blob
+	// The position of the new buffer must start inside the blob range, but its length may extend beyond it
+	// Existing data will be overwritten
+	void BLB_write(thread_db* tdbb, const offset_t position, const void* buffer, ULONG length);
```
hvlad: Are these two functions supposed to be a public API ? I see no changes in interfaces.
If not - it should not start with `BLB_`

src/jrd/blb.h
```diff
@@ -142,6 +151,41 @@ class blb : public pool_alloc<type_blb>
 	void insert_page(thread_db*);
 	void destroy(const bool purge_flag);
 
+	// Modify only existing data. Throw error on side violation
+	void modifyExistingData(thread_db* tdbb, offset_t position, const void* buffer, const ULONG length);
```
hvlad: Sorry, am I missed a way to modify non-existing data ? :)

src/jrd/blb.h
```diff
@@ -142,6 +151,41 @@ class blb : public pool_alloc<type_blb>
...
+
+	// Modify existing data
+	// Output:
+	//     true: the input range is only inside the blob data
+	//     false: the input range is extends beyond existing data. Modify `buffer` and `length` to return only non-written data
+	template<class BufferType, class SizeType>
+	requires((std::is_same_v<BufferType, void> || std::is_same_v<BufferType, UCHAR>) && std::is_integral_v<SizeType>)
+	bool modifyDataMoveBuffer(thread_db* tdbb, const offset_t position, const BufferType*& buffer, SizeType& length)
```
hvlad: DataMoveBuffer  ? What is it ? We already have MoveBuffer used in MOV\CVT, and it is completely different thing. 

Why template here ? Why it is inline routine ?

src/jrd/blb.cpp
```diff
@@ -90,6 +90,82 @@ static void move_to_string(Jrd::thread_db*, dsc*, dsc*);
...
+	DataModifyHelper(thread_db* tdbb, const vcl* blb_pages, const offset_t position, const void* buffer, const ULONG length) :
+		m_newData(buffer), m_newLength(length),
+		m_level1Pages(*blb_pages),
+		m_pageDataLength(tdbb->getDatabase()->dbb_page_size - BLP_SIZE)
+	{
+		m_level1PageId = position / m_pageDataLength; // Number of used pages
+		m_offset = position % m_pageDataLength; // Position in the page
+	}
+
+	// Get data from blob data page and replace data on it
```
hvlad: Get data ?

src/jrd/blb.cpp
```diff
@@ -90,6 +90,82 @@ static void move_to_string(Jrd::thread_db*, dsc*, dsc*);
...
+		UCHAR* data = reinterpret_cast<UCHAR*>(page->blp_page);
+		const ULONG dataLength = std::min<ULONG>(m_pageDataLength - m_offset, m_newLength - m_written);
+		fb_assert(dataLength <= m_pageDataLength);
+
+		memcpy(data + m_offset, reinterpret_cast<const UCHAR*>(m_newData) + m_written, dataLength);
+		m_written += dataLength;
+		m_offset = 0; // Offset only in the first page
+	};
+
+	// Move child page Id from level 1 to level 2
```
hvlad: This is completely unclear. 
There is no way to "move" "page Id" (which "Id" ???) between "levels".

src/jrd/blb.cpp
```diff
@@ -90,6 +90,82 @@ static void move_to_string(Jrd::thread_db*, dsc*, dsc*);
...
+		return pageId % pagesOnRootPage;				// 100000 % 8000 = 4000 // level2 page number
+	}
+
+	// Get level 1 or level 2 page
+	inline ULONG getNextLevel1PageId() noexcept
+	{
+		return m_level1Pages[m_level1PageId++];
+	}
+
+	// Pages are over, write to buffer
```
hvlad: You mean - pointer pages ? Data pages ? Anything else ?

src/jrd/blb.cpp
```diff
@@ -1573,6 +1649,21 @@ void blb::BLB_put_data(thread_db* tdbb, const UCHAR* buffer, SLONG length)
 	SET_TDBB(tdbb);
 	const BLOB_PTR* p = buffer;
 
+	// BLB_put_segment will remove the flag after teh first call so replace the data here
```
hvlad: teh -> the

src/jrd/blb.cpp
```diff
@@ -90,6 +90,82 @@ static void move_to_string(Jrd::thread_db*, dsc*, dsc*);
...
+	// Move child page Id from level 1 to level 2
+	inline ULONG setLevel2(const USHORT pagesOnRootPage)
+	{
+		const auto pageId = m_level1PageId;
+		m_level1PageId = pageId / pagesOnRootPage;		// 100000 / 8000 = 12	// level1 page number
+		return pageId % pagesOnRootPage;				// 100000 % 8000 = 4000 // level2 page number
+	}
+
+	// Get level 1 or level 2 page
+	inline ULONG getNextLevel1PageId() noexcept
```
hvlad: Unclear as above.

src/jrd/blb.cpp
```diff
@@ -1944,6 +2046,131 @@ void blb::scalar(thread_db*		tdbb,
...
+	//           To locate and modify the required page:
+	//           1. Calculate the page offset: OFFSET = position / <page size>.
+	//           2. Determine the target level 2 page: FIRST = OFFSET / <number of IDs per page>.
+	//           3. Compute the byte offset within the level 2 page:
+	//              BytesOffset = position % <page size>.
+	//           4. Modify the first relevant level 2 page, then move to the next one.
+	//           5. If no more level 2 pages are available, advance to the next level 1 page,
+	//              read its first level 2 page, and continue modifying subsequent level 2 pages.
+	//           6. If all pages have been processed but there is still unmodified data, update the <buffer>.
+
```
hvlad, line 2096: Thanks, but.. where you get the terminology you used ? I'm sorry, but it is very, very hard to read - I need constantly translate into familiar terms. 

There is no level of _blob pages_. There is level of _blob_. 

There are _blob pointer pages_ and _blob data pages_, and _blob record_ contains _blob data_ (for level-0 blobs) or _array of pointers_ (for non-0 level blobs).

src/jrd/blb.cpp
```diff
@@ -1944,6 +2046,131 @@ void blb::scalar(thread_db*		tdbb,
...
+	if (blb_flags & BLB_large_scan)
+	{
+		window.win_flags = WIN_large_scan;
+		window.win_scans = 1;
+	}
+
+	auto releasePage = [&tdbb, &window](const bool mark)
+	{
+		if (mark)
+			CCH_MARK(tdbb, &window); // Mark as dirty
```
hvlad: Page must be marked **before** any attempt to change the contents!

src/jrd/blb.cpp
```diff
@@ -3033,3 +3260,22 @@ void blb::BLB_cancel()
...
+{
+	// Mode 0 - from start
+	BLB_lseek(0, position);
+	return BLB_get_data(tdbb, reinterpret_cast<UCHAR*>(buffer), length, false);
+}
+
+void blb::BLB_write(thread_db* tdbb, const offset_t position, const void* buffer, ULONG length)
+{
+	if (!(blb_flags & BLB_temporary) || (blb_flags & BLB_closed))
+		ERR_post(Arg::Gds(isc_cannot_update_old_blob)); // Cannot update existing blob
```
hvlad: This check is duplicated many times, worth to move it in to separate routine.

src/include/firebird/impl/msg/jrd.h
```diff
@@ -1020,3 +1020,4 @@ FB_IMPL_MSG(JRD, 1017, dsql_agg_non_agg_context, -104, "42", "000", "Aggregate f
 FB_IMPL_MSG(JRD, 1018, dsql_agg_param_not_accum, -204, "42", "000", "Aggregate function input parameters may be referenced only in ON ACCUMULATE DO")
 FB_IMPL_MSG(JRD, 1019, dsql_agg_exit_group, -204, "42", "000", "EXIT is not allowed in ON GROUP DO section of aggregate function")
 FB_IMPL_MSG(JRD, 1020, dsql_agg_return, -204, "42", "000", "RETURN is not allowed in ON START DO, ON ACCUMULATE DO or ON FINISH DO sections of aggregate function; use EXIT instead")
+FB_IMPL_MSG(JRD, 1021, blob_out_of_length_write, -204, "42", "000", "Cannot write to blob. Position @1 is out of blob length @2")
```
hvlad: `blob_write_out_of_bounds` or `blob_write_after_the_end` ?

Same for the message text.

hvlad (changes requested):

hvlad:
Don't get me wrong - the code itself is not bad, but it is very hard to read, understand and verify.

src/jrd/blb.h
Noremos: So that's what that prefix means. I thought it was just style. Since the new write and read functions are just shorthand for `BLB_lseek` + `BLB_get_data`/`BLB_put_data`, I don't think there's any point in adding them to the API.
The prefix has been removed.

src/jrd/blb.h
Noremos: Yes... I wanted to clarify that the input range is specified correctly, but, yes, the name doesn't make sense.
Changed the name to `modifyData`

src/jrd/blb.h
Noremos: DataMoveBuffer - I was thinking of `modify data and move buffer`, but I understand it's not entirely clear. I changed it to "modifyBlobChunk." It's also not entirely clear, but I can't figure out the better option.

The tempalte is used here becase `BLB_put_segment` has `void*,USHORT` args, `BLB_put_data` has `UCHAR*, SLONG` args and blb::write has  `void*, ULONG` args. So at least 3 unique combinations.

And it inlined because of the template. The arguments are known, yes, but the implementation of template functions in C++ is extremely inconvenient.

src/jrd/blb.cpp
Noremos: Corrected the description

src/jrd/blb.cpp
Noremos: I changed the code to make it more clear

src/jrd/blb.cpp
Noremos: I added a separate class for iterating over page IDs, hopefully this will make code easier to understand.

src/jrd/blb.cpp
Noremos: Fixed, thanks

src/jrd/blb.cpp
Noremos: Sorry for the inaccuracy; I have virtually no experience working with pages, so I used the existing code as an example. It has a `blob_page` class for both the `BLOB data page` and the `BLOB pointer page`, so I misnamed it.
Thanks to your clarification, all the terminology is now correct (I hope).

src/jrd/blb.cpp
Noremos: Thanks for pointing it out. Fixed

src/jrd/blb.cpp
Noremos: I've moved the check to a new function: `verifyBlobModifiable`. But my eternal enemy, naming, is back.

src/include/firebird/impl/msg/jrd.h
Noremos: Changed to blob_write_after_the_end as suggested

PR #9069, opened 2026-06-22
Add package constant type to metacache

Noremos:
Implementation of the [proposed](https://github.com/FirebirdSQL/firebird/pull/9065#discussion_r3447028209) improvement


src/dsql/ExprNodes.cpp
```diff
@@ -6480,15 +6480,18 @@ ValueExprNode* FieldNode::internalDsqlPass(DsqlCompilerScratch* dsqlScratch, Rec
 		{
 			dsqlScratch->qualifyExistingName(constantName, obj_package_constant);
 
-			if (PackageReferenceNode::constantExists(tdbb, dsqlScratch->getTransaction(), constantName))
+			dsc consatntDsc{};
```
AlexPeshkoff: Mistype: constantDsc

Noremos: Fixed here and in other places

src/dsql/PackageNodes.epp
```diff
@@ -434,20 +437,23 @@ DmlNode* PackageReferenceNode::parse(thread_db* tdbb, MemoryPool& pool, Compiler
...
 				auto package = MetadataCache::getVersioned<Cached::Package>(tdbb, fullName.getSchemaAndPackage(), CacheFlag::AUTOCREATE);
 				if (package)
 				{
 					node->m_package = csb->csb_resources->packages.registerResource(package->getPermanent());
+					constant = package->findConstant(fullName);
 				}
 			}
 
-			if (!node->m_package)
+			if (!node->m_package || constant == nullptr)
```
AlexPeshkoff: constant == nullptr check should be enough

Noremos: Fixed as suggested

dyemanov:
Please resolve the conflicts.

Noremos:
> Please resolve the conflicts.

Resolved. Sorry for the delay

src/dsql/ExprNodes.cpp
```diff
@@ -6480,15 +6480,18 @@ ValueExprNode* FieldNode::internalDsqlPass(DsqlCompilerScratch* dsqlScratch, Rec
 		{
 			dsqlScratch->qualifyExistingName(constantName, obj_package_constant);
 
-			if (PackageReferenceNode::constantExists(tdbb, dsqlScratch->getTransaction(), constantName))
+			dsc constantDsc{};
+			auto transaction = dsqlScratch->getTransaction();
+			if (PackageReferenceNode::constantExists(tdbb, transaction, constantName, &constantDsc))
```
dyemanov: Given `transaction` is not used outside this call, I'd pass `dsqlScratch->getTransaction()` directly as you did in another `constantExists` call below.

src/jrd/Package.h
```diff
@@ -60,25 +60,30 @@ class ConstantValue final : public Firebird::PermanentStorage
...
 	bool hash(thread_db* tdbb, Firebird::sha512& digest) const;
 
-	static dsc getDesc(thread_db* tdbb, Jrd::jrd_tra* transaction, const QualifiedName& name);
-
 	static void genConstantBlr(thread_db* tdbb, DsqlCompilerScratch* dsqlScratch,
 		ValueExprNode* constExpr, dsql_fld* type, const MetaName& schema);
 
-
-	void updateValue(const dsc typeDesc)
+	void setType(const dsc typeDesc)
```
dyemanov:
```suggestion
	void setType(const dsc& typeDesc)
```

src/dsql/PackageNodes.epp
```diff
@@ -370,10 +370,14 @@ void PackageItemsHolder::clear()
...
-PackageReferenceNode::PackageReferenceNode(MemoryPool& pool, const QualifiedName& fullName, const UCHAR itemType)
+PackageReferenceNode::PackageReferenceNode(MemoryPool& pool, const QualifiedName& fullName, const UCHAR itemType,
+	const dsc* type)
 		: TypedNode<ValueExprNode, ExprNode::TYPE_PACKAGE_REFERENCE>(pool),
 		  m_fullName(fullName), m_itemType(itemType)
-{}
+{
+	if (type)
+		m_cachedDsc = *type;
+}
```
dyemanov: Given that address of variable (e.g. `&constantDsc`) is always passed as `type`, the `if` looks unnecessary and could be replaced with an assertion. Or even better `type` should be `const dsc&`.  For that, `PackageReferenceNode::parse()` should be adjusted to create `node` after `constant` has been found -- which looks like a good thing anyway.

src/dsql/PackageNodes.epp
```diff
@@ -474,13 +480,12 @@ void PackageReferenceNode::setParameterName(dsql_par* parameter) const
 
 void PackageReferenceNode::make(DsqlCompilerScratch* dsqlScratch, dsc* desc)
 {
-	jrd_tra* transaction = dsqlScratch->getTransaction();
-	thread_db* tdbb = JRD_get_thread_data();
-	*desc = ConstantValue::getDesc(tdbb, transaction, m_fullName);
+	fb_assert(m_cachedDsc.dsc_dtype != dtype_unknown);
```
dyemanov:
```suggestion
	fb_assert(!m_cachedDsc.isUnknown());
```

src/dsql/PackageNodes.epp
```diff
@@ -500,18 +505,22 @@ bool PackageReferenceNode::constantExists(thread_db* tdbb, Jrd::jrd_tra* transac
...
+		*outConstantDesc = value->getDesc();
+
 	return true;
 }
 
-void PackageReferenceNode::getDesc(thread_db* tdbb, CompilerScratch*, dsc* desc)
+void PackageReferenceNode::getDesc(thread_db*, CompilerScratch*, dsc* desc)
 {
-	*desc = ConstantValue::getDesc(tdbb, tdbb->getTransaction(), m_fullName);
+	fb_assert(m_cachedDsc.dsc_dtype != dtype_unknown);
```
dyemanov:
```suggestion
	fb_assert(!m_cachedDsc.isUnknown());
```

src/dsql/ExprNodes.cpp
Noremos: This was done primarily to improve readability. Passing `dsqlScratch->getTransaction()` directly increases the line length to 119 characters, which is very close to the recommended line limit. If this isn't an issue, I will make the changes as suggested.

src/jrd/Package.epp
```diff
@@ -55,8 +55,8 @@ static_assert(DSC_SIZE_TO_HASH - offsetof(dsc, dsc_flags) == sizeof(dsc::dsc_fla
 
 bool ConstantValue::hash(thread_db* tdbb, Firebird::sha512& digest) const
 {
-	fb_assert(m_value.vlu_desc.dsc_dtype != dtype_unknown);
-	digest.process(DSC_SIZE_TO_HASH, &m_value.vlu_desc);
+	fb_assert(m_type.dsc_dtype != dtype_unknown);
```
dyemanov:
```suggestion
	fb_assert(!m_type.isUnknown());
```

src/jrd/Package.h
```diff
@@ -91,8 +96,12 @@ class ConstantValue final : public Firebird::PermanentStorage
 	// Lock in case of makeing value during the execute state
 	Firebird::RWLock m_makeValueLock{};
 
-	// Keep type to gen hash (when not commited - we cannot read it from system table)
-	// Keep value when scanning and after the first execution
+	// The constant type and constant value type may differed
+	// For example, the the defiend type is int but value type is short)
+	// So keep it explicit
```
dyemanov:
```suggestion
	// The constant type and constant value type may be different
	// (e.g. the defined type is int, but the value type is short),
	// so keep it explicit
```

dyemanov (changes requested):

src/dsql/ExprNodes.cpp
dyemanov: I don't mind, but then please change it in both places and make `transaction` a `const auto`.

src/jrd/Package.h
Noremos: As far as I know, it is better to pass small structures (16 bytes or less) by value. However, the difference is negligible: https://quick-bench.com/q/tjak27_P1jOlDCnSom1mV0iCpcQ

src/dsql/PackageNodes.epp
Noremos: Changed as suggested

src/dsql/ExprNodes.cpp
Noremos: It turns out that `transaction` is no longer used in `PackageReference Node::constant Exists`, so I removed that argument.

src/jrd/Package.h
dyemanov: True for 8 bytes and maybe for 16 bytes as well, but `sizeof(dsc)` may suddenly increase and we'll loose the fast track.

dyemanov (approved):

PR #9091, opened 2026-07-14
Fix segfault when passing invalid `include` path in fbtrace config via fbtracemgr

Noremos:
Using the `include` directive allows you to insert configuration from a separate file. However, passing an invalid path causes the server to crash (for user sessions only). Using recursive include also leads to a crash

**Steps to reproduce**
1. create user NT password '123';
2. Create a file /tmp/trace_include_missing.conf with the following contents:
```
database
{
	enabled = true
}
include /a/b/c/d
```

3. Execute:
`./fbtracemgr -se <server ip/port>:service_mgr -user NT -password 123 -start -name nytrace -config /tmp/trace_include_missing.conf`

After this, a segfault will occur.


src/common/utils_proto.h
```diff
@@ -333,6 +333,29 @@ namespace fb_utils
...
+		}
+
+		constexpr bool has_value() const noexcept
+		{
+			return m_value != nullptr;
+		}
+
+	private:
+		T* m_value;
+	};
```
dyemanov, line 360: Why a new class instead of `std::optional<const char*>`?

Noremos: std::optional does not guarantee that the contained pointer is non-null. You can still accidentally wrap a nullptr in an std::optional or forget to validate the pointer before storing it. As a result, you often end up having to perform two checks: `optPointer.has_value() && optPointer.value() != nullptr` (or `optPointer.has_value(), fb_assert(optPointer.value() != nullptr`)

This makes std::optional dangerous to use for pointer types. A SafePointer, on the other hand, guarantees that the contained pointer is never nullptr, eliminating the need for the second check.

aafemt: Single check: `optPointer.value_or(nullptr) != nullptr`.

aafemt: And generally, if you check for `nullptr` anyway - a dumb pointer is enough. If you want `nullptr` to be unacceptable - a reference is used.

Noremos: > Single check: `optPointer.value_or(nullptr) != nullptr`.

But you can still accidentally call `value()` and get a null pointer. Why create unnecessary problems when they can easily be prevented with a simple wrapper?

> And generally, if you check for nullptr anyway - a dumb pointer is enough. If you want nullptr to be unacceptable - a reference is used.

This code requires passing pointers to functions. But overall, I agree: using references would be a good idea.

aafemt: First of all, these functions are not engraved in stone. They can be refactored.
Second, what exactly was the problem with a dumb pointer that was used before introduction of this helper?

src/common/config/config_file.cpp
```diff
@@ -518,19 +520,20 @@ bool ConfigFile::translate(const char* fileName, const String& from, String& to)
...
 	else if (from == "this")
 	{
-		if (!fileName)
+		if (!fileName.has_value())
 		{
 			return false;
 		}
 
-		PathName tempPath(fileName);
+		const char* fileNameData = fileName.value_or("");
```
aafemt, line 528: Here usage of `value_or` is meaningless because fileName without value cannot reach this point.

src/common/utils_proto.h
Noremos: > First of all, these functions are not engraved in stone. They can be refactored. Second, what exactly was the problem with a dumb pointer that was used before introduction of this helper?

 The problem was with null pointer dereference in the error message
`(Arg::Gds(isc_conf_include) << currentFileName << parPath << Arg::Gds(isc_include_miss)).raise();`

My simple fix prevents such errors once and for all. Without any hard refactoring. Just one new class and its done. Why are you getting so nervous about this?

src/common/config/config_file.cpp
```diff
@@ -753,24 +756,26 @@ void ConfigFile::parse(Stream* stream)
...
+		(Arg::Gds(isc_conf_include) << fileNameForError << parPath << Arg::Gds(isc_include_depth)).raise();
 	}
 
 	// for relative paths first of all prepend with current path (i.e. path of current conf file)
 	PathName path;
 	if (PathUtils::isRelative(parPath))
 	{
 		PathName dummy;
-		PathUtils::splitLastComponent(path, dummy, currentFileName);
+		PathUtils::splitLastComponent(path, dummy, currentFileName.value_or(""));
```
aafemt, line 781: The question is: how can you parse file without name?

aafemt (commented):
So, the only usage of the only member of this helper is meaningless.

src/common/config/config_file.cpp
Noremos: The answer is: look at the TextStreem class
https://github.com/Noremos/firebird/blob/abe42e79a49090362c7904003871de66919db7b4/src/common/config/config_file.cpp#L165-L168

src/common/utils_proto.h
aafemt: Your PR fixes only this exactly case. Other possible cases where `nullptr` is used in construction of an error vector are untouched. If the dereference was the problem, it had to be fixed in the point where the dereference happen, i.e. in `operator<<`. It would cover all possible cases, not just this one.

src/common/config/config_file.cpp
Noremos: Yes, but I decided not to include the `value` function to avoid potential errors.

src/common/utils_proto.h
Noremos: What exacly cases are you talking? I described the case, i fixed it. If you see other porblems with TextStream - let me known. I will include it into this fix. But if you just want to prove your point - please don't. I am tired of this pointless conversation

src/common/config/config_file.cpp
aafemt: So, whole this PR can be reduced to
```
        const char* getFileName() const
        {
-               return NULL;
+               return "Passed text";
        }
```

src/common/utils_proto.h
aafemt: In short: you found that some code crashes when `nullptr` is put into status vector as a value for `isc_arg_string`. That code must be fixed at first hand, this code (where `nullptr` is really used) is secondary.

src/common/config/config_file.cpp
Noremos: Probably, but would it be a good fix?

`getFileName` implements a `ConfigFile::Stream` class. If someone adds a new implementation without a proper `nullptr` check, the segfault can happen again. The current return type of `getFileName` allows `nullptr`, so the issue is still possible.

Changing the return type to something like `std::string_view` could solve the problem, but that would be more of a refactoring than a simple fix.

Since this can lead to server crashes, I think it is worth adding stronger pointer checks in this area. But its @dyemanov to decide

aafemt: A good fix would fix the source of the problem (`getFileName` in this case) and immediate reason for the crash (handling of `nullptr` in a status vector in this case). This PR massages something in between, that's why I don't consider it as a good fix.

aafemt:
BTW, looking into source I see that we probably have another problem here: path from `currentFileName` is appended to relative path in `include` directive and in the case of empty file name it will end up relative to the current directory which is often `system32` on Windows and root on Linux. It allows a remote attacker to make engine to read any file which is, IMHO, undesirable.

Noremos:
> BTW, looking into source I see that we probably have another problem here: path from `currentFileName` is appended to relative path in `include` directive and in the case of empty file name it will end up relative to the current directory which is often `system32` on Windows and root on Linux. It allows a remote attacker to make engine to read any file which is, IMHO, undesirable.

This issue is already being addressed privately with the Firebird team

Noremos:
This PR should fix a segfault on MacOS in the Unit Test IncludeInUserSessionBug

dyemanov (approved):

hvlad:
> This PR should fix a segfault on MacOS in the Unit Test IncludeInUserSessionBug

v5.0-release still crashes

Noremos:
> > This PR should fix a segfault on MacOS in the Unit Test IncludeInUserSessionBug
> 
> v5.0-release still crashes

Hello. This fix hasn't been backported to version 5 yet. Dmitry Yemanov wanted to do it, but I can create a PR myself.

hvlad:
Do it, please

Noremos:
> Do it, please

https://github.com/FirebirdSQL/firebird/pull/9138

hvlad:
Thanks. 
Please, set lables on both PR's

Noremos:
> Thanks. Please, set lables on both PR's

Unfortunately, I don't have access to change lables

hvlad:
> > Thanks. Please, set lables on both PR's
> 
> Unfortunately, I don't have access to change lables

Ok, I've set some, let @dyemanov set more if required.

PR #9092, opened 2026-07-14
fix build

Ev3nt:


sim1984:
This fixes Firebird's building, but there's more than one such iterator, and if we want to use standard algorithm functions elsewhere, we should do so for all Firebird containers. It's a good idea to review all containers in `src/common/classes`.

`Firebird:Array` works, but it should be updated to a more modern format.
`Firebird::Stack` - it also lacks categories.
`Firebird::HashTable` - it also lacks categories.

PR #9093, opened 2026-07-14
refactor(cast format): Improve error message about invalid format

TreeHunter9:
The issue was reported by Pavel in [devel](https://groups.google.com/g/firebird-devel/c/aNkUxNtur88).


pavel-zotov:
::: QA note :::
See test for #8501

PR #9100, opened 2026-07-20
fix(client): Clear internal client buffer on failed packet send

TreeHunter9:
Previously a failed send left stale bytes in the buffer, which got prepended to the next packet.

### DIFF
--- a/src/remote/client/interface.cpp
+++ b/src/remote/client/interface.cpp
@@ -9801,6 +9801,15 @@ static void send_packet(rem_port* port, PACKET* packet)
  *
  **************************************/
 
+	auto buffer_cleanup = [](RemoteXdr* xdr)
+	{
+		if (xdr == nullptr)
+			return;
+
+		xdr->x_handy += xdr->x_private - xdr->x_base;
+		xdr->x_private = xdr->x_base;
+	};
+
 	RefMutexGuard guard(*port->port_write_sync, FB_FUNCTION);
 
 	if (port->port_flags & PORT_detached || port->port_state == rem_port::BROKEN)
@@ -9823,8 +9832,12 @@ static void send_packet(rem_port* port, PACKET* packet)
 			if (!p->sent)
 			{
 				if (!port->send_partial(&p->packet))
+				{
+					buffer_cleanup(port->port_send);
+
 					(Arg::Gds(isc_net_write_err) <<
 					 Arg::Gds(isc_random) << "send_packet/send_partial").raise();
+				}
 
 				p->sent = true;
 			}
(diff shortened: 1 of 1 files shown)


TreeHunter9:
The incorrect behavior can be seeing in this example:
```c
#include "ibase.h"
#include <stdio.h>
#include <string.h>

static void print_status_vector(const ISC_STATUS *status_vector) {
	char buffer[512] = {};
	while (fb_interpret(buffer, 512, &status_vector))
		printf("%s\n", buffer);
}

int main(int argc, char **argv) {
	const char *db  = "localhost:employee", 
			   *usr = "sysdba",
			   *pwd = "masterkey";

	ISC_STATUS_ARRAY st = {0};
	isc_db_handle dbh = 0;
	isc_tr_handle tr = 0;
	char dpb[256]; int dl = 0;
	dpb[dl++] = isc_dpb_version1;
	dpb[dl++] = isc_dpb_user_name;
	dpb[dl++] = (char)strlen(usr);
	memcpy(dpb + dl, usr, strlen(usr));
	dl += strlen(usr);
	dpb[dl++] = isc_dpb_password;
	dpb[dl++] = (char)strlen(pwd);
	memcpy(dpb + dl, pwd, strlen(pwd));
	dl += strlen(pwd);

	if (isc_attach_database(st, strlen(db), (char *)db, &dbh, dl, dpb)) {
		print_status_vector(st);
		return 1;
	}
	if (isc_start_transaction(st, &tr, 1, &dbh, 0, NULL)) {
		print_status_vector(st);
		return 1;
	}

	ISC_ARRAY_DESC desc;
	desc.array_desc_dtype = 8;
	desc.array_desc_scale = 0;
	desc.array_desc_length = 4;
	memcpy(desc.array_desc_field_name, "A", 1);
	memcpy(desc.array_desc_relation_name, "ARR", 3);
	desc.array_desc_dimensions = 1;
	desc.array_desc_flags = 0;
	desc.array_desc_bounds[0].array_bound_lower = 1;
	desc.array_desc_bounds[0].array_bound_upper = 3;

	ISC_QUAD aid = {0, 1};

	unsigned char sdl[64]; int n = 0;
	sdl[n++] = 2;

	char slice[256] = {0}; ISC_LONG slen = 0;
	// Will throw an error in the client lib, packet will not be send, but it is stored in buffer.
	isc_get_slice(st, &dbh, &tr, &aid, (unsigned short)n, (const ISC_UCHAR *)sdl,
				0, NULL, (ISC_LONG)sizeof(slice), slice, &slen);
	print_status_vector(st);

	// Will send two packets, incomplete `isc_get_slice` and `isc_commit_transaction`.
	// The client will get an error and will think that it is from `isc_commit_transaction`, but it actually from `isc_get_slice` API call.
	isc_commit_transaction(st, &tr);
	print_status_vector(st);
	isc_detach_database(st, &dbh);
	print_status_vector(st);
	return 0;
}
```

PR #9106, opened 2026-07-25
Implement a way to build an index without blocking concurrent data modifications

### LINKED ISSUE #7967: Implement a way to build an index without blocking concurrent data modifications
Currently, when index is build, any modifications of table data is not allowed. This is required to create correct index not missing new keys inserted during the build.

The time when such read lock is required could be significantly shortened.

hvlad:


AlexPeshkoff:
On 7/25/26 20:30, Vlad Khorsun wrote:

>   * 91ac1e6
>     <https://github.com/FirebirdSQL/firebird/pull/9106/commits/91ac1e6b66a6105bbbdcc5e7a3fa59adef7a82b5>
>     A way to build an index without blocking concurrent data
>     modifications.
>

Vlad, there is one main problem with your approach - option CONCURRENTLY 
silently affects ability to have NOT VALIDATED index instead fatal error 
for unique indices. I see the following problems:
- users often miss warnings, i.e. once upon a time using CONCURRENTLY 
option they can get active but not validated index and do not know about 
it - specially taking into an account that new records are forced to be 
unique;
- correct logic with hard job, done for index creation, that should 
better be not repeated applies to traditional (without option 
CONCURRENTLY) index creation too, but getting NOT VALIDATED index is not 
supported in this case.

What about adding one more option that controls ability to get not 
validated index? As far as I can see it's not dependent from concurrent 
creation and may be used separately.

hvlad:
> Vlad, there is one main problem with your approach - option CONCURRENTLY 
silently affects ability to have NOT VALIDATED index instead fatal error 
for unique indices. I see the following problems:
>- users often miss warnings, i.e. once upon a time using CONCURRENTLY 
option they can get active but not validated index and do not know about 
it - specially taking into an account that new records are forced to be 
unique;

It depends on tool used. We can't control third party tools. 
Our isql show warnings, while it is hard to distinguish them from errors.

From the other side: the concurrent index creation is a **new** feature with a **new** 
syntax and its behaviour re. unique indices is explicitly documented. I.e. if one
uses CONCURRENTLY, it is supposed (s)he read the docs before it ;)

>- correct logic with hard job, done for index creation, that should 
better be not repeated applies to traditional (without option 
CONCURRENTLY) index creation too, but getting NOT VALIDATED index is not 
supported in this case.
>
>What about adding one more option that controls ability to get not 
validated index? As far as I can see it's not dependent from concurrent 
creation and may be used separately.

Good idea, I support it. Could be useful for restore too.

AlexPeshkoff:
On 7/27/26 14:42, Vlad Khorsun wrote:
> *hvlad* left a comment (FirebirdSQL/firebird#9106) 
> <https://github.com/FirebirdSQL/firebird/pull/9106#issuecomment-5090850260>
>
>     Vlad, there is one main problem with your approach - option
>     CONCURRENTLY
>     silently affects ability to have NOT VALIDATED index instead fatal
>     error
>     for unique indices. I see the following problems:
>
>       * users often miss warnings, i.e. once upon a time using
>         CONCURRENTLY
>         option they can get active but not validated index and do not
>         know about
>         it - specially taking into an account that new records are
>         forced to be
>         unique;
>
> It depends on tool used. We can't control third party tools.
> Our isql show warnings, while it is hard to distinguish them from errors.
>
> From the other side: the concurrent index creation is a *new* feature 
> with a *new*
> syntax and its behaviour re. unique indices is explicitly documented. 
> I.e. if one
> uses CONCURRENTLY, it is supposed (s)he read the docs before it ;)
>

Read docs before? May be this happens on the dark side of a moon. :D :D

Getting serious - if one can control arrival of not validated index 
separately from concurrency that becomes much less dangerous. One who 
wants concurrency will use CONCURRENTLY, if new option is added 
explicitly - OK, user should be ready to get warning instead error.

>       * correct logic with hard job, done for index creation, that should
>         better be not repeated applies to traditional (without option
>         CONCURRENTLY) index creation too, but getting NOT VALIDATED
>         index is not
>         supported in this case.
>
>     What about adding one more option that controls ability to get not
>     validated index? As far as I can see it's not dependent from
>     concurrent
>     creation and may be used separately.
>
> Good idea, I support it. Could be useful for restore too.
>

Certainly it will be useful.

doc/sql.extensions/README.index_concurrently.md
```diff
@@ -0,0 +1,131 @@
...
+
+create index using `CONCURRENTLY` clause
+
+	create index idx_tab_f1 on tab(f1) concurrently;
+
+create unique index on column with duplicates
+
+	create unique index idx_tab_id on tab(id);
+
+it trows an error:
```
asfernandes, line 86:
```suggestion
it throws an error:
```

src/dsql/DdlNodes.epp
```diff
@@ -13052,6 +13048,65 @@ Cached::Relation* ModifyIndexNode::getRelByIndex(thread_db* tdbb, const Qualifie
...
+			if (IDX.RDB$INDEX_INACTIVE != 0)
+				status_exception::raise(Arg::Gds(isc_random) << "Can't validate inactive index");
+
+			if (IDX.RDB$UNIQUE_FLAG.NULL || !(IDX.RDB$UNIQUE_FLAG & IDX_UNIQUE))
+				status_exception::raise(Arg::Gds(isc_random) << "Can't validate non-unique index");
+
+			fb_assert(IDX.RDB$UNIQUE_FLAG & IDX_UNIQUE);
+
+			if (!(IDX.RDB$UNIQUE_FLAG & IDX_NOT_VALIDATED))
+				status_exception::raise(Arg::Gds(isc_random) << "Index uniqueness already validated");
```
asfernandes, line 13076: These all should be message codes.

src/dsql/DdlNodes.epp
```diff
@@ -13052,6 +13048,65 @@ Cached::Relation* ModifyIndexNode::getRelByIndex(thread_db* tdbb, const Qualifie
...
+			}
+		}
+		END_FOR
+	}
+	catch (const Exception& ex)
+	{
+		Arg::StatusVector err(ex);
+
+		string msg;
+		msg.printf("Validation of unique index %s failed.", indexName.toQuotedString().c_str());
```
asfernandes, line 13098: This one too, maybe.

src/jrd/btr.h
```diff
@@ -332,6 +336,10 @@ class BtrPageGCLock : public Lock
 
 // Struct used for index creation
 
+class IdxCreationHelper;
+
+enum class IdxCreate { AtOnce, ForRollback, Concurrently };
```
asfernandes: And another standard for enum names is born...

src/jrd/idx.cpp
```diff
@@ -208,6 +210,286 @@ bool IDX_check_master_types(thread_db* tdbb, index_desc& idx, Cached::Relation*
...
+		BTR_delete_tree(m_tdbb, m_creation.relation->getId(), m_creation.index->idx_id, m_tempRoot);
+	}
+
+	bool lockWrites(SSHORT wait)
+	{
+		bool ok = true;
+
+		ThreadStatusGuard guard(m_tdbb);
+
+		/*ok = */m_protectRelations.lock();
```
asfernandes, line 466: SHould remove this comment?

src/jrd/idx.cpp
```diff
@@ -1596,6 +1945,71 @@ void IDX_store(thread_db* tdbb, record_param* rpb, jrd_tra* transaction)
...
+
+				if (!scanner.checkRecord(tdbb, &rpb))
+					continue;
+
+				index_desc* index = scanner.getIndexDesc();
+
+				for (RecordStack::iterator rec(stack); rec.hasData(); ++rec)
+				{
+					if (cmpRecordKeys(tdbb, rec.object(), relation, index,
+						rpb.rpb_record, relation, index))
```
asfernandes, line 1996:
```suggestion
							rpb.rpb_record, relation, index))
```

doc/sql.extensions/README.index_concurrently.md
hvlad: done

src/dsql/DdlNodes.epp
hvlad: done

src/dsql/DdlNodes.epp
hvlad: done too

src/jrd/idx.cpp
hvlad: Not now

src/jrd/idx.cpp
hvlad: done

PR #9110, opened 2026-07-31
Remove no-op isc_info_svc_dump_pool_info handler

Pyolar:
Service::query() and Service::query2() both handle isc_info_svc_dump_pool_info
by reading a 2-byte length from the SPB, advancing items past it, then
calling memcpy / strncpy to copy data into fname.

length2 is only checked against sizeof(fname) to prevent destination
overflow.  There is no check that items + length2 stays within the SPB
buffer (end_items / end_items2).  A crafted SPB with a dump_pool_info
clumplet whose declared length exceeds the remaining buffer causes
out-of-bounds read from the items pointer.

Fix: remove both dead case branches.

### DIFF
--- a/src/jrd/svc.cpp
+++ b/src/jrd/svc.cpp
@@ -1250,18 +1250,6 @@ ISC_STATUS Service::query2(thread_db* /*tdbb*/,
 			}
 			break;
 
-		case isc_info_svc_dump_pool_info:
-			{
-				char fname[MAXPATHLEN];
-				size_t length2 = gds__vax_integer(items, sizeof(USHORT));
-				if (length2 >= sizeof(fname))
-					length2 = sizeof(fname) - 1; // truncation
-				items += sizeof(USHORT);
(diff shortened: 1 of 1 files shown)


hvlad:
Really correct handling should return corresponding error, not silently ignore the value passed.

But, this internal and not documented SPB item is no-op for a long time, and better will be to remove it or to complete implementation, imho.

This is not demand for you to go in this way, just call for opinions - should we fix it properly or remove it finally.

aafemt:
Taking into account that `items` is not advanced after copying, I would say that this code never worked.

+1 for complete removal.

Pyolar:
Removed both dead `isc_info_svc_dump_pool_info` case branches in `Service::query()` and `Service::query2()`. If the preference is to fix it properly instead of deleting, I will update this PR accordingly.

hvlad:
Update title, please

Pyolar:
@hvlad Done, title and description updated.

hvlad (approved):

PR #9117, opened 2026-08-07
Fix non-virtual destructor in TraceDescriptors

Foxpunk:
Problem

`TraceDescriptors` declares a pure virtual method (`fillParams()`) and
is inherited by several derived classes (`TraceDscFromValues`,
`TraceDscFromMsg`, `TraceDscFromDsc`), but its destructor was not
virtual. Deleting a derived object through a `TraceDescriptors*`
pointer would result in undefined behavior, since the derived class's
destructor would not be called.

Fix

Added an explicit `virtual ~TraceDescriptors() = default;` destructor.

### DIFF
--- a/src/jrd/trace/TraceObjects.h
+++ b/src/jrd/trace/TraceObjects.h
@@ -545,7 +545,7 @@ public:
 	{}
 
 protected:
-	void fillParams();
+	void fillParams() override;
 
 private:
 	Request* const m_request;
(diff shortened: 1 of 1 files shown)


sim1984:
At the same time, mark `fillParams` in the inherited classes with the keyword `override`.

Foxpunk:
@sim1984 I added override.

hvlad (approved):

PR #9118, opened 2026-08-07
Add asserts in trace

Foxpunk:
Problem

Several switch statements over enum-like fields, and return values of
time(), fseek() and snprintf(), were not checked, which could lead to
silent incorrect behavior.

Fix

    Added explicit default: fb_assert(false) branches to switch
    statements to catch unexpected values early.
    Saved the return values of time(), fseek() and snprintf() into
    [[maybe_unused]] variables and added fb_assert checks on them.

### DIFF
--- a/src/jrd/trace/TraceCmdLine.cpp
+++ b/src/jrd/trace/TraceCmdLine.cpp
@@ -434,6 +434,8 @@ void fbtrace(UtilSvc* uSvc, TraceSvcIntf* traceSvc)
 			case IN_SW_TRACE_RESUME:
 				usage(uSvc, isc_trace_switch_param_miss, "ID", action_sw->in_sw_name);
 				break;
+			default:
+				fb_assert(false);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


hvlad:
> Problem
> 
> Several switch statements over enum-like fields, and return values of time(), fseek() and snprintf(), were not checked, which could lead to silent incorrect behavior.
> 
> Fix
> 
> Added explicit default: fb_assert(false) branches to switch
> statements to catch unexpected values early.

You should better look at the code and get understanding what it does.

> Saved the return values of time(), fseek() and snprintf() into
> [[maybe_unused]] variables and added fb_assert checks on them.

How this visual spam prevents "silent incorrect behavior" ? Sorry, can't resist.

Foxpunk:
@hvlad I can drop the fb_assert(false) default branches from the switch
statements. I just thought it might help during testing, but I don't
have a strong enough argument for keeping them.

Can I keep the fb_assert checks on the time()/fseek()/snprintf() return
values, since those actually verify something concrete?

I'll also clean up the PR description to be more to the point.

PR #9121, opened 2026-08-08
Validate external file path at metadata load time

Pyolar:
External table paths are stored in RDB$EXTERNAL_FILE at CREATE TABLE time without any check against ExternalFileAccess. The path was only validated when the table was opened, so with a restricted directory list a user could create a table pointing outside the allowed directories and the bad path stayed in metadata.

This moves the check into checkExternalFileAccess(), shared by the metadata loader (met.epp) and ExternalFile::open(). Escaping paths are now rejected at CREATE TABLE with isc_conf_access_denied. Existing databases containing such tables fail on first access instead of keeping the invalid path silently.

Tested with ExternalFileAccess = Restrict: CREATE TABLE with a path outside the list fails with SQLSTATE 28000, tables inside the list work unchanged.

Companion tests: FirebirdSQL/firebird-qa#39

### DIFF
--- a/src/jrd/ext.cpp
+++ b/src/jrd/ext.cpp
@@ -115,6 +115,27 @@ namespace Jrd
 
 using namespace Jrd;
 
+namespace Jrd {
+
+// Validate that an external file path is allowed by the configured
+// ExternalFileAccess policy.  Raises isc_conf_access_denied otherwise.
+// Used at metadata load time (met.epp) and at file open time so that
+// paths escaping the allowed directory list are rejected consistently,
+// regardless of whether the external table was just created or loaded
+// from an existing database.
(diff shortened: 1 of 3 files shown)


pavel-zotov:
::: QA NOTES :::
Two tests have been implemented by sunliqiang (08.08.2026) for this ticket.
I've slightly changed them and put here: [ext_access_allowed](https://github.com/FirebirdSQL/firebird-qa/blob/master/tests/functional/table/external/external_file_access_allowed_test.py) ;  [ext_access_denied](https://github.com/FirebirdSQL/firebird-qa/blob/master/tests/functional/table/external/external_file_access_denied_test.py)

Currently they both are marked to be SKIPPED until this ticked will be fixed.
Enabling these tests requires change to firebird.conf and re-implementing of several other (old) tests -- see notes.

PR #9123, opened 2026-08-13
Correct assert in constantExists function

Noremos:
Schema of a constant may not be specified if there is no search path


pavel-zotov:
::: QA note :::
i can't understand how to reproduce difference before and after this PR has been committed.
Following snippet shows the same:
```
create package pg_test as
begin
    constant PHONE varchar(12) = '+11111111111';
end
;

create schema stock;
create package stock.pg_test as
begin
    constant PHONE varchar(12) = '+22222222222';
end
;

set bail off;
select pg_test.PHONE as phone from rdb$database;
set search_path to stock;
select pg_test.PHONE as phone from rdb$database;
```

```
PHONE                           +11111111111
PHONE                           +22222222222
```

dyemanov:
@pavel-zotov Only the debug build was affected (while running core_6469_test.py).

PR #9124, opened 2026-08-13
postfix: Decrement blr length after reading value from blr in `parse_format` for `blr_int128`

TreeHunter9:


aafemt:
Perhaps, there is also a problem in `PARSE_messages`: `blr_length` is not decremented on consumption of message number, so `parse_format` receives value one bigger than actual blr length.

TreeHunter9:
> Perhaps, there is also a problem in PARSE_messages: blr_length is not decremented on consumption of message number, so parse_format receives value one bigger than actual blr length.

As I can see, `blr_length` is decremented before reading blr value:
```cpp
		if (blr_length-- == 0) // <-- here
		{
			error = true;
			break;
		}

		const USHORT msg_number = *blr++;
```
Or were you referring to some other place?

aafemt:
This decrement is paired with consumption of `blr_message` verb one line above.

Then `msg_number` is consumed, but corresponding decrement of `blr_length` is only 14 lines below:
```
==>		const USHORT msg_number = *blr++;

		rem_fmt* const format = parse_format(blr, blr_length);
		if (!format)
		{
			error = true;
			break;
		}

		RMessage* next = FB_NEW RMessage(format->fmt_length);
		next->msg_next = message;
		message = next;
		message->msg_address = reinterpret_cast<UCHAR*>(format);
		message->msg_number = msg_number;

==>		if (blr_length-- == 0)
		{
			error = true;
			break;
		}

```

TreeHunter9:
> This decrement is paired with consumption of blr_message verb one line above.

Hm, I though it is paired differently:
```cpp
	if (blr_length < 3)
		return NULL;
	blr_length -= 3;  // <--- We need to read 3 u8 values from blr --->

	const SSHORT version = *blr++;  // <--- 1 --->
	if (version != blr_version4 && version != blr_version5)
		return NULL;

	if (*blr++ != blr_begin)  // <--- 2 --->
		return NULL;

	RMessage* message = NULL;

	bool error = false;

	while (*blr++ == blr_message)  // <--- 3 --->
	{
		if (blr_length-- == 0)  // <--- We need to read 1 u8 values from blr --->
		{
			error = true;
			break;
		}

		const USHORT msg_number = *blr++;  // <--- 1 --->

		rem_fmt* const format = parse_format(blr, blr_length);
		if (!format)
		{
			error = true;
			break;
		}

		RMessage* next = FB_NEW RMessage(format->fmt_length);
		next->msg_next = message;
		message = next;
		message->msg_address = reinterpret_cast<UCHAR*>(format);
		message->msg_number = msg_number;

		if (blr_length-- == 0)  // <--- It checks the next `while (*blr++ == blr_message)` iteration --->
		{
			error = true;
			break;
		}
	}
```

aafemt:
Ah, you are right, I'm sorry.

dyemanov (approved):

PR #9134, opened 2026-08-27
Adjust ICU version range to lookup for 53-99 major versions

asfernandes:
ICU 53 is from 2014 and is supposed to be more thread sharing friendly

https://icu.unicode.org/design/cpp/shared_object


aafemt:
Still this lookup takes too long. I would limit it to version explicitly mentioned in fbintl.conf only.

PR #9157, opened 2026-09-19
Improvement: Rewrite hash join hash table for better scalability

sim1984:
# Replace hash join table with a flat bucket-array structure

## Summary

The old `HashJoin` implementation used a fixed-size hash table of 1009 buckets, with a per-bucket `CollisionList` object holding a `SortedArray` of `{hash, position}` entries. On inner streams beyond a few hundred thousand records, average chain lengths reached thousands and lookups degenerated into scattered binary searches over heap-allocated buffers. Attempts to make the table grow dynamically did not help: cardinality estimates for table functions and stored procedures use a fixed default of 1000, so rehashing either never fired or fired when it was already too late, and each rehash added allocation and copying overhead on top of the lookup cost.

This change replaces the table with a flat per-stream bucket array. During the build phase, `put()` only appends raw `{hash, position}` pairs into a single contiguous `Array<Entry>`. Once all inner streams have been read, `build()` picks the final table size from the exact number of collected entries and performs a counting sort over the entries array: one pass to count bucket sizes, one prefix-sum pass to compute offsets, and one redistribution pass into a temporary buffer. After `build()`, records sharing a bucket occupy a contiguous range in a single flat array, sorted by hash. Lookup is a binary search inside that range.

No `CollisionList` objects, no per-bucket allocations, no rehashing at any point.

## What changed

- `HashJoin::HashTable` is rewritten around a `Stream` inner class holding `entries`, `bucketStart`, and an iterator cursor.
- `put()` is a plain append; bucket assignment is deferred until `build()`.
- `build()` chooses the table size from the actual number of entries (`HASH_SIZES` up to 1000003) and performs the counting sort described above.
- `bucketStart` has `tableSize + 1` elements so that a bucket can be expressed as a half-open range without a special case for the last one.
- `setup()` checks that every stream has a non-empty bucket for the target hash, then primes each stream's cursor via `reset()`.
- `HashJoin::maxCapacity()` now returns `MAX_HASH_SIZE * HASH_LOAD_FACTOR`, which corresponds to the actual capacity of the flat layout.

The lookup API (`setup` / `reset` / `iterate`) and its usage from `fetchRecord()` are unchanged, so no changes were needed in the rest of `HashJoin` or in `RecordSource`.

## Test results

A test suite covering pure `GENERATE_SERIES` joins, integer keys with uniform / skewed / low-cardinality / NULL distributions, string keys, composite keys, and indexed variants is attached (`hash-join-tests.sql`). Both builds were run on the same machine, with the same database configuration (`DefaultDbCachePages = 32K`, `TempCacheLimit = 1G`) and freshly created data. All `COUNT(*)` results are identical between the two builds and match the expected values.

Elapsed times, in seconds:

| Test | Old (1009) | New (flat) | Δ |
|---|---|---|---|
| GS 1M × 1k | 0.083 | 0.075 | −10% |
| GS 1M × 100k | 0.123 | 0.101 | **−18%** |
| GS 1M × 1M | 0.448 | 0.326 | **−27%** |
| GS 1M × 10M | 2.799 | 1.690 | **−40%** |
| GS 10M × 10M | 8.728 | 4.875 | **−44%** |
| GS mod-join (duplicates) | 0.559 | 0.508 | −9% |
| GS 3-way chain | 0.088 | 0.077 | −12% |
| 1k × 1k | 0.000 | 0.001 | ~0 |
| 100k × 100k | 0.054 | 0.049 | −9% |
| 1M × 1M | 0.627 | 0.506 | **−19%** |
| 1M × 1k | 0.159 | 0.154 | −3% |
| 1M × 100k | 0.222 | 0.191 | **−14%** |
| Skewed (10k-row hot key) | 0.225 | 0.191 | **−15%** |
| Low cardinality (100 ids) | 0.186 | 0.166 | **−11%** |
| NULL in join key | 0.049 | 0.045 | −8% |
| Composite (int, int) | 0.021 | 0.021 | ~0 |
| Two hashed streams | 0.161 | 0.153 | −5% |
| SEMI (EXISTS) | 0.016 | 0.016 | ~0 |
| VARCHAR(36) join | 0.083 | 0.073 | −12% |
| VARCHAR(36) SEMI | 0.079 | 0.070 | −11% |
| (int, varchar) key | 0.090 | 0.079 | −12% |
| Indexed 100k | 0.054 | 0.050 | −7% |
| Indexed 1M | 0.649 | 0.507 | **−22%** |
| Indexed VARCHAR | 0.078 | 0.072 | −8% |

Summary: parity on inputs below ~10k records, 9–18% faster on 100k, 19–27% faster on 1M, 40–44% faster on 10M. No regressions in any test, including the duplicate-heavy and skewed cases that used to give the old layout trouble.

Query plans are also unchanged between the two builds, including the indexed variants where the optimizer may pick an index-based plan instead of a hash join.

## Memory

Structural memory usage is identical between the two builds — `Delta memory` in the test output matches to the byte for every test. The only difference is the transient peak during `build()`: the flat layout allocates a temporary buffer of the same size as the entries array for the redistribution pass, so `Max memory` on the 10M test grows from ~749 MB to ~839 MB. The temporary buffer is released immediately after `build()` returns; the additional allocation does not persist across queries.

For deployments where a single query runs multiple hash joins concurrently, this peak is worth keeping in mind. A possible future improvement is an in-place redistribution using cycle-following and a bitmask of visited positions, which would eliminate the temporary buffer at the cost of more random memory access during build.

[hash-join-tests.sql](https://github.com/user-attachments/files/32413299/hash-join-tests.sql)


src/jrd/recsrc/HashJoin.cpp
```diff
@@ -43,210 +46,330 @@ using namespace Jrd;
...
+	// off by orders of magnitude; getting the size wrong is exactly
+	// what a dynamic resizing scheme would be paying for later.
+	// Since all inner streams are fully materialized by the time
+	// build() runs, we can afford to decide once and for all.
+	ULONG total = 0;
+	for (auto* s : m_streams)
+		total += s->entries.getCount();
+
+	const ULONG desired = total / (m_streams.getCount() * HASH_LOAD_FACTOR);
+	m_tableSize = nextHashSize(desired);
```
aafemt, line 430: IMHO, there is no need to round up to the next prime. The main purpose of prime's usage is to flatten distribution of keys among array. Hash functions already aimed to even distribution, so usage of primes hardly can improve situation.

sim1984: Using a prime number for the hash table size is an attempt to improve the situation when the hash function is poor.
When calculating the hash, we use this:

```cpp
InternalHash::hash(sub.totalKeyLength, keyBuffer);
```

You can view the implementation of this function here
https://github.com/FirebirdSQL/firebird/blob/master/src/common/classes/Hash.cpp

This leads to two scenarios: on x86, we likely use the hardware implementation of CRC32 (which is significantly better, though not ideal), whereas on ARM, we use a rather poor hash function, `basicHash()`.

In principle, we could check for hardware CRC32 support on ARM as well and utilize it. We could also improve `basicHash()` itself, but that is outside the scope of this PR.

If high-quality hash functions are used, we could switch to hash table sizes that are powers of two. Even CRC32 is not particularly well-suited for hash tables.

TreeHunter9:
Hi, I not very familiar with the part of the code related to the optimizer, but this was very interesting to read.
Looking at the algorithm made me think of a few questions:
1. Do you consider other implementations of hash join, e.g. swiss table?
2. I have seen a few articles about using the Bloom filter in hash joins. Maybe it’s worth trying to implement that?

The 2. question goes beyond the scope of this PR, it's more of a general improvement to the hash join algorithm. Has anyone considered this?

sim1984:
> Hi, I not very familiar with the part of the code related to the optimizer, but this was very interesting to read. Looking at the algorithm made me think of a few questions:
> 
> 1. Do you consider other implementations of hash join, e.g. swiss table?
> 2. I have seen a few articles about using the Bloom filter in hash joins. Maybe it’s worth trying to implement that?
> 
> The 2. question goes beyond the scope of this PR, it's more of a general improvement to the hash join algorithm. Has anyone considered this?

The goal of this PR is to ensure that Hash Join performs just as efficiently with large tables (containing around a million records) as it does with small ones. Note that I modified only the hash table implementation itself, leaving the hash function untouched—though, frankly speaking, our current hash function isn't particularly well-suited for hash tables.

I’ve incorporated some ideas from "Swiss Tables" here. For instance, there are no separate entities representing collision chains; instead, both the collision handling and the hash table itself consist simply of a couple of arrays. As I see it, the major limitation is that we cannot allocate an unlimited amount of memory for the hash table, given that we might have a thousand users executing such queries simultaneously.

There are a few other ideas I haven't tried yet. For example:

1. Experiment with the hash function. If the hash function is effective, the hash table size could be set to a power of 2. This would make calculating the slot index slightly less computationally expensive.

2. Make the table construction process slightly more complex in exchange for saving on peak memory usage.

src/jrd/recsrc/HashJoin.cpp
```diff
@@ -43,210 +46,330 @@ using namespace Jrd;
...
+		for (ULONG k = 0; k < n; k++)
 		{
-			if (const auto collisions = m_collisions[i])
-				collisions->sort();
+			const Entry& e = entries[k];
+			temp[next[e.hash % tableSize]++] = e;
 		}
 
+		entries.assign(temp);
+	}
```
TreeHunter9, line 369: When I looked at the code, I came up with the algorithm that is works without creating a copy:
```cpp
		Array<ULONG> next(getPool(), tableSize);
		next.grow(tableSize);
		memcpy(next.begin(), bucketStart.begin(), tableSize * sizeof(ULONG));

		// Try to rearrange array by placing entry in correct bucket without making copy.
		// `tableSize - 1` is because after iterating over all buckets except the last one, the lasting elements will be
		// already placed at the correct (last) bucket.
		for (ULONG bucket = 0; bucket < tableSize - 1; bucket++)
		{
			// Iterate bucket by bucket and place entry in the correct spot.
			// After placing entry, the bucket start point will be incremented, so when we are starting to process next bucket,
			// we will skip entries, that already at the right bucket.
			for (ULONG k = next[bucket]; k < bucketStart[bucket + 1]; k++)
			{
				Entry& e = entries[k];
				const auto targetBucket = e.hash % tableSize;

				// Already placed at correct bucket, just increment the pointer.
				if (targetBucket == bucket)
				{
					++next[bucket];
					continue;
				}

				// Place entry in correct bucket by swapping current entry with entry that is taking our spot.
				std::swap(e, entries[next[targetBucket]++]);

				// Due to the swap operation, we need to process newly arrived entry at the current position once more.
				--k;
			}
		}
```
To be honest, I have not validate it correctness by running some queries, but for me it looks solid (I hope I'm not missing some edge cases and it should work).

sim1984: Thank you. At first glance, everything is correct. I will try to run tests with this solution.

sim1984: Unfortunately, your approach involving on-the-spot allocation is significantly slower. However, memory conservation is still necessary when dealing with large streams being hashed. That is why a hybrid approach was developed.

The flat hash table built from a temporary buffer during `Stream::build()` allocated a second array the size of the entries array while redistributing records into their buckets. For a 10M-record inner stream this was an extra 80 MB of transient memory per hash join, held only during `build()` but multiplied by the number of concurrent connections running hash joins at the same time.

Introduce a hybrid strategy: below `IN_PLACE_THRESHOLD` (1M records) keep the temporary buffer, because at that size it fits comfortably in L3 and the sequential read path is faster. Above the threshold, use an in-place redistribution by swapping entries into their target bucket within the same array. This eliminates the second allocation entirely for the cases where it matters most, at the cost of random memory access during build.

Measured on the standard test suite (100K page cache, second run of each test after warm-up):

  GS 1M x 1M        0.334s vs 0.476s  (-30%)
  GS 1M x 10M       2.469s vs 2.773s  (-11%)
  GS 10M x 10M      5.772s vs 8.290s  (-30%)
  Table 1M x 1M     0.535s vs 0.669s  (-20%)
  Table 1M x 100k   0.194s vs 0.227s  (-15%)
  Hot key 1M x 100k 0.195s vs 0.237s  (-18%)
  UUID 1M x 1M      0.778s vs 0.909s  (-14%)

No regressions in any test. Structural memory usage (Delta memory) is identical to the fixed-size implementation. Peak memory on 10M-record joins drops from ~839 MB (pure temp buffer) to ~766 MB, matching the old 1009-bucket layout.

The threshold is chosen so that small and medium joins keep the cache-friendly path, while large joins trade a small amount of build time for a significant reduction in transient memory. 

A future separate dataaccess method `BatchHashJoin` may address the memory question more fundamentally; this change is a bounded improvement in the meantime.

TreeHunter9: Yeah, I missed the point of performance consideration, the "copy" path is very fast on the current OOO (Out of Order) CPU architecture and it can process multiple for loop at the same time due to minimal data dependencies on each iteration. The "update-in-place" path is struggling with it, a lot of dependencies on each iteration.
I will try to improve the algorithm performance, but I think that, due to the nature of the algorithm itself, it still won't be able to outperform the “copy” approach.

sim1984: > I will try to improve the algorithm performance, but I think that, due to the nature of the algorithm itself, it still won't be able to outperform the “copy” approach.

The question is, is this actually necessary? In its current form, our Hash Join is a single-pass algorithm. By design, it is not intended to handle datasets with 50 million records. For such cases, we need to switch to a multi-pass Hash Join that utilizes batch processing. While it will be slower than the current implementation, the switch is only worthwhile when joining truly massive tables; the current algorithm will suffice for medium-sized tables.

The concept behind BatchHashJoin is as follows: we estimate the cardinality of the inner (probe) table. If the inner table is large—for instance, if its cardinality exceeds 4 million records—we use BatchHashJoin; otherwise, we use the standard Hash Join. The BatchHashJoin algorithm proceeds as follows:

1. A hash function is calculated for the keys of both the outer (build) and inner (probe) streams. The least significant bits of the resulting hash determine which batch the record belongs to. Each batch for both the outer and inner tables is written to a temporary file. Consequently, we end up with N temporary files containing the datasets for both tables, with the batch number corresponding to the hash's least significant bits.

2. We take the i-th temporary file from the inner table and build a hash table from its records. We then read the corresponding file from the outer table and probe the hash table for each record, outputting a result if a match is found. This process continues until all records in the current batch of the outer table have been processed. We then move on to the next batch for both the outer and inner tables.

Thus, BatchHashJoin never consumes excessive memory. Although the algorithm itself is slower due to the need to read and write temporary files, it remains faster than Merge Join and operates within a limited memory footprint.

This is how hash joins for large tables are implemented in PostgreSQL and MariaDB.

TreeHunter9: > the current algorithm will suffice for medium-sized tables.

I guess you right. I tried to improve it, and the best I could get was ~20% improvements in performance of `build()` function (numbers from `perf` profiling using your test), but since `build()` is only a small part of the total runtime, the overall gain is minimal.
I took your test and comparison looks like this:
```
1.4: 7.441s -> 7.182s
1.5: 13.775s -> 13.326s

For other queries it almost non-existing.
```
Here's the new algorithm, but apparently a few percent improvement doesn't change much overall, so it's up to you whether to use it or not.
<details>
  <summary>New algorithm</summary>

```cpp
		// Try to rearrange array by placing entry in correct bucket without making copy.
		// `tableSize - 1` is because after iterating over all buckets except the last one, the lasting elements will be
		// already placed at the correct (last) bucket.
		for (ULONG bucket = 0; bucket < tableSize - 1; bucket++)
		{
			// Iterate bucket by bucket and place entry in the correct spot.
			// After placing entry, the bucket start point will be incremented, so when we are starting to process next bucket,
			// we will skip entries, that already at the right bucket.
			const auto bucketEnd = bucketStart[bucket + 1];

			// Do several passes of iterating over bucket until this bucket is filled out with correct entries.
			// This is done in such way because of Out-of-Order execution pipeline of modern CPU, so by having
			// minimal data dependencies as possible in for loop is it possible to allow CPU to execute multiple
			// iteration in parallel.
			while (true)
			{
				const auto bucketStart = next[bucket];
				if (bucketStart == bucketEnd)
					break;

				// Handle special case where one entry is left in bucket, so focus on that entry.
				// Basically this is small optimization to eliminate unnecessary `while + for` loop for single entry.
				if (bucketEnd - bucketStart == 1)
				{
					Entry& e = entries[bucketStart];
					while (true)
					{
						const auto targetBucket = e.hash % tableSize;
						std::swap(e, entries[next[targetBucket]++]);

						// Our last entry was found and placed at correct spot, proceed to the next bucket.
						if (targetBucket == bucket)
							break;
					}
					break;
				}

				for (ULONG k = bucketStart; k < bucketEnd; k++)
				{
					Entry& e = entries[k];
					const auto targetBucket = e.hash % tableSize;

					// Place entry in correct bucket by swapping current entry with entry that is taking our spot.
					std::swap(e, entries[next[targetBucket]++]);
				}
			}
		}
```

</details>

sim1984: I applied your algorithm but modified it slightly. Then I ran it against these tests. Here are the results. 
[hash-join-tests.sql](https://github.com/user-attachments/files/32574295/hash-join-tests.sql)


## Test results

Compared against the previous fixed-size 1009-bucket implementation on the same database configuration (`DefaultDbCachePages = 100K`, `TempCacheLimit = 1G`).
Each test is executed twice in a row; the second run is reported, with the page cache warmed up and `Reads = 0` in all join tests.

Elapsed times, in seconds:

| Test | Old (1009) | New (flat + hybrid) | Δ |
|---|---|---|---|
| GS 1M × 1k | 0.081 | 0.076 | −6% |
| GS 1M × 100k | 0.126 | 0.104 | −17% |
| GS 1M × 1M | 0.476 | 0.345 | **−28%** |
| GS 1M × 10M | 2.773 | 1.826 | **−34%** |
| GS 10M × 10M | 8.290 | 5.120 | **−38%** |
| GS mod-join (duplicates) | 0.583 | 0.529 | −9% |
| GS 3-way chain | 0.086 | 0.077 | −10% |
| Table 100k × 100k | 0.060 | 0.048 | −20% |
| Table 1M × 1M | 0.669 | 0.504 | **−25%** |
| Table 1M × 1k | 0.162 | 0.152 | −6% |
| Table 1M × 100k | 0.227 | 0.191 | −16% |
| Skewed (10k-row hot key) | 0.237 | 0.188 | **−21%** |
| Low cardinality (100 ids) | 0.184 | 0.165 | −10% |
| NULL in join key | 0.050 | 0.044 | −12% |
| Composite (int, int) | 0.023 | 0.022 | −4% |
| Two hashed streams | 0.170 | 0.156 | −8% |
| SEMI (EXISTS) | 0.016 | 0.016 | 0% |
| UUID 100k × 100k | 0.080 | 0.071 | −11% |
| UUID 1M × 1M | 0.909 | 0.743 | **−18%** |
| VARCHAR(36) join | 0.082 | 0.074 | −10% |
| VARCHAR(36) SEMI | 0.083 | 0.070 | −16% |
| (int, varchar) key | 0.099 | 0.080 | −19% |
| (int, int) key | 0.023 | 0.020 | −13% |
| Multi-column SEMI | 0.087 | 0.077 | −11% |
| Indexed 100k | 0.063 | 0.048 | −24% |
| Indexed 1M | 0.671 | 0.505 | **−25%** |
| Indexed VARCHAR | 0.084 | 0.071 | −15% |
| Indexed composite | 0.092 | 0.079 | −14% |

Summary: parity on small inputs, 10–25% faster on medium joins, and up to 38% faster on large ones. No regressions in any test.

### Memory

Structural memory (`Delta memory`) is identical to the old layout in every test. Peak memory on 10M-record joins drops from ~839 MB (pure temp buffer) to ~766 MB, matching the old 1009-bucket layout — the in-place redistribution path avoids the second allocation entirely above `IN_PLACE_THRESHOLD`.

### Correctness

All `COUNT(*)` results match the expected values and are identical between the two builds, including NULL keys, duplicate-heavy joins, composite keys, and the multi-stream case. Query plans are unchanged, including the indexed variants where the optimizer may prefer an index-based plan.

